PR #1, opened 2019-05-31
refactor: stop using localStorage

obiyankenobi:
The lib now doesn't require localStorage and clients should pass a storage to the lib using `setStorage()`.

### DIFF
--- a/setupTests.js
+++ b/setupTests.js
@@ -5,42 +5,37 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-let hathorMemoryStorage = {};
 // Creating memory storage to be used in the place of localStorage
-const storageFactory = {
+class MemoryOnlyStore {
+  constructor() {
+    this.hathorMemoryStorage = {};
+  }
+
   getItem(key) {
-    const ret = hathorMemoryStorage[key];
+    const ret = this.hathorMemoryStorage[key];
     if (ret === undefined) {
       return null
     }
     return ret;
-  },
+  }
 
   setItem(key, value) {
-    hathorMemoryStorage[key] = value;
-  },
+    this.hathorMemoryStorage[key] = value;
+  }
 
   removeItem(key) {
-    delete hathorMemoryStorage[key];
-  },
+    delete this.hathorMemoryStorage[key];
+  }
 
   clear() {
-    hathorMemoryStorage = {};
-  },
-
-  key(n) {
-    return Object.keys(hathorMemoryStorage)[n] || null;
-  },
-
-  getAll() {
-    return hathorMemoryStorage;
-  },
+    this.hathorMemoryStorage = {};
+  }
 }
 
 // Mocking localStorage for tests
 import 'jest-localstorage-mock';
 const storage = require('./src/storage').default;
-storage.setStore(storageFactory);
+storage.setStore(new MemoryOnlyStore());
 
 // Mocking WebSocket for tests
 import { Server, WebSocket } from 'mock-socket';
--- a/setupTests.js
+++ b/setupTests.js
@@ -5,8 +5,42 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+let hathorMemoryStorage = {};
+// Creating memory storage to be used in the place of localStorage
+const storageFactory = {
+  getItem(key) {
+    const ret = hathorMemoryStorage[key];
+    if (ret === undefined) {
+      return null
+    }
+    return ret;
+  },
+
+  setItem(key, value) {
+    hathorMemoryStorage[key] = value;
+  },
+
+  removeItem(key) {
+    delete hathorMemoryStorage[key];
+  },
+
+  clear() {
+    hathorMemoryStorage = {};
+  },
+
+  key(n) {
+    return Object.keys(hathorMemoryStorage)[n] || null;
+  },
+
+  getAll() {
+    return hathorMemoryStorage;
+  },
+}
+
 // Mocking localStorage for tests
 import 'jest-localstorage-mock';
+const storage = require('./src/storage').default;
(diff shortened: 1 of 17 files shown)


src/storage.js
```diff
@@ -0,0 +1,41 @@
+class Storage {
+  constructor() {
+    if (!Storage.instance) {
+      this.store = null;
+      this.memory = false;
```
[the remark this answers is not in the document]
obiyankenobi: Nope. Removed it.

setupTests.js
```diff
@@ -5,16 +5,50 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+let hathorMemoryStorage = {};
```
msbrogli: Shouldn't it be one regular implementation of a storage (instead of being here only for tests)?

setupTests.js
```diff
@@ -5,16 +5,50 @@
...
+// Creating memory storage to be used in the place of localStorage
+const storageFactory = {
```
msbrogli: Isn't it better to use a class instead of an object? We know they are almost the same, but the implementation would be easier to read and understand.

src/helpers.js
```diff
@@ -8,6 +8,8 @@
 import { GENESIS_BLOCK, DECIMAL_PLACES, DEFAULT_SERVER } from './constants';
 import path from 'path';
 
+const storage = require('./storage').default;
```
msbrogli: I'd rather something like `import { defaultStorage as storage } from './storage';`.

src/storage.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return Storage.instance;
+  }
+
+  /**
+   * Set the underlying storage object
+   *
+   * @param {Object} myStorage The storage to be used
+   */
+  setStorage(myStorage) {
```
msbrogli: I'd rather call it `setStore`.

src/storage.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * This storage should have the same methods as localStorage (from the browser) and has
+ * to support storaing js objects. It should handle any serialization needed.
+ *
+ * @class
+ * @name Storage
+ */
+class Storage {
+  constructor() {
+    if (!Storage.instance) {
```
msbrogli: I don't feel it should be a singleton.

src/storage.js
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Remove all key and associated objects from storage
+   */
+  clear() {
+    return this.store.clear();
+  }
+
+  /**
+   * Return the name of the nth key in storage
```
msbrogli: Why do we use it?

src/helpers.js
msbrogli: Sorry. I think the correct one would be `import defaultStorage from './storage'`;

src/storage.js
obiyankenobi: I don't think we actually use it. I just implemented so we're fully compatible with the localStorage interface, which is the original store. I'll remove this.

setupTests.js
obiyankenobi: This is a regular implementation of the storage. It's just a in-memory storage, without persistence.

setupTests.js
obiyankenobi: Not sure I follow. Why is it easier to read and understand?

src/storage.js
obiyankenobi: We use a singleton to make sure the same store is used across the lib, just like WebSocketHandler.

setupTests.js
msbrogli: I know. Maybe we should let this implementation available for whoever wants to use it. And I think we should rename it to `MemoryOnlyStore`.

setupTests.js
msbrogli: You had to create a global variable and set the attributes there. I think the encapsulation form is simpler and less error prone (even though this code is very simple).

src/storage.js
msbrogli: You already do it exporting the instance instead of the class, don't you?

src/storage.js
msbrogli: Interesting. I didn't know that. I think we can either keep it or remove it. Your call.

src/helpers.js
obiyankenobi: Done

src/storage.js
obiyankenobi: Done

src/storage.js
obiyankenobi: Remove it

setupTests.js
obiyankenobi: Ok, I created the class but did NOT move into a separate file so other people can use. Using memory storage is very dangerous and users should not do it unless they know exactly what they're doing. In that case, they can just implement their own.

setupTests.js
obiyankenobi: Done

src/storage.js
obiyankenobi: You're correct. Change it.

msbrogli:
> I think we should have a place in the README explaining how to set a storage and what we expect (a `getItem`, a `setItem`, ...)

I agree with it.

setupTests.js
```diff
@@ -5,42 +5,37 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-let hathorMemoryStorage = {};
 // Creating memory storage to be used in the place of localStorage
-const storageFactory = {
+class MemoryOnlyStore {
+  constructor() {
+    this.hathorMemoryStorage = {};
```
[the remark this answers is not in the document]
obiyankenobi: This is just internal, not used outside the class.

PR #2, opened 2019-06-07
feat: creating methods for token management (mint, melt, delegate authority, destroy authority)

pedroferreira1:


src/tokens.js
```diff
@@ -271,101 +272,288 @@ const tokens = {
...
+    if (result.inputsAmount > amount) {
+      // Need to create change output
+      const newAddress = wallet.getAddressToUse();
+      outputs.push({'address': newAddress, 'value': result.inputsAmount - amount, 'tokenData': 1});
+    }
+
+    if (createAnotherMelt) {
+      // New melt authority for this wallet
+      const newAddress = wallet.getAddressToUse();
+      outputs.push({'address': newAddress, 'value': TOKEN_MELT_MASK, 'tokenData': AUTHORITY_TOKEN_DATA});
```
msbrogli: Would it be `AUTHORITY_TOKEN_DATA | token_id`?

src/tokens.js
```diff
@@ -271,101 +272,288 @@ const tokens = {
...
+    // Input targeting the output that contains the mint authority output
+    const input = {'tx_id': txID, 'index': index, 'token': token, 'address': addressSpent};
+
+    // Output1: Mint token amount
+    const outputs = [{'address': address, 'value': amount, 'tokenData': 1}];
+
+    if (createAnotherMint) {
+      // Output2: new mint authority for this wallet
+      const newAddress = wallet.getAddressToUse();
+      outputs.push({'address': newAddress, 'value': TOKEN_MINT_MASK, 'tokenData': AUTHORITY_TOKEN_DATA});
```
msbrogli: Would it be `AUTHORITY_TOKEN_DATA | token_id`?

src/tokens.js
```diff
@@ -271,101 +272,288 @@ const tokens = {
...
+    }
+
+    if (createMelt) {
+      // We create a melt output for this wallet when creating the token
+      const newAddress2 = wallet.getAddressToUse();
+      outputs.push({'address': newAddress2, 'value': TOKEN_MELT_MASK, 'tokenData': AUTHORITY_TOKEN_DATA});
```
msbrogli: Would it be `AUTHORITY_TOKEN_DATA | token_id`?

src/tokens.js
```diff
@@ -271,101 +272,288 @@ const tokens = {
...
+   * @param {string} address Destination address of the delegated authority output
+   * @param {boolean} createAnother If should create another authority output for this wallet, after delegating this one
+   * @param {string} type Authority type to be delegated ('mint' or 'melt')
+   *
+   * @return {Object} Delegate authority data {'inputs', 'outputs', 'tokens'}
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  getDelegateAuthorityData(txID, index, addressSpent, token, address, createAnother, type) {
```
msbrogli: I would rename it to `createDelegateAuthorityData` (instead of get).

src/tokens.js
```diff
@@ -271,101 +272,288 @@ const tokens = {
...
+   * @param {number} amount Amount of the token that will be minted
+   * @param {boolean} createAnotherMint If should create another mint output after spending this one
+   * @param {boolean} createMelt If should create a melt output (useful when creating a new token)
+   *
+   * @return {Object} Mint data {'inputs', 'outputs', 'tokens'}
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  getMintData(txID, index, addressSpent, token, address, amount, createAnotherMint, createMelt) {
```
msbrogli: I would rename it to `createMintData` (instead of get).

src/tokens.js
```diff
@@ -271,101 +272,288 @@ const tokens = {
...
+   * @param {string} token Token uid to be melted
+   * @param {number} amount Amount of the token to be melted
+   * @param {boolean} createAnotherMelt If should create another melt output after spending this one
+   *
+   * @return {Object} Melt data {'inputs', 'outputs', 'tokens'}
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  getMeltData(txID, index, addressSpent, token, amount, createAnotherMelt) {
```
msbrogli: I would rename it to `createMeltData` (instead of get).

src/tokens.js
pedroferreira1: No, actually the `AUTHORITY_TOKEN_DATA` is in https://github.com/HathorNetwork/hathor-wallet-lib/blob/feat/token-management/src/constants.js#L121.

This value is the `TOKEN_AUTHORITY_MASK` with 1 (that is the index of the token we are managing - mint, melt, delegate, destroy). We are assuming at this point that those transactions that execute those actions are only for one token

src/tokens.js
pedroferreira1: Done

src/tokens.js
pedroferreira1: Done

src/tokens.js
pedroferreira1: Done

src/tokens.js
msbrogli: I checked the constants file, and this assumption is not there. I think we should let `AUTHORITY_TOKEN_DATA = 0b10000000`, and we should create and use a new constant called `AUTHORITY_TOKEN_DATA_FIRST` (or something like that).

pedroferreira1: We already have `TOKEN_AUTHORITY_MASK = 0b10000000`.
Maybe I should just rename the variable to `AUTHORITY_TOKEN_DATA_FIRST ` and explain this in the comments. What do you think?

msbrogli: Maybe just improve its comment in the constants file.

I'd suggest something like "Token data for an authority output of the first token in a transaction. As most transactions with authority outputs have only one token, it may be directly used, as a shortcut."

And maybe change `0b10000001` to `TOKEN_AUTHORITY_MASK | 1`. What do you think?

pedroferreira1: Sounds good! Changed it

src/tokens.js
```diff
@@ -271,101 +272,287 @@ const tokens = {
...
-        } else {
-          reject(response.message);
-        }
-      }, (e) => {
-        // Error in request
-        console.log(e);
-        reject(e.message);
+      const txPromise = transaction.sendTransaction(txData, pin);
+      txPromise.then((response) => {
+        // Save in storage and redux new token configuration
```
obiyankenobi: Remove redux from comment

src/tokens.js
```diff
@@ -271,101 +272,287 @@ const tokens = {
...
+  meltTokens(txID, index, addressSpent, token, amount, pin, createAnotherMelt) {
+    // Get melt data
+    let newTxData = this.createMeltData(txID, index, addressSpent, token, amount, createAnotherMelt);
+    return transaction.sendTransaction(newTxData, pin);
+  },
+
+  /**
+   * Get inputs from the amount to be melted
+   *
+   * @param {number} amount Amount of the token to get the inputs
```
obiyankenobi: This is amount of tokens to be melted?

src/tokens.js
```diff
@@ -271,101 +272,287 @@ const tokens = {
...
+      outputs.push({'address': newAddress, 'value': outputValue, 'tokenData': AUTHORITY_TOKEN_DATA});
+    }
+
+    // Create new data
+    const newTxData = {'inputs': [input], 'outputs': outputs, 'tokens': [token]};
+    return newTxData;
+  },
+
+  /**
+   * Delegate authority outputs for and address (mint or melt authority)
```
obiyankenobi: and or an?

src/wallet.js
```diff
@@ -460,7 +460,7 @@ const wallet = {
    * @memberof Wallet
    * @inner
    */
-  hasTokenAndAddress(tx, selectedToken, walletData) {
+  hasTokenAndAddress(tx, selectedToken, walletData, acceptAuthority) {
```
obiyankenobi: Missing `acceptAuthority ` on docstring

src/wallet.js
```diff
@@ -505,12 +505,12 @@ const wallet = {
    * @memberof Wallet
    * @inner
    */
-  filterHistoryTransactions(historyTransactions, selectedToken) {
+  filterHistoryTransactions(historyTransactions, selectedToken, acceptAuthority) {
```
obiyankenobi: Missing acceptAuthority on docstring

src/tokens.js
pedroferreira1: Yes, I've improved the comment

src/tokens.js
pedroferreira1: Done

src/tokens.js
pedroferreira1: Fixed, thanks

src/wallet.js
pedroferreira1: Added, thanks

src/wallet.js
pedroferreira1: Added, thanks

PR #3, opened 2019-06-19
fix: should generate new address only if haven't reached the gap limit yet

pedroferreira1:


src/wallet.js
```diff
@@ -441,7 +441,7 @@ const wallet = {
     // Updating address because the last one was used
     if (this.hasNewAddress()) {
       this.getNextAddress();
-    } else {
+    } else if (this.canGenerateNewAddress()) {
       this.generateNewAddress();
     }
```
msbrogli: I think you must add an else block that throws an exception. What do you think?

pedroferreira1: This method is supposed to return an address to be used (to create change outputs, for example). I don't think we should raise an exception only because we've reached the gap limit, we should just return the current address (without generating a new one), until it receives a tx.

msbrogli: It feels strange to stop generating new addresses. But I think we will review all these policies in the next weeks. So, for now, let's do this patch.

PR #7, opened 2019-07-02
chore: adding license file

pedroferreira1:


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
msbrogli: Shouldn't we add a header in every source-code of the project?

msbrogli: Something like:
```
/**
 * Copyright (c) Hathor Labs and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 */
```

pedroferreira1: We already have this header in all files, only the `LICENSE` file in the root dir that was missing.

PR #9, opened 2019-07-14
feat: creating methods to handle address update

pedroferreira1:


src/wallet.js
```diff
@@ -500,14 +500,36 @@ const wallet = {
...
+    return storage.getItem('wallet:address');
+  },
+
+  /**
+   * Update current address, if has not yet reached the GAP_LIMIT
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  updateCurrentAddress() {
```
msbrogli: I don't like the method name. Maybe `nextAddress`? Or `getNextAddress`?

src/wallet.js
```diff
@@ -500,14 +500,36 @@ const wallet = {
...
     if (this.hasNewAddress()) {
       this.getNextAddress();
     } else if (this.canGenerateNewAddress()) {
       this.generateNewAddress();
     }
-    return address;
```
msbrogli: I think it should move to the next address and return it. What do you think?

src/wallet.js
```diff
@@ -500,14 +500,36 @@ const wallet = {
...
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getCurrentAddress() {
+    return storage.getItem('wallet:address');
+  },
+
+  /**
+   * Update current address, if has not yet reached the GAP_LIMIT
```
msbrogli: Suggestion:

Move to the next address in the derivation chain and return it.

It may not move to the next address if the number of unused addresses has reached the GAP_LIMIT. In this case, it returns the same as `getCurrentAddress`.

src/wallet.js
pedroferreira1: I changed to `getNextAddress `

src/wallet.js
pedroferreira1: Since the method was changed to `getNextAddress `, so updated the method and docs, to include the return

PR #10, opened 2019-07-15
feat: change PIN method

pedroferreira1:


pedroferreira1:
That's true! It was done in the last commit. Can you check again, please? @obiyankenobi

src/wallet.js
```diff
@@ -394,6 +394,39 @@ const wallet = {
...
+    }
+
+    const accessData = storage.getItem('wallet:accessData');
+
+    // Get new PIN hash
+    const newHash = this.hashPassword(newPin);
+    // Update new PIN data in storage
+    accessData['hash'] = newHash.key.toString();
+    accessData['salt'] = newHash.salt;
+    storage.setItem('wallet:accessData', accessData);
```
obiyankenobi: You should move this `setItem` to the end of the method, after decrypting/encrypting the `mainKey`.

src/wallet.js
```diff
@@ -394,6 +394,39 @@ const wallet = {
     }
   },
 
+  /**
+   * Validates old PIN and change it for the new one
```
msbrogli: Typo: Validate old PIN...

src/wallet.js
```diff
@@ -394,6 +394,39 @@ const wallet = {
...
+   *
+   * @param {string} oldPin
+   * @param {string} newPin
+   *
+   * @return {boolean} if PIN was changed
```
msbrogli: Suggestion: True if the PIN was successfully changed.

PR #11, opened 2019-07-20
feat: new create token tx

pedroferreira1:


src/tokens.js
```diff
@@ -271,25 +271,23 @@ const tokens = {
...
     // Create tx data
-    let txData = {'inputs': [input], 'outputs': [authorityOutput, output], 'tokens': [tokenUID]};
+    let txData = {'inputs': [], 'outputs': [authorityOutput], 'tokens': []};
     const promise = new Promise((resolve, reject) => {
       const txPromise = transaction.sendTransaction(txData, pin);
       txPromise.then((response) => {
         // Save in storage new token configuration
-        this.addToken(response.tx.tokens[0], name, symbol);
-        const mintPromise = this.mintTokens(response.tx.hash, 0, address, response.tx.tokens[0], address, mintAmount, pin, {
+        this.addToken(response.tx.hash, name, symbol);
```
obiyankenobi: I'd just create a variable `const tokenUid = response.tx.hash` so there's no doubt why using the response tx hash.

src/tokens.js
```diff
@@ -271,25 +271,23 @@ const tokens = {
   createToken(input, output, address, name, symbol, mintAmount, pin) {
     // Create authority output
     // First the tokens masks that will be the value for the authority output
-    const tokenMasks = TOKEN_CREATION_MASK | TOKEN_MINT_MASK | TOKEN_MELT_MASK;
-    // Create token uid
-    const tokenUID = this.getTokenUID(input.tx_id, input.index);
-    const authorityOutput = {'address': address, 'value': tokenMasks, 'tokenData': AUTHORITY_TOKEN_DATA};
+    const tokenMasks = TOKEN_CREATION_MASK;
```
msbrogli: I think we should include the `TOKEN_MINT_MASK | TOKEN_MELT_MASK` flags. But we should discuss it in the RFC, then we change here according to the decision there.

pedroferreira1:
This merge request does not make sense anymore. This feature has changed a lot and the new create token tx was implemented in the PR (#25)

PR #15, opened 2019-07-26
fix: tests passing and configuring Travis CI

obiyankenobi:


__tests__/date.test.js
```diff
@@ -7,12 +7,11 @@
 
 import dateFormatter from '../src/date';
 
-
 test('Parse timestamp', () => {
   // 2019-05-08 11:43:49
   const timestamp = 1557312229;
-  const formatted = dateFormatter.parseTimestamp(timestamp);
-  expect(formatted).toBe('5/8/2019 11:43:49 AM');
+  const formatted = dateFormatter.parseTimestamp(timestamp, 'UTC');
```
msbrogli: As it returns a string, shouldn't we call it `timestampToString` or something like it?

obiyankenobi: I believe this is not related to this PR.

This would also create additional tasks on wallet desktop and wallet lib.

msbrogli: Ok. I think we should open an issue related to it, and we can skip it in this PR.

__tests__/date.test.js
```diff
@@ -7,12 +7,11 @@
 
 import dateFormatter from '../src/date';
 
-
 test('Parse timestamp', () => {
   // 2019-05-08 11:43:49
```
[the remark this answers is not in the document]
obiyankenobi: Done

PR #17, opened 2019-08-02
feat: add function to set default server

obiyankenobi:


src/helpers.js
```diff
@@ -172,11 +172,15 @@ const helpers = {
    * @inner
    */
   getServerURL() {
-    let server = storage.getItem('wallet:server');
-    if (server === null) {
-      server = DEFAULT_SERVER;
+    const server = storage.getItem('wallet:server');
+    const defaultServer = storage.getItem('wallet:defaultServer');
```
msbrogli: Isn't it confusing to have both a server and a default server?

obiyankenobi: Server is the one it's currently connected. Default server is the one to connect if we don't set any (and if it's not set we'll use the default from the lib).

Initially I wasn't gonna add this and only use `wallet:server`, but when the user resets the wallet the behavior is to remove the selected server but keep the default server.

msbrogli: I understand the use case, but I'm not sure whether this is the best design. We can go forward and change it later if necessary.

PR #18, opened 2019-08-07
feat: token deposit

obiyankenobi:


src/tokens.js
```diff
@@ -611,7 +700,36 @@ const tokens = {
...
       return myIndex + 1;
     }
-  }
+  },
+
+  /**
+   * Get inputs and outputs for token mint deposit. The output will be the difference
+   * between our inputs and the deposit amount.
+   *
+   * @param {int} mintAmount Amount of tokens to mint
```
pedroferreira1: I think it's `number` and not `int`

src/tokens.js
```diff
@@ -268,28 +274,51 @@ const tokens = {
...
+  createToken(address, name, symbol, mintAmount, pin) {
     const promise = new Promise((resolve, reject) => {
+      // get the input used to create token UID.
+      // We use getMintDepositInfo here as it'll raise an error if we don't have enough HTR tokens
+      // for minting the requested amount, before the token uid creation tx happens
+      let inputs, inputsAmount, outputs, outputChange;
+      try {
+        ({inputs, inputsAmount, outputs} = this.getMintDepositInfo(mintAmount));
+        outputChange = wallet.getOutputChange(inputsAmount, 0);
+      } catch (e) {
```
pedroferreira1: You should catch the specific error

src/tokens.js
```diff
@@ -303,34 +332,66 @@ const tokens = {
...
+      // select HTR deposit inputs
+      const depositInfo = this.getMintDepositInfo(amount);
+      inputs.push(...depositInfo.inputs);
+      outputs.push(...depositInfo.outputs);
+    } else {
+      // user chose HTR deposit inputs
+      inputs.push(...depositInputs.inputs);
+      // create change output, if needed
+      const depositAmount = helpers.getDepositAmount(amount);
+      if (depositInputs.amount - depositAmount > 0) {
```
pedroferreira1: If `depositInputs.amount - depositAmount < 0` you should raise an exception because the `mintInput` selected does not have sufficient amount for the deposit

src/tokens.js
```diff
@@ -345,48 +406,71 @@ const tokens = {
...
       minimumTimestamp: 0,
     }, options);
     // Get mint data
-    let newTxData = this.createMintData(txID, index, addressSpent, token, address, amount, fnOptions);
-    return transaction.sendTransaction(newTxData, pin, fnOptions);
+    const promise = new Promise((resolve, reject) => {
+      let newTxData;
+      try {
+        newTxData = this.createMintData(mintInput, token, address, amount, depositInputs, fnOptions);
+      } catch (e) {
```
pedroferreira1: What error you are expecting to catch here? I didn't see any `throw` in the `createMintData ` method. 

And you should catch the specific error

src/tokens.js
```diff
@@ -422,6 +505,10 @@ const tokens = {
       outputs.push({'address': newAddress, 'value': TOKEN_MELT_MASK, 'tokenData': AUTHORITY_TOKEN_DATA});
     }
 
+    // withdraw HTR tokens
+    const withdrawAmount = helpers.getDepositAmount(amount);
```
pedroferreira1: Isn't it supposed to be `getWithdrawAmount`?

src/tokens.js
msbrogli: I agree.

src/tokens.js
obiyankenobi: Done

src/tokens.js
obiyankenobi: Done

src/tokens.js
obiyankenobi: `createMintData ()` uses `getMintDepositInfo()`, which throws.

Changed it to catch specific error.

src/tokens.js
obiyankenobi: Thanks!!

src/tokens.js
obiyankenobi: Done

__tests__/tokens_utils.js
```diff
@@ -47,65 +49,72 @@ test('Token UID', () => {
...
 
+test('Insufficient funds', async (done) => {
+  await loadWallet();
+  const tokenName = 'TestCoin';
+  const tokenSymbol = 'TTC';
+  try {
+    // we only have 100 tokens on wallet, so minting 2000000 should fail (deposit = 20000)
+    tokens.createToken(address, tokenName, tokenSymbol, 2000000, pin);
+    done.fail('Should have rejected');
+  } catch (e) {
```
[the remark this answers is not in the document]
obiyankenobi: Done

src/errors.js
```diff
@@ -26,3 +26,11 @@ export class AddressError extends Error {}
...
  */
 export class OutputValueError extends Error {}
+
+/**
+ * Error thrown when we have insufficient tokens
+ *
+ * @memberof Errors
+ * @inner
+ */
+export class InsufficientTokensError extends Error {}
```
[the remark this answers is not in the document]
obiyankenobi: Done

src/tokens.js
```diff
@@ -22,6 +23,13 @@ import { HATHOR_TOKEN_CONFIG, TOKEN_CREATION_MASK, TOKEN_MINT_MASK, TOKEN_MELT_M
  */
 
 const tokens = {
+
+  /*
+   * This is a default value for the token deposit percentage. We should update
+   * using the version API
+   */
+  depositPercentage: 0.01,
```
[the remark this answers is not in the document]
obiyankenobi: We don't raise an exception when any of the other values from version API is not set. Seems it would be an inconsistent behavior to do this just for `depositPercentage`. Either we change for all or none. If we decide to do it, we'd probably have to update more places than the scope of this PR.

The way things are, I prefer to have a default. If we don't get the value from server and the percentage is wrong, the mint tx won't be accepted anyway.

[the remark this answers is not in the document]
pedroferreira1: I don't think it's a problem to do only for `depositPercentage ` right now. I would prefer doing for this value only and then from now on we always do.

@msbrogli had a big problem because the API values were not set, at least this would have given him a hint.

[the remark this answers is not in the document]
obiyankenobi: Opened #21 for dealing with constants not set. When that is merged to dev, I'll update this branch and use the same logic for `depositPercentage `.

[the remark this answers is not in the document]
obiyankenobi: Changes made. Please review again.

PR #21, opened 2019-08-16
refactor: save tx weight constants in memory and throw error when not set

obiyankenobi:


src/transaction.js
```diff
@@ -363,6 +363,8 @@ const transaction = {
    *  'timestamp': 1,
    * }
    *
+   * @throws {ConstantNotSet} If the weight constants are not set yet
```
obiyankenobi: This function does not throw `ConstantNotSet` but it uses one that does and we don't catch the exception. 

Should we have this docstring here? @jansegre your input is also appreciated.

pedroferreira1: I think in this case where you are not catching the exception you should add in the docstring.

src/transaction.js
```diff
@@ -600,13 +604,23 @@ const transaction = {
    *
    * @return {Object} Object with the parameters {'txMinWeight', 'txWeightCoefficient', 'txMinWeightK'}
    *
+   * @throws {ConstantNotSet} If the weight constants are not set yet
+   *
    * @memberof Transaction
    * @inner
    */
   getTransactionWeightConstants() {
-    return {'txMinWeight': parseFloat(storage.getItem('wallet:txMinWeight')),
```
msbrogli: Why don't we use a single key to all of them?
The basic usage would be:
```
const txParams = getTransactionWeightContants();
txParams.txMinWeight
```

src/errors.js
```diff
@@ -26,3 +26,11 @@ export class AddressError extends Error {}
...
  */
 export class OutputValueError extends Error {}
+
+/**
+ * Error thrown when a constant that we get from the server is not set
+ *
+ * @memberof Errors
+ * @inner
+ */
+export class ConstantNotSet extends Error {}
```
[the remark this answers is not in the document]
obiyankenobi: https://github.com/HathorNetwork/hathor-wallet-lib/issues/22

src/transaction.js
obiyankenobi: Done

Issue #20, opened 2019-08-15
Rejection exception object not exposed


obiyankenobi:
In several places the exception object is not passed to `reject()`, just the exception's message.

We do stuff like this:
```
try {
  ...
} catch (e) {
  reject(e.message);
}
```

Passing the exception object is convenient so clients can catch specific errors.

If we decide to change this, it'll impact both wallets.

PR #25, opened 2019-09-13
feat: updating create token method for the new creation token tx

### LINKED ISSUE #24: Add support for new token structure
The new token structure is described in Hathor RFC 4 [1].

The new transaction has different serialization and the initial mint is done on the very first transaction (as opposed to the 2 transactions we have nowadays).

[1] RFC 4 - Token: https://gitlab.com/HathorNetwork/rfcs/blob/token-creation/text/0004-tokens.md

pedroferreira1:
Closes #24

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -69,6 +69,11 @@ export const DEFAULT_SERVERS = [
 export const DEFAULT_SERVER = DEFAULT_SERVERS[0];
 
 /**
+ * Block version field
+ */
+export const BLOCK_VERSION = 0;
+
+/**
  * Transaction version field
  */
 export const DEFAULT_TX_VERSION = 1;
--- a/src/constants.js
+++ b/src/constants.js
@@ -111,12 +111,12 @@ export const TOKEN_INDEX_MASK = 0b01111111
 export const TOKEN_AUTHORITY_MASK = 0b10000000
 
 /**
- * Mask to check if it's mint UTXO (second to last bit indicates it)
+ * Mask to check if it's mint UTXO (last bit indicates it)
  */
 export const TOKEN_MINT_MASK = 0b00000001
 
 /**
- * Mask to check if it's melt UTXO (third bit from right to left indicates it)
+ * Mask to check if it's melt UTXO (second to last bit indicates it)
  */
 export const TOKEN_MELT_MASK = 0b00000010
 
--- a/src/constants.js
+++ b/src/constants.js
@@ -112,16 +112,19 @@ export const TOKEN_INDEX_MASK = 0b01111111
 
 /**
  * Mask to check if it's authority output (first bit indicates it)
+ * For further information: https://gitlab.com/HathorNetwork/rfcs/blob/master/text/0004-tokens.md
  */
 export const TOKEN_AUTHORITY_MASK = 0b10000000
 
 /**
  * Mask to check if it's mint UTXO (last bit indicates it)
+ * For further information: https://gitlab.com/HathorNetwork/rfcs/blob/master/text/0004-tokens.md
  */
 export const TOKEN_MINT_MASK = 0b00000001
 
 /**
  * Mask to check if it's melt UTXO (second to last bit indicates it)
+ * For further information: https://gitlab.com/HathorNetwork/rfcs/blob/master/text/0004-tokens.md
  */
 export const TOKEN_MELT_MASK = 0b00000010
 
--- a/src/constants.js
+++ b/src/constants.js
@@ -71,8 +71,18 @@ export const DEFAULT_SERVER = DEFAULT_SERVERS[0];
 /**
  * Transaction version field
  */
-// FIXME tx version should not be hardcoded
-export const DEFAULT_TX_VERSION  = 1;
+export const DEFAULT_TX_VERSION = 1;
+
+/**
+ * Create token transaction version field
+ */
+export const CREATE_TOKEN_TX_VERSION = 2;
+
+/**
+ * Create token information version
+ * so far we expect name and symbol
+ */
+export const TOKEN_INFO_VERSION = 1;
 
 /**
  * Max value (inclusive) before having to use 8 bytes: 2147483648 ~= 2.14748e+09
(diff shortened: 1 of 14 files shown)


src/constants.js
```diff
@@ -100,20 +110,15 @@ export const TOKEN_INDEX_MASK = 0b01111111
  */
 export const TOKEN_AUTHORITY_MASK = 0b10000000
 
-/**
- * Mask to check if it's token id creation UTXO (last bit indicates it)
- */
-export const TOKEN_CREATION_MASK = 0b00000001
-
 /**
  * Mask to check if it's mint UTXO (second to last bit indicates it)
```
obiyankenobi: Need to update comment (not 'second to last' anymore).

src/constants.js
```diff
@@ -100,20 +110,15 @@ export const TOKEN_INDEX_MASK = 0b01111111
...
  */
-export const TOKEN_MINT_MASK = 0b00000010
+export const TOKEN_MINT_MASK = 0b00000001
 
 /**
  * Mask to check if it's melt UTXO (third bit from right to left indicates it)
```
obiyankenobi: Update comment.

src/transaction.js
```diff
@@ -636,6 +680,40 @@ const transaction = {
...
+   * {
+   *  'name': 'TokenName',
+   *  'symbol': 'TKN',
+   * }
+   *
+   * @return {Array} array of bytes
+   * @memberof Transaction
+   * @inner
+   */
+  createTokenTxInfo(txData) {
```
obiyankenobi: The name `createTokenTxInfo` seems misleading to me, as we're not creating anything. I prefer `serializeTokenInfo` or something along those lines.

pedroferreira1: Agreed! Thanks

src/constants.js
```diff
@@ -101,19 +111,14 @@ export const TOKEN_INDEX_MASK = 0b01111111
 export const TOKEN_AUTHORITY_MASK = 0b10000000
 
 /**
- * Mask to check if it's token id creation UTXO (last bit indicates it)
- */
-export const TOKEN_CREATION_MASK = 0b00000001
-
-/**
- * Mask to check if it's mint UTXO (second to last bit indicates it)
+ * Mask to check if it's mint UTXO (last bit indicates it)
```
msbrogli: Maybe add a link to the RFC for further information.

src/constants.js
```diff
@@ -101,19 +111,14 @@ export const TOKEN_INDEX_MASK = 0b01111111
...
  */
-export const TOKEN_MINT_MASK = 0b00000010
+export const TOKEN_MINT_MASK = 0b00000001
 
 /**
- * Mask to check if it's melt UTXO (third bit from right to left indicates it)
+ * Mask to check if it's melt UTXO (second to last bit indicates it)
```
msbrogli: Maybe add a link to the RFC for further information.

src/transaction.js
```diff
@@ -435,11 +450,28 @@ const transaction = {
...
+   *
+   * @memberof Transaction
+   * @inner
+   */
+  setWeight(data) {
     // Calculate tx weight if needed.
-    if (incompleteTxData.weight == 0) {
-      let minimumWeight = this.calculateTxWeight(incompleteTxData);
-      incompleteTxData['weight'] = minimumWeight;
+    if (data.weight === 0) {
```
msbrogli: Why would we do this `if`? Maybe add a `force` parameter? Or create a `setWeightIfNeeded`?

pedroferreira1: The `if` is in case the user does not want to set the weight, we calculate it, otherwise we use what the user set.

Maybe could check if `if ('weight' in data)`. What do you think?

src/constants.js
pedroferreira1: Should we add the current link? The one still in the MR? Or wait if it's merged?

src/transaction.js
msbrogli: I think that if the user wants to calculate the weight, they just skip calling this method. Or they could call another method called `setWeightIfNeeded`. But I don't like skipping the weight calculation when I call `setWeight`. I think this might be error prone. What do you think?

src/constants.js
msbrogli: In this case, I think we can merge the RFC first (since it is ready or almost ready to be merged).

src/transaction.js
pedroferreira1: I agree. But currently we don't have any case where we always want to set weight. So I will just update the name of this method to `setWeightIfNeeded` to avoid confusion. Is it ok?

src/constants.js
pedroferreira1: @obiyankenobi do you think we will merge the RFC soon?

Or should I add a #TODO in the code to add this link in the future?

obiyankenobi: There are just 2 discussions open there: https://gitlab.com/HathorNetwork/rfcs/merge_requests/10

msbrogli: I've replied in both discussions in the RFC.

src/transaction.js
msbrogli: I think it is. Thanks!

src/tokens.js
```diff
@@ -93,16 +93,19 @@ const tokens = {
...
    *
    * @memberof Tokens
    * @inner
    */
   unregisterToken(uid) {
-    const tokens = this.getTokens();
-    const filteredTokens = tokens.filter((token) => token.uid !== uid);
-    this.saveToStorage(filteredTokens);
-    return filteredTokens;
+    const promise = new Promise((resolve, _) => {
```
obiyankenobi: I don't think we should use an exception here. There are no async operations.

pedroferreira1: If the user decides to use an async storage (e.g. https://github.com/callstack/async-storage) it might have.

pedroferreira1: I agree that nothing is prepared, we still need a big refactor. But if we start preparing every code we change it's better.

It's like this issue here: https://github.com/HathorNetwork/hathor-wallet-lib/issues/20. We still have lot of places that we need to change but the new ones (or when we change something), we are already updating.

To be honest the error handling is needed. This way or using a `try ... catch`. I don't see anything bad (it's a little bit slower but almost nothing because it's a simple code)

obiyankenobi: In this case I don't think it makes sense because we're indicating the operation is async, when it's clearly not. In other operations where we only access the storage (like `tokenInfoExists()` or `getTokens()`), we don't use promises. This would make the lib more confusing to users.

msbrogli: We are indicating that the operation may be async, which is true. We know that we need a larger refactor to allow async operations in the whole storage, but I feel new methods should do it from the beginning. I don't think it is more confusing for users because all classes have some sync and some async methods.

obiyankenobi: We're indicating the operation *IS* async.

>The Promise object represents the eventual completion (or failure) of an asynchronous operation, and its resulting value.

We just need to make the standard in our lib then to always use promises instead of try/catch.

src/constants.js
pedroferreira1: Added

Issue #22, opened 2019-08-19
Create base class HathorLibError for errors and exceptions


obiyankenobi:
We should extend all errors in the lib from this base class so it's easy to catch them all.

PR #26, opened 2019-09-19
feat: improvements token info

pedroferreira1:
- Adding api to get token info;
- Adding extra validations when a token is registered. Verifying if name and symbol match with the one in the DAG and checking if there is no other token with this name, or symbol already registered;
- Using version field to get transaction type (block, transaction, token creation tx).

### DIFF
--- a/src/api/wallet.js
+++ b/src/api/wallet.js
@@ -71,6 +71,36 @@ const walletApi = {
       return Promise.reject(res);
     });
   },
+
+  /**
+   * Call get token transaction history API
+   *
+   * @param {string} uid Token uid to get the info
+   * @param {number} count Quantity of elements to be returned
+   * @param {string} hash Hash of transaction as reference in pagination
+   * @param {number} timestamp Timestamp of transaction as reference in pagination
+   * @param {string} page The button clicked in the pagination ('previous' or 'next')
+   * @param {function} resolve Method to be called after response arrives
+   *
+   * @return {Promise}
+   * @memberof ApiWallet
+   * @inner
+   */
+  getTokenHistory(uid, count, hash, timestamp, page, resolve) {
+    const data = {id: uid, count};
+
+    if (hash) {
+      data['hash'] = hash
+      data['timestamp'] = timestamp
+      data['page'] = page
+    }
+
+    return createRequestInstance(resolve).get('thin_wallet/token_history', {'params': data}).then((res) => {
+      resolve(res.data)
+    }, (res) => {
+      return Promise.reject(res);
+    });
+  },
 };
 
 export default walletApi;
--- a/src/api/wallet.js
+++ b/src/api/wallet.js
@@ -53,6 +53,24 @@ const walletApi = {
     });
   },
 
+  /**
+   * Call get token info API
+   *
+   * @param {string} uid Token uid to get the info
+   * @param {function} resolve Method to be called after response arrives
+   *
+   * @return {Promise}
+   * @memberof ApiWallet
+   * @inner
+   */
+  getTokenInfo(uid, resolve) {
+    const data = {id: uid};
+    return createRequestInstance(resolve).get('thin_wallet/token', {'params': data}).then((res) => {
+      resolve(res.data)
+    }, (res) => {
+      return Promise.reject(res);
+    });
+  },
 };
 
 export default walletApi;
--- a/src/api/wallet.js
+++ b/src/api/wallet.js
@@ -54,16 +54,16 @@ const walletApi = {
   },
 
   /**
-   * Call get token info API
+   * Call get general token info API
    *
-   * @param {string} uid Token uid to get the info
+   * @param {string} uid Token uid to get the general info
    * @param {function} resolve Method to be called after response arrives
    *
    * @return {Promise}
    * @memberof ApiWallet
    * @inner
    */
-  getTokenInfo(uid, resolve) {
+  getGeneralTokenInfo(uid, resolve) {
     const data = {id: uid};
     return createRequestInstance(resolve).get('thin_wallet/token', {'params': data}).then((res) => {
       resolve(res.data)
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -9,7 +9,7 @@ import path from 'path';
 
 import storage from './storage';
 import tokens from './tokens';
-import { GENESIS_BLOCK, DECIMAL_PLACES, DEFAULT_SERVER } from './constants';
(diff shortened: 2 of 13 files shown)


src/tokens.js
```diff
@@ -112,45 +112,72 @@ const tokens = {
...
    */
-  validateTokenToAddByUid(uid) {
-    const existedToken = this.tokenExists(uid);
-    if (existedToken) {
-      return {success: false, message: `You already have this token: ${uid} (${existedToken.name})`};
-    }
+  validateTokenToAddByUid(uid, name, symbol) {
+    const promise = new Promise((resolve, reject) => {
+      // Validate if token uid was already added
+      const existedToken = this.tokenExists(uid);
```
obiyankenobi: existedToken?

src/tokens.js
```diff
@@ -112,45 +112,72 @@ const tokens = {
...
-    if (tokenData === null) {
-      return {success: false, message: 'Invalid configuration string'};
-    }
-    if (uid && uid !== tokenData.uid) {
-      return {success: false, message: `Configuration string uid does not match: ${uid} != ${tokenData.uid}`};
-    }
+    const promise = new Promise((resolve, reject) => {
+      const tokenData = this.getTokenFromConfigurationString(config);
+      if (tokenData === null) {
+        reject('Invalid configuration string');
```
obiyankenobi: We have issue #22 open so that we reject the error object, not just the message. This way, clients using the lib can catch specific errors. I think we can start doing in the parts of the lib we are refactoring.

If you expose the error, you can later test it instead of expecting a generic string `rejects.toEqual(expect.any(String))`. We do something like this in the test 'Insufficient funds' for `__tests__/token_utils.js`.

We would need to create additional error classes, like TokenValidationError and so on.

src/tokens.js
pedroferreira1: Reading again it seems a bad var name, changed to `token`. Thanks

src/tokens.js
pedroferreira1: Agreed! Now the error is being rejected

pedroferreira1: I did the test a little bit different than the 'Insufficient funds' one, do you think is it a problem how it's done now? 

https://github.com/HathorNetwork/hathor-wallet-lib/pull/26/files#diff-616b0e865ab989f74898b8625441a9d3R183

obiyankenobi: Yeah, looks even better than the other code.

src/api/wallet.js
```diff
@@ -53,6 +53,54 @@ const walletApi = {
...
+   * Call get token info API
+   *
+   * @param {string} uid Token uid to get the info
+   * @param {function} resolve Method to be called after response arrives
+   *
+   * @return {Promise}
+   * @memberof ApiWallet
+   * @inner
+   */
+  getTokenInfo(uid, resolve) {
```
msbrogli: I think we are using "tokenInfo" for two different things. It is both the information of the token inside the create token transaction and the general token info from the API, right? Maybe we should rename one of them.

src/helpers.js
```diff
@@ -56,7 +56,11 @@ const helpers = {
     if (this.isBlock(tx)) {
       return 'Block';
     } else {
-      return 'Transaction';
+      if (tx.version === DEFAULT_TX_VERSION) {
+        return 'Transaction';
+      } else if (tx.version === CREATE_TOKEN_TX_VERSION) {
+        return 'Create token transaction';
+      }
     }
```
msbrogli: I think we should include a `Unknown` if there is no match.

src/tokens.js
```diff
@@ -259,6 +286,30 @@ const tokens = {
...
+   *
+   * @return {Object|null} Token if name or symbol already exists, else null
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  tokenInfoExists(name, symbol) {
+    const tokens = this.getTokens();
+    for (const token of tokens) {
+      if (token.name === name) {
```
msbrogli: Maybe we should do some cleanup, such as removing spaces, etc. Maybe we should prohibit some of them in the CreateTokenTransaction.

src/tokens.js
```diff
@@ -259,6 +286,30 @@ const tokens = {
...
+        return {token, key: 'name'};
+      }
+      if (token.symbol === symbol) {
```
msbrogli: The same here. A simple comparison seems too simple and trivial to be bypassed.

src/tokens.js
pedroferreira1: I've done the strip cleanup, do you think we should do more? Maybe begin with this only?

src/tokens.js
pedroferreira1: I've done the strip cleanup, do you think we should do more? Maybe begin with this only?

src/api/wallet.js
pedroferreira1: I agree. I've renamed this API to to generalTokenInfo, what do you think?

src/tokens.js
msbrogli: Maybe we should create a cleanup method. I would be concerned about:
1) Uppercase vs Lowercase
2) Spaces at the beginning or at the end
3) Multiples spaces in the middle (in face, should we allow any space at all in the symbol?)
4) Special characters with unexpected results.

src/tokens.js
msbrogli: Maybe we should create a cleanup method. I would be concerned about:
1) Uppercase vs Lowercase
2) Spaces at the beginning or at the end
3) Multiples spaces in the middle (in face, should we allow any space at all in the symbol?)
4) Special characters with unexpected results.

src/tokens.js
pedroferreira1: I've done the cleanup taking care of 1, 2 and 3. Maybe we could discuss more if we should take care of special chars for now. Do you think it's enough for now and I create the task to discuss this point 4?

src/api/wallet.js
msbrogli: Seems good. Thanks!

src/tokens.js
msbrogli: I think it is enough for now. Let's create a new issue and discuss it further before implementing different rules.

pedroferreira1: Ok, perfect

src/tokens.js
pedroferreira1: Same as conversation https://github.com/HathorNetwork/hathor-wallet-lib/pull/26#discussion_r330072548

PR #33, opened 2019-12-03
feat: adding new methods for hardware wallet

pedroferreira1:
- [x] Tests for new methods

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -7,7 +7,7 @@
 
 import { GAP_LIMIT, LIMIT_ADDRESS_GENERATION, HATHOR_BIP44_CODE, NETWORK, TOKEN_MINT_MASK, TOKEN_MELT_MASK, TOKEN_INDEX_MASK, HATHOR_TOKEN_INDEX, HATHOR_TOKEN_CONFIG, MAX_OUTPUT_VALUE, HASH_KEY_SIZE, HASH_ITERATIONS } from './constants';
 import Mnemonic from 'bitcore-mnemonic';
-import { HDPublicKey, Address } from 'bitcore-lib';
+import { HDPublicKey, Address, crypto } from 'bitcore-lib';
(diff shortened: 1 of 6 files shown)


src/wallet.js
```diff
@@ -1097,6 +1116,16 @@ const wallet = {
     storage.setItem('wallet:closed', true);
   },
 
+  /*
+   * Set in storage as not closed
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  afterOpen() {
```
msbrogli: I don't get it. When is it called? What does `'wallet:closed'` mean?

src/wallet.js
```diff
@@ -134,12 +134,30 @@ const wallet = {
...
+   * @param {string} xpub Xpub string of the wallet being started
+   * @param {boolean} loadHistory if should load the history from the generated addresses
+   *
+   * @return {Promise} Promise that resolves when finishes loading address history, in case loadHistory = true, else returns null
+   * @memberof Wallet
+   * @inner
+   */
+  startWallet(accessData, xpub, loadHistory) {
+    WebSocketHandler.setup();
+    this.afterOpen();
```
msbrogli: Shouldn't we call it after the connection has been established?

src/wallet.js
obiyankenobi: It's used to indicate if Hathor Wallet was closed, as we lock it. So next time user opens it, it will ask for pin.

src/wallet.js
obiyankenobi: This does not refer to the websocket being open, but to the wallet state, as I explained on the other comment.

I change the method name to `setWalletAsOpen`.

msbrogli: Great. Thanks!

src/wallet.js
msbrogli: Ok. Thanks!

PR #37, opened 2020-01-14
feat: methods to validate block reward lock when choosing inputs of a tx

### LINKED ISSUE #35: Handle block reward locks when choosing inputs of a tx and calculating balance
The wallet-lib will get the height of the network when calculating if an utxo can be spent (to check if the block reward is locked). Besides that, we will calculate the balance in real time considering the height of the network.

To do it we must update codes in 4 repositories: full node, wallet lib, wallet desktop and wallet mobile.

- **Full node:** we need to have an API where we could get the current height of the network. Besides that we need also to have the height sent in the metrics ws message.
  - **Issue:** https://github.com/HathorNetwork/hathor-core/issues/45
  - **MRs:**
    - **Get height from API:** https://gitlab.com/HathorNetwork/hathor-python/merge_requests/359
    - **Send height on ws:** https://gitlab.com/HathorNetwork/hathor-python/merge_requests/360

- **Wallet lib:** we need to update the method that check if an utxo can be spent to consider the height of the network getting the value from the API. We also need to consider the height when calculating the balance.
  - **Issue:** https://github.com/HathorNetwork/hathor-wallet-lib/issues/35
  - **PR:** https://github.com/HathorNetwork/hathor-wallet-lib/pull/40

- **Wallet desktop:** now methods to prepare data to be sent is async because of the API request to get network height, so we need to do this refactor. Besides that we need to get current height from API and save in redux to calculate balance in real time.
  - **PR:** Prepare send tokens data and calculate balance:
    - https://github.com/HathorNetwork/hathor-wallet/pull/65

- **Wallet mobile:** now methods to prepare data to be sent is async because of the API request to get network height, so we need to do this refactor. Besides that we need to get current height from API and save in redux to calculate balance in real time.
  - **PR:** Prepare send tokens data and calculate balance:
    - https://github.com/HathorNetwork/hathor-wallet-mobile/pull/39

pedroferreira1:
Closes #35

### DIFF
--- a/src/tokens.js
+++ b/src/tokens.js
@@ -334,26 +334,30 @@ const tokens = {
         createMelt: true,
       };
 
-      const txData = this.createMintData(null, null, address, mintAmount, null, mintOptions);
+      const dataPromise = this.createMintData(null, null, address, mintAmount, null, mintOptions);
 
-      // Set create token tx version value
-      const createTokenTxData = Object.assign(txData, {
-        version: CREATE_TOKEN_TX_VERSION,
-        name,
-        symbol,
-      });
+      dataPromise.then((txData) => {
+        // Set create token tx version value
+        const createTokenTxData = Object.assign(txData, {
+          version: CREATE_TOKEN_TX_VERSION,
+          name,
+          symbol,
+        });
 
-      const txPromise = transaction.sendTransaction(createTokenTxData, pin);
-      txPromise.then((response) => {
-        // Save in storage new token configuration
-        const tokenUid = response.tx.hash;
-        this.addToken(tokenUid, name, symbol);
-        resolve({uid: tokenUid, name, symbol});
-      }, (message) => {
-        // I need to reject an error because we've changed the createMintData to reject an error
-        // Changing sendTransaction method to reject an error also would require refactor in other methods
-        // We already have an issue to always reject an error but while we don't do it, we need this
-        reject(new Error(message));
+        const txPromise = transaction.sendTransaction(createTokenTxData, pin);
+        txPromise.then((response) => {
+          // Save in storage new token configuration
+          const tokenUid = response.tx.hash;
+          this.addToken(tokenUid, name, symbol);
+          resolve({uid: tokenUid, name, symbol});
+        }, (message) => {
+          // I need to reject an error because we've changed the createMintData to reject an error
+          // Changing sendTransaction method to reject an error also would require refactor in other methods
+          // We already have an issue to always reject an error but while we don't do it, we need this
+          reject(new Error(message));
+        });
+      }, (e) => {
+        reject(e);
       });
     });
     return promise;
@@ -386,12 +390,12 @@ const tokens = {
    *
    * @throws {InsufficientFundsError} If not enough tokens for deposit
    *
-   * @return {Object} Mint data {'inputs', 'outputs', 'tokens'}
+   * @return {Promise} Promise resolved with object containing mint data {'inputs', 'outputs', 'tokens'}
    *
    * @memberof Tokens
    * @inner
    */
-  createMintData(mintInput, token, address, amount, depositInputs, options) {
+  async createMintData(mintInput, token, address, amount, depositInputs, options) {
     const fnOptions = Object.assign({
       createAnotherMint: true,
       createMelt: false,
@@ -403,7 +407,7 @@ const tokens = {
 
     if (!depositInputs) {
       // select HTR deposit inputs
-      const depositInfo = this.getMintDepositInfo(amount);
+      const depositInfo = await this.getMintDepositInfo(amount);
       inputs.push(...depositInfo.inputs);
       outputs.push(...depositInfo.outputs);
     } else {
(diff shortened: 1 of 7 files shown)


src/version.js
```diff
@@ -29,6 +30,7 @@ const version = {
       versionApi.getVersion((data) => {
         // Update transaction weight constants
         transaction.updateTransactionWeightConstants(data.min_tx_weight, data.min_tx_weight_coefficient, data.min_tx_weight_k);
+        wallet.updateRewardLockConstant(data.reward_spend_min_blocks);
```
msbrogli: We should update the minimum API version.

src/wallet.js
```diff
@@ -690,24 +696,49 @@ const wallet = {
...
+   * @param {number} currentHeight Height of the last block of the network. If not passed, we get it from the API.
+   *
+   * @return {Promise} Promise resolving a boolean if the utxo can be spent
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  async canUseUnspentTx(unspentTx, blockHeight, currentHeight) {
+    if (currentHeight === undefined && blockHeight) {
+      const response = await walletApi.getMiningInfoRequest();
```
msbrogli: You can check if current height is enough. If it is, then you can return without running the `getMiningInfoRequest`, otherwise you do the request.

src/wallet.js
```diff
@@ -1235,14 +1274,16 @@ const wallet = {
...
    * @memberof Wallet
    * @inner
    */
-  prepareSendTokensData(data, token, chooseInputs, historyTransactions, allTokens) {
+  async prepareSendTokensData(data, token, chooseInputs, historyTransactions, allTokens) {
     // Get the data and verify if we need to select the inputs or add a change output
-
+    // Need to use a async/await because we get the current height of the blockchain
+    // from api so we can check if a block reward can be spent
+    const response = await walletApi.getMiningInfoRequest();
```
msbrogli: You're calling it in multiple places. Maybe you should schedule an update every N seconds, and create a `getMiningInfoRequestIfNeeded()` that will only call if at most X seconds has passed. What do you think?

src/version.js
pedroferreira1: Yes. I usually update this MIN_API_VERSION after merging the full node MR's and updating the VERSION field there.

src/wallet.js
pedroferreira1: I just call it if is not passed in the parameter. In some methods this is useful because we call `canUseUnspentTx` many times, so I execute `getMiningInfoRequest` only once and pass as parameter.

See method `prepareSendTokensData`. I call `canUseUnspentTx` in a loop, so instead of requesting the data from the API every call I execute one time in `prepareSendTokensData` and pass as parameter.

src/wallet.js
pedroferreira1: I only execute the API call if it's not passed as parameter.

src/tokens.js
```diff
@@ -386,12 +390,12 @@ const tokens = {
...
    * @throws {InsufficientFundsError} If not enough tokens for deposit
    *
-   * @return {Object} Mint data {'inputs', 'outputs', 'tokens'}
+   * @return {Promise} Promise resolved with object containing mint data {'inputs', 'outputs', 'tokens'}
    *
    * @memberof Tokens
    * @inner
    */
-  createMintData(mintInput, token, address, amount, depositInputs, options) {
+  async createMintData(mintInput, token, address, amount, depositInputs, options) {
```
obiyankenobi: It looks weird to me that we're mixing `async/await` with promises, especially in the same function.

src/tokens.js
```diff
@@ -755,16 +763,16 @@ const tokens = {
...
    * @throws {InsufficientFundsError} If not enough tokens for deposit
    *
-   * @return {Object} Mint inputs/outputs data {'inputs', 'outputs'}
+   * @return {Promise} Promise resolved with object containing mint inputs/outputs data {'inputs', 'outputs'}
    *
    * @memberof Tokens
    * @inner
    */
-  getMintDepositInfo(mintAmount) {
+  async getMintDepositInfo(mintAmount) {
```
obiyankenobi: We also have some functions that are `async` and others are not. Personally, I prefer to use `async` as the code becomes easier to read. Maybe we should open an issue to refactor the lib and only use one style.

src/wallet.js
```diff
@@ -936,7 +975,7 @@ const wallet = {
           return ret;
         }
         if (txout.spent_by === null && txout.token === selectedToken && this.isAddressMine(txout.decoded.address, data)) {
-          if (this.canUseUnspentTx(txout)) {
+          if (await this.canUseUnspentTx(txout, tx.height, currentHeight)) {
```
obiyankenobi: Why not `canUseUnspentTxSync`?

src/wallet.js
obiyankenobi: I think using a loop won't be needed as we get the information on websocket regularly, so we'll probably always call the methods with the height parameter. But the code to get this info from websocket is on the wallet PR, right?

src/wallet.js
```diff
@@ -1282,7 +1326,7 @@ const wallet = {
         }
 
         const output = utxo.output;
-        if (this.canUseUnspentTx(output)) {
+        if (await this.canUseUnspentTx(output, utxo.height, response.data.blocks)) {
```
obiyankenobi: Why not `canUseUnspentTxSync`?

src/wallet.js
pedroferreira1: Humm your idea would be to get the ws message from the full node in the lib and update internally so we wouldn't need to get this data from the API anymore?

This could work and the code would be much easier. When I opened this PR we didn't have this ws message but now that we have it might be better.

obiyankenobi: The idea was to use this info which is already available on the wallet. But "intercepting" it directly on the lib seems like a better way to handle it.

PR #38, opened 2020-01-14
feat: adding validation of max number of inputs and outputs of a tx

### LINKED ISSUE #36: Add client validation to maximum number of inputs and outputs
PR: https://github.com/HathorNetwork/hathor-wallet-lib/pull/38

pedroferreira1:
Closes #36

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -615,8 +615,8 @@ const transaction = {
   },
 
   /**
-   * Execute some verifications if this data can become a transaction
-   * Validates maximum number of inputs and outputs
+   * Validate transaction information.
+   * For now, we only verify the maximum number of inputs and outputs.
    *
    * @param {Object} data Object with inputs and outputs
    * {
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -8,7 +8,7 @@
 import { OP_GREATERTHAN_TIMESTAMP, OP_DUP, OP_HASH160, OP_EQUALVERIFY, OP_CHECKSIG, OP_PUSHDATA1 } from './opcodes';
 import { DECIMAL_PLACES, CREATE_TOKEN_TX_VERSION, DEFAULT_TX_VERSION, TOKEN_INFO_VERSION, MAX_OUTPUT_VALUE_32, MAX_OUTPUT_VALUE, TOKEN_AUTHORITY_MASK } from './constants';
 import { HDPrivateKey, crypto, encoding, util } from 'bitcore-lib';
-import { AddressError, OutputValueError, ConstantNotSet, CreateTokenTxInvalid } from './errors';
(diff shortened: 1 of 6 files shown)


src/transaction.js
```diff
@@ -600,6 +614,35 @@ const transaction = {
...
+   * @throws {MaximumNumberOutputsError} If the tx has more outputs than the maximum allowed
+   *
+   * @memberof Transaction
+   * @inner
+   */
+  verifyTxData(data) {
+    const maxNumberInputs = transaction.getMaxInputsConstant();
+    const maxNumberOutputs = transaction.getMaxOutputsConstant();
+
+    if (data.inputs.length > maxNumberInputs) {
```
msbrogli: What if `maxNumberInputs` is null? I think it should be impossible, but would it raise an exception?

src/version.js
```diff
@@ -30,6 +30,8 @@ const version = {
         // Update transaction weight constants
         transaction.updateTransactionWeightConstants(data.min_tx_weight, data.min_tx_weight_coefficient, data.min_tx_weight_k);
         tokens.updateDepositPercentage(data.token_deposit_percentage);
+        transaction.updateMaxInputsConstant(data.max_number_inputs);
```
msbrogli: Shouldn't we increase the minimum API version to ensure it will work?

src/transaction.js
```diff
@@ -600,6 +614,35 @@ const transaction = {
     }
   },
 
+  /**
+   * Execute some verifications if this data can become a transaction
+   * Validates maximum number of inputs and outputs
```
obiyankenobi: This docstring (the first sentence) doesn't help much. I would suggest:

```
Validate transaction information. For now, we only verify the maximum number of inputs and outputs.
```

src/transaction.js
pedroferreira1: Yes, it would raise an exception before at `getMaxInputsConstant `. This exception is captured in all methods that send a transaction to show a pretty message to the user

src/version.js
pedroferreira1: Yes. I usually update this `MIN_API_VERSION` after merging the full node MR's and updating the `VERSION` field there. But could do it here, what do you think?

src/transaction.js
pedroferreira1: Yeah, I agree it's better. Updated it, thanks.

src/version.js
msbrogli: I think we can do it here.

src/transaction.js
msbrogli: But where is it raised?

pedroferreira1: The methods `getMaxInputsConstant` and `getMaxOutputsConstant` throw the error in case it's null.

src/version.js
pedroferreira1: Done

src/transaction.js
msbrogli: Ok. Thanks!

PR #40, opened 2020-01-16
feat: handling block reward lock in inputs choosing and balance

### LINKED ISSUE #35: Handle block reward locks when choosing inputs of a tx and calculating balance

pedroferreira1:
Closes #35

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -986,7 +986,13 @@ const wallet = {
    * @param {number} index Output index to search
    * @param {string} selectedToken UID of the token to check existence
    *
-   * @return {Object} {success: boolean, message: Error message in case of failure, output: output object in case of success}
+   * @return {Object}
+   *
+   *  {
+   *    exists: boolean,
+   *    message: Error message in case of failure,
+   *    output: output object with 'height' key as the height of the tx of this output (in case of success only)
+   *  }
    *
    * @memberof Wallet
    * @inner
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -1696,7 +1696,7 @@ const wallet = {
    */
   handleWebsocketDashboard(data) {
     // So far we just use the height of the network and update in the variable
-    wallet.updateNetworkHeight(data.height);
+    wallet.updateNetworkHeight(data.best_block_height);
   },
 
   /**
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -12,7 +12,7 @@ import CryptoJS from 'crypto-js';
 import walletApi from './api/wallet';
 import tokens from './tokens';
 import helpers from './helpers';
-import { OutputValueError } from './errors';
+import { ConstantNotSet, OutputValueError } from './errors';
 import version from './version';
 import storage from './storage';
 import network from './network';
@@ -58,7 +58,7 @@ const wallet = {
   /*
    * Should never be accessed directly, only through get method
    */
-  _networkHeight: null,
+  _networkHeight: 0,
 
   /**
    * Verify if words passed to generate wallet are valid. In case of invalid, returns message
@@ -1024,6 +1024,8 @@ const wallet = {
       if (txout.spent_by !== null) {
         return {exists: false, message: `Output [${index}] of transaction [${txId}] is already spent`};
       }
+      // Set txout height as block height (if it's tx it will be undefined and the lib will handle it)
+      txout.height = tx.height;
       return {exists: true, 'output': txout};
     }
     // Requests txId does not exist in historyTransactions
@@ -1691,7 +1693,7 @@ const wallet = {
    */
   handleWebsocketDashboard(data) {
     // So far we just use the height of the network and update in the variable
-    this.updateNetworkHeight(data.height);
+    wallet.updateNetworkHeight(data.height);
   },
 
   /**
@@ -1704,8 +1706,8 @@ const wallet = {
    */
   updateNetworkHeight(networkHeight) {
     if (networkHeight !== this._networkHeight) {
-      WebSocketHandler.emit('height_updated', networkHeight);
       this._networkHeight = networkHeight;
+      WebSocketHandler.emit('height_updated', networkHeight);
     }
   },
 
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -1703,7 +1703,7 @@ const wallet = {
    * @inner
    */
   updateNetworkHeight(networkHeight) {
-    if networkHeight !== this._networkHeight {
+    if (networkHeight !== this._networkHeight) {
       WebSocketHandler.emit('height_updated', networkHeight);
       this._networkHeight = networkHeight;
     }
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -898,7 +898,7 @@ const wallet = {
     transaction.clearTransactionWeightConstants();
     tokens.clearDepositPercentage();
     this.clearRewardLockConstant();
-    this.clearNetowokrBestChainHeight();
+    this.clearNetworkBestChainHeight();
     storage.removeItem('wallet:started');
     storage.removeItem('wallet:backup');
     storage.removeItem('wallet:locked');
@@ -1738,7 +1738,7 @@ const wallet = {
    * @memberof Wallet
    * @inner
    */
-  clearNetowokrBestChainHeight() {
+  clearNetworkBestChainHeight() {
     this._networkBestChainHeight = 0;
   },
 }
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -50,6 +50,16 @@ import _ from 'lodash';
  * @namespace Wallet
  */
 const wallet = {
+  /*
+   * Should never be accessed directly, only through get method
+   */
+  _rewardSpendMinBlocks: null,
+
+  /*
+   * Should never be accessed directly, only through get method
+   */
+  _networkHeight: null,
+
   /**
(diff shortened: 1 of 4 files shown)


src/wallet.js
```diff
@@ -1586,6 +1601,127 @@ const wallet = {
...
+   * Return the network height
+   *
+   * @return {number} Network height
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getNetworkHeight() {
+    return this._networkHeight;
+  },
```
obiyankenobi: There's no `clearNetworkHeight`? We should do it when resetting the wallet.

src/wallet.js
```diff
@@ -1586,6 +1601,127 @@ const wallet = {
...
+  /**
+   * Return the network height
+   *
+   * @return {number} Network height
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getNetworkHeight() {
+    return this._networkHeight;
```
obiyankenobi: Should throw `ConstantNotSet` if not set.

src/wallet.js
```diff
@@ -1586,6 +1601,127 @@ const wallet = {
...
+
+  /**
+   * Update network height variable
+   *
+   * @param {number} networkHeight
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  updateNetworkHeight(networkHeight) {
```
obiyankenobi: I think we need to do an API call to get the height in `executeGenerateWallet`, otherwise the wallet won't have it set for a few moments after starting.

src/wallet.js
```diff
@@ -1586,6 +1601,127 @@ const wallet = {
...
+    this._rewardSpendMinBlocks = null;
+  },
+
+  /**
+   * Method called when websocket connection is opened
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  websocketOpened() {
```
msbrogli: I'd rename it to either `onWebsocketReady()` or `onWebsocketOpened()`.

src/wallet.js
```diff
@@ -1586,6 +1601,127 @@ const wallet = {
...
+   * Right now we just update the network height
+   *
+   * @param {Object} data Metrics object ws data with 'height' key
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  handleWebsocketDashboard(data) {
+    // So far we just use the height of the network and update in the variable
+    wallet.updateNetworkHeight(data.height);
```
msbrogli: Maybe we should check the type of the data before calling it. It seems that there's only one type of data, but others may be included later.

src/wallet.js
```diff
@@ -1586,6 +1601,127 @@ const wallet = {
...
+    this.addMetricsListener();
+  },
+
+  /**
+   * Method called when websocket connection is closed
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  websocketClosed() {
```
msbrogli: I'd rename it to `onWebsocketBeforeClose()`. If it is after close, there's no point trying to remove the listener.

src/wallet.js
```diff
@@ -50,6 +50,16 @@ import _ from 'lodash';
...
 const wallet = {
+  /*
+   * Should never be accessed directly, only through get method
+   */
+  _rewardSpendMinBlocks: null,
+
+  /*
+   * Should never be accessed directly, only through get method
+   */
+  _networkHeight: 0,
```
msbrogli: I think we need a better definition of what it means. Is it the highest height (even if the block is not in the best blockchain) or is it the height of the best blockchain's head?

src/wallet.js
pedroferreira1: It's not exactly a constant, that's why I am handling it differently. I don't want a bad ws connection to make the app crashes when calculating balance, that's why I didn't raise the exception.

I feel it's not necessary and wouldn't be bad for the wallet to calculate the balance considering networkHeight 0 instead of throwing an exception.

obiyankenobi: If we make the API call when we start the wallet, it'll never crash. It's kind of the same treatment we use for the variables we get from version API. We require them to be set.

pedroferreira1: It's easy to add this API call but I still think it's a waste. This variable is totally different from the constants. If the ws is not working the wallet won't work because it will show as 'Offline' and if it's working we will be receiving and updating the height value.

src/wallet.js
pedroferreira1: It always depends on the implementation. It's like if we lose connection to the ws, when we connect again it will update the height value and update. And the update will be propagated to the wallets that are listening to it. I don't see any reason why we should add complexity to the code with one more API call. Nothing is sync in the wallet, so it's not a problem to update the height half a second later, like we do when receiving new transactions

src/wallet.js
obiyankenobi: Do we get the height immediately after opening the ws or only when there's a new block? If it's the latter, it could take a while before getting the height. Average time is 30s, but it's not uncommon to take more than one minute

src/wallet.js
pedroferreira1: Updated the variable name to `_networkBestChainHeight` and add a comment explaining it

src/wallet.js
pedroferreira1: You mean to validate we have `data.height` and that it's a number?

src/wallet.js
pedroferreira1: Done

src/wallet.js
pedroferreira1: The metrics ws sends data every 5 seconds, so that's the maximum time we would be outdated. I say it depends on the implementation because in the wallet desktop we start the ws connection before starting the wallet (this might not be good be when we start the wallet we always have the ws data updated). In the mobile wallet we start the ws connection when loading txs, so it's basically the time to load the txs from the server is the time to receive the first ws message.

obiyankenobi: Ah ok, then I think it's fine.

src/wallet.js
```diff
@@ -1011,6 +1027,8 @@ const wallet = {
       if (txout.spent_by !== null) {
         return {exists: false, message: `Output [${index}] of transaction [${txId}] is already spent`};
       }
+      // Set txout height as block height (if it's tx it will be undefined and the lib will handle it)
+      txout.height = tx.height;
```
msbrogli: Why are settings it to the txout? Is it an internal thing?

pedroferreira1: Yes, this method is used when I choose the inputs of a transaction, so I have to check if the output exists and return it, in case of success. This output is then checked to see if it's valid to be spent (lock checks), so I need to have the tx height of it. That's why I set the height of the output as the height of it's transaction, so I can use elsewhere.

src/wallet.js
msbrogli: I misunderstood. Sorry. I though we could receive different types of packets.

src/wallet.js
msbrogli: Great! Thanks!

src/wallet.js
msbrogli: I got it. Maybe we should include in the documentation that the txout may have a height that is a copy of its block height. What do you think?

pedroferreira1: I am still not sure where we should add this information. We currently don't have any centralized place to describe the objects. I've added this to the method docstring. Do you think it's enough for now, or maybe add elsewhere also?

msbrogli: It seems good for now. Maybe we should create an issue to describe the objects. I think it is important to keep track of their properties and types.

pedroferreira1: I agree. I think one of the things we should do in the lib refactor is to create classes for each of the objects we have, like a model. So it will be described and designed in the big refactor we will do

msbrogli: Ok. I took note of this idea in the Issue description. https://github.com/HathorNetwork/hathor-wallet-lib/issues/41

PR #42, opened 2020-01-20
Small changes to support new UI for explorer using new APIs

pedroferreira1:
- Considering merged mined block version when checking if is block;
- Adding method to get data from tokens list API.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -57,13 +57,13 @@ const helpers = {
       if (tx.version === BLOCK_VERSION) {
         return 'Block';
       } else if (tx.version === MERGED_MINED_BLOCK_VERSION) {
-        return 'Merged mined block';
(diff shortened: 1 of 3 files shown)


src/helpers.js
```diff
@@ -54,7 +54,11 @@ const helpers = {
    */
   getTxType(tx) {
     if (this.isBlock(tx)) {
-      return 'Block';
+      if (tx.version === BLOCK_VERSION) {
+        return 'Block';
+      } else if (tx.version === MERGED_MINED_BLOCK_VERSION) {
+        return 'Merged mined block';
```
msbrogli: I'd say "Merged Mining Block" or "MM Block". What do you think?

pedroferreira1: Changed to "Merged Mining Block"

PR #43, opened 2020-01-21
fix: handling success false when getting token info

pedroferreira1:


__tests__/tokens_utils.js
```diff
@@ -45,7 +45,8 @@ mock.onGet('thin_wallet/token').reply((config) => {
     'melt': [],
     'name': token1.name,
     'symbol': token1.symbol,
-    'total': 100
+    'total': 100,
+    'success': true,
```
msbrogli: Shouldn't we add a test with `success: false`?

pedroferreira1: Done!

PR #44, opened 2020-01-21
fix(websocket): WebSocket was not working in Node.js

msbrogli:


src/WebSocketHandler.js
```diff
@@ -23,6 +25,8 @@ class WS extends EventEmitter {
   constructor(){
     super();
 
+    this.WebSocket = _WebSocket;
```
obiyankenobi: I think it's worth leaving a comment explaining why this is done/the problem we had.

msbrogli: Fixed. Thanks!

PR #47, opened 2020-01-24
feat: address history pagination

pedroferreira1:


src/wallet.js
```diff
@@ -224,21 +224,65 @@ const wallet = {
...
+   * Since this API is paginated, we enter a loop getting all data and return only after all requests are finished
+   *
+   * @param {Array} addresses Array of addresses (string) to get history
+   *
+   * @return {Promise} Promise that resolves when all addresses history requests finish
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  async getHistory(addresses) {
```
msbrogli: Maybe rename to `getTxHistory` or `getAddressHistoryNewApi`?

src/wallet.js
```diff
@@ -224,21 +224,65 @@ const wallet = {
 
       storage.setItem('wallet:data', dataJson);
 
-      walletApi.getAddressHistory(addresses, (response) => {
```
msbrogli: Maybe we should mark `getAddressHistory` as deprecated.

src/api/wallet.js
```diff
@@ -34,6 +34,35 @@ const walletApi = {
...
+   * Call API to get address history
+   *
+   * XXX Our current method to allow retry a request demands that we create an axios
+   * instance with a resolve callback, which will be used in case of failure and the
+   * user decides to retry. Because of that, it's impossible to use the old method (getAddressHistory)
+   * to get data with async/await, only with promises. Because of the pagination,
+   * we are in a looping getting data while not finished, so the code with async/await is
+   * much cleaner.
+   *
+   * So, right now to use async/await we must use this method and it's not possible to
```
msbrogli: But would the user be able to retry the whole operation? What happens if a problem occur during the fetch?

src/api/wallet.js
```diff
@@ -34,6 +34,35 @@ const walletApi = {
     });
   },
 
+  /**
```
msbrogli: We may have to reduce the time limit in `getAddressHistory` in this file. MR #52 may have changed the time limit to 40 seconds and we may need to reduce it to the default value.

src/api/wallet.js
pedroferreira1: The user would need to reset the wallet and start again. Our retry process depends on the promise resolve method, which does not exist on the async/await call. My opinion is that we should move forward with this and create an issue to handle it.

src/wallet.js
pedroferreira1: You mean keep the old method `getAddressHistory` and mark as deprecated in the docstring?

src/api/wallet.js
pedroferreira1: Done

src/wallet.js
pedroferreira1: Renamed to `getTxHistory`

src/wallet.js
msbrogli: Here, in the lib, we can remove the old method if we'd like to. But on `hathor-core`, we cannot change the API otherwise all wallets will stop working unless they are updated.

src/api/wallet.js
msbrogli: You mean "reset the wallet" or "restart the wallet"? Either way, I don't think we should move forward in this case. It seems an awful user experience.

src/wallet.js
```diff
@@ -263,21 +263,65 @@ const wallet = {
...
         reject(e);
       });
+
     });
     return promise;
   },
 
+  /**
+   * Asynchronous method to get history of an array of transactions
+   * Since this API is paginated, we enter a loop getting all data and return only after all requests are finished
```
msbrogli: Typo: ... after all requests have finished.

src/wallet.js
pedroferreira1: We are keeping the old API resource and removed the old lib method.

src/api/wallet.js
pedroferreira1: The user is able to close the wallet and open again to try, or click on the button "Reset Wallet" when the error modal opens.

I agree it's bad but I don't think we should do something in this PR. It would required a full rebuild in the retry request method.

We can discuss that we will only release a new wallet version once this is solved, or not. But not in this PR.  And my opinion is that right now this PR would help lot of people. We could release a new version and work on solving this asap.

msbrogli: I agree. Let's open an issue to solve this retry issue.

pedroferreira1: Done

PR #48, opened 2020-01-29
refactor: Split sendTransaction into two steps: prepareTransaction and SendPreparedTransaction

msbrogli:


src/transaction.js
```diff
@@ -643,6 +643,32 @@ const transaction = {
     }
   },
 
+  /**
+   * Generate a transaction and returns it, without propagating to the network.
+   **/
```
pedroferreira1: Still missing docstring about parameters, return and memberof

msbrogli: Fixed. Thanks!

PR #50, opened 2020-01-31
refactor: headless incorporation

pedroferreira1:
Still want to find a better place and name to use for this "new version" of the lib

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ var axios = require('./lib/api/axiosInstance');
 var storage = require('./lib/storage');
 var network = require('./lib/network');
 var MemoryStore = require('./lib/memory_store');
+var Wallet = require('./lib/new/wallet');
 
 module.exports = {
   helpers: helpers.default,
@@ -32,4 +33,5 @@ module.exports = {
   storage: storage.default,
   MemoryStore: MemoryStore.default,
   network: network.default,
+  Wallet: Wallet.default,
 }
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ var axios = require('./lib/api/axiosInstance');
 var storage = require('./lib/storage');
 var network = require('./lib/network');
 var MemoryStore = require('./lib/memory_store');
-var Wallet = require('./lib/new/wallet');
+var HathorWallet = require('./lib/new/wallet');
 
 module.exports = {
   helpers: helpers.default,
@@ -33,5 +33,5 @@ module.exports = {
   storage: storage.default,
   MemoryStore: MemoryStore.default,
   network: network.default,
-  Wallet: Wallet.default,
+  HathorWallet: HathorWallet.default,
 }
--- a/src/new/README.md
+++ b/src/new/README.md
@@ -1,0 +1,60 @@
+# New Wallet Lib
+
+### Start wallet
+
+*Parameters:*
+
+- network: Network you want the wallet connected to. E.g., 'testnet' or 'mainnet'.
+- server: Server you want to connect to. E.g., 'https://node1.mainnet.hathor.network/v1a/'.
+- seed: 24 words of this wallet in a string separated by space.
+- tokenUid: Optional parameter if you want this wallet to work with an specific token only. E.g., '0000035f21f454686b3a01b59ae1345011501d0fcd8973d25d3d6ebe6f1c00bb'.
+
+```
+const data = {
+    network: 'mainnet',
+    server: 'https://node1.mainnet.hathor.network/v1a/',
+    seed: '24 words',
+    tokenUid: '0000035f21f454686b3a01b59ae1345011501d0fcd8973d25d3d6ebe6f1c00bb',
+}
+
+const wallet = Wallet(data);
+
+console.log(wallet.state); // 0 = CLOSED
+
+wallet.start(); // This will start the wallet and load its data from the server
+```
+
+### Wallet addresses
+
+```
+wallet.getAddressToUse(); // Return address and mark as used
+
+wallet.getCurrentAddress(); // Return address
+```
+
+### Wallet balance
+
+```
+wallet.getBalance(); // Will return an object with 'available' and 'locked' in integer (no decimals).
+                     // E.g., if the avaible is 28.13, it will return {available: 2813, locked: 0}.
+
+// PS: If you want the balance of a token different from the wallet instance, you can pass as parameter the tokenUid.
+// E.g. wallet.getBalance('00') will get the balance of HTR
+```
+
+### Send tokens
+
+```
+wallet.sendTransaction(address, value); // This method allow only one output, and it will send the token of the wallet instance
+                                        // address is an string on base58 and value is an integer. E.g. if you want to send 10.23
+                                        // you need to pass 1023 to the method.
+
+// PS: If you want to send a different than the one of the wallet instance, you can pass as parameter the token object.
+// E.g. wallet.sendTransaction(address, value, {uid: '00', name: 'Hathor', symbol: 'HTR'}) will send HTR and not the instance token.
+```
+
+### Wallet events
+
+- state: every time the wallet state changes. The event parameter is the new state.
+- new-tx: every time the wallet receives a new transaction. The event parameter is the new transaction.
+- update-tx: every time the wallet receives an old transaction update. The event parameter is the updated transaction.
--- a/src/new/README.md
+++ b/src/new/README.md
@@ -24,6 +24,30 @@ console.log(wallet.state); // 0 = CLOSED
 wallet.start(); // This will start the wallet and load its data from the server
 ```
 
+The wallet has 4 possible states: CLOSED, CONNECTING, SYNCING, READY.
+
+1. CLOSED: before starting the wallet.
+2. CONNECTING: after starting the wallet and before connecting the websocket.
+3. SYNCING: after connecting the websocket and before loading data from full node.
+4. READY: after finish loading data from full node.
+
+If websocket connection is lost we automatically reconnects it.
+
+### Wallet events
+
+The wallet emits events so any change can be handled in real time.
+
+- state: every time the wallet state changes. The event parameter is the new state.
+- new-tx: every time the wallet receives a new transaction. The event parameter is the new transaction.
+- update-tx: every time the wallet receives an old transaction update. The event parameter is the updated transaction.
+
+```
+wallet.on('state', (state) => {
+  console.log(`State changed to: ${Wallet.getHumanState(state)}`);
+});
+
+```
+
 ### Wallet addresses
 
 ```
@@ -52,9 +76,3 @@ wallet.sendTransaction(address, value); // This method allow only one output, an
 // PS: If you want to send a different than the one of the wallet instance, you can pass as parameter the token object.
 // E.g. wallet.sendTransaction(address, value, {uid: '00', name: 'Hathor', symbol: 'HTR'}) will send HTR and not the instance token.
 ```
-
-### Wallet events
-
-- state: every time the wallet state changes. The event parameter is the new state.
-- new-tx: every time the wallet receives a new transaction. The event parameter is the new transaction.
-- update-tx: every time the wallet receives an old transaction update. The event parameter is the updated transaction.
--- a/src/new/README.md
+++ b/src/new/README.md
@@ -1,78 +1,115 @@
-# New Wallet Lib
+# Hathor Wallet Lib
 
-### Start wallet
+This library can be used to easily create a new Hathor Wallet. It handles all communication with the server and exposes an easy-to-use API. It is used by Hathor Wallet for iOS, Android, Linux, macOS, and Windows. It is also used by a headless Hathor Wallet.
 
-*Parameters:*
+## How to install?
 
-- network: Network you want the wallet connected to. E.g., 'testnet' or 'mainnet'.
-- server: Server you want to connect to. E.g., 'https://node1.mainnet.hathor.network/v1a/'.
-- seed: 24 words of this wallet in a string separated by space.
-- tokenUid: Optional parameter if you want this wallet to work with an specific token only. E.g., '0000035f21f454686b3a01b59ae1345011501d0fcd8973d25d3d6ebe6f1c00bb'.
+`npm install @hathor/wallet-lib`
 
-```
-const data = {
-    network: 'mainnet',
-    server: 'https://node1.mainnet.hathor.network/v1a/',
-    seed: '24 words',
-    tokenUid: '0000035f21f454686b3a01b59ae1345011501d0fcd8973d25d3d6ebe6f1c00bb',
-}
+## How to use?
 
-const wallet = Wallet(data);
+The wallet requires a backend server with both an HTTP API and a WebSocket Server. Usually, Hathor's full-node implement both services and is used as a backend server.
 
-console.log(wallet.state); // 0 = CLOSED
+The connection has four possible states:
+
+- `CLOSED`: When it is not connected to the server, and it is not trying to connect.
+- `CONNECTING`: When it is still trying to connect to the server.
+- `SYNCING`: When it has already connected but it is still downloading the transaction history.
+- `READY`: When the wallet is fully synced and ready to be used.
+
+The wallet automatically reconnects when the connection is down.
+
+### Initializing your wallet
+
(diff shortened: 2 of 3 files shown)


src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+ *
+ * You can subscribe for the following events:
+ * - state: Fired when the state of the Wallet changes.
+ * - new-tx: Fired when a new tx arrives.
+ * - update-tx: Fired when a known tx is updated. Usually, it happens when one of its outputs is spent.
+ * - more-addresses-loaded: Fired when loading the history of transactions. It is fired multiple times,
+ *                          one for each request sent to the server.
+ **/
+class Wallet extends EventEmitter {
+  constructor({ network, server, seed, tokenUid }) {
```
obiyankenobi: Missing parameters documentation.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    // tokenUid is optional so we can get the token of the wallet
+    this.token = null;
+    this.tokenUid = tokenUid;
+
+
+    this.passphrase = '';
+    // XXX Update it so we don't have fixed pin/password
+    this.pinCode = '123';
```
obiyankenobi: Maybe we can create an issue to discuss what to do about the pin and password. I'm not sure the lib should be handling them. It looks like something that should be one layer above. Whoever uses the lib decides what safety mechanism he wants to implement.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    wallet.updateHistoryData(historyTransactions, allTokens, [newTx], null, walletData);
+
+    if (isNewTx) {
+      this.emit('new-tx', newTx);
+    } else {
+      this.emit('update-tx', newTx);
+    }
+    return;
+
+    // TODO XXX Uncomment the following block to keep track of the balance.
```
obiyankenobi: Do we need to keep this commented code?

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    for (const d of data) {
+      const dataOutput = {'address': d.address, 'value': d.value, 'tokenData': tokens.getTokenIndex(allTokens, d.token.uid)};
+      const partialData = {'inputs': [], 'outputs': [dataOutput]}
+      const result = wallet.prepareSendTokensData(partialData, d.token, true, historyTxs, allTokens);
+
+      if (!result.success) {
+        console.log('Error sending tx:', result.message);
+        return;
```
obiyankenobi: Why not throw an exception?

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    this.server = server;
+    if (!this.server) {
+      // We set the default server of the lib
+      this.server = DEFAULT_SERVER;
+
+    }
+
+    // tokenUid is optional so we can get the token of the wallet
+    this.token = null;
+    this.tokenUid = tokenUid;
```
obiyankenobi: If the user does not pass one, we could have it use HTR by default. In this case, there would be no exceptions if the user later calls something like `getBalance` without a parameter.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    const walletData = wallet.getWalletData();
+    const historyTxs = 'historyTransactions' in walletData ? walletData.historyTransactions : {};
+    const ret = wallet.prepareSendTokensData(data, txToken, true, historyTxs, [txToken]);
+
+    if (!ret.success) {
+      return Promise.reject(ret.message);
+    }
+
+    return transaction.sendTransaction(ret.data, this.pinCode);
+  }
```
obiyankenobi: I think we're missing something like `updateOutputs` like we have on `sendMultiTokenTransaction`.

obiyankenobi: Actually, I think it's easier if this function just calls `sendMultiTokenTransaction`. We'd have just one code to maintain.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    storage.setItem('local:tokens:history', newTokensHistory);
+    storage.setItem('local:tokens:balance', newTokensBalance);
+    */
+  };
+
+  /**
+   * Send a transaction with multi tokens
+   * @param {Object} data Array of {'address', 'value', 'token'}
+   */
+  sendMultiTokenTransaction(data) {
```
obiyankenobi: This name seems misleading. I think the main thing is that is sends multiple outputs, not necessarily multiple tokens. I'd suggest naming this `sendTransaction`. As for the other function already named like this, I suggested refactoring it. We can:
- change name to `sendOne` or something like that;  
- remove it all together. People would use this function with an array with only one item.

index.js
```diff
@@ -32,4 +33,5 @@ module.exports = {
   storage: storage.default,
   MemoryStore: MemoryStore.default,
   network: network.default,
+  Wallet: Wallet.default,
```
msbrogli: I think there is a `wallet` here. So, we should rename it to something else. Maybe `simpleWallet`, or `HathorWallet`, or `WalletManager`. What do you think?

src/new/README.md
```diff
@@ -0,0 +1,60 @@
+# New Wallet Lib
+
+### Start wallet
+
+*Parameters:*
+
+- network: Network you want the wallet connected to. E.g., 'testnet' or 'mainnet'.
+- server: Server you want to connect to. E.g., 'https://node1.mainnet.hathor.network/v1a/'.
+- seed: 24 words of this wallet in a string separated by space.
+- tokenUid: Optional parameter if you want this wallet to work with an specific token only. E.g., '0000035f21f454686b3a01b59ae1345011501d0fcd8973d25d3d6ebe6f1c00bb'.
```
msbrogli: I think it should be a map of `tokenSymbol` to `tokenUid`. So, methods can receive the `tokenSymbol` instead of the `tokenUid`. What do you think?

src/new/README.md
```diff
@@ -0,0 +1,60 @@
...
+wallet.sendTransaction(address, value); // This method allow only one output, and it will send the token of the wallet instance
+                                        // address is an string on base58 and value is an integer. E.g. if you want to send 10.23
+                                        // you need to pass 1023 to the method.
+
+// PS: If you want to send a different than the one of the wallet instance, you can pass as parameter the token object.
+// E.g. wallet.sendTransaction(address, value, {uid: '00', name: 'Hathor', symbol: 'HTR'}) will send HTR and not the instance token.
+```
+
+### Wallet events
+
```
msbrogli: We should explain how the events work, how to subscribe and unsubscribe, and also explain the four possible states: CLOSED, CONNECTING, SYNCING, and READY.

src/new/wallet.js
msbrogli: I think HTR should be the default no matter what is passed here. Or we should rename it to `defaultToken`.

src/new/wallet.js
msbrogli: I agree. Current implementation requires a pinCode and passphrase to store sensitive data. We should refactor the internals to be able to change it here.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+        throw error;
+      })
+    } else {
+      // CONNECTING or CLOSED?
+      this.serverInfo = null;
+      this.setState(Wallet.CONNECTING);
+    }
+  }
+
+  getAddressToUse() {
```
msbrogli: I'd include a `markAsUsed` in the `getCurrentAddress` because it seems more intuitive.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+    }
+  }
+
+  reloadData() {
+    // TODO Reload data?
+    console.log('reloadData');
+  }
+
+  getBalance(tokenUid) {
+    if (!tokenUid && !this.token) {
```
msbrogli: I don't like it. Would we need to pass `"HTR_TOKEN"` on every call?

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+      throw Error('Token must be set in the class instance or passed as parameter in the method.')
+    }
+    const uid = tokenUid || this.token.uid;
+    const historyTransactions = this.getTxHistory();
+    return wallet.calculateBalance(Object.values(historyTransactions), uid);
+  }
+
+  getTxHistory() {
```
msbrogli: I think we should filter per token, similar to the `getBalance` method.

src/new/wallet.js
msbrogli: I think we will remove it soon. It updates the balance of each token.

src/new/wallet.js
msbrogli: I agree.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+  /**
+   * Send tokens to only one address.
+   *
+   * @param {String} address Address to send the tokens
+   * @param {number} value Amount of tokens to be sent
+   * @param {Object} token Token object {'uid', 'name', 'symbol'}. Optional parameter if user already set on the class
+   *
+   * @return {Promise} Promise that resolves when transaction is sent
+   **/
+  sendTransaction(address, value, token) {
```
msbrogli: We can rename it to `sendSimpleTransaction`. But  I think the `sendMultiTokenTransaction` (renamed to `sendTransaction`) is enough and it may be removed.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+  }
+}
+
+// State constants.
+Wallet.CLOSED =  0;
+Wallet.CONNECTING = 1;
+Wallet.SYNCING = 2;
+Wallet.READY = 3;
+
+Wallet.getHumanState = (state) => {
```
msbrogli: It depends on the language and the words each project would like to use.
I'd either remove it or rename it to `getHumanFriendlyState`.

src/new/wallet.js
```diff
@@ -0,0 +1,441 @@
...
+  const map = {
```
msbrogli: This can be out of the method.

src/new/wallet.js
msbrogli: It should be inside the `class` definition. Even though it seems the same to set it here, it is different. When it is inside the `class`, the method is defined in the prototype. When we put it here, it is out of the prototype.

src/new/README.md
pedroferreira1: I don't think we should do it because we could have in the network two different tokens with the same symbol, so the UID is the only unique data we have.

src/new/wallet.js
pedroferreira1: Makes sense, will do it

src/new/wallet.js
pedroferreira1: No. If you start the wallet and have `tokenUid` as '00' you don't need to pass any parameter in this method. You only need to pass parameter if you want something different from the instance.

The behaviour you said you don't like is exactly what was before.

src/new/wallet.js
pedroferreira1: I agree also, it should be in the refactor of the lib issue. But I think we should have this disabled but available as a default in the lib for apps to use it. So we also can use in both desktop and mobile wallets.

Like our memory storage, that we have a default and, if the user wants to change, they are able to do it.

pedroferreira1: Added to the lib issue

src/new/wallet.js
pedroferreira1: `updateOutputs` was used in `sendMultiTokenTransaction` so we can have multiple outputs of the same token without choosing the same input. In this method we update the `spent_tx` when the created tx arrives on the ws. It might not be the best behaviour but it's what we've been using so far in the wallet and wallet lib. We can discuss changing it on the wallet lib refactor.

src/new/README.md
pedroferreira1: Done

src/new/wallet.js
pedroferreira1: I agree, it has this problem and others already written on the code. This method deserves a refactor to be used. Created an issue to do it later. https://github.com/HathorNetwork/hathor-wallet-lib/issues/51

src/new/wallet.js
pedroferreira1: I will keep it this way until we refactor `sendMultiTokenTransaction` to a usable method.

src/new/wallet.js
pedroferreira1: Makes sense, removed it and added to wallet-headless again

pedroferreira1: Removed and added to wallet headless

src/new/wallet.js

src/new/wallet.js
pedroferreira1: I am returning a `Promise.reject` now, since this method returns promises. What do you think?

src/new/wallet.js
pedroferreira1: It's the default token uid now

index.js
pedroferreira1: I agree, it's on the description of the PR that I would like a better name for it. I like `HathorWallet`. Changed it

src/new/wallet.js
pedroferreira1: I agree. I think we should keep it there so @msbrogli can remember about it when doing the bigger refactor.

src/new/wallet.js
```diff
@@ -0,0 +1,439 @@
...
+
+    storage.setItem('local:tokens:history', newTokensHistory);
+    storage.setItem('local:tokens:balance', newTokensBalance);
+    */
+  };
+
+  /**
+   * Send a transaction with multi tokens
+   * @param {Object} data Array of {'address', 'value', 'token'}
+   */
```
obiyankenobi: Missing return type/value.

src/new/README.md
msbrogli: As it is a wallet, we can go both ways because only registered tokens would be allowed, and we don't allow to have two registered tokens with the same symbol.

But I agree that it is safer to give the `tokenUid`.

src/new/wallet.js
msbrogli: I recently found an easy way to do it using es6: `getAddressToUse({ markAsUsed = false } = {}) { ... }`.

This would require a named argument and would automatically handle the default parameter.

src/new/wallet.js
msbrogli: As we will discuss the best interface later, I think we can move forward.

src/new/wallet.js
msbrogli: We can do it after we discuss the best interface for this class.

src/new/wallet.js
```diff
@@ -0,0 +1,439 @@
...
+    }
+  }
+
+  reloadData() {
+    // TODO Reload data?
+    console.log('reloadData');
+  }
+
+  getBalance(tokenUid) {
+    const uid = tokenUid || this.token.uid;
```
obiyankenobi: Should use `this.tokenUid`.

src/new/wallet.js
```diff
@@ -0,0 +1,439 @@
...
+
+    return balance;
+  }
+
+  getTokenData() {
+    if (this.tokenUid === HATHOR_TOKEN_CONFIG.uid) {
+      // Hathor token we don't get from the full node
+      this.token = HATHOR_TOKEN_CONFIG;
+    } else {
+      // Get token info from full node
```
obiyankenobi: I'm not sure we need to do this. We don't seem to use the token name and symbol in the lib. If the user starts the wallet with a UID, but he probably already has this information. I think we should only have `this.tokenUid`.

src/new/wallet.js
pedroferreira1: Added.

src/new/wallet.js
pedroferreira1: It's the same value.

src/new/wallet.js
pedroferreira1: We do use the token name and symbol in the `prepareSendTokensData` method. We use it to return a pretty error message.

If we decide not to do this, we would require the user to pass the full token configuration when calling `sendTransaction`. If we refactor the `prepareSendTokensData` method, we could remove it but this would require another big refactor. I don't see why do this right now.

src/new/wallet.js
obiyankenobi: Ah, I thought we would not use `this.token` as per my other comment.

src/new/wallet.js
obiyankenobi: Ok, let's refactor it later.

PR #52, opened 2020-02-05
fix: loading wallet could crash

obiyankenobi:


src/wallet.js
```diff
@@ -1571,7 +1573,10 @@ const wallet = {
    */
   txExists(txData) {
     const data = this.getWalletData();
-    return txData.tx_id in data['historyTransactions'];
+    if (data && data['historyTransactions']) {
```
msbrogli: When can `data` or `data.historyTransactions` be undefined? Maybe we should add a test that fails without this fix.

obiyankenobi: It happens before we load the wallet.

Added the test.

src/api/wallet.js
```diff
@@ -27,7 +27,7 @@ const walletApi = {
    */
   getAddressHistory(addresses, resolve) {
     const data = {addresses};
-    return createRequestInstance(resolve).get('thin_wallet/address_history', {'params': data}).then((res) => {
+    return createRequestInstance(resolve, 40000).get('thin_wallet/address_history', {'params': data}).then((res) => {
```
msbrogli: I don't think we should increase the time limit.

obiyankenobi: Unless we get the pagination PR (#47) as well, we need to increase it.

msbrogli: Ok. We can change it, but we need to undo it after the PR #47 is merged. So, I left a comment in PR #47 to reduce this time limit.

PR #55, opened 2020-02-12
fix: move xpubkey to wallet:accessData on storage

obiyankenobi:
We actually need xpubkey to always be persisted. As with the latest change in the wallet, `wallet:data` is kept in memory, so it would not persist across restarts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hathor/wallet-lib",
-  "version": "0.9.1",
+  "version": "0.9.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hathor/wallet-lib",
-  "version": "0.9.1",
+  "version": "0.9.2",
```
msbrogli: It seems to add new features, so I think we should bump to 0.10.0.

PR #57, opened 2020-02-26
feat: split send transaction methods into prepareData and sendTx

pedroferreira1:


src/new/wallet.js
```diff
@@ -290,6 +290,25 @@ class HathorWallet extends EventEmitter {
    * @return {Promise} Promise that resolves when transaction is sent
    **/
   sendTransaction(address, value, token) {
+    const ret = this.prepareTransaction(address, value, token);
+
+    if (ret.success) {
+      return transaction.sendTransaction(ret.data, this.pinCode);
+    } else {
+      return Promise.reject(ret.message);
+    }
```
obiyankenobi: As tx has already been prepared, can't we just `return this. sendPreparedTransaction(ret.data)`?

pedroferreira1: You mean change this whole `if ... else` block for this? If it's that, we can't.

We could change

`return transaction.sendTransaction(ret.data, this.pinCode);`
to
`return this. sendPreparedTransaction(ret.data);`

I didn't feel it would make much difference. But I can change it, if you prefer.

obiyankenobi: Ah, I thought we would be able to replace the whole `if/else`, didn't notice we were checking the result of `this.prepareTransaction`.

Still, I think using `return this.sendPreparedTransaction(ret.data);` is better. Otherwise we'll be calling `transaction.prepareData` twice, once in `this.prepareTransaction` and also in `transaction.sendTransaction`.

pedroferreira1: Done!

PR #60, opened 2020-03-04
fix: when stopping wallet we should use endConnection

pedroferreira1:


pedroferreira1:
No need for this PR anymore. This change is already done in a bigger PR (https://github.com/HathorNetwork/hathor-wallet-lib/pull/61)

Issue #30, opened 2019-10-28
[Design]: refactor: Make it easier to use


obiyankenobi:
Although the lib is functional, we've debated several improvements that can be done. It's still not easy for developers to start a new project, quickly add it and start interacting with Hathor's network. Right now, developers need some knowledge of the lib's inner workings to be able o use it.

We need a lib with a clear interface and classes for the different elements in Hathor-land: transactions, blocks, inputs, outputs, etc.

@msbrogli took a step on that direction when implementing the twitter bot (check wallet.js): https://github.com/HathorNetwork/twitter-hathor-bot

pedroferreira1:
@obiyankenobi can you list here the first ideas you had when thinking about this refactor?

obiyankenobi:
I wrote them in this document: https://docs.google.com/document/d/1tD7D07PvDHAbv2N0UqmDmSOO7WxE7dIAD2gyijIgjqw/edit

One of the problems has already been fixed. The other 2 are:
- hide lib internal state: nowadays, the dev still needs some knowledge about the lib's state and update it depending on the event it receives. 
- hide websocket class: we still do stuff like this `hathorLib.WebSocketHandler.on('reload_data', this.fetchDataFromServer);`

Both points are kind of addressed on the facade @msbrogli did, especially the second.

Regarding the first, I believe the lib should handle all state "under the hood", without interaction from the user.

obiyankenobi:
One other thing is removing some data from storage that could be kept in memory, like `wallet:closed` flag.

obiyankenobi:
Copying from #44 that was a duplicate issue:
"Create classes for each object type, so we would have a full description of their attributes and types."

pedroferreira1:
We should also consider removing PIN/Password handling from the lib (maybe just let wallets use what we have now as default but allow it not to be required)

Comment: https://github.com/HathorNetwork/hathor-wallet-lib/pull/50#discussion_r373858260

pedroferreira1:
> One other thing is removing some data from storage that could be kept in memory, like `wallet:closed` flag.

Some data could be kept in memory but we should take care when removing them from storage. The one from your example `wallet:closed` must be kept on storage because we use it after opening the app. We could do it with IPC_RENDERER messages to send data from electron main process to the renderer process but it was easier and more readable to do like this.

We just need to do this refactor carefully.

msbrogli:
# Design for this issue

## Problem

[...]

PR #61, opened 2020-03-11
feat: Add support for multiple wallets using a single connection and separate storages

msbrogli:
It partially implements #30.

We focused on the connection and storage part because they are necessary to allow the creation of multiple independent wallets at the same time in `hathor-wallet-headless`.


src/new/connection.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    networkInstance.setNetwork(network);
+    this.network = network;
+
+    this.state = Connection.CLOSED;
+    this.serverInfo = null;
+
+    this.onConnectionChange = this.onConnectionChange.bind(this);
+    this.handleWebsocketMsg = this.handleWebsocketMsg.bind(this);
```
obiyankenobi: Where's this method?

src/new/connection.js
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * This is a Connection that may be shared by one or more wallets.
+ *
+ * It has the following states:
+ * - CLOSED: When it is disconnected from the server.
+ * - CONNECTING: When it is connecting to the server.
+ * - CONNECTED: When it is connected.
+ *
+ * You can subscribe for the following events:
+ * - update-state: Fired when the state of the Wallet changes.
```
obiyankenobi: Currently you emit `state`.

src/new/connection.js
```diff
@@ -0,0 +1,130 @@
...
+ * - websocket-msg: Fired when a new message arrive from the websocket.
```
obiyankenobi: Not fired anywhere.

src/new/connection.js
pedroferreira1: This is an old code, it's already fixed.

src/new/connection.js
pedroferreira1: Updated.

src/new/connection.js
pedroferreira1: This was an old event, updated it

index.js
```diff
@@ -23,7 +23,6 @@ module.exports = {
   transaction: transaction.default,
   version: version.default,
   wallet: wallet.default,
-  WebSocketHandler: WebSocketHandler.default,
```
msbrogli: Shouldn't we leave it here for compatibility?

src/helpers.js
```diff
@@ -178,8 +178,8 @@ const helpers = {
    * @inner
    */
   getServerURL() {
-    const server = storage.getItem('wallet:server');
-    const defaultServer = storage.getItem('wallet:defaultServer');
+    const server = StorageProxy.getStorage().getItem('wallet:server');
```
msbrogli: I think it is best to create a `const wallet = StorageProxy.getStorage()` to avoid duplicate code.

src/new/wallet.js
```diff
@@ -89,28 +82,33 @@ class HathorWallet extends EventEmitter {
...
+   * @param {Number} newState Enum of new state after change
    **/
-  onConnectionChange(value) {
-    if (value) {
+  onConnectionChangedState(newState) {
+    if (newState === Connection.CONNECTED) {
+      StorageProxy.setStorage(this.storage);
       this.setState(HathorWallet.SYNCING);
-      wallet.loadAddressHistory(0, GAP_LIMIT).then(() => {
+      wallet.loadAddressHistory(0, GAP_LIMIT, this.conn).then(() => {
```
msbrogli: Even though our strategy should work most of the time, there may be race conditions where it fails. As `loadAddressHistory` is async, multiple wallets may mix the storages while it is loading.

Our strategy seems to work perfectly for sync methods, but async methods may randomly fail (sometimes without throwing an exception).

src/storage.js
```diff
@@ -74,6 +74,4 @@ class Storage {
   }
 }
 
-const instance = new Storage();
-
-export default instance;
+export default Storage;
```
msbrogli: This change breaks compatibility. Is it safe to do it?

src/websocket.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  /**
+   * Set in redux if websocket is online
+   *
+   * @param {*} value Can be true|false|undefined
+   */
+  setIsOnline(value) {
+    // Save in redux
+    // Need also to keep the value in 'this' because I was accessing redux store
```
msbrogli: Is it an outdated comment? I think this library shouldn't depend on redux (even though we can have a `redux.js` to make redux integration easier).

src/wallet.js
```diff
@@ -869,17 +870,21 @@ const wallet = {
...
   /**
    * Subscribe to receive updates from an address in the websocket
    *
    * @param {string} address
    */
-  subscribeAddress(address) {
+  subscribeAddress(address, connection = null) {
     const msg = JSON.stringify({'type': 'subscribe_address', 'address': address});
-    WebSocketHandler.sendMessage(msg);
+    WebSocketHandler.ws.sendMessage(msg);
```
msbrogli: It seems we introduced an incompatibility because of the `.ws.` part. I think we shouldn't do it otherwise we will have to refactor our wallets as well.

src/wallet.js
```diff
@@ -869,17 +870,21 @@ const wallet = {
...
+
+    if (connection) {
+      connection.websocket.sendMessage(msg);
```
msbrogli: Why are we subscribing to two connections?

src/wallet.js
```diff
@@ -973,13 +978,13 @@ const wallet = {
...
    */
   cleanLoadedData() {
-    storage.removeItem('wallet:accessData');
-    storage.removeItem('wallet:data');
-    storage.removeItem('wallet:address');
-    storage.removeItem('wallet:lastSharedIndex');
-    storage.removeItem('wallet:lastGeneratedIndex');
-    storage.removeItem('wallet:lastUsedIndex');
-    storage.removeItem('wallet:lastUsedAddress');
+    StorageProxy.getStorage().removeItem('wallet:accessData');
```
msbrogli: I think we should create a `const storage = StorageProxy.getStorage()` to avoid duplicate code.

src/wallet.js
```diff
@@ -1000,12 +1005,12 @@ const wallet = {
     tokens.clearDepositPercentage();
     this.clearRewardLockConstant();
     this.clearNetworkBestChainHeight();
-    storage.removeItem('wallet:started');
-    storage.removeItem('wallet:backup');
-    storage.removeItem('wallet:locked');
-    storage.removeItem('wallet:tokens');
-    storage.removeItem('wallet:sentry');
-    storage.removeItem('wallet:type');
+    StorageProxy.getStorage().removeItem('wallet:started');
```
msbrogli: I think we should create a `const storage = StorageProxy.getStorage()` to avoid duplicate code.

src/tokens.js
```diff
@@ -9,7 +9,7 @@ import buffer from 'buffer';
 import { crypto, util } from 'bitcore-lib';
 import transaction from './transaction';
 import wallet from './wallet';
-import storage from './storage';
+import StorageProxy from './storage_proxy';
```
msbrogli: I think the name `StorageProxy` is misleading because it seems we are dealing with a class. I think we should rename it to `storageProxy` in all imports because we are dealing with an instance.

src/helpers.js
```diff
@@ -7,7 +7,7 @@
 
 import path from 'path';
 
-import storage from './storage';
+import StorageProxy from './storage_proxy';
```
msbrogli: `storageProxy`

src/transaction.js
```diff
@@ -11,7 +11,7 @@ import { HDPrivateKey, crypto, encoding, util } from 'bitcore-lib';
 import { AddressError, OutputValueError, ConstantNotSet, CreateTokenTxInvalid, MaximumNumberInputsError, MaximumNumberOutputsError } from './errors';
 import dateFormatter from './date';
 import helpers from './helpers';
-import storage from './storage';
+import StorageProxy from './storage_proxy';
```
msbrogli: `storageProxy`

src/wallet.js
```diff
@@ -14,12 +14,12 @@ import tokens from './tokens';
 import helpers from './helpers';
 import { ConstantNotSet, OutputValueError, WalletTypeError } from './errors';
 import version from './version';
-import storage from './storage';
+import StorageProxy from './storage_proxy';
```
msbrogli: `storageProxy`

src/new/wallet.js
```diff
@@ -6,16 +6,16 @@
...
 import transaction from '../transaction';
 import tokens from '../tokens';
 import version from '../version';
-import ws from '../WebSocketHandler';
-import storage from '../storage';
-import MemoryStore from '../memory_store';
 import walletApi from '../api/wallet';
+import Storage from '../storage';
+import MemoryStore from '../memory_store';
+import StorageProxy from '../storage_proxy';
```
msbrogli: `storageProxy`

src/storage.js
pedroferreira1: This requires some changes in both wallets. Since right now we can't have a unique storage, I didn't find a way to keep compatibility, I had to return a class, instead of an instance.

In the wallets we call `hathorLib.storage.setStore(new LocalStorageStore());`, which used to make sense, because `hathorLib.storage` used to return a storage instance but right now it will return a class.

We also have direct storage calls, like `hathorLib.storage.setItem('wallet:version', VERSION);`, this would also require a refactor to support these changes.

Have you thought about a way I could keep compatibility?

src/new/wallet.js
pedroferreira1: Yes, that makes sense. I could receive a storage as parameter in async calls, (like I did with `this.conn`, so I know all the sequence of calls will be executed in the same storage).

What do you think?

src/wallet.js
pedroferreira1: The old wallet code still uses the `WebSocketHandler`, which has a websocket connection. Our new code has a connection instance, which has a different websocket connection. To receive the message in both connections I should subscribe both.

index.js
pedroferreira1: Yes, makes sense. Done

src/tokens.js
pedroferreira1: Done

src/helpers.js
pedroferreira1: Done

src/wallet.js
pedroferreira1: Done

src/new/wallet.js

src/transaction.js
pedroferreira1: Done

src/wallet.js
pedroferreira1: Done

src/wallet.js
pedroferreira1: Done

src/websocket.js
pedroferreira1: Yes, it's an old code. We used to have this code in the wallet (where we use redux). Removed it

src/helpers.js
pedroferreira1: Done

src/wallet.js
pedroferreira1: Yes, the only way to keep the compatibility is if we keep the `WebSocketHandler` as it is. But since we are already breaking compatibility with Storage in the wallets (unless we think about a different design), I think I can refactor the wallet desktop and mobile with these changes. Or do you thinks it's risky?

src/storage_proxy.js
```diff
@@ -0,0 +1,46 @@
...
+ * Any storage access will be done by StorageProxy.getStorage().method
+ * and this allow many storage instances to exist.
+ *
+ * Anytime you want to change the current storage instance to be used,
+ * just need to call StorageProxy.setStorage(storage)
+ *
+ * @class
+ * @name StorageProxy
+ */
+class StorageProxy {
```
obiyankenobi: I don't understand why we had to create this. Couldn't we use `storage.js` for this? We'd use `storage.setStore()` instead of `storageProxy.setStorage`.

src/websocket.js
```diff
@@ -0,0 +1,256 @@
...
+import _WebSocket from 'isomorphic-ws';
+
+const WS_READYSTATE_READY = 1;
+
+
+/**
+ * Handles websocket connections and message transmission
+ *
+ * @class
+ * @name WebSocketHandler
```
obiyankenobi: WebSocketHandler?

src/WebSocketHandler.js
```diff
@@ -5,246 +5,35 @@
...
-    this.started = false;
-    this.connected = false;
-    this.setIsOnline(false);
-    wallet.onWebsocketBeforeClose();
-    if (this.ws) {
-      this.ws.close();
-      this.ws = null;
-    }
+import helpers from './helpers';
+import WS from './websocket';
```
obiyankenobi: What is the goal of having `WebSocketHandler.js` and `websocket.js`? Is it for keeping the interface?

If that's the case, this should still be an event emitter and we should emit all events. For instance, this is a snippet from our wallet code:
```js
hathorLib.WebSocketHandler.on('height_updated', this.handleHeightUpdated);
```

This won't work after this change. I'm not saying that we should keep compatibility. Actually, we could discuss it. Maybe it's easier to break it already, as I don't believe other people are using the lib.

src/new/wallet.js
msbrogli: I think it may solve the problem, but we may need a better solution for the future. Is it the only place with async calls?

src/storage.js
msbrogli: I think we can refactor our wallets as long as it is a simple task. If the refactor is not really simple, maybe we should keep exporting an instance instead of the class.

src/wallet.js
msbrogli: I don't think it is risky, but I think it should be simple. As we will refactor the wallets, why don't we just remove the `WebSockerHandler.js` and directly use the new connection class?

src/wallet.js
msbrogli: I got it. But it won't be needed if we remove `WebSocketHandler`. I think you should either drop a comment about it or remove the `WebSocketHandler` from the lib and update the wallets.

src/new/wallet.js
```diff
@@ -342,21 +342,18 @@ class HathorWallet extends EventEmitter {
...
-    const store = new MemoryStore();
-    storage.setStore(store);
-    storage.setItem('wallet:server', this.server);
-
-    ws.on('is_online', this.onConnectionChange);
-    ws.on('reload_data', this.reloadData);
-    ws.on('addresses_loaded', this.onAddressesLoaded);
-    ws.on('wallet', this.handleWebsocketMsg);
+    storageProxy.setStorage(this.storage);
+    this.conn.start();
```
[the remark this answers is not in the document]
msbrogli: > In light of the next comment I just made, I think it's better to always call `conn.start()` so we can handle the result. If it's already started, it should resolve the promise immediately.

I agree.

src/new/wallet.js
```diff
@@ -388,13 +384,13 @@ class HathorWallet extends EventEmitter {
...
   stop() {
+    storageProxy.setStorage(this.storage);
     // TODO Double check that we are properly cleaning things up.
     // See: https://github.com/HathorNetwork/hathor-wallet-headless/pull/1#discussion_r369859701
-    ws.stop()
-    ws.removeListener('is_online', this.onConnectionChange);
-    ws.removeListener('reload_data', this.reloadData);
-    ws.removeListener('addresses_loaded', this.onAddressesLoaded);
-    ws.removeListener('wallet', this.handleWebsocketMsg);
+    this.conn.stop()
```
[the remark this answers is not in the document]
msbrogli: I agree that the best solution is to have a counter in the connection. Each call to `.start()` increases the counter, and each call to `.stop()` decreases it. The connection is only dropped when the counter is zero.

src/storage_proxy.js
msbrogli: The problem is that the `storage` is used all across the code. Before a major refactor to stop doing it, we need to be able to change the storage without while we keep the lib compatible to legacy code in our wallets.

obiyankenobi: I still don't understand the problem.

pedroferreira1: We have many storage instances, we need a place to store what storage should be used in the moment.

For example:

`w1 = HathorWallet()` -> w1 has a connection (c1) and a storage (s1)
`w2 = HathorWallet()` -> w2 has a connection (c2) and a storage (s2)

When c1 receives a new history in the ws, it should update s1 and not s2, so it updates the current storage in the `StorageProxy`, so all methods we already have that were not refactored will use the correct storage.

src/wallet.js
pedroferreira1: The `WebSocketHandler` wasn't removed because it's not used only in the wallets but in our old code of the lib as well. We agreed not to refactor that code right now.

src/storage.js
pedroferreira1: Talking with @obiyankenobi about the `StorageProxy` I removed it and kept only our old storage and store classes. With this I don't need to refactor anything because it's still the same code for the storage.

Can you see if it's good?

src/WebSocketHandler.js
```diff
@@ -5,246 +5,35 @@
...
+
+class WebSocketHandler {
+  constructor() {
+    this.ws = null;
+    /* Right after importing the modules helpers and wallet, they are
+     * not available and are still undefined.
+     * This is probably caused by a cyclic import (wallet -> helpers -> tokens -> wallet)
+     * but still need more study.
+     * For now the setTimeout is used, so we can use the helpers module properly.
+     */
```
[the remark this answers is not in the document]
pedroferreira1: With the current solution for `WebSocketHandler` and `Storage` we won't need to refactor any wallet that uses the lib. I think we should create an issue to handle this after, so it's a small refactor, rather than just mix everything here. What do you think?

src/new/wallet.js
```diff
@@ -89,28 +82,33 @@ class HathorWallet extends EventEmitter {
...
-   *
-   * @param {Object} result {historyTransactions, allTokens, newSharedAddress, newSharedIndex, addressesFound}
-   **/
-  onAddressesLoaded(result) {
-    this.emit('more-addresses-loaded', result);
+    this.storage = null;
+    if (storage) {
+      this.storage = storage;
+    } else {
+      // Settings storage data
```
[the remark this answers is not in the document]
pedroferreira1: It was for an old code, removed it.

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: When I was doing this PR my thoughts was that we connection was not going to be shared in this first version. We need to think about bigger problems (not just start/stop) if we decide to allow shared connections in this PR.

For instance, one of the problems not solved in this PR is that two HathorWallet instances would receive the event for one new tx, which instance should update the storage?

That's why I've decided not to handle this case now.

src/storage_proxy.js
pedroferreira1: @msbrogli I've removed the StorageProxy and kept only the storage as a simpler solution. Can you see what you think?

src/websocket.js
pedroferreira1: Updated, thanks

src/WebSocketHandler.js
pedroferreira1: Done that, can you see what you think?

src/new/wallet.js
```diff
@@ -367,13 +364,12 @@ class HathorWallet extends EventEmitter {
     const promise = new Promise((resolve, reject) => {
       version.checkApiVersion().then((info) => {
```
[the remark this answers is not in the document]
pedroferreira1: I agree we should handle only in the wallet. Updated it.

src/new/wallet.js
pedroferreira1: In the new wallet code yes. 

I've added a piece of code to handle it, can you see if it's good?

My first solution was to add an optional parameter for `setItem` and `getItem` in the storage, however the chain of method calls in `loadAddressHistory` is so big that I had to change many methods. I ended up using the solution in the commit because it's less complex to understand and to refactor, if needed.

src/new/wallet.js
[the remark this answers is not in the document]
obiyankenobi: All right, I agree we postpone connection sharing for now.

src/WebSocketHandler.js
obiyankenobi: Looks good.

src/WebSocketHandler.js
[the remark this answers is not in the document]
obiyankenobi: We don't need to update the wallets right after this change. 

Let's say we move these functions as I suggested. Next time we have to update a wallet and use this new version of the lib, we'll get an error as there won't be `getDepositAmount` and `getWithdrawAmount` in `helpers.js`. We'll only need to change where we import them from. Seems like a simple really change to me and we don't need this hack here.

src/wallet.js
obiyankenobi: We could have on `WebSocketHandler`, so no refactor on `wallet.js` is needed for this.
```js
sendMessage = msg => {
    return this.ws.sendMessage(msg);
}
```

src/WebSocketHandler.js
```diff
@@ -5,246 +5,60 @@
...
-   * Method called to send a message to the server
-   *
-   * @param {string} msg Message to be sent to the server (usually JSON stringified)
-   */
-  sendMessage(msg) {
-    if (!this.started) {
-      this.setIsOnline(false);
-      return;
-    }
+      this.ws = new WS({ wsURL: helpers.getWSServerURL });
```
obiyankenobi: Missing `()` on function call.

pedroferreira1: The parameter `wsURL` can be either a function or a string.

src/wallet.js
pedroferreira1: I don't mind changing this on the lib, as long as the wallet code does not change. If I create this `sendMessage`, I would also need to override the `emit` method from `WebSocketHandler` because we have some calls like:

`WebSocketHandler.ws.emit('addresses_loaded', result);`

That's why I prefer keeping the refactor as it is.

src/WebSocketHandler.js
msbrogli: To keep the current behaviour, it must be a function. So, if the server is changed in the storage, it will try to connect to the new server on the next call.

obiyankenobi: Ah, got it. Thanks.

src/new/wallet.js
```diff
@@ -367,13 +361,13 @@ class HathorWallet extends EventEmitter {
...
```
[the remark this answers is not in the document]
obiyankenobi: I believe this refactor will not take into account sharing the connection. If the user does that, there might be errors.

src/websocket.js
```diff
@@ -0,0 +1,256 @@
...
+   *
+   * @param {Object} evt Event that has data (evt.data) sent in the websocket
+   */
+  onMessage(evt) {
+    const message = JSON.parse(evt.data)
+    const _type = message.type.split(':')[0]
+    if (_type === 'pong') {
+      this.onPong();
+    }
+    this.emit(_type, message)
```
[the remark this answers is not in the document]
pedroferreira1: Maybe it's useless but I don't see why we should stop emitting it. Just to 'save' a message every 3 seconds?

src/WebSocketHandler.js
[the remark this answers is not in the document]
pedroferreira1: I've removed the cyclic import and also the `setTimeout` hack. When I test this version in the wallets I just need to change where we get `getDepositAmount` and `getWithdrawAmount`

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: Yes, that's the idea. To double check this, I've added a validation of the connection in the constructor, so we never get into this situation. Can you see if it's good?

src/storage.js
msbrogli: Seems good. Thanks!

src/wallet.js
msbrogli: I agree. As lon as the wallet code does not change, it is good to go.

src/new/wallet.js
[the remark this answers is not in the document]
msbrogli: I agree we can postpone because we have to test all these cases. But, specifically about your point, I think all wallets will receive all transactions and they will filter them. I'm not sure if this criteria is enough, but a starting filter criteria would be to have at least one input or output that contains an address of that wallet.

src/storage_proxy.js
msbrogli: Seems good. Thanks!

src/websocket.js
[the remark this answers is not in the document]
msbrogli: I agree. Maybe the wallet can use it in the future.

src/new/wallet.js
```diff
@@ -89,28 +85,32 @@ class HathorWallet extends EventEmitter {
...
-    this.emit('more-addresses-loaded', result);
+    this.store = null;
+    if (store) {
+      this.store = store;
+    } else {
+      // Creating default store
+      this.store = new MemoryStore();
+    }
+    storage.setStore(this.store);
+    storage.setItem('wallet:server', this.conn.currentServer);
```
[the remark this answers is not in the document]
msbrogli: I think it is because the store is empty at this point.

src/websocket.js
```diff
@@ -0,0 +1,256 @@
...
+      // TODO Throw error?
+      return;
+    }
+
+    if (this.ws) {
+      const dt = (new Date() - this.latestSetupDate) / 1000;
+      if (dt < this.openConnectionTimeout) {
+        return;
+      }
+      this.ws.close();
```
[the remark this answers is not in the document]
msbrogli: Exactly. This method is called every N seconds, no matter the state of the connection.

src/wallet.js
```diff
@@ -159,10 +159,11 @@ const wallet = {
     this.setWalletAccessData(accessData);
     this.setWalletData(walletData);
 
+    WebSocketHandler.setup();
```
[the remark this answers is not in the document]
msbrogli: I agree that we shouldn't change it back. One can run it without `loadHistory`, which would not require a connection.

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: I agree about the wallets, however nowadays we already update some storage data from `WebSocketHandler`, which would receive events from more than one wallet. We still need to validate that all event handled by this class can share the storage.

src/wallet.js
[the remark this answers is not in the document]
pedroferreira1: Makes sense, but in our case we are running without `loadHistory` but we are calling `loadHistory` manually later. I will add a manual call to `WebSocketHandler.setup();`.

src/new/connection.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  /**
+   * Called when a new wallet message arrives from websocket.
+   *
+   * @param {Object} wsData Websocket message data
+   **/
+  handleWalletMessage(wsData) {
+    this.emit('wallet-update', wsData);
```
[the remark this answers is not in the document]
pedroferreira1: I will open a different issue to do it. I think this PR is already too big.
https://github.com/HathorNetwork/hathor-wallet-lib/issues/66

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: There is no reason to do it in the constructor but makes no difference, actually. I moved to the `start` method.

src/websocket.js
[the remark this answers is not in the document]
pedroferreira1: Added the comment

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: We've agreed to postpone the share connection problems for now.

src/wallet.js
pedroferreira1: I don't think we should think about removing the `WebSocketHandler` for now. It's used in the whole library and wallets. It will be removed soon in the future. For now I left a comment in the code, as you suggested.

src/websocket.js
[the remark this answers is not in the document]
pedroferreira1: Yes, it will. When we create `this.ws` instance, we also set the callback for `this.ws.onClose`

```
this.ws.onclose = () => {
    this.onClose();
}
```

src/wallet.js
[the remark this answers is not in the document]
pedroferreira1: I moved back to the place where it was before and now I am manually calling `WebSocketHandler.setup` in the `HathorWallet` before calling `loadAddressHistory`

src/new/wallet.js
```diff
@@ -89,28 +86,32 @@ class HathorWallet extends EventEmitter {
...
+   * @param {Number} newState Enum of new state after change
    **/
-  onConnectionChange(value) {
-    if (value) {
+  onConnectionChangedState(newState) {
+    if (newState === Connection.CONNECTED) {
+      storage.setStore(this.store);
       this.setState(HathorWallet.SYNCING);
-      wallet.loadAddressHistory(0, GAP_LIMIT).then(() => {
+      WebSocketHandler.setup();
```
obiyankenobi: I thought that if `newState === Connection.CONNECTED` the websocket was already setup. Wouldn't it be a problem to call this?

src/websocket.js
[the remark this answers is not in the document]
obiyankenobi: Our `onClose` will try to restart the connection then:
```
setTimeout(() => {
      this.setup()
    }, this.retryConnectionInterval);
```

Don't we need to do `this.ws.onclose = () => {};`?

src/new/wallet.js
pedroferreira1: It's a different websocket connection. When the refactor is done we will completely remove `WebSocketHandler` but for now it's needed to support the old library code.

In our normal sequence used nowadays in the wallet we call
`generateWallet` -> `WebSocketHandler.setup` -> `loadAddressHistory`

It's exactly what I did, except that I had to manually call the `WebSocketHandler.setup` because we don't automatically load the history. So it's not a problem, we are doing exactly what we do right now.

src/websocket.js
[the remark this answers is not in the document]
pedroferreira1: It really makes no difference, the method `this.setup` is prepared to handle the cases where it's called multiple times. I will add this because maybe it's more correct semantically but it's not exactly a fix.

src/new/wallet.js
msbrogli: I think it works because the javascript main loop always wait for methods to finish. So, each method will run using the correct store.

src/new/wallet.js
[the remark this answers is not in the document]
msbrogli: I think it is good. We need to make it clear in the docs that the connection cannot be shared yet.

msbrogli:
Approved.

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: I've updated the README to add this information

PR #70, opened 2020-04-14
fix: websocket setup might be called multiple times

pedroferreira1:


src/WebSocketHandler.js
```diff
@@ -42,7 +51,9 @@ class WebSocketHandler extends EventEmitter {
 
   endConnection() {
     // To keep compatibility with methods previously used in this singleton
-    return this.ws.endConnection();
+    if (this.ws !== null) {
+      return this.ws.endConnection();
```
msbrogli, line 55: Should we set `this.ws = null`?

src/WebSocketHandler.js
```diff
@@ -12,28 +12,37 @@ import WS from './websocket';
...
   setup() {
-    this.ws = new WS({ wsURL: helpers.getWSServerURL });
-
-    this.on('is_online', (value) => {
-      if (value) {
-        wallet.onWebsocketOpened();
-        this.ws.emit('reload_data');
-      } else {
-        wallet.onWebsocketBeforeClose();
+    if (this.ws === null) {
```
msbrogli, line 22: What if it is not null? What do we do?

src/WebSocketHandler.js
```diff
@@ -12,28 +12,37 @@ import WS from './websocket';
...
-     * In the wallets we capture some events from it, so
-     * this following code is to emit all events emitted from this.ws
-     */
-    this.oldEmit = this.ws.emit;
-    this.ws.emit = (type, ...args) => {
-      this.emit(type, ...args);
-      return this.oldEmit(type, ...args);
     }
 
     // To keep compatibility with methods previously used in this singleton
```
obiyankenobi, line 48: Do we need to execute this even if `this.ws` was not null? I imagine the setup would be done already.

pedroferreira1: It's important to do the `setup` again. Look at the `endConnection` code.

```
  endConnection() {
    this.setIsOnline(undefined);
    this.started = false;
    this.connected = undefined;
    if (this.ws) {
      this.ws.onclose = () => {};
      this.ws.close();
      this.ws = null;
    }
    clearInterval(this.heartbeat);
  }
```

If you do an `endConnection` in the `src/websocket.js` file and then `setup`, `this.ws` won't be `null` but we would need to execute the setup all over again.

However, I think I will follow @msbrogli suggestion in the other comment and set `this.ws = null` when ending the connection, so this should never happen.

I believe it doesn't hurt to keep like this only to protect in case this happen again. This `WebSocketHandler` is the file that will be removed in the next lib refactor.

src/WebSocketHandler.js
pedroferreira1: Thinking more about this situation, I believe it's better. Because if we are changing the server it's important to start the new `this.ws` in the new server.

src/WebSocketHandler.js
pedroferreira1: If it's not null we only execute the last line of the method

`return this.ws.setup();`

Run the setup normally

src/WebSocketHandler.js
pedroferreira1: Done. Tested with server change and everything is working fine.

src/WebSocketHandler.js
obiyankenobi: Ok

src/WebSocketHandler.js
msbrogli: I got it. Thanks!

PR #71, opened 2020-04-28
feat: updating txs count while paginating address history request

pedroferreira1:
ezgif-1-4785841e98f7

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -286,21 +286,7 @@ const wallet = {
       this.setWalletData(dataJson);
       storage.setStore(oldStore);
 
-      this.getTxHistory(addresses, connection, store).then((history) => {
-        let oldStore = _.clone(storage.store);
-        if (store) {
-          storage.setStore(store);
-        }
-        const data = this.getWalletData();
-        // Update historyTransactions with new one
-        const historyTransactions = 'historyTransactions' in data ? data['historyTransactions'] : {};
-        const allTokens = 'allTokens' in data ? data['allTokens'] : [];
-        const result = this.updateHistoryData(historyTransactions, allTokens, history, resolve, data, reject, connection, store);
-        WebSocketHandler.ws.emit('addresses_loaded', result);
-        storage.setStore(oldStore);
-      }, (e) => {
-        reject(e);
-      });
+      this.getTxHistory(addresses, resolve, reject, connection, store);
 
     });
     return promise;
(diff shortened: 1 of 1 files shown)


src/wallet.js
```diff
@@ -341,6 +342,15 @@ const wallet = {
...
           addressesToSearch = addressesToSearch.slice(addrIndex);
+
+          // Update lib data so we can get the partial transaction count while the pagination doesn't finish
+          const data = this.getWalletData();
+          const historyTransactions = 'historyTransactions' in data ? data['historyTransactions'] : {};
+          const allTokens = 'allTokens' in data ? data['allTokens'] : [];
+          const ret = this.updateHistoryData(historyTransactions, allTokens, result.history, null, data, null, connection, store);
+          WebSocketHandler.ws.emit('addresses_loaded', ret);
+        } else {
+          history = result.history;
```
obiyankenobi, line 353: Won't `history` return just the transactions of the last request, when `hasMore == false`?

obiyankenobi: Now I see that just for the last request the data in storage won't be updated here, but on the function that handles this promise. Is there a reason for that?

src/wallet.js
```diff
@@ -286,7 +286,7 @@ const wallet = {
       this.setWalletData(dataJson);
       storage.setStore(oldStore);
 
-      this.getTxHistory(addresses).then((history) => {
+      this.getTxHistory(addresses, connection, store).then((history) => {
```
obiyankenobi, line 289: As per my comment on `getTxHistory`, we seem to be updating the storage there already. I think we either do everything there or everything here.

src/wallet.js
obiyankenobi: I forgot how bad our code was here 🤣

It looks like `updateHistoryData` is doing a lot. We could probably refactor the code that merges new txs to the ones in storage to a new method and call it when we don't need `updateHistoryData` to call `resolve()`.

msbrogli: Another way to go would be to call a callback method every page. Then the callback would update the history data, and we can remove it from the promise resolution. What do you think?

pedroferreira1: I've done a refactor and separated the part of the `updateHistoryData` that updated the storage on another method, so I can call it when I've downloaded some txs but still need to finish the pagination.

When it's the last page, I just call the normal method, update the storage and continue calling the sequence of `loadAddress`.

Even though I agree this deserves a bigger refactor, my opinion is that we shouldn't wast much time on doing it right now, since this is a part of the lib we will probably refactor soon.

What do you think? @msbrogli @obiyankenobi

src/wallet.js
pedroferreira1: It's doing everything there right now.

PR #72, opened 2020-04-30
feat: improving send tokens timeout message

pedroferreira1:
image

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -6,7 +6,7 @@
  */
 
 import { OP_GREATERTHAN_TIMESTAMP, OP_DUP, OP_HASH160, OP_EQUALVERIFY, OP_CHECKSIG, OP_PUSHDATA1 } from './opcodes';
-import { DECIMAL_PLACES, CREATE_TOKEN_TX_VERSION, DEFAULT_TX_VERSION, TOKEN_INFO_VERSION, MAX_OUTPUT_VALUE_32, MAX_OUTPUT_VALUE, TOKEN_AUTHORITY_MASK } from './constants';
+import { DECIMAL_PLACES, CREATE_TOKEN_TX_VERSION, DEFAULT_TX_VERSION, TOKEN_INFO_VERSION, MAX_OUTPUT_VALUE_32, MAX_OUTPUT_VALUE, TOKEN_AUTHORITY_MASK, STRATUM_TIMEOUT_RETURN_CODE } from './constants';
 import { HDPrivateKey, crypto, encoding, util } from 'bitcore-lib';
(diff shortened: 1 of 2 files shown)


src/transaction.js
```diff
@@ -768,7 +768,13 @@ const transaction = {
         if (response.success) {
           resolve(response);
         } else {
-          reject(response.message);
+          let message = '';
+          if (response.return_code === 'stratum_timeout') {
```
msbrogli, line 772: I know it's been used only here, but I think we should make it a constant.

src/transaction.js
```diff
@@ -768,7 +768,13 @@ const transaction = {
...
+            message = 'All transactions need to resolve proof-of-work as an anti spam mechanism. Currently Hathor Network provides their full nodes to execute this task for free. However they may be fully loaded right now.\n\nThe bigger the transactions, the harder it is to resolve the proof of work.\n\nPlease, wait a few moments and try again soon and, if you can split in more transactions it will be easier to send them.';
```
msbrogli, line 773: I'd only change to: "and try again soon. If you can split in two or more smaller transactions, it will be easier to send them. Notice that what matters most is the number of inputs (instead of the amount being transferred)."

Is it confusing?

obiyankenobi: - We don't seem to say anywhere "Timeout solving transaction's proof-of-work". Or do we say it on the wallet modal? This is the most important information.

- Should we use "Hathor Network" or "Hathor Labs"?

- @msbrogli I think your suggestion is a bit confusing. Even if they have this information, there's not much they can do. People don't deal with the number of inputs, just the amount. Unless they manually select the inputs, which is very unlikely.

- My suggestion:
```
Timeout solving transaction's proof-of-work.

All transactions need to solve a proof-of-work as an anti spam mechanism. Currently, Hathor Labs provides this service for free, but their servers may be fully loaded right now.

The bigger the transactions, the harder it is to solve the proof-of-work. Please, wait a few moments and try again soon. If you can split your transaction into smaller ones it might be easier to send them.
```

pedroferreira1: I agree with @obiyankenobi that the inputs information is confusing, I've updated to his suggestion.

image

src/transaction.js
pedroferreira1: Done

src/transaction.js
msbrogli: I think the last part "If you split your transaction into smaller ones it might be easier to send them" may mislead to the understanding that the amount of the transaction is important. On the other hand, I also agree that talking about inputs is more confusing.

Ok. I agree that we can move forward.

PR #74, opened 2020-05-27
feat: using tx mining API to send transactions

pedroferreira1:
It's still a draft because there are some things to be done:

- [x] Add comments and docstring;
- [x] Handle some error cases (marked with TODO in the code)
- [x] Fix tests

### DIFF
--- a/src/api/txMining.js
+++ b/src/api/txMining.js
@@ -20,13 +20,17 @@ const txMiningApi = {
    * @param {String} tx Data in hexadecimal
    * @param {boolean} propagate If should propagate tx after the job is completed
    * @param {boolean} add_parents If should return the parents
+   * @param {Number} timeout Optional parameter to define the timeout of the submit job in seconds
    *
    * @return {Promise}
    * @memberof txMiningApi
    * @inner
    */
-  submitJob(tx, propagate, add_parents, resolve) {
-    const postData = {tx, propagate, add_parents};
+  submitJob(tx, propagate, add_parents, timeout, resolve) {
+    let postData = {tx, propagate, add_parents};
+    if (timeout) {
+      postData.timeout = timeout;
+    }
     return txMiningRequestClient(resolve).post('submit-job', postData).then((res) => {
       resolve(res.data)
     }, (res) => {
--- a/src/api/txMining.js
+++ b/src/api/txMining.js
@@ -55,6 +55,24 @@ const txMiningApi = {
       return Promise.reject(res);
     });
   },
+
+  /**
+   * Cancel a job
+   *
+   * @param {String} job Job id
+   *
+   * @return {Promise}
+   * @memberof txMiningApi
+   * @inner
+   */
+  cancelJob(job, resolve) {
+    const data = {'job-id': job};
+    return txMiningRequestClient(resolve).post('cancel-job', data).then((res) => {
+      resolve(res.data)
+    }, (res) => {
+      return Promise.reject(res);
+    });
+  },
 };
 
 export default txMiningApi;
--- a/src/api/txMining.js
+++ b/src/api/txMining.js
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import txMiningRequestClient from './txMiningAxios';
+
+/**
+ * Api calls for tx mining
+ *
+ * @namespace ApiTxMining
+ */
+
+const txMiningApi = {
+  /**
+   * Submit tx to be mined
+   *
+   * @param {String} tx data in hexadecimal
+   *
+   * @return {Promise}
+   * @memberof txMiningApi
+   * @inner
+   */
+  submitJob(tx, propagate, add_parents, resolve) {
+    const postData = {tx, propagate, add_parents};
+    return txMiningRequestClient(resolve).post('submit-job', postData).then((res) => {
+      resolve(res.data)
+    }, (res) => {
+      return Promise.reject(res);
+    });
+  },
+
+  /**
+   * Get job status
+   *
+   * @param {String} job Job id
+   *
+   * @return {Promise}
+   * @memberof txMiningApi
+   * @inner
+   */
+  getJobStatus(job, resolve) {
+    const data = {'job-id': job};
+    return txMiningRequestClient(resolve).get('job-status', {'params': data}).then((res) => {
+      resolve(res.data)
+    }, (res) => {
+      return Promise.reject(res);
+    });
+  },
+};
+
+export default txMiningApi;
--- a/src/api/txMining.js
+++ b/src/api/txMining.js
@@ -17,7 +17,9 @@ const txMiningApi = {
   /**
    * Submit tx to be mined
    *
-   * @param {String} tx data in hexadecimal
+   * @param {String} tx Data in hexadecimal
+   * @param {boolean} propagate If should propagate tx after the job is completed
+   * @param {boolean} add_parents If should return the parents
    *
    * @return {Promise}
    * @memberof txMiningApi
--- a/src/constants.js
+++ b/src/constants.js
@@ -182,4 +182,4 @@ export const MIN_POLLING_INTERVAL = 0.5;
 /**
  * Base URL for the tx mining api
  */
-export const TX_MINING_URL = 'http://localhost:8001/';
+export const TX_MINING_URL = 'http://txmining.mainnet.hathor.network/';
--- a/src/constants.js
+++ b/src/constants.js
@@ -177,7 +177,7 @@ export const STRATUM_TIMEOUT_RETURN_CODE = 'stratum_timeout';
 /**
  * Minimum job status poll to update job data when mining a tx
  */
-export const MIN_POLL = 0.5;
+export const MIN_POLLING_INTERVAL = 0.5;
 
 /**
  * Base URL for the tx mining api
--- a/src/constants.js
+++ b/src/constants.js
@@ -173,3 +173,13 @@ export const HASH_KEY_SIZE = 256;
  * Return code of the send_tokens response when there is a stratum timeout
  */
 export const STRATUM_TIMEOUT_RETURN_CODE = 'stratum_timeout';
+
+/**
+ * Minimum job status poll to update job data when mining a tx
+ */
+export const MIN_POLL = 0.5;
+
+/**
+ * Base URL for the tx mining api
+ */
+export const TX_MINING_URL = 'http://localhost:8001/';
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -9,6 +9,7 @@ import path from 'path';
 
 import storage from './storage';
 import { BLOCK_VERSION, CREATE_TOKEN_TX_VERSION, DEFAULT_TX_VERSION, MERGED_MINED_BLOCK_VERSION, GENESIS_BLOCK, DECIMAL_PLACES, DEFAULT_SERVER } from './constants';
+import { AddressError, OutputValueError, ConstantNotSet, CreateTokenTxInvalid, MaximumNumberInputsError, MaximumNumberOutputsError } from './errors';
 
 /**
  * Helper methods
@@ -323,7 +324,32 @@ const helpers = {
    */
   cleanupString(string) {
     return string.replace(/\s\s+/g, ' ').trim().toLowerCase();
-  }
+  },
+
+  /**
+   * Handle error for method transaction.prepareData
+   * Check if error is one of the expected and return the message
+   * Otherwise, throws the unexpected error
+   *
+   * @param {Error} e Error thrown
+   *
+   * @return {string} Error message
+   * @memberof Helpers
+   * @inner
+   */
+  handlePrepareDataError(e) {
+    if (e instanceof AddressError ||
+        e instanceof OutputValueError ||
+        e instanceof ConstantNotSet ||
+        e instanceof CreateTokenTxInvalid ||
+        e instanceof MaximumNumberOutputsError ||
+        e instanceof MaximumNumberInputsError) {
+      return e.message;
+    } else {
+      // Unhandled error
+      throw e;
+    }
+  },
 }
 
 export default helpers;
--- a/src/new/sendTransaction.js
+++ b/src/new/sendTransaction.js
@@ -43,6 +43,9 @@ class SendTransaction extends EventEmitter {
     // Job ID
     this.jobID = null;
 
+    // Error to be shown in case of no miners connected
+    this.noMinersError = 'There are no miners to resolve the proof of work of this transaction.';
+
     // Promise that resolves when push tx finishes with success
     // or rejects in case of an error
     this.promise = new Promise((resolve, reject) => {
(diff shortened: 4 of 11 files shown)


src/api/txMining.js
```diff
@@ -0,0 +1,54 @@
...
+ * Api calls for tx mining
+ *
+ * @namespace ApiTxMining
+ */
+
+const txMiningApi = {
+  /**
+   * Submit tx to be mined
+   *
+   * @param {String} tx data in hexadecimal
```
obiyankenobi, line 20: Missing other parameters

src/api/txMining.js
```diff
@@ -0,0 +1,54 @@
...
+   * Get job status
+   *
+   * @param {String} job Job id
+   *
+   * @return {Promise}
+   * @memberof txMiningApi
+   * @inner
+   */
+  getJobStatus(job, resolve) {
+    const data = {'job-id': job};
```
obiyankenobi, line 45: We probably want to have a standard for json naming. I'd say we should use `job_id` here to follow what we've been using.

src/new/sendTransaction.js
```diff
@@ -0,0 +1,94 @@
...
+          this.emit('estimation-updated', {estimation: response.expected_total_time});
+          this.handleJobStatus();
+        }
+      });
+    }, poll_time);
+  }
+
+  handlePushTx() {
+    const txHex = transaction.getTxHexFromData(this.data);
+    txApi.pushTx(txHex, false, (response) => {
```
obiyankenobi, line 71: Need to handle the error case and probably emit a corresponding event.

src/api/txMining.js
pedroferreira1: I will add all docstrings soon and then remove from draft

src/api/txMining.js
pedroferreira1: Actually I am following what we've been using in the new events in the classes. Check `src/new/wallet.js` and all new events are like `job-id`. We can decide to change but I think it's better to keep like this now, since it's our standard in the lib.

src/new/sendTransaction.js
pedroferreira1: My first idea what to return only one event with the request response, so people can check the success. But thinking about it now, maybe it's not the best approach. I will emit an event for success with the sent tx data and an event for error, with the error message.

Is that it, right?

src/api/txMining.js
obiyankenobi: Ah, my bad, didn't see that on your original post.

src/api/txMining.js
obiyankenobi: But this is not an event, this is the API parameter.

I mixed things up, it's not json as it's a get request. Anyway, I think we've been using `_` on all requests and `-` for local events.

src/new/sendTransaction.js
obiyankenobi: Actually, I meant that `txApi.pushTx()` can return a promise rejection, which is not being handled.

What you're saying is that the `tx-sent` event would have both the success and error messages?

src/api/txMining.js
pedroferreira1: Ahhh, my bad! I thought this was another file, sorry.

Actually this `job-id` was 'defined' in the design https://github.com/HathorNetwork/rfcs/issues/24 (it does not say that the GET parameter is this, but almost). And this is what @msbrogli has been using in his code, I am just expecting here what he is sending there.

This code won't be merged before his PR, so I will keep this conversation opened, so we can finish this discussion in his PR (and I change here, if needed).

src/new/sendTransaction.js
pedroferreira1: Yes, but not anymore. I've changed that to emit different events (for success and error) and added the promise reject error event also.

src/api/txMining.js
obiyankenobi: Got it, thanks.

src/new/sendTransaction.js
obiyankenobi: Looks good now.

src/api/txMining.js
msbrogli: I implemented following the design in the `tx-mining-service` as well.

src/api/txMining.js
```diff
@@ -0,0 +1,56 @@
...
+    return txMiningRequestClient(resolve).get('job-status', {'params': data}).then((res) => {
+      resolve(res.data)
+    }, (res) => {
+      return Promise.reject(res);
+    });
+  },
```
msbrogli, line 53: Should we add the `cancel-job` API?

src/constants.js
```diff
@@ -173,3 +173,13 @@ export const HASH_KEY_SIZE = 256;
  * Return code of the send_tokens response when there is a stratum timeout
  */
 export const STRATUM_TIMEOUT_RETURN_CODE = 'stratum_timeout';
+
+/**
+ * Minimum job status poll to update job data when mining a tx
+ */
+export const MIN_POLL = 0.5;
```
msbrogli, line 180: Maybe rename to `MIN_POLLING_INTERVAL`.

src/new/sendTransaction.js
```diff
@@ -0,0 +1,112 @@
...
+   * If the job is done, emits 'job-done' event, complete and send the tx
+   * Otherwise, schedule again the job status request and emits 'estimation-updated' event.
+   */
+  handleJobStatus() {
+    // this.estimation and MIN_POLL are in seconds
+    const poll_time = Math.max(this.estimation / 2, MIN_POLL)*1000;
+
+    setTimeout(() => {
+      txMiningApi.getJobStatus(this.jobID, (response) => {
+        if (response.status === 'done') {
```
msbrogli, line 73: Even though it should never happen, you may get a 404 if you take too long to check the status. Current tx-mining-server keeps the result for 5 minutes.

src/new/sendTransaction.js
```diff
@@ -0,0 +1,112 @@
...
+ * - Update mining time estimation from time to time;
+ * - Get back mining response;
+ * - Push tx to the network;
+ *
+ * It emits the following events:
+ * 'job-submitted': after job was submitted;
+ * 'estimation-updated': after getting the job status;
+ * 'job-done': after job is finished;
+ * 'send-success': after push tx succeeds;
+ * 'send-error': if an error happens;
```
msbrogli, line 28: Maybe we should add an event for unexpected errors.

src/new/sendTransaction.js
```diff
@@ -0,0 +1,112 @@
...
+ **/
+class SendTransaction extends EventEmitter {
```
msbrogli, line 30: Nice class. We may have a promise that is solved after the push tx. So, users might choose the best approach for their specific case.

src/new/wallet.js
```diff
@@ -342,11 +354,23 @@ class HathorWallet extends EventEmitter {
    *
    * @param {Object} data Full transaction data
    *
-   * @return {Promise} Promise that resolves when transaction is sent
+   * @return {Object} bject with {success: true, sendTransaction, promise}, where sendTransaction is a
+   * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
    **/
   sendPreparedTransaction(data) {
```
msbrogli, line 360: We should add an option to get the `SendTransaction` object instead of the promise.

src/transaction.js
```diff
@@ -731,21 +738,30 @@ const transaction = {
...
-          e instanceof CreateTokenTxInvalid ||
-          e instanceof MaximumNumberOutputsError ||
-          e instanceof MaximumNumberInputsError) {
-        return Promise.reject(e.message);
-      } else {
-        // Unhandled error
-        throw e;
-      }
+      const message = helpers.handlePrepareDataError(e);
+      return Promise.reject(e.message);
```
msbrogli, line 742: Should it be just `message`?

src/api/txMining.js
pedroferreira1: When are we going to use this API? Or do you think we should have it on the lib in case someone wants to use it?

src/new/sendTransaction.js
pedroferreira1: I will handle this case. What do you think we should return? Unexpected error? Because it shouldn't happen

src/new/sendTransaction.js
pedroferreira1: I will and all rejected promises will return it

src/new/sendTransaction.js
pedroferreira1: Interesting, I will add it.

src/transaction.js
pedroferreira1: Not in this method. This is the old send transaction method and it always returns a promise.

src/constants.js
pedroferreira1: Done!

src/new/wallet.js
pedroferreira1: This method returns an object (in case of success) with `{'success': true, promise, sendTransactionObject}`, so who is using it can decide if want to get the feedback events, or the promise.

src/api/txMining.js
obiyankenobi: This is something I didn't notice when reviewing the design, but I think it makes sense to stick to one "convention". We've been using mostly `_`, so I think we should continue.

src/constants.js
```diff
@@ -173,3 +173,13 @@ export const HASH_KEY_SIZE = 256;
...
+
+/**
+ * Minimum job status poll to update job data when mining a tx
+ */
+export const MIN_POLLING_INTERVAL = 0.5;
+
+/**
+ * Base URL for the tx mining api
+ */
+export const TX_MINING_URL = 'http://localhost:8001/';
```
obiyankenobi, line 185: Need to update this before merging. Do we have a live test API already?

src/tokens.js
```diff
@@ -332,35 +333,43 @@ const tokens = {
...
+      name,
+      symbol,
+    });
+
+    let preparedData = null;
+    try {
+      preparedData = transaction.prepareData(createTokenTxData, pin);
+    } catch (e) {
+      const message = helpers.handlePrepareDataError(e);
+      return {success: false, message: e.message};
```
obiyankenobi, line 355: Why do we need `const message = helpers.handlePrepareDataError(e); `? We don't use it.

src/tokens.js
```diff
@@ -702,15 +718,48 @@ const tokens = {
...
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  handleSendTransaction(data, pin, options) {
+    let preparedData = null;
+    try {
+      preparedData = transaction.prepareData(data, pin, options);
+    } catch (e) {
+      const message = helpers.handlePrepareDataError(e);
```
obiyankenobi, line 757: Why do we need `const message = helpers.handlePrepareDataError(e); `? We don't use it.

src/transaction.js
obiyankenobi: Why do we need `const message = helpers.handlePrepareDataError(e); `? We don't use it.

src/helpers.js
```diff
@@ -323,7 +324,32 @@ const helpers = {
...
+   * Check if error is one of the expected and return the message
+   * Otherwise, throws the unexpected error
+   *
+   * @param {Error} e Error thrown
+   *
+   * @return {string} Error message
+   * @memberof Helpers
+   * @inner
+   */
+  handlePrepareDataError(e) {
```
obiyankenobi, line 340: This is currently called in a lot of places where it might not be needed. It probably needs to be called in 1 place and we could keep the code there so it's easier to understand.

src/api/txMining.js
msbrogli: I think we should use it if user clicks on cancel, but I don't think we have this choice right now.

src/new/sendTransaction.js
msbrogli: I feel that unexpected error is a good one. It should definitely never happen. Maybe we should tell the user to check if the transaction has been sent looking into the history and try again if it hasn't.

src/new/sendTransaction.js
msbrogli: I don't think so.

src/new/wallet.js
msbrogli: Great! Thanks!

src/constants.js
msbrogli: We do. The live test API is available at http://txmining.mainnet.hathor.network/.

src/new/sendTransaction.js
pedroferreira1: Added

src/constants.js
pedroferreira1: Updated

src/helpers.js
pedroferreira1: This code was used before on the wallet desktop, I just moved here, in order to prevent duplicating code.

The method I should change to handle the errors is `transaction.prepareData`, however this is called in many places and this change would break compatibility, so I decided to keep the method unchanged and handle the errors like this. I don't see any problems with it, since it's being reused and without code duplication.

src/tokens.js
pedroferreira1: My bad, the returned message was supposed to be `message` and not `e.message`. Fixed it.

src/tokens.js
pedroferreira1: My bad, the returned message was supposed to be `message` and not `e.message`. Fixed it.

src/transaction.js
pedroferreira1: It's like @msbrogli said, it should be `message`. Fixed it, thanks.

src/new/sendTransaction.js
pedroferreira1: Yeah, it's better. Added this to the unexpected error

src/api/txMining.js
pedroferreira1: Yes, we never allow the user to cancel a transaction while it's sending. Do you think we should add the api endpoint even if it's not used? I prefer not to add it.

src/api/txMining.js
pedroferreira1: I agree that we should stick to a convention, however we are already using more than one, e.g. https://api.economics.hathor.network/circulating-supply.

So we should first define which convention we will use. I think we should change the endpoints on the tx-mining-api if we decide to use `_`

obiyankenobi: Ok, agreed

src/helpers.js
obiyankenobi: Ok, now `const message = helpers.handlePrepareDataError(e);` is being used in the other places, so it makes sense to keep this code here.

src/api/txMining.js
msbrogli: As it is Hathor's standard wallet library, I think we should; maybe someone will need it in the future.

pedroferreira1: Done

src/new/wallet.js
```diff
@@ -387,11 +403,14 @@ class HathorWallet extends EventEmitter {
    *
    * @param {Object} data Full transaction data
    *
-   * @return {Promise} Promise that resolves when transaction is sent
+   * @return {Object} bject with {success: true, sendTransaction, promise}, where sendTransaction is a
```
obiyankenobi, line 406: bject -> Object

PR #75, opened 2020-06-09
feat: method to get xpubkey using PIN to decrypt data

pedroferreira1:


src/wallet.js
```diff
@@ -2035,6 +2035,21 @@ const wallet = {
...
   },
+
+  /**
+   * Get xpubkey from storage xpriv (assumes PIN is correct)
+   *
+   * @param {String} pin User PIN used to encrypt xpriv on storage
+   *
+   * @return {String} Wallt xpubkey
+   */
+  getXPubFromPin(pin) {
```
msbrogli, line 2046: I'd name it `getXPubKeyFromXPrivKey(int)`. What do you think?

src/wallet.js
```diff
@@ -2035,6 +2035,21 @@ const wallet = {
     const key = hdpubkey.derive(index);
     return key.publicKey.toBuffer();
   },
+
+  /**
+   * Get xpubkey from storage xpriv (assumes PIN is correct)
+   *
+   * @param {String} pin User PIN used to encrypt xpriv on storage
+   *
+   * @return {String} Wallt xpubkey
```
obiyankenobi, line 2044: typo `Wallt`

src/wallet.js
pedroferreira1: Changed!

src/wallet.js
pedroferreira1: Thanks

PR #78, opened 2020-07-17
feat: creating method to send a transaction from an array of outputs

pedroferreira1:
Even though it's more logical to use `sendTransaction` I didn't want to break compatibility with the other method in case other projects are already using it.

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -347,6 +347,7 @@ class HathorWallet extends EventEmitter {
 
   /**
    * Send a transaction from its outputs
+   * Currently does not have support to send custom tokens, only HTR
    *
    * @param {Array} outputs Array of outputs with each element as an object with {'address', 'value'}
    *
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -295,7 +295,7 @@ class HathorWallet extends EventEmitter {
     const ret = this.prepareTransaction(address, value, token);
 
     if (ret.success) {
-      return this.sendPreparedTransaction(ret.data);
+      return transaction.sendTransaction(ret.data, this.pinCode);
     } else {
       return Promise.reject(ret.message);
     }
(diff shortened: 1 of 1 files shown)


src/new/wallet.js
```diff
@@ -325,15 +325,52 @@ class HathorWallet extends EventEmitter {
...
-    if (!ret.success) {
-      return ret;
+  /**
+   * Send a transaction from its outputs
+   *
+   * @param {Array} outputs Array of outputs with each element as an object with {'address', 'value'}
+   *
+   * @return {Promise} Promise that resolves when transaction is sent
+   **/
+  sendManyOutputsTransaction(outputs) {
```
obiyankenobi, line 355: Isn't this the same as `sendMultiTokenTransaction`?

src/new/wallet.js
```diff
@@ -325,15 +325,52 @@ class HathorWallet extends EventEmitter {
...
+    storage.setStore(this.store);
+    const data = {
+      tokens: [], // For now does not support custom tokens
+      inputs: [],
+      outputs: [],
+    };
+
+    for (const output of outputs) {
+      data.outputs.push({address: output.address, value: output.value, tokenData: 0})
```
obiyankenobi, line 364: It's always HTR?

src/new/wallet.js
pedroferreira1: No. The `sendMultiTokenTransaction` is more than this one and has lot of workarounds that I don't like (e.g. manually update outputs with fixed values). It was done in a hurry for the twitter bot and it's not well tested. I was about to delete this method but I decided to leave it there and keep just the new method in the PR.

src/new/wallet.js
pedroferreira1: Yes. There is a comment above this line saying that

```
    const data = {
      tokens: [], // For now does not support custom tokens
      inputs: [],
      outputs: [],
    };
```

I started doing the multi token method but it's more complicated than that, so I decided to keep this way just to solve the problem we have now with Euroline. This code will completely change after the wallet service, so there is no reason to support that now.

obiyankenobi: Got it, we just need to add this comment to the docstring.

pedroferreira1: Done

PR #82, opened 2020-07-22
fix: when reloading data we should clean storage data before loading new data

pedroferreira1:


src/wallet.js
```diff
@@ -1281,10 +1281,13 @@ const wallet = {
...
    * Reload data in the storage
    *
+   * @param {Connection} connection Connection object to subscribe for the addresses (optional)
+   * @param {Store} store Store object to save the data (optional)
+   *
    * @memberof Wallet
    * @inner
    */
-  reloadData() {
+  reloadData(connection = null, store = null) {
```
msbrogli, line 1290: To keep it flexible and safe, I think we should pass an object, e.g. `{ connection = null, store = null }`. This will require any call to explicitly name the options.

pedroferreira1: Done.

PR #88, opened 2020-08-25
refactor: keep only one websocket connection when using new Hathor Wallet code

pedroferreira1:
**The problem:**

We were getting some weird logs with two websocket connections in a single wallet headless, where one of the connections was opened and receiving messages while the other was closed for a long time.

**The change:**

I changed the old code to use a specific connection, which can be changed by a method. By default it uses the `WebSocketHandler` to continue supporting wallets desktop and mobile, while the new code sets the Connection object to be used in the old code, so only one websocket connection is created.

**Future refactor:**

With this change the wallet headless will not work with more than one wallet anymore, which might not be a problem because nowadays we are having problems even with only one wallet. So after this change I would like to continue debug Euroline's logs while I execute the next refactor, which is to support more than one wallet connected at the same time in the headless, while each one has only one connection.

### DIFF
--- a/src/new/connection.js
+++ b/src/new/connection.js
@@ -10,6 +10,7 @@ import networkInstance from '../network';
 import { DEFAULT_SERVERS } from '../constants';
 import version from '../version';
 import helpers from '../helpers';
+import wallet from '../wallet';
 import WS from '../websocket';
 
 /**
@@ -51,6 +52,7 @@ class Connection extends EventEmitter {
     this.currentServer = this.servers[0];
 
     this.websocket = new WS({ wsURL: helpers.getWSServerURL(this.currentServer) });
+    wallet.setConnection(this);
   }
 
   /**
@@ -105,6 +107,22 @@ class Connection extends EventEmitter {
     this.websocket.removeListener('is_online', this.onConnectionChange);
     this.websocket.removeListener('wallet', this.handleWalletMessage);
     this.setState(Connection.CLOSED);
+  }
+
+  /**
+   * Call websocket endConnection
+   * Needed for compatibility with old src/wallet code
+   **/
+  endConnection() {
+    this.websocket.endConnection();
+  }
+
+  /**
+   * Call websocket setup
+   * Needed for compatibility with old src/wallet code
+   **/
+  setup() {
+    this.websocket.setup();
   }
 }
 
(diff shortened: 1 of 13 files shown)


src/new/connection.js
```diff
@@ -51,6 +52,7 @@ class Connection extends EventEmitter {
     this.currentServer = this.servers[0];
 
     this.websocket = new WS({ wsURL: helpers.getWSServerURL(this.currentServer) });
+    wallet.setConnection(this);
```
obiyankenobi, line 55: This doesn't seem right to me. Why would the old wallet code need to be aware of this connection? From what I understand, Connection is only used with the new Wallet facade.

This is the way I imagined, but I might be missing something:
`src/wallet.js`: uses `WebSocketHandler`;
`src/new/wallet.js`: uses `Connection`;

src/wallet.js
```diff
@@ -162,7 +164,7 @@ const wallet = {
     let promise = null;
     if (loadHistory) {
       // Load history from address
-      WebSocketHandler.setup();
+      this._connection.setup();
```
obiyankenobi, line 167: Besides the other comment about being aware of the Connection class, we're changing it to have just 1 ws connection. In that case, using the singleton seems even more appropriate.

obiyankenobi (changes requested):

src/new/connection.js
pedroferreira1: This wouldn't work the way it is for the reason that the wallet headless must support different wallets (hence different connections) using the same lib code, and the singleton would have a single ws connection for all wallets.

This refactor here is the first of two connection refactors (I first wanted to fix the double connection problem to help Euroline) and the second refactor is to add again support to have many wallets connected.

src/wallet.js
obiyankenobi: Discussed this on a call. `WebSocketHandler` is the default if no `Connection` is set.

src/new/connection.js
obiyankenobi: We discussed this on a call. As updated on the description, this PR removes support for multiple connections, which will be re-added in the future.

obiyankenobi (approved):

pedroferreira1:
This PR won't be merged. The changes will be integrated in another PR with fix to work also with multiples wallets

PR #90, opened 2020-08-28
feat: improve seed words split

pedroferreira1:


src/wallet.js
```diff
@@ -67,24 +67,28 @@ const wallet = {
    *
    * @param {string} words Words (separated by space) to generate the HD Wallet seed
    *
-   * @return {Object} {'valid': boolean, 'message': string}
+   * @return {Object} {'valid': boolean, 'message': string, 'words': cleanWords}
```
obiyankenobi, line 70: `words` is type `string`, right?

Also, it might be more useful to just return an array, no?

obiyankenobi (approved):

src/wallet.js
```diff
@@ -67,24 +67,28 @@ const wallet = {
...
    * @memberof Wallet
    * @inner
    */
   wordsValid(words) {
+    let newWordsString = '';
     if (_.isString(words)) {
-      if (words.split(' ').length !== 24) {
+      // 1. Remove one or more spaces (or line breaks) before and after the 24 words
+      // 2. Substitute more then one space (or line break) for a single space
+      newWordsString = words.trim(/\s+/).replace(/\s+/g, ' ');
```
msbrogli, line 79: Maybe we should include `\r`, `\n`, and `\t`.

msbrogli (changes requested):

src/wallet.js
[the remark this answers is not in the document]
pedroferreira1: Would be something like this including this comment and the other below?

`words.trim(/(\s\t\n)+/g).replace(/[^\x00-\x7F]/g, " ").replace(/\s+/g, ' ');`?

src/wallet.js
pedroferreira1: Yeah, I improved this comment. Thanks

src/wallet.js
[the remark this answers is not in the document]
msbrogli: Maybe this is simpler: `words.replace(/[^A-Za-z0-9]/g, ' ').replace(/\s+/g, ' ').trim()`. What do you think?

msbrogli (approved):

src/wallet.js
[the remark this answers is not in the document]
pedroferreira1: Yeah, that also works and it's more readable. Changed to that

pedroferreira1: Resolved in the other comments

PR #94, opened 2020-09-09
refactor: keep only one websocket connection when using new Hathor Wallet code

pedroferreira1:


src/WebSocketHandler.js
```diff
@@ -15,47 +15,47 @@ class WebSocketHandler extends EventEmitter {
...
 
   endConnection() {
     // To keep compatibility with methods previously used in this singleton
-    if (this.ws !== null) {
-      this.ws.endConnection();
-      this.ws.emit = () => {};
-      this.ws = null;
+    if (this.websocket !== null) {
+      this.websocket.endConnection();
+      this.websocket.emit = () => {};
```
obiyankenobi, line 57: I know it's old code, but why is `this.websocket.emit = () => {}` after `this.websocket.endConnection()`? I thought it'd be the contrary.

src/wallet.js
```diff
@@ -972,13 +988,21 @@ const wallet = {
...
+  cleanWallet({endConnection = true, connection = null} = {}) {
+    this.unsubscribeAllAddresses({connection});
     this.cleanLoadedData();
-    WebSocketHandler.endConnection();
+    if (endConnection) {
+      if (connection === null) {
+        this._connection.endConnection();
+      } else {
+        connection.endConnection();
+      }
```
obiyankenobi, line 1004: It might be worth creating a method:
```js
_getConnection(connection) {
  return !connection ? this._connection : connection;
}
```

src/wallet.js
```diff
@@ -337,14 +339,23 @@ const wallet = {
...
           // If it's the last page, we update the storage and call the next loadAddress (if needed)
           // This is all done on updateHistoryData
+          let oldStore = _.clone(storage.store);
+          if (store) {
+            storage.setStore(store);
+          }
           const data = this.getWalletData();
           const historyTransactions = 'historyTransactions' in data ? data['historyTransactions'] : {};
           const allTokens = 'allTokens' in data ? data['allTokens'] : [];
+          storage.setStore(oldStore);
```
obiyankenobi, line 349: Would be good to have some comments on all these storage changes.

src/WebSocketHandler.js
pedroferreira1: The idea here is to execute a full cleanup on `this.websocket`, so does not make difference the order. We are not trying to prevent something that might happen in `this.websocket.endConnection()`.

src/wallet.js
pedroferreira1: Added some comments explaining why we need to set the other stores.

src/wallet.js
pedroferreira1: Done, can you see if it's good?

obiyankenobi (approved):

src/new/wallet.js
```diff
@@ -100,6 +101,8 @@ class HathorWallet extends EventEmitter {
 
     this.onConnectionChangedState = this.onConnectionChangedState.bind(this);
     this.handleWebsocketMsg = this.handleWebsocketMsg.bind(this);
+
+    this.firstConnection = true;
```
msbrogli, line 105: Add comments explaining what it is used for.

src/wallet.js
```diff
@@ -337,14 +345,22 @@ const wallet = {
         } else {
           // If it's the last page, we update the storage and call the next loadAddress (if needed)
           // This is all done on updateHistoryData
+          let oldStore = _.clone(storage.store);
```
msbrogli, line 348: Why do we have to get it cloned?

msbrogli (approved):
The code is getting more and more confusing. We must refactor it as soon as possible.

src/wallet.js
pedroferreira1: I save the `oldStore` to set it back after the actions are done. We don't know the `store` that will be implemented for each case, so I clone it to prevent an error if the store is passed as reference, to guarantee when I set it back later it will really be the old store.

src/new/wallet.js
pedroferreira1: Done!

src/wallet.js
pedroferreira1: As discussed privately, we will merge this and handle the `clone` problem in another PR

PR #98, opened 2020-09-14
feat: return error message with words that are wrong

pedroferreira1:


src/wallet.js
```diff
@@ -79,12 +79,31 @@ const wallet = {
...
       } else if (!Mnemonic.isValid(newWordsString)) {
-        // Invalid sequence of words
-        return {'valid': false, 'message': 'Invalid sequence of words'};
+        // Check if there is a word that does not belong to the list of possible words
+        const wordlist = Mnemonic.Words.ENGLISH;
+        const wordsLowerCase = wordsArray.map((value) => value.toLowerCase());
+        const errorList = [];
+
+        for (let i = 0; i < wordsLowerCase.length; i += 1) {
+          const w = wordsLowerCase[i];
```
obiyankenobi, line 93: Why not `for (const word of wordsLowerCase)`?

obiyankenobi (approved):

msbrogli (approved):

src/wallet.js
pedroferreira1: No particular reason. Changed to `for ... of ...`, it's better

PR #102, opened 2020-09-21
fix: should add new listener when setting new connection as wallet default

pedroferreira1:


obiyankenobi (approved):

src/wallet.js
```diff
@@ -2111,7 +2115,9 @@ const wallet = {
    * @inner
    */
   setConnection(connection) {
+    this.removeMetricsListener();
```
msbrogli, line 2118: As the other connections will kept running, we might lose messages when we remove the listener because the websocket's pubsub won't queue the messages for late processing.

One way to go is to register the listeners on the connection and create an internal buffer. So, if nobody's listening, we can wait for the next subscription and dispatch the buffer when it happens.

What do you think?

msbrogli (changes requested):

src/wallet.js
pedroferreira1: I think I understand your concern that we might lose some messages and would have inconsistent information on the lib, is that right?

If that's the case, I think we should keep this as simple as possible for the use we have now because the lib will change a lot soon.

This metrics listener receives the messages from the full node about all the metrics (the same one we listen on the explorer to update the number of blocks, txs and hash rate). This message is sent every 5 seconds and in the lib we use it to update the current network height to know which blocks can already be spent. If we lose one or two messages is not a big deal here because 5 seconds later we have another one with the updated data. Even though I think your solution would prevent some problems in case we add different messages here, I think we should keep it as simple as possible for now.

msbrogli: I thought we would lost some messages. I didn't know a new update will arrive 5s later, so it would update anyway. I'd suggest us to add a comment explaining that some messages might be lost but new ones will arrive 5s after, so it won't generate any bug.

msbrogli (approved):

src/wallet.js
pedroferreira1: Done!

Issue #103, opened 2020-09-22
Refactor lib code


pedroferreira1:
#TODO add a more detailed description

1. Separate methods used by the wallets
2. Separate helpers methods, e.g. validation of words, serialization of transactions
3. Create interface to interact with wallet service
4. Add object models, e.g. Transaction, Output, Input, ...

obiyankenobi:
> Add object models, e.g. Transaction, Output, Input. (I would start with only those for now)

On the wallet service, I've created two interfaces for the outputs, following our return data from the APIs: `DecodedOutput` and `TxOutputs`. To be honest, I still have to think if it's better to merge these two types or keep them separated. And we probably want to create proper classes so there are methods.

```
export interface DecodedOutput {
  type: string;
  address: string;
  timelock: number | null;
  value: number;
  token_data?: number;
}

export interface TxOutput {
  value: number;
  script: string;
  token: string;
  decoded: DecodedOutput;
  spent_by: string | null;
  token_data?: number;
}
```

> I think those methods could be in another lib package and the wallet-lib would use them but for now my suggestion is just to separate them in a different module so it's easier to do some tests and interact with them.

I'm a bit confused about your suggestion. You mean keep them in the same lib or no? I'm not sure what you call lib and module.

Anyway, under "Methods to separate in a new module" we need to debate some of them a bit better. From what I understood, these should be methods on a standalone lib, not related directly to a wallet. If that's the case, there are some which should not be there, in my opinion. For instance, "createMintTokenTx (integration with wallet-service)". You can only interact with the wallet service if you started a wallet there (and the wallet service loads all addresses and marks the corresponding UTOXs as belonging to a wallet). So these kinds of methods should only be on a wallet-lib.

Moreover, some methods will probably move to a class, like `txToBytes`, `calculateTxWeight`, etc will all move to the `Transaction` class (defined as the first task).

pedroferreira1:
> On the wallet service, I've created two interfaces for the outputs, following our return data from the APIs: DecodedOutput and TxOutputs

I don't really see the benefits of having the `TxOutput` for the lib, this has some calculated fields using full node data (`spent_by`). I created the `Output` class as simple as possible with `value`, `address`, `tokenData` and `timelock`.

We have a method to create output script, so it's as simple as `output.createScript()` to get this. What do you think? My idea is to keep simple (and complete). And if we need more, we can add it in the future.

[...]

PR #110, opened 2020-10-02
refactor: creating first models classes and moving methods to it

pedroferreira1:
There are some decisions that were made in this PR that should be discussed:

1. We used to get some constants from the full node when starting the wallet. I've added those constants to our `constants.js` file because it should be possible by anyone to create and test a `Transaction` before connecting to any server.

2. I've decided to keep all older files (`src/wallet.js`, `src/helpers.js`, ...) to keep compatibility if any user is directly using them, however they will all be useless after `wallet-service` integration. Should we delete them? This includes the tests files, I ended up "duplicating" some tests to separate all of them.

One other "problem" from that is the export name I had to use because `src/wallet.js` already exports for `wallet`, then I had to use `walletUtils` for `src/utils/wallet.ts`.

3. I've added `"noImplicitAny": false` to the `tsconfig.json` for now so we don't need to take care of external types, like `bitcore-lib` types. Is it ok for now or should we make it strict?

4. To make our classes constructor more flexible I had to write them like this:

```
type optionsType = {
  version?: number,
  weight?: number,
  // All other fields
};

const defaultOptions: optionsType = {
  version: DEFAULT_TX_VERSION,
  weight: 0,
  // All other fields
}

class Transaction {
  inputs: Input[];
  outputs: Output[];
  version: number;
  weight: number;
  // All other fields

  constructor(inputs: Input[], outputs: Output[], options: optionsType = defaultOptions) {
    const newOptions = Object.assign(defaultOptions, options);
    const { version, weight, nonce, timestamp, parents, tokens, hash, name, symbol } = newOptions;
    this.inputs = inputs;
    this.outputs = outputs;
    this.version = version;
    this.weight = weight;
    // All other fields
  }
```

Even though I like it very much and makes really flexible for all optional fields, I think it's still a bit verbose and we have to write a lot of similar code. Can we do it better?


obiyankenobi:
> 1. We used to get some constants from the full node when starting the wallet. I've added those constants to our `constants.js` file because it should be possible by anyone to create and test a `Transaction` before connecting to any server.

Seems good to me, as long as they are identified in the file as potentially causing problems if different from the server. For example, wrong weight params could create a tx with weight too low. I'll take a look in the code to see how you did it.

> 2. I've decided to keep all older files (`src/wallet.js`, `src/helpers.js`, ...) to keep compatibility if any user is directly using them, however they will all be useless after `wallet-service` integration. Should we delete them? This includes the tests files, I ended up "duplicating" some tests to separate all of them.
>
> One other "problem" from that is the export name I had to use because `src/wallet.js` already exports for `wallet`, then I had to use `walletUtils` for `src/utils/wallet.ts`.

On #103 we briefly talked about keeping it on the same lib or creating a different one. We agreed to keep it in the same one for simplicity, but maybe it will be simpler to split it already? Is it much work? Any other downsides other than the work?

src/models/transaction.ts
```diff
@@ -0,0 +1,348 @@
...
+  hash?: string | null,
+  name?: string | null,
+  symbol?: string | null,
+};
+
+const defaultOptions: optionsType = {
+  version: DEFAULT_TX_VERSION,
+  weight: 0,
+  nonce: 0,
+  timestamp: null,
```
obiyankenobi, line 32: Maybe it should be the current timestamp?

src/models/output.ts
```diff
@@ -0,0 +1,155 @@
...
+class Output {
+  // Output value as an integer
+  value: number;
+  // Address object of the value destination
+  address: Address;
+  // tokenData of the output
+  tokenData: number;
+  // Timestamp of the timelock of the output
+  timelock: number | null;
+
```
obiyankenobi, line 35: Might be worth having a `tokenUid` field here (optional). From the perspective of a user, I want to create an outputs like `const output = new Output(value, address, tokenUid)`.

Later on, I create a transaction with it: `const tx = new Transaction(inputs, [output])`. And it also would be good not having to deal with the token data and tokens array manually. So something like this would set the tokens array and the corresponding token data for each output: `tx.finalizeTokenData()`.

A user that wants to send a tx could use it like this:
```
const o1 = new Output(value1, address1, tokenA);
const o2 = new Output(value2, address2, tokenB);
const o3 = new Output(value3, address3, tokenA);

const tx = new Transaction(inputs, [o1, o2, o3]);
tx.finalizeTokenData();
// still need to calculate weight, maybe finalize should do it all
send(tx);
```

obiyankenobi (changes requested):

src/models/address.ts
```diff
@@ -0,0 +1,105 @@
...
+   * @memberof Address
+   * @inner
+   */
+  validateAddress(): boolean {
+    const addressBytes = this.decode();
+    const errorMessage = `Invalid address: ${this.base58}`;
+
+    // Validate address length
+    if (addressBytes.length !== 25) {
+      throw new AddressError(errorMessage);
```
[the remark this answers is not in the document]
msbrogli: I agree we should personalize the error messages with details of the error. If anyone is using it and would like to show a custom message to the user, they can catch the exception and raise a new one.

src/models/address.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    // Validate address checksum
+    const checksum = addressBytes.slice(-4);
+    const addressSlice = addressBytes.slice(0, -4);
+    const correctChecksum = helpers.getChecksum(addressSlice);
+    if (!util.buffer.equals(checksum, correctChecksum)) {
+      throw new AddressError(errorMessage);
```
msbrogli, line 93: Same here.

src/models/address.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    // Validate version byte. Should be the p2pkh or p2sh
+    const firstByte = addressBytes[0];
+    if (firstByte !== this.network.versionBytes.p2pkh && firstByte !== this.network.versionBytes.p2sh) {
+      throw new AddressError(errorMessage);
```
msbrogli, line 99: Same here.

src/models/network.ts
```diff
@@ -0,0 +1,125 @@
...
+const mainnet = Networks.add({
+  name: 'mainnet',
+  alias: 'production',
+  pubkeyhash: versionBytes['mainnet']['p2pkh'],
+  privatekey: 0x80,
+  scripthash: versionBytes['mainnet']['p2sh'],
+  xpubkey: 0x0488b21e,
+  xprivkey: 0x0488ade4,
+  networkMagic: 0xf9beb4d9,
+  port: 8333,
```
msbrogli, line 45: Isn't it Bitcoin's default port?!

src/models/network.ts
```diff
@@ -0,0 +1,125 @@
...
+  mainnet
+}
+
+type versionBytesType = {
+  p2pkh: number,
+  p2sh: number,
+}
+
+
+class Network {
```
msbrogli, line 73: Maybe we should add the constants related to the network here.

src/models/output.ts
msbrogli: I'd rather keep only the original fields and add some methods like `setTokenUid(...)`, `getTokenUid(...)`, etc. We can add a `tokenUid` to the constructor as well.

I don't like to create a specific `tokenUid` field because it might conflict with the `tokenData`. It will probably be a misuse but we should prevent users from making mistakes.

src/models/output.ts
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @memberof Output
+   * @inner
+   */
+  isMelt(): boolean {
+    return this.isAuthority() && ((this.value & TOKEN_MELT_MASK) > 0);
+  }
+
+  /**
+   * Get index of token list of the output
```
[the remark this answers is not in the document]
msbrogli: I agree that it must be clarified in the docstrings.

src/models/transaction.ts
msbrogli: I think it should be null here. We can initialize in the constructor or create a `setCurrentTimestamp()` method.

src/models/transaction.ts
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  /**
+   * Return transaction data to sign in inputs
+   * 
+   * @return {Buffer}
+   * @memberof Transaction
+   * @inner
+   */
+  getDataToSign(): Buffer {
```
msbrogli, line 90: Should we cache it? I mean, the common use is to sign/verify the transaction, and all inputs will get the same data.

src/models/input.ts
```diff
@@ -0,0 +1,42 @@
...
+const defaultOptions = {
+  data: null
+}
+
+class Input {
+  // Hash of the transaction is being spent
+  hash: string;
+  // Index of the outputs array from the output being spent
+  index: number;
+  // Input signed data
```
msbrogli, line 21: Not necessarily "signed data". Maybe change to redeemScript?

src/models/transaction.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    // Tokens data
+    for (const token of this.tokens) {
+      arr.push(new encoding.BufferReader(token).buf);
+    }
+
+    for (const inputTx of this.inputs) {
+      arr.push(util.buffer.hexToBuffer(inputTx.hash));
+      arr.push(helpers.intToBytes(inputTx.index, 1));
+      // Input data will be fixed to 0 for now
```
msbrogli, line 114: We should clarify that we don't use the redeemScript to sign/verify the transaction (and add a link to The Anatomy of a Transaction RFC).

src/models/transaction.ts
```diff
@@ -0,0 +1,350 @@
...
+    // We need to take into consideration the decimal places because it is inside the amount.
+    // For instance, if one wants to transfer 20 HTRs, the amount will be 2000.
+    const amount = sumOutputs / (10 ** DECIMAL_PLACES);
+
+    let weight = (TX_WEIGHT_CONSTANTS.txWeightCoefficient * Math.log2(txSize) + 4 / (1 + TX_WEIGHT_CONSTANTS.txMinWeightK / amount) + 4);
+
+    // Make sure the calculated weight is at least the minimum
+    weight = Math.max(weight, TX_WEIGHT_CONSTANTS.txMinWeight)
+    // FIXME precision difference between backend and frontend (weight (17.76246721531992) is smaller than the minimum weight (17.762467215319923))
+    return weight + 1e-6;
```
msbrogli, line 181: We can explain that `1e-6` have no practical effect when mining the transaction.

src/models/transaction.ts
```diff
@@ -0,0 +1,350 @@
...
+   * For now, we only verify the maximum number of inputs and outputs.
+   *
+   * @throws {MaximumNumberInputsError} If the tx has more inputs than the maximum allowed
+   * @throws {MaximumNumberOutputsError} If the tx has more outputs than the maximum allowed
+   *
+   * @memberof Transaction
+   * @inner
+   */
+  validate() {
+    if (this.inputs.length > MAX_INPUTS) {
```
msbrogli, line 289: Why don't we add other validations?

src/models/transaction.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  /**
+   * Create Transaction object from inputs and outputs
+   *
+   * @return {Transaction} Transaction object
+   * @memberof Transaction
+   * @inner
+   */
+  static createUnsignedTx(inputs: Input[], outputs: Output[]): Transaction {
+    return new Transaction(inputs, outputs);
```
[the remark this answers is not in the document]
msbrogli: I agree that we can remove this static method.

src/utils/date.ts
```diff
@@ -0,0 +1,72 @@
...
+   * From seconds transform into days, hours, minutes and seconds
+   *
+   * @param {number} uptime Seconds of uptime
+   *
+   * @return {string} Formatted uptime seconds
+   *
+   * @memberof Date
+   * @inner
+   */
+  uptimeFormat(uptime: number): string {
```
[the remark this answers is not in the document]
msbrogli: I like it. But we can easily replace it by [the moment lib](https://github.com/moment/moment).

src/utils/helpers.ts
```diff
@@ -0,0 +1,260 @@
...
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  getTxType(tx: Transaction): string {
+    if (this.isBlock(tx)) {
+      if (tx.version === BLOCK_VERSION) {
+        return 'Block';
+      } else if (tx.version === MERGED_MINED_BLOCK_VERSION) {
+        return 'Merged Mining Block';
```
msbrogli, line 37: Spaced strings? Isn't it better to return an ENUM? Or put them in constants?

src/utils/helpers.ts
```diff
@@ -0,0 +1,260 @@
...
+   * Return the checksum of the bytes passed
+   * Checksum is calculated as the 4 first bytes of the double sha256
+   * 
+   * @param {Buffer} bytes Data from where the checksum is calculated
+   *
+   * @return {Buffer}
+   * @memberof Address
+   * @inner
+   */
+  getChecksum(bytes: Buffer): Buffer {
```
msbrogli, line 255: I'd rename it to `getAddressChecksum(...)`.

src/utils/wallet.ts
```diff
@@ -0,0 +1,220 @@
...
+    let newWordsString = '';
+    if (_.isString(words)) {
+      // 1. Replace all non ascii chars by a single space
+      // 2. Remove one or more spaces (or line breaks) before and after the 24 words
+      // 3. Set text to lower case
+      newWordsString = words.replace(/[^A-Za-z0-9]/g, ' ').replace(/\s+/g, ' ').trim().toLowerCase();
+      const wordsArray = newWordsString.split(' ');
+      if (wordsArray.length !== 24) {
+        // Must have 24 words
+        return {'valid': false, 'message': 'Must have 24 words'};
```
msbrogli, line 39: I'd raise an exception and let the user decide what to do with it. It works better for i18n as well.

src/utils/wallet.ts
```diff
@@ -0,0 +1,220 @@
...
+
+        for (const word of wordsArray) {
+          if (wordlist.indexOf(word) < 0) {
+            errorList.push(word);
+          }
+        }
+
+        let errorMessage = '';
+        if (errorList.length > 0) {
+          errorMessage = `Invalid words: ${errorList.join(' ')}`;
```
msbrogli, line 53: I'd return a list of invalid words and let the user decide what to do with it.

src/utils/wallet.ts
```diff
@@ -0,0 +1,220 @@
...
+   *
+   * @param {Buffer} pubkey Uncompressed public key
+   *
+   * @return {Buffer} Compressed public key
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  toPubkeyCompressed(pubkey: Buffer): Buffer {
+    const x = pubkey.slice(1, 33);
```
msbrogli, line 119: Assert the size of `pubKey` before slicing?

src/utils/wallet.ts
```diff
@@ -0,0 +1,220 @@
...
+   * Get public key for specific key index derivation
+   *
+   * @param {number} index Index of the key to derive
+   *
+   * @return {Buffer} Public key
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getPublicKeyFromXpub(xpubkey: string, index: number): Buffer {
```
msbrogli, line 135: Should we receive a path?

msbrogli (changes requested):

src/constants.js
```diff
@@ -189,6 +189,31 @@ export const MIN_POLLING_INTERVAL = 0.5;
  */
 export const TX_MINING_URL = 'https://txmining.mainnet.hathor.network/';
 
+/**
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
pedroferreira1: Yeah, removed it, thanks

src/models/address.ts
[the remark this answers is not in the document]
pedroferreira1: I agree about the error messages being more detailed. I just don't understand having this if it's going to be a private method, as the comment above to expose only one of the methods. The user won't be able to catch this exception message.

src/models/address.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Check if address is a valid string
+   *
+   * @return {boolean} If address is valid
+   * @memberof Address
+   * @inner
+   */
+  isValid(): boolean {
```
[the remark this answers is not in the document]
pedroferreira1: I agree. `validateAddress` is now private.

src/models/address.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { AddressError } from '../errors';
+import { encoding, util } from 'bitcore-lib';
+import defaultNetwork from '../network';
```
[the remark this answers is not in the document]
pedroferreira1: Agreed. Done

src/models/output.ts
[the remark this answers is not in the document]
pedroferreira1: It was not considering the 1. I've added the `- 1` to the code and explained in the docstring.

src/models/transaction.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    for (const outputTx of this.outputs) {
+      arr.push(outputTx.valueToBytes());
+      // Token data
+      arr.push(helpers.intToBytes(outputTx.tokenData, 1));
+
+      const outputScript = outputTx.createScript();
+      arr.push(helpers.intToBytes(outputScript.length, 2));
+      arr.push(outputScript);
+    }
```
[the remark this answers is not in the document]
pedroferreira1: Yeah, it's better with each class serializing the object.

src/utils/helpers.ts
pedroferreira1: It was exactly this in the beginning but we don't use this checksum only for addresses. We use it to calculate the custom token configuration string as well. That's why I changed the name. What do you think?

src/models/transaction.ts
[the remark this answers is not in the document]
pedroferreira1: I agree. Removed

src/utils/date.ts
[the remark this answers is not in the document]
pedroferreira1: Yeah, this is used only in the explorer. I will remove it and create there.

src/utils/helpers.ts
```diff
@@ -0,0 +1,260 @@
...
+   * Check if object is a block or a transaction
+   *
+   * @param {Transaction} tx Transaction to be checked
+   *
+   * @return {boolean} true if object is a block, false otherwise
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  isBlock(tx: Transaction): boolean {
```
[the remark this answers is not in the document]
pedroferreira1: Agreed

src/models/transaction.ts
pedroferreira1: What's the difference between initializing here or in the constructor? I think it would be ok for the default value to be the current timestamp, just like the version. I would just need to take care to execute the timestamp in the moment the constructor is created, not when I set this object. If it gets too complicated I can set in the constructor.

src/models/output.ts
pedroferreira1: I like the idea of `tokenUid` to be honest, it's much more intuitive to create an output with this, instead of requiring the user to understand which `tokenData` to use. However it's not that simple to create a `finalizeTokenData` method, it would require us to add many other optional fields. The normal case if easy, HTR is `tokenData = 0` and each `tokenUid` you increment one but if the output is a mint or melt, you will never be able to set the correct `tokenData`. For that we need to have more optional fields in the output, e.g. `isMint` or `isMelt`.

I suggest we keep simple as it is right now and when we start using these models we can improve it the way we think is better.

__tests__/models/output.test.js
```diff
@@ -0,0 +1,70 @@
...
+import buffer from 'buffer';
+import { AUTHORITY_TOKEN_DATA, TOKEN_MINT_MASK, TOKEN_MELT_MASK, MAX_OUTPUT_VALUE } from '../../src/constants';
+
+
+test('Validate value', () => {
+  const o1 = new Output(1000, new Address('WZ7pDnkPnxbs14GHdUFivFzPbzitwNtvZo'));
+  expect(o1.valueToBytes()).toBeInstanceOf(buffer.Buffer);
+
+  // Negative value is invalid
+  const o2 = new Output(-1000, new Address('WZ7pDnkPnxbs14GHdUFivFzPbzitwNtvZo'));
```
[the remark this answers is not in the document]
pedroferreira1: Great! Added!

src/utils/wallet.ts
pedroferreira1: Done!

src/utils/wallet.ts
pedroferreira1: I agree. Done

src/utils/wallet.ts
pedroferreira1: I just did it but I am thinking about being consistent. For all the other errors I am trowing an exception and letting the user decide what to do. In this specific case I will return `{'valid': false, 'invalidWords': ['walk']}`.

Is it consistent or can we do it better?

src/utils/wallet.ts
pedroferreira1: This method is a bit different. Here we expect the `xpub` of the specific key you want to derive. I think a method that we receive a root xpub and derive the path is valid but should be discussed. Should we accept a different an xpub from a different source than the root? Should we accept a partial path?

I think we will need a method like this eventually and then we can add. What do you think?

src/constants.js
```diff
@@ -189,6 +189,31 @@ export const MIN_POLLING_INTERVAL = 0.5;
...
+export const TX_WEIGHT_CONSTANTS = {
+  txMinWeight: 14,
+  txWeightCoefficient: 1.6,
+  txMinWeightK: 100
+};
+
+/**
+ * Maximum number of inputs
+ */
+export const MAX_INPUTS = 255;
```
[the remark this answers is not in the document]
pedroferreira1: I am not sure we should do that. Right now the transaction does not depend on the network object, only the address (which we could consider the same).

There are 4 constants that depend on the full node that were added to this constants file in order to have this lib independent from any full node. The only one that is not related to transaction is the `TOKEN_DEPOSIT_PERCENTAGE`. Should we make everything related to the network object? Or maybe make those constants easily editable?

src/models/address.ts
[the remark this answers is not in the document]
msbrogli: User can call this method directly, right? I imagine a situation in which an error occurs, user logs the transaction and one is inspecting what's wrong. In this case, one might call this method directly and one will get a friendly exception message.

src/models/transaction.ts
msbrogli: This code is executed when the module is loaded. The constructor is executed each time an object is created.

src/utils/helpers.ts
msbrogli: Ok. I usually don't like generic names because it is hard to figure out where it's been used.

src/models/transaction.ts
pedroferreira1: Yeah, makes sense. I will keep it null for now and we can change later if we think it's better.

src/models/address.ts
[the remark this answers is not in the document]
pedroferreira1: Not exactly, after this discussion here https://github.com/HathorNetwork/hathor-wallet-lib/pull/110#discussion_r501795240 (where I think we 3 agreed), this method became private. Or have I misunderstood the discussion?

src/models/network.ts
pedroferreira1: Yeah, this is not really important for us. It's some internal value for the bitcore lib

src/models/transaction.ts
pedroferreira1: Makes sense! Done

src/models/input.ts
pedroferreira1: Just made clear that it's a signed data for P2PKH and redeemScript for P2SH

src/models/transaction.ts
pedroferreira1: Done!

src/models/transaction.ts
pedroferreira1: Done

src/models/transaction.ts
pedroferreira1: Which ones? I've added the ones that were already in the lib. We could add almost all validations from full node, like the sum of inputs/outputs and others but is it useful? It's a bunch of code because there are special cases with authorities.

src/utils/helpers.ts
pedroferreira1: Created an ENUM

__tests__/utils/wallet.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+  expect(wallet.getXPubKeyFromXPrivKey(xpriv.xprivkey)).toBe(xpriv.xpubkey);
+
+  const derivedXpriv = xpriv.derive(`m/44'/${HATHOR_BIP44_CODE}'/0'/0`);
+  const chainCode = derivedXpriv._buffers.chainCode;
+  const fingerprint = derivedXpriv._buffers.parentFingerPrint;
+  const derivedXpub = wallet.xpubFromData(derivedXpriv.publicKey.toBuffer(), chainCode, fingerprint, 'testnet');
+  expect(derivedXpub).toBe(derivedXpriv.xpubkey);
+
+  // TODO test toPubkeyCompressed and getPublicKeyFromXpub
```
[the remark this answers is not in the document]
pedroferreira1: Added the missing tests

src/constants.js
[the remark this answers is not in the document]
msbrogli: I agree it is good to have constants to make it independent of the server. But I feel we should require an initialization from the user. Something like:

```
initFromServer(host)
initManually({network: "mainnet"})
```

Hence, users can decide what's the best way to configure their environment.

src/models/address.ts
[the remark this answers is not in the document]
msbrogli: I'd expose `validateAddress` as well. I feel both methods are useful and used in difference situations.

src/models/address.ts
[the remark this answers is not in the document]
msbrogli: I don't think `validateAddress()` should be private, and I feel it should personalize the error messages.

src/models/output.ts
msbrogli: But there's no such thing as `tokenUid`. I agree that we should keep it simple for users, so we should have methods to do that properly (e.g., `setTokenUid(...)`, `getTokenUid(...)`).

src/models/transaction.ts
msbrogli: I agree that we can do it later. Maybe we should just create an issue to do that.

The very first next step can be to validate transactions with no authority inputs/outputs. This validation is very simple.

src/utils/wallet.ts
msbrogli: There's no way to know whether the xpub is the root or a derived path. I agree with the current format. Maybe we should just update the docs to let the user know that they should derivate the path and send the xpub just after doing it.

src/models/network.ts
msbrogli: Ok. Maybe add a comment explaining it.

src/utils/wallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (errorList.length > 0) {
+          return {'valid': false, 'invalidWords': errorList};
+        } else {
+          // Invalid sequence of words
+          throw new InvalidWords('Invalid sequence of words.')
+        }
+      }
+    } else {
+      // Must be string
+      throw new InvalidWords('Words must be a string.')
```
msbrogli, line 62: I'd reverse the `if` condition and put this at the beginning. I feel it is way easier to read non-indented code (and we don't need the `else` in this case).

src/utils/wallet.ts
```diff
@@ -0,0 +1,225 @@
...
+    let newWordsString = '';
+    if (_.isString(words)) {
+      // 1. Replace all non ascii chars by a single space
+      // 2. Remove one or more spaces (or line breaks) before and after the 24 words
+      // 3. Set text to lower case
+      newWordsString = words.replace(/[^A-Za-z0-9]/g, ' ').replace(/\s+/g, ' ').trim().toLowerCase();
+      const wordsArray = newWordsString.split(' ');
+      if (wordsArray.length !== 24) {
+        // Must have 24 words
+        throw new InvalidWords('Must have 24 words.')
```
msbrogli, line 40: Shouldn't we validate the given words and return which ones are invalid? It seems useful when user is still typing the words and we'd like to advance potential errors.

src/utils/wallet.ts
```diff
@@ -0,0 +1,225 @@
...
+
+        for (const word of wordsArray) {
+          if (wordlist.indexOf(word) < 0) {
+            errorList.push(word);
+          }
+        }
+
+        let errorMessage = '';
+        if (errorList.length > 0) {
+          return {'valid': false, 'invalidWords': errorList};
```
msbrogli, line 54: I'd throw an exception with the `errorList` there. Is it possible?

src/utils/wallet.ts
```diff
@@ -0,0 +1,225 @@
...
+        } else {
+          // Invalid sequence of words
+          throw new InvalidWords('Invalid sequence of words.')
```
msbrogli, line 57: It seems a bug to raise this exception here. I feel that `errorList.length == 0` means that no error has been found, isn't it?

msbrogli (changes requested):

src/utils/wallet.ts
msbrogli: I feel it is weird to sometimes raise an exception and other times return `{'valid': false}`. I agree that we must be consistent in the way to handle errors. I usually prefer to raise exceptions because it stops the execution if the user does not handle it.

msbrogli: I also reviewed the new method and created new discussion points.

src/models/output.ts
pedroferreira1: As we've discussed on a call we will keep the `Output` class as it is, with the `tokenData` field and discuss about how to make it easier for users to create txs/outputs. 

Issue created for further discussion: https://github.com/HathorNetwork/hathor-wallet-lib/issues/159

src/models/transaction.ts
pedroferreira1: Issue created: https://github.com/HathorNetwork/hathor-wallet-lib/issues/160

src/utils/wallet.ts
pedroferreira1: Done

src/models/network.ts
pedroferreira1: Done

src/constants.js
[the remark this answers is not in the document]
pedroferreira1: As we've discussed, I've created an issue to handle this configuration in another PR.

src/models/network.ts
pedroferreira1: Issue created to handle this in another PR: https://github.com/HathorNetwork/hathor-wallet-lib/issues/161

src/utils/wallet.ts
pedroferreira1: Done

src/utils/wallet.ts
pedroferreira1: Actually not. This code block is inside `!Mnemonic.isValid(newWordsString)`, and the `errorList.length == 0` means that all words are valid in the mnemonic dictionary but they are in the wrong order. That's why the error is 'Invalid sequence of words.'.

src/utils/wallet.ts
pedroferreira1: Can you see this commit? https://github.com/HathorNetwork/hathor-wallet-lib/pull/110/commits/e15d98df6aa37801ae854384cf60ecc451e8ca98

I've done this change and the one below about throwing the error.

src/utils/wallet.ts
pedroferreira1: Can you see this commit? https://github.com/HathorNetwork/hathor-wallet-lib/pull/110/commits/e15d98df6aa37801ae854384cf60ecc451e8ca98

It's the same as the other. That's what you thought?

src/utils/wallet.ts
pedroferreira1: I've replied in the other discussions

src/utils/wallet.ts
msbrogli: That's great! Thanks!

src/utils/wallet.ts
msbrogli: I got it. Thanks for the explanation.

msbrogli (approved):

src/utils/wallet.ts
```diff
@@ -0,0 +1,239 @@
...
+   * Validate an xpubkey.
+   *
+   * @param {string} xpubkey The xpubkey
+   *
+   * @return {boolean} true if it's a valid xpubkey, false otherwise
+   * @memberof Wallet
+   * @inner
+   */
+  isXpubKeyValid(xpubkey: string): boolean {
+    try {
```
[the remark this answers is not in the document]
pedroferreira1: Done here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/110/commits/638232982dfa6976ab5460d24519a7fd085ddb74

andreabadesso (approved):

PR #111, opened 2020-11-08
feat: add methods to create token, mint and melt to the new wallet class

pedroferreira1:


src/new/wallet.js
```diff
@@ -511,6 +518,151 @@ class HathorWallet extends EventEmitter {
...
+        continue;
+      }
+
+      for (const [index, output] of tx.outputs.entries()) {
+        // This output is not mine
+        if (!wallet.isAddressMine(output.decoded.address, walletData)) {
+          continue;
+        }
+
+        // This token is not the one of this screen
```
obiyankenobi, line 633: Update comment for lib.

src/new/wallet.js
```diff
@@ -511,6 +518,151 @@ class HathorWallet extends EventEmitter {
...
+        if (output.spent_by) {
+          continue;
+        }
+
+        if (wallet.isMeltOutput(output)) {
+          meltInput = {tx_id, index, address: output.decoded.address};
+          break
+        }
+      }
+    }
```
obiyankenobi, line 648: It might be better to refactor this part as it shares a lot of code with the mint code. So we could just pass the functions `wallet.isMeltOutput` and `wallet. isMintOutput` as params. For instance:
```js
const meltOutput = selectAuthorityUtxo(wallet.isMeltOutput, walletData);
```

obiyankenobi (approved):

src/new/wallet.js
```diff
@@ -511,6 +518,151 @@ class HathorWallet extends EventEmitter {
...
+        continue;
+      }
+
+      for (const [index, output] of tx.outputs.entries()) {
+        // This output is not mine
+        if (!wallet.isAddressMine(output.decoded.address, walletData)) {
+          continue;
+        }
+
+        // This token is not the one of this screen
```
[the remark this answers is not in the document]
pedroferreira1: Done!

src/new/wallet.js
pedroferreira1: Done

src/new/wallet.js
pedroferreira1: Done!

obiyankenobi (approved):

src/new/wallet.js
```diff
@@ -384,22 +384,29 @@ class HathorWallet extends EventEmitter {
    * Currently does not have support to send custom tokens, only HTR
    *
    * @param {Array} outputs Array of outputs with each element as an object with {'address', 'value'}
+   * @param {Object} token Token object {'uid', 'name', 'symbol'}. Optional parameter if user already set on the class
```
obiyankenobi, line 387: Not ideal here that we have to send the full object, instead of just the `uuid`. But I was looking into this and it seems it's required in the end by `wallet.prepareSendTokensData`, even though we just use the `uuid` there also. The name and symbol are only used for log messages. But I guess there's no better way to do it now.

obiyankenobi (approved):

src/new/wallet.js
pedroferreira1: I totally agree with you, my first idea was to expect only the token UID but the lib method expects the full object as you said. I thought about receiving the UID and then sending a request to the full node to get the full data but I thought it would be too much only because of log messages, then I left like this

obiyankenobi: Yeah, I don't think there's much we can do now.

src/new/wallet.js
```diff
@@ -399,27 +399,34 @@ class HathorWallet extends EventEmitter {
...
    * @param {Array} inputs Array of inputs with each element as an object with {'hash', 'index'}
    *
    * @return {Promise} Promise that resolves when transaction is sent
    **/
-  sendManyOutputsTransaction(outputs, inputs = []) {
+  sendManyOutputsTransaction(outputs, token = null, inputs = []) {
```
msbrogli, line 407: I wouldn't change the order of the parameters because of compatibility. I think it is better to put as the last parameter.

src/new/wallet.js
```diff
@@ -531,6 +538,131 @@ class HathorWallet extends EventEmitter {
...
+
+  /**
+   * Select authority utxo for mint or melt. Depends on the callback received as parameter
+   *
+   * @param {String} tokenUid UID of the token to select the authority utxo
+   * @param {function} isUtxoCallback Callback to check if the output is the authority I wanet (isMeltOutput or isMintOutput)
+   *
+   * @return {Object} Object with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
+   **/
+  selectAuthorityUtxo(tokenUid, isUtxoCallback) {
```
msbrogli, line 575: I'd name it `filterUTXOs()`. We could add an `{options}` parameter to allow common filters (e.g. mint authority, melt authority, p2pkh).

msbrogli (changes requested):

src/new/wallet.js
pedroferreira1: Yes, I totally agree with you. I made this confusion in a merge conflict. Fixed it

src/new/wallet.js
pedroferreira1: Do you think we should add this `options` parameter right now? I see it as adding more confusion to the code right now (we would need a map from the option to the callback method).

Anyway, I agree with you that the name `filterUTXOs` is better than `isUtxoCallback`, just want to make sure if I should go with the callback method or the `options` parameter? In my opinion we should keep only the callback, what do you think?

src/new/wallet.js
msbrogli: Thanks! I think it is in use by one of the wallets. Remember to update it there as well, ok?

src/new/wallet.js
msbrogli: We can keep the callback and refactor it later to include more options. Just leave a comment so we know how to improve it later.

msbrogli (approved):

src/new/wallet.js
pedroferreira1: Yes, it was updated!

src/new/wallet.js
pedroferreira1: Done!

PR #115, opened 2020-12-17
feat: add tx mining testnet URL and use it depending on the network selected

pedroferreira1:


obiyankenobi (approved):

jansegre (approved):

src/api/txMiningAxios.js
```diff
@@ -23,7 +24,13 @@ import axiosWrapperCreateRequestInstance from './axiosWrapper';
  * @param {number} timeout Timeout in milliseconds for the request
  */
 const txMiningRequestClient = (resolve, timeout) => {
-  return axiosWrapperCreateRequestInstance(TX_MINING_URL, resolve, timeout);
+  let txMiningURL;
```
msbrogli, line 27: Whoever is using the library should be able to set any custom tx mining service url.

src/constants.js
```diff
@@ -183,3 +183,8 @@ export const MIN_POLLING_INTERVAL = 0.5;
  * Base URL for the tx mining api
  */
 export const TX_MINING_URL = 'https://txmining.mainnet.hathor.network/';
+
+/**
+ * Base URL for the tx mining api in the testnet
+ */
+export const TX_MINING_TESTNET_URL = 'https://txmining.testnet.hathor.network/';
```
msbrogli, line 190: I feel that we should validate the the tx-mining-service is connected to the same network that we are dealing with.

src/api/txMiningAxios.js
pedroferreira1: I agree with you. I think it's a different task from the one in this PR, even though I believe the URL should be customizable. 

I think we can go on with this PR and the other one from the wallet desktop to make it work for both mainnet and testnet only and then start another PR to make it customizable by anyone. What do you think?

PR #120, opened 2020-12-23
feat: accept inputs array as one of the paramers for the sendManyOutputsTransaction method on the new wallet class

pedroferreira1:


jansegre (approved):

src/new/wallet.js
```diff
@@ -409,6 +415,10 @@ class HathorWallet extends EventEmitter {
       data.outputs.push({address: output.address, value: output.value, tokenData: 0})
     }
 
+    for (const input of inputs) {
```
obiyankenobi, line 418: Is `inputs` required now? Won't it break compatibility? Maybe it should be null by default and we should check here the before iterating.

obiyankenobi (changes requested):

src/new/wallet.js
pedroferreira1: It won't break compatibility because it's only used in the headless, which is where we've changed to have the inputs as required. But it won't be a bad idea to have a default of `[]` so I don't need to send inputs when I want them automatically selected.

obiyankenobi (approved):

PR #123, opened 2020-12-24
feat: when the gap limit is increased we use now a POST http API to get the address history

pedroferreira1:


obiyankenobi (approved):

src/wallet.js
```diff
@@ -351,7 +351,15 @@ const wallet = {
     let history = null;
 
     while (hasMore === true) {
-      const response = await walletApi.getAddressHistoryForAwait(addressesToSearch, firstHash);
+      let response;
+      // Using the GET request we end up getting a 414 (URI too large) if we customize the gap limit to a
+      // huge number. I am using the POST API because only new full node versions will have this API, so
+      // it's good to keep compatibility with older behaviour
+      if (addressesToSearch.length > GAP_LIMIT) {
```
jansegre, line 358: This doesn't seem like the correct condition. This `GAP_LIMIT` is always the default gap limit right? In that case it works in practice, but I still think we should either always try to use the POST (and maybe fallback to GET if it returns 404/403) or have add a custom const for this, like `MAX_ADDRESSES_GET`. Otherwise we might change `GAP_LIMIT` and break this for no obvious reasons (I think it might make sense to have a larger `GAP_LIMIT`, like Bitcoin has 1000 by default, even though the BIP suggests 20).

jansegre (approved):

src/wallet.js
pedroferreira1: The idea with this `if` was to keep the old behaviour for common wallets, in case someone does not update the full node (to have the POST API) and update the wallet.

pedroferreira1: So anyone that uses the default `GAP_LIMIT` would still requesting a GET and not a POST

jansegre: I understand the backwards compatibility, but we can do the same if we create a `MAX_ADDRESSES_GET = 20` and use that here. The problem I see is in the future if we want to change `GAP_LIMIT = 100`, then it will break the backwards compatibility instantly, and it won't be obvious.

pedroferreira1: My bad, now I understand your comment about us changing our default `GAP_LIMIT`. It makes sense. Have just pushed a new commit that adds the `MAX_ADDRESSES_GET = 20` constant to be used here.

jansegre (approved):

PR #126, opened 2020-12-31
feat: limit the number of addresses to search on address_history request

pedroferreira1:


src/wallet.js
```diff
@@ -345,65 +345,64 @@ const wallet = {
...
-            throw Error("Invalid address returned from the server.");
+        const result = response.data;
+        let ret = null;
+
+        if (result.success) {
+          hasMore = result.has_more;
+
+          if (hasMore || i !== lastChunkIndex) {
+            // Update storage data with new history
+            ret = this.saveNewHistoryOnStorage(null, null, result.history, undefined, connection, store);
```
obiyankenobi, line 369: I'm wondering if we could just "accumulate" the history from all calls and later just call once to update storage. But I guess we're changing the behavior, so better do it on a different PR.

obiyankenobi (approved):

src/wallet.js
pedroferreira1: I guess this could work and would make the code simpler. But I agree with you that we should keep this PR only for the gap limit change. I will leave a comment in the code with this suggestion

src/wallet.js
```diff
@@ -1970,6 +1979,7 @@ const wallet = {
   addMetricsListener() {
     if (this._connection && this._connection.websocket) {
       this._connection.websocket.on('dashboard', this.handleWebsocketDashboard);
+      this._connection.websocket.on('subscribe_address', this.onSubscribeAddress);
```
msbrogli, line 1982: We also need to unsubscribe but I couldn't find this piece of code.

src/wallet.js
```diff
@@ -1985,6 +1995,21 @@ const wallet = {
...
+   *
+   * @param {Object} data {success, type, message}
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  onSubscribeAddress(data) {
+    // If an address subscription fails, we stop the service
+    if (data.success === false) {
+      throw new Error(data.message)
```
msbrogli, line 2009: What does it do exactly? Log the error? Stop the wallet?

src/websocket.js
```diff
@@ -144,6 +144,11 @@ class WS extends EventEmitter {
     const _type = message.type.split(':')[0]
     if (_type === 'pong') {
       this.onPong();
+    } else {
+      if (this.timeoutTimer) {
+        clearTimeout(this.timeoutTimer);
```
msbrogli, line 149: Add a comment explaining why we need this.

src/wallet.js
```diff
@@ -345,65 +342,72 @@ const wallet = {
...
+
+            if (hasMore) {
+              // If has more data we set the first_hash of the next search
+              // and update the addresses array with only the missing addresses
+              firstHash = result.first_hash;
+              const addrIndex = addressesToSearch.indexOf(result.first_address);
+              if (addrIndex === -1) {
+                throw Error("Invalid address returned from the server.");
+              }
+
```
msbrogli, line 384: Empty line?

msbrogli (changes requested):

src/wallet.js
pedroferreira1: Yes, the wallet is stopped.

src/wallet.js
pedroferreira1: Added that, thanks!

src/wallet.js
pedroferreira1: Removed

src/websocket.js
pedroferreira1: Done

msbrogli (approved):

PR #132, opened 2021-01-21
fix: avoid summing the parents size twice when parents are sent in tx…

andreabadesso:
This is needed by hathor-wallet-service because we need to validate the weight that is being sent via API alongside with the parents.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -434,9 +434,11 @@ const transaction = {
   calculateTxWeight(txData) {
     let txSize = this.txToBytes(txData).length;
 
-    // XXX Parents are calculated only in the server but we need to consider them here
-    // Parents are always two and have 32 bytes each
-    txSize += 64
+    // If parents are not in txData, we need to consider them here
+    if (!txData.parents || !txData.parents.length || txData.parents.length === 0) {
+      // Parents are always two and have 32 bytes each
+      txSize += 64;
+    }
 
     let sumOutputs = this.getOutputsSum(txData.outputs);
     // Preventing division by 0 when handling authority methods that have no outputs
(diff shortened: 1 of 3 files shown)


pedroferreira1 (approved):

src/transaction.js
```diff
@@ -434,9 +434,11 @@ const transaction = {
   calculateTxWeight(txData) {
     let txSize = this.txToBytes(txData).length;
 
-    // XXX Parents are calculated only in the server but we need to consider them here
-    // Parents are always two and have 32 bytes each
-    txSize += 64
+    // If parents are not in txData, we need to consider them here
+    if (!txData.parents || !txData.parents.length || txData.parents.length === 0) {
+      // Parents are always two and have 32 bytes each
+      txSize += 64;
```
msbrogli, line 440: I think we should raise an exception if `txData.parents.length > 2`, and use `txSize += 64 - 32 * txData.parents.length`.

msbrogli, line 440: I think we should raise an exception if `txData.parents.length > 2`, and use `txSize += 64 - 32 * txData.parents.length`.

msbrogli (changes requested):

src/transaction.js
andreabadesso: Done.

Also added a test case for that error.

Should we also throw an error when `parents.length == 1`?

msbrogli: Great! Initially I thought about it but maybe the application will fill the other parent later. As more than two parents is always invalid, I feel it is enough to check this condition. What do you think?

andreabadesso: I agree

A possible use case for single parent validation would be someone wanting to "chain" transactions using a single parent, leaving the other one for automatic selecion

I think this is ready to merge, what do you think?

msbrogli (approved):

src/transaction.js
msbrogli: Yes, it is.

PR #136, opened 2021-02-02
fix: should not erase server when resetting wallet

pedroferreira1:


andreabadesso (approved):

src/wallet.js
```diff
@@ -1122,7 +1122,6 @@ const wallet = {
    */
   resetWalletData() {
     this.cleanWallet();
-    this.cleanServer();
```
msbrogli, line 1125: I feel we should add a test to confirm the server is not being erased.

msbrogli (approved):

src/wallet.js
pedroferreira1: Done!

PR #138, opened 2021-02-03
feat: add change address as optional parameter when selecting utxos of a tx

pedroferreira1:


src/new/wallet.js
```diff
@@ -316,15 +316,16 @@ class HathorWallet extends EventEmitter {
...
    * @param {Object} token Token object {'uid', 'name', 'symbol'}. Optional parameter if user already set on the class
+   * @param {String} changeAddress Optional parameter with address for the change output
    *
    * @return {Object} In case of success, an object with {success: true, sendTransaction, promise}, where sendTransaction is a
    * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
    * In case of error, an object with {success: false, message}
    *
    **/
-  sendTransaction(address, value, token) {
+  sendTransaction(address, value, token, changeAddress = null) {
```
msbrogli, line 326: I'd use a parameters called options that should receive an object. It will allow simple future extension with more options. So, it would be `sendTransaction(address, value, token, { changeAddress = null }) {...}`.

src/new/wallet.js
```diff
@@ -339,10 +340,11 @@ class HathorWallet extends EventEmitter {
    * @param {String} address Address to send the tokens
    * @param {number} value Amount of tokens to be sent
    * @param {Object} token Token object {'uid', 'name', 'symbol'}. Optional parameter if user already set on the class
+   * @param {String} changeAddress Optional parameter with address for the change output
    *
    * @return {Object} Object with {success: false, message} in case of an error, or {success: true, data}, otherwise
    **/
-  prepareTransaction(address, value, token) {
+  prepareTransaction(address, value, token, changeAddress = null) {
```
msbrogli, line 347: Same here.

src/new/wallet.js
```diff
@@ -357,18 +359,19 @@ class HathorWallet extends EventEmitter {
...
    * Complete transaction data with inputs
    *
    * @param {Object} data Partial data that will be completed with inputs
    * @param {Object} token Token object {'uid', 'name', 'symbol'}. Optional parameter if user already set on the class
+   * @param {String} changeAddress Optional parameter with address for the change output
    *
    * @return {Object} Object with 'success' and completed 'data' in case of success, and 'message' in case of error
    **/
-  completeTxData(partialData, token) {
+  completeTxData(partialData, token, changeAddress = null) {
```
msbrogli, line 374: Same here.

src/new/wallet.js
```diff
@@ -401,10 +404,11 @@ class HathorWallet extends EventEmitter {
    * @param {Array} outputs Array of outputs with each element as an object with {'address', 'value'}
    * @param {Array} inputs Array of inputs with each element as an object with {'hash', 'index'}
    * @param {Object} token Token object {'uid', 'name', 'symbol'}. Optional parameter if user already set on the class
+   * @param {String} changeAddress Optional parameter with address for the change output
    *
    * @return {Promise} Promise that resolves when transaction is sent
    **/
-  sendManyOutputsTransaction(outputs, inputs = [], token = null) {
+  sendManyOutputsTransaction(outputs, inputs = [], token = null, changeAddress = null) {
```
msbrogli, line 411: Same here.

src/wallet.js
```diff
@@ -1195,17 +1195,19 @@ const wallet = {
...
+   * @param {String} changeAddress Optional parameter with address for the change output
    *
    * @return {Object} {'address': string, 'value': number, 'tokenData': number, 'isChange': true}
    *
    * @memberof Wallet
    * @inner
    */
-  getOutputChange(value, tokenData) {
-    const address = this.getAddressToUse();
+  getOutputChange(value, tokenData, changeAddress = null) {
```
msbrogli, line 1209: Same here. In this case, I feel we should just call it `address`.

src/wallet.js
```diff
@@ -1498,13 +1500,14 @@ const wallet = {
...
    * @param {Object} Array with all tokens already selected in the send tokens
+   * @param {String} changeAddress Optional parameter with address for the change output
    *
    * @return {Object} {success: boolean, message: error message in case of failure, data: prepared data in case of success}
    *
    * @memberof Wallet
    * @inner
    */
-  prepareSendTokensData(data, token, chooseInputs, historyTransactions, allTokens) {
+  prepareSendTokensData(data, token, chooseInputs, historyTransactions, allTokens, changeAddress = null) {
```
msbrogli, line 1510: Same here.

msbrogli (approved):

src/new/wallet.js
pedroferreira1: Done but the method signature ended up being a bit different.
```
sendTransaction(address, value, token, options = { changeAddress: null }) {
```

What do you think?

src/wallet.js
```diff
@@ -1194,18 +1194,34 @@ const wallet = {
...
-    const address = this.getAddressToUse();
-    return {'address': address, 'value': value, 'tokenData': tokenData, 'isChange': true};
+  getOutputChange(value, tokenData, options = { address: null }) {
+    const { address } = options;
+    if (address) {
+      if (!transaction.isAddressValid(address)) {
+        throw new AddressError('Change address is invalid.');
+      }
+      if (!this.isAddressMine(address)) {
+        throw new AddressError('Change address is not from loaded wallet.');
```
msbrogli, line 1219: Minor thing: The library doesn't know there might be multiple wallets. I'd change the error message to "Change address is not from this wallet.".

msbrogli (approved):

src/new/wallet.js
msbrogli: Great! Thanks!

src/wallet.js
pedroferreira1: I agree. Updated

thr0wn (approved):

PR #141, opened 2021-02-14
feat: mark input as selected as soon as its used, so we dont try to reuse it

pedroferreira1:

PR #144, opened 2021-02-18
feat: utxo consolidation

thr0wn:
- New method `getUtxos` on `new/wallet.js` (get utxos given a set of filter options)
- New method `consolidateUtxos` on `new/wallet.js` (consolidate utxos given a set of filter options)
- New method `filterUtxo` on `wallet.js` (filter or not a utxo given a set of filter options)


src/constants.js
```diff
@@ -106,6 +106,11 @@ export const MAX_OUTPUT_VALUE_32 = 2 ** 31 - 1;
  */
 export const MAX_OUTPUT_VALUE = 2 ** 43;
 
+/**
+ * Max number of inputs in a transaction
+ */
+export const MAX_INPUTS = 255;
```
pedroferreira1, line 112: No need to add this. We already have this value that we get from the server. You can access with `transaction.getMaxInputsConstant()`

src/new/wallet.js
```diff
@@ -184,6 +184,102 @@ class HathorWallet extends EventEmitter {
...
   }
 
+  /**
+   * Get utxos of the wallet addresses
+   *
+   * @param {object} options Consolidation options shared with filterUtxos and consolidateUtxos (https://github.com/HathorNetwork/hathor-wallet-headless/issues/44)
+   *
+   * @return {object} { total_amount_available: number, total_utxos_available: number, total_amount_locked: number, total_utxos_locked: number, utxos: Array[] } utxos and meta information about it
+   */
+  getUtxos(options) {
```
pedroferreira1, line 194: It's important that the first line of this method is `storage.setStore(this.store);` to use the correct store if we have more than one wallet started.

src/new/wallet.js
```diff
@@ -184,6 +184,102 @@ class HathorWallet extends EventEmitter {
...
+
+  /**
+   * Consolidates many utxos into a single one
+   *
+   * @param {string} destinationAddress Address of the consolidated utxos
+   * @param {object} options Consolidation options (https://github.com/HathorNetwork/hathor-wallet-headless/issues/44)
+   *
+   * @return {object} { success: boolean, promise: Promise } indicates that the transaction is sent or not
+   */
+  consolidateUtxos(destinationAddress, options) {
```
pedroferreira1, line 257: It's important that the first line of this method is `storage.setStore(this.store);` to use the correct store if we have more than one wallet started.

src/new/wallet.js
```diff
@@ -184,6 +184,102 @@ class HathorWallet extends EventEmitter {
...
+      // Orphan transactions should be ignored
+      if (transaction.is_voided) {
+        continue;
+      }
+
+      // Iterate through outputs
+      for (let j = 0; j < transaction.outputs.length; j++) {
+        const output = transaction.outputs[j];
+
+        const is_unspent = output.spent_by === null;
```
pedroferreira1, line 217: I think after this line you can already have 

```
if (!is_unspent) continue
```

Because it's already spent, is not an utxo anymore so we don't waste time filtering this output.

src/new/wallet.js
```diff
@@ -184,6 +184,102 @@ class HathorWallet extends EventEmitter {
...
+   * @param {string} destinationAddress Address of the consolidated utxos
+   * @param {object} options Consolidation options (https://github.com/HathorNetwork/hathor-wallet-headless/issues/44)
+   *
+   * @return {object} { success: boolean, promise: Promise } indicates that the transaction is sent or not
+   */
+  consolidateUtxos(destinationAddress, options) {
+    const utxoDetails = this.getUtxos(options);
+    const inputs = [];
+    let value = 0;
+    for (let i = 0; i < utxoDetails.utxos.length; i++) {
```
pedroferreira1, line 261: Inside this `for` you can use only the utxos that are not locked. Maybe we could add a new filter option to add only locked values maybe. What do you think?

src/new/wallet.js
```diff
@@ -184,6 +184,102 @@ class HathorWallet extends EventEmitter {
...
+        index: utxo.index,
+      })
+      value += utxo.amount;
+    }
+    const outputs = [{
+      address: destinationAddress,
+      value,
+    }];
+    const token = {
+      uid: options.token || HATHOR_TOKEN_CONFIG.uid,
```
pedroferreira1, line 277: You need to at least add `name: '', symbol: ''` in this object.

src/wallet.js
```diff
@@ -876,6 +876,51 @@ const wallet = {
...
+      filter_address: null,
+      amount_smaller_than: Infinity,
+      amount_bigger_than: 0,
+      maximum_amount: Infinity
+    }, options);
+
+    // Filter by address, if options.filter_address is specified
+    const is_address_valid = filterOptions.filter_address === null || filterOptions.filter_address === output.decoded.address;
+    // Filter by maximum_amount (sum of utxos amounts), if options.maximum_amount is specified
+    const is_max_amount_valid = filterOptions.maximum_amount >= utxoDetails.total_amount_available;
```
pedroferreira1, line 901: Here you must validate if the `filterOptions.maximum_amount >= utxoDetails.total_amount_available + output.value` because you must validate if this utxo can be added.

src/wallet.js
```diff
@@ -876,6 +876,51 @@ const wallet = {
...
+    // Filter more utxos than options.max_utxos (default: MAX_INPUTS)
+    const is_max_utxos_valid = filterOptions.max_utxos >= utxoDetails.utxos.length;
```
pedroferreira1, line 903: Just like the previous comment. You must check if `filterOptions.max_utxos >= utxoDetails.utxos.length + 1`

pedroferreira1 (changes requested):

__tests__/utxo.test.js
```diff
@@ -0,0 +1,111 @@
...
+  getTxHistory() {
+    return txHistoryFixture;
+  }
+}
+
+describe("UTXO Consolidation", () => {
+  let hathorWallet;
+  const destinationAddress = "WVGxdgZMHkWo2Hdrb1sEFedNdjTXzjvjPi";
+  beforeAll(() => {
+    transaction.updateMaxInputsConstant(255);
```
msbrogli, line 25: Move `255` to a local constant in this file.

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
...
+      total_utxos_locked: 0,
+      utxos: [],
+    };
+
+    // Iterate through transactions
+    for (let i = 0; i < historyTransactions.length; i++) {
+      const transaction = historyTransactions[i];
+
+      // Orphan transactions should be ignored
+      if (transaction.is_voided) continue;
```
msbrogli, line 210: Fix code format:

```javascript
if (...) {
  continue;
}
```

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
...
+      total_amount_locked: 0,
+      total_utxos_locked: 0,
+      utxos: [],
+    };
+
+    // Iterate through transactions
+    for (let i = 0; i < historyTransactions.length; i++) {
+      const transaction = historyTransactions[i];
+
+      // Orphan transactions should be ignored
```
msbrogli, line 209: The correct is "Voided transactions"...

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
     return historyTransactions;
   }
 
+  /**
+   * Get utxos of the wallet addresses
+   *
+   * @param {object} options Consolidation options shared with filterUtxos and consolidateUtxos (https://github.com/HathorNetwork/hathor-wallet-headless/issues/44)
+   *
+   * @return {object} { total_amount_available: number, total_utxos_available: number, total_amount_locked: number, total_utxos_locked: number, utxos: Array[] } utxos and meta information about it
```
msbrogli, line 192: Can we format it with multiple lines? It is hard to read the way it is.

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
     return historyTransactions;
   }
 
+  /**
+   * Get utxos of the wallet addresses
+   *
+   * @param {object} options Consolidation options shared with filterUtxos and consolidateUtxos (https://github.com/HathorNetwork/hathor-wallet-headless/issues/44)
```
msbrogli, line 190: We should document here the options here.

src/wallet.js
```diff
@@ -876,6 +876,51 @@ const wallet = {
     return data;
   },
 
+  /**
+   * Filter an utxo based on the specified options
+   *
+   * @param {object} output Transaction output to be filtered/validated
+   * @param {object} utxoDetails utxos and meta information
+   * @param {object} options Consolidation options shared with getUtxos and consolidateUtxos (https://github.com/HathorNetwork/hathor-wallet-headless/issues/44)
```
msbrogli, line 884: We should document here the options here.

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
...
+          }
+        }
+      }
+    }
+
+    return utxoDetails;
+  }
+
+  /**
+   * Consolidates many utxos into a single one
```
msbrogli, line 254: Suggestion: ... for either HTR or exactly one custom token.

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
...
+      address: destinationAddress,
+      value: total_amount,
+    }];
+    const token = {
+      uid: options.token || HATHOR_TOKEN_CONFIG.uid,
+      name: '',
+      symbol: ''
+    };
+
+    const result = this.sendManyOutputsTransaction(outputs, inputs, token);
```
msbrogli, line 294: Can you refactor and split it this method in two: (i) the first one to build the transaction, and (ii) the second one to send the transaction? It's easier to test and it might be useful for debugging purposes.

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
...
+
+    if (!result.success) {
+      throw new Error(result.message);
+    }
+
+    const tx = await Promise.resolve(result.promise);
```
msbrogli, line 300: Why do you await? I feel that the user should decide whether to await or not. So, I feel that we should return a promise here.

src/wallet.js
```diff
@@ -876,6 +876,51 @@ const wallet = {
     return data;
   },
 
+  /**
+   * Filter an utxo based on the specified options
```
msbrogli, line 880: Add a note of what functions are using it.

msbrogli (changes requested):

src/new/wallet.js
```diff
@@ -184,6 +184,129 @@ class HathorWallet extends EventEmitter {
...
+      // Iterate through outputs
+      for (let j = 0; j < transaction.outputs.length; j++) {
+        const output = transaction.outputs[j];
+
+        const is_unspent = output.spent_by === null;
+        if (!is_unspent) {
+          // No other checks required
+          continue;
+        }
+
```
msbrogli, line 221: Shouldn't we check whether it's been used by another transactions that hasn't been received by the websocket yet (see https://github.com/HathorNetwork/hathor-wallet-lib/pull/141)?

msbrogli (changes requested):

src/new/wallet.js
thr0wn: Following your idea, I added `only_available_utxos` to grab only available (not locked) utxos. So,  `consolidateUtxos` calls `getUtxos` with `{... only_available_utxos: true}`.

__tests__/utxo.test.js
thr0wn: Done

src/new/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: Done

src/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: Done. New method `prepareConsolidateUtxosData` added.

src/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: `async/await` are just syntax sugar for Promises, so someone calling `consolidateUtxos` (which is an async method and already return a promise) can either do:
```js
const promise = hathorWallet.consolidateUtxos(...);
...
// Code that does not wait for utxo consolidation
...
promise.then(...).catch(...);
```
or
```js
const result = await hathorWallet.consolidateUtxos(...);
...
// Code that waits for utxo consolidation
```

If you recommend not proceed with `async/await` here, I can do something more similar to `sendManyOutputsTransaction` that returns a `Promise|{success: false, message: '...'}`. In my opinion, we should use `async/await` in asynchronous code whenever possible, mainly because it is easier to read.

pedroferreira1 (changes requested):

src/new/wallet.js
thr0wn: Hi @msbrogli. Talking with @pedroferreira1 he explained that [`canUseUnspentTx`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/3ab23e174560b6140dedc512995a851739eb2811/src/wallet.js#L997) prevents issues with utxo being used by another transaction. So, as I used `canUseUnspentTx` here, this issue will not happen.

src/new/wallet.js
```diff
@@ -184,6 +184,176 @@ class HathorWallet extends EventEmitter {
...
+        if (!is_unspent || (locked && options.only_available_utxos)) {
+          // No other filtering required
+          continue;
+        }
+
+        const filters = wallet.filterUtxos(output, utxoDetails, options);
+        const is_authority = wallet.isAuthorityOutput(output);
+
+        // Max amount reached or max utxos.length, no more utxo should be added
+        if (!filters.is_max_amount_valid || !filters.is_max_utxos_valid) {
```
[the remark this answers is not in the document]
thr0wn: You are right. Fixed.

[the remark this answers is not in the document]
pedroferreira1: Great!

pedroferreira1 (approved):

src/new/wallet.js
```diff
@@ -184,6 +184,182 @@ class HathorWallet extends EventEmitter {
...
+
+      // Voided transactions should be ignored
+      if (transaction.is_voided) continue;
+
+      // Iterate through outputs
+      for (let j = 0; j < transaction.outputs.length; j++) {
+        const output = transaction.outputs[j];
+
+        const is_unspent = output.spent_by === null;
+        const locked = !wallet.canUseUnspentTx(output, j);
```
pedroferreira1, line 234: This second parameter is the transaction height, in case of a block. Why is it `j` here?

pedroferreira1 (changes requested):

src/new/wallet.js
msbrogli: Great! I think we should add a note explaining that `canUseUnspentTx()` already handles this case.

src/new/wallet.js
thr0wn: Fixed. Now I'm using `transaction.height`.

src/new/wallet.js
msbrogli: I agree that this syntax is way better to read and less error prone. Thanks!

src/new/wallet.js
```diff
@@ -184,6 +184,182 @@ class HathorWallet extends EventEmitter {
...
+  }
+
+  /**
+   * Consolidates many utxos into a single one for either HTR or exactly one custom token.
+   *
+   * @typedef {Object} ConsolidationResult
+   * @property {number} total_utxos_consolidated - Number of utxos consolidated
+   * @property {number} total_amount - Consolidated amount
+   * @property {number} tx_id - Consolidated transaction id
+   * @property {{ address: string, amount: number, tx_id: string, locked: boolean, index: number }[]} utxos - Array of utxos
```
msbrogli, line 331: Only the utxos used in this consolidation, right?

msbrogli (approved):

pedroferreira1 (approved):

src/new/wallet.js
thr0wn: Done

src/new/wallet.js
thr0wn: Yes. I adjusted the docs to clarify it is an array of **consolidated** utxos, thanks!

src/new/wallet.js
```diff
@@ -184,6 +184,183 @@ class HathorWallet extends EventEmitter {
...
+   *
+   * @param {string} destinationAddress Address of the consolidated utxos
+   * @param {UtxoOptions} options Utxo filtering options
+   *
+   * @return {Promise<ConsolidationResult>} Indicates that the transaction is sent or not
+   *
+   */
+  async consolidateUtxos(destinationAddress, options = {}) {
+    storage.setStore(this.store);
+    const { outputs, inputs, token, utxos, total_amount } = this.prepareConsolidateUtxosData(destinationAddress, options);
```
pedroferreira1, line 342: Before `prepareConsolidateUtxosData` we should validate that `destinationAddress` is from the same wallet.

thr0wn: Done, thanks.

PR #147, opened 2021-02-24
fix: handle special case where we could be reloading wallet data while we expect to update a key on storage

pedroferreira1:


src/new/sendTransaction.js
```diff
@@ -187,7 +187,11 @@ class SendTransaction extends EventEmitter {
 
     // Before sending the tx to be mined, we iterate through the inputs and set selected_as_input
     for (const input of this.data.inputs) {
-      historyTransactions[input.tx_id].outputs[input.index]['selected_as_input'] = selected;
+      if (input.tx_id in historyTransactions) {
+        historyTransactions[input.tx_id].outputs[input.index]['selected_as_input'] = selected;
+      } else {
+        console.log(`updateOutputSelected: Error updating output as selected=${selected}. ${input.tx_id} is not in the storage data. Transactions history length: ${Object.values(historyTransactions).length}`);
```
msbrogli, line 193: Maybe we should add more info such as "WS Connection Uptime".

src/wallet.js
```diff
@@ -1554,7 +1554,7 @@ const wallet = {
 
       if (newData.inputsAmount < outputsAmount) {
         // Don't have this amount of token
-        return {success: false, message:  `Token ${token.symbol}: Insufficient amount of tokens`};
+        return {success: false, message:  `Token ${token.symbol}: Insufficient amount of tokens. You have a total amount of ${newData.inputsAmount} distributed in ${newData.inputs.length} utxos and you need a total amount of ${outputsAmount}.`};
```
msbrogli, line 1557: Maybe add a `debug` field with several fields to debug?

msbrogli (changes requested):

msbrogli (approved):

PR #154, opened 2021-02-28
feat: new method to get a transaction from the wallet

pedroferreira1:


src/new/wallet.js
```diff
@@ -205,6 +205,22 @@ class HathorWallet extends EventEmitter {
     return historyTransactions;
   }
 
+  /**
+   * Get a transaction data from the wallet
+   *
+   * @param {String} id Hash of the transaction to get data from
+   *
+   * @return {Object} Data from the transaction to get. Can be null if the wallet does not contain the tx.
```
msbrogli, line 213: Maybe add here which pieces of data are provided?

msbrogli (approved):

src/new/wallet.js
pedroferreira1: Done

PR #162, opened 2021-03-05
chore: github workflow

pedroferreira1:


andreabadesso (changes requested):
Maybe also add a step to code lint?

pedroferreira1:
> Maybe also add a step to code lint?

We have no linter in the lib right now. But it's something we should add for sure in the future, then we can add to the checks

jansegre (approved):
Looks good, I'd recommend making a fork and pushing it to the dev branch to test a run.

msbrogli (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
...
+    branches:
+    - dev
+    - master
+jobs:
+  test:
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 40  # default is 360
+    strategy:
+      matrix:
+        node-version: [8.x]
```
msbrogli, line 19: Add `10.x`?

PR #163, opened 2021-03-05
Tests now work for v8, v10, v12 and v14

pedroferreira1:


msbrogli (approved):

__tests__/tokens_utils.js
```diff
@@ -138,6 +138,8 @@ test('New token', async (done) => {
   const ret = tokens.createToken(address, tokenName, tokenSymbol, 200, pin);
   const sendTransaction = ret.sendTransaction;
   sendTransaction.start();
+  // We must do this, otherwise the setTimeout method will be called with a null localStorage
+  sendTransaction.updateOutputSelected(false);
```
msbrogli, line 142: Is it related to the different versions of nodejs?

msbrogli (approved):

__tests__/tokens_utils.js
pedroferreira1: It's not. Without this the test was not failing but an async exception is raised because of a `setTimeout` that is called after the test is finished. This prevents this `setTimeout` from being called.

msbrogli: Ok. Thanks for the clarification.

PR #164, opened 2021-03-09
fix: Handle promise exception in wallet.getTxHistory

msbrogli:


jansegre (approved):

src/new/wallet.js
```diff
@@ -718,7 +719,10 @@ class HathorWallet extends EventEmitter {
     this.conn.on('state', this.onConnectionChangedState);
     this.conn.on('wallet-update', this.handleWebsocketMsg);
 
-    wallet.executeGenerateWallet(this.seed, this.passphrase, this.pinCode, this.password, false);
+    const ret = wallet.executeGenerateWallet(this.seed, this.passphrase, this.pinCode, this.password, false);
+    if (ret !== null) {
+      throw "What?!";
```
pedroferreira1, line 724: This was used for debug right? I think we should delete it or throw a nicer error.

pedroferreira1 (approved):

src/new/wallet.js
msbrogli: Fixed. Thanks!

pedroferreira1 (approved):

jansegre (approved):

PR #172, opened 2021-03-12
feat: create method to get the index of an address

pedroferreira1:


jansegre (approved):

src/wallet.js
```diff
@@ -1972,6 +1972,26 @@ const wallet = {
...
+   * @param {string} address Address to check
+   *
+   * @return {Number | null}
+   * @memberof Wallet
+   * @inner
+   */
+  getAddressIndex(address) {
+    const data = this.getWalletData();
+
+    if (this.isAddressMine(address, data)) {
```
msbrogli, line 1988: Don't all addresses in `data.keys` belong to the wallet?

msbrogli (approved):

src/wallet.js
pedroferreira1: Yes. I just verify this to safely access `data.keys[address]`. Because if `!this.isAddressIsMine`, I don't have the key `address` in `data.keys`

msbrogli: Ok. I just checked the `isAddressMine()` code and it is very simple. Thanks!

PR #176, opened 2021-03-19
feat: improve lib usability for other applications with more events being emitted and fixed stop not properly cleaning wallet storage

pedroferreira1:


andreabadesso (approved):

src/new/sendTransaction.js
```diff
@@ -193,6 +193,11 @@ class SendTransaction extends EventEmitter {
     }
 
     const walletData = wallet.getWalletData();
+    if (walletData === null) {
```
msbrogli, line 196: It is a race condition, right? Nice catch! What if a new wallet is started in the between? Would it add this transaction to the history of the wrong wallet?

src/new/wallet.js
```diff
@@ -159,9 +168,14 @@ class HathorWallet extends EventEmitter {
         console.error('Error loading wallet', {error});
       })
     } else {
-      // CONNECTING or CLOSED?
       this.serverInfo = null;
-      this.setState(HathorWallet.CONNECTING);
+      if (this.firstConnection) {
+        // If firstConnection = true, then a stop() was executed
```
msbrogli, line 173: Why? What if the connection fails for a while, then it reconnects, and finally we call a stop()?

src/new/connection.js
```diff
@@ -97,6 +98,18 @@ class Connection extends EventEmitter {
     this.websocket.on('is_online', this.onConnectionChange);
     this.websocket.on('wallet', this.handleWalletMessage);
 
+    this.websocket.on('reload_data', () => {
+      this.emit('reload-data');
+    });
+
+    this.websocket.on('height_updated', (height) => {
+      this.emit('height-updated', height);
```
msbrogli, line 106: I'd rename the event to `best-block-update`.

src/new/connection.js
```diff
@@ -69,6 +69,7 @@ class Connection extends EventEmitter {
     } else {
       this.setState(Connection.CONNECTING);
     }
+    this.emit('is-online', value);
```
msbrogli, line 72: Why do we need this event? Can't the app use the `state` event?

src/new/connection.js
```diff
@@ -97,6 +98,18 @@ class Connection extends EventEmitter {
...
+    this.websocket.on('reload_data', () => {
+      this.emit('reload-data');
+    });
+
+    this.websocket.on('height_updated', (height) => {
+      this.emit('height-updated', height);
+    });
+
+    this.websocket.on('addresses_loaded', (data) => {
+      this.emit('addresses-loaded', data);
```
msbrogli, line 110: Why do we have to propagate this event? What does it mean for the app?

src/new/connection.js
```diff
@@ -97,6 +98,18 @@ class Connection extends EventEmitter {
     this.websocket.on('is_online', this.onConnectionChange);
     this.websocket.on('wallet', this.handleWalletMessage);
 
+    this.websocket.on('reload_data', () => {
+      this.emit('reload-data');
```
msbrogli, line 102: Why do we have to propagate this event? What does it mean for the app?

msbrogli (changes requested):

src/new/connection.js
pedroferreira1: This event is not needed. It was useless, then I added the `beforeReload` callback

src/new/sendTransaction.js
pedroferreira1: Yeah, this is a condition I caught in the app when I send the transaction, it succeeds, then I reset the wallet before the `setTimeout` is triggered. Then the `walletData` would be `null`.

We have this `if (input.tx_id in historyTransactions)` and if it's the wrong wallet we wouldn't set this value.

src/new/connection.js
pedroferreira1: Makes sense, done.

src/new/connection.js
pedroferreira1: Yes, it can. Makes sense

src/new/wallet.js
pedroferreira1: When we call a stop we set `firstConnection = true`. That's why I created this `if`. Before we were changing to 'CONNECTING', then to 'CLOSED' and depending on a race condition it could change to 'CLOSED' first and then 'CONNECTING', which would be wrong because it should be only 'CLOSED'.

src/new/sendTransaction.js
msbrogli: Don't we block the UX until the transaction is broadcasted to the network?

src/new/connection.js
msbrogli: I got it. Shouldn't we rename it to something like `wallet-load-partial-update`?

src/new/wallet.js
msbrogli: I still don't understand why we use the `firstConnection` to differentiate both situations. Isn't it better to let the `.stop()` mark a flag (e.g. `stopRequested`) and check it here?

msbrogli (approved):

src/new/sendTransaction.js
pedroferreira1: Yeah, but the `setTimeout` is execute 60s after the tx is sent to be mined. We could clear the `setTimeout` on success as well. But I would keep this protection anyway to prevent any race condition.

src/new/connection.js
pedroferreira1: Done

src/new/wallet.js
pedroferreira1: Makes sense, I created `walletStopped` flag to check in this `if`

src/new/sendTransaction.js
msbrogli: Ok. I agree to keep this protection here but we definitely should clear the `setTimeout` on success.

pedroferreira1: Done

msbrogli (approved):

PR #177, opened 2021-03-23
 feat: getAddresses method should expect the account xpub derivation and receive the changeDerivation

pedroferreira1:


src/helpers.js
```diff
@@ -405,9 +406,12 @@ const helpers = {
       network.setNetwork(networkName);
     }
 
+    // The xpub is in the account derivation step, we must derive until the change path
+    const derivedXpub = xpub.derive(changeDerivation);
```
msbrogli, line 410: I don't get why we need to derive here.

src/utils/wallet.ts
```diff
@@ -211,7 +212,7 @@ const wallet = {
    * @memberof Wallet
    * @inner
    */
-  getAddresses(xpubkey: string, startIndex: number, quantity: number, networkName: string = 'mainnet'): Object {
+  getAddresses(xpubkey: string, changeDerivation: number, startIndex: number, quantity: number, networkName: string = 'mainnet'): Object {
```
msbrogli, line 215: I feel I've already seen a similar change in this PR. Why do we need the `changeDerivation` here?

msbrogli (changes requested):
I don't understand what these changes are doing and why they are useful.

pedroferreira1:
This PR is not needed anymore, as discussed privately.

PR #178, opened 2021-03-24
fix: fix tx weight calculus when receiving parents in txData

pedroferreira1:


andreabadesso (approved):

src/transaction.js
```diff
@@ -440,7 +440,9 @@ const transaction = {
       throw new MaximumNumberParentsError(`Transaction has ${txData.parents.length} parents and can have at most 2.`);
     }
 
-    txSize += 64 - (32 * (get(txData, 'parents') || 0));
+    // Parents may not be sent in txData
+    const parents = get(txData, 'parents');
```
msbrogli, line 444: Why don't we do `const parents = txData.parents || []`?

msbrogli (changes requested):

src/transaction.js
pedroferreira1: Using `get` we prevent any error thrown if `txData` is `null` or `undefined`. We are not adding any new dependencies to the project, so I think it's ok to use the `get`.

msbrogli: Ok. But we can still use `|| []` to make sure it is never null, right? Using this we don't have to use the conditional on the line below.

msbrogli (approved):

src/transaction.js
pedroferreira1: Makes sense. Added it

PR #179, opened 2021-03-24
feat: create methods to derive an address at specific index and derive xpub child

pedroferreira1:


src/utils/wallet.ts
```diff
@@ -228,6 +228,37 @@ const wallet = {
...
+   * @param {number} changeDerivationIndex Change derivation index
+   * @param {number} addressIndex Index of address to generate
+   * @param {string} networkName 'mainnet' or 'testnet'
+   *
+   * @return {string} Address at the requested index
+   * @throws {XPubError} In case the given xpub key is invalid
+   * @memberof Wallet
+   * @inner
+   */
+  getAddressAtIndex(xpubkey: string, changeDerivationIndex: number, addressIndex: number, networkName: string = 'mainnet'): Object {
```
msbrogli, line 246: Same thing about the `changeDerivationIndex` we discussed today.

msbrogli (changes requested):

src/utils/wallet.ts
pedroferreira1: Yes, changed it.

andreabadesso (approved):

src/utils/wallet.ts
```diff
@@ -228,6 +228,57 @@ const wallet = {
       addrMap[address.toString()] = index;
     }
     return addrMap;
+  },
+
+  /**
+   * Get Hathor address at specific index
+   *
+   * @param {string} xpubkey The xpubkey. We expect the xpub in the last derivation path (change).
```
msbrogli, line 236: Suggestion: "The xpubkey in the last derivation path (change level according to BIP0044)".

msbrogli (approved):

Issue #161, opened 2021-03-04
Create an easy way to init/edit configuration variables


pedroferreira1:
We should make it easy for the user to init the lib from a server, from the mainnet/testnet default (offline), from a custom set of configurations.

Then we would have some constants, e.g. `HATHOR_BIP44_CODE` and others that can be edited, e.g. `MAX_OUTPUTS`, `GAP_LIMIT`, ...

Discussion in PR: https://github.com/HathorNetwork/hathor-wallet-lib/pull/110#discussion_r574666976

luislhl:
I think we fixed this in https://github.com/HathorNetwork/hathor-wallet-lib/pull/265

Do you think there is any additional change needed for this issue, @pedroferreira1 ?

PR #184, opened 2021-04-06
feat: wallet service integration

pedroferreira1:


src/models/transaction.ts
```diff
@@ -108,7 +108,7 @@ class Transaction {
     arr.push(helpers.intToBytes(this.version, 2))
 
     // Len tokens
-    if (this.tokens.length) {
+    if (this.version !== CREATE_TOKEN_TX_VERSION) {
```
msbrogli, line 111: I feel we shouldn't mix both cases in a single serialization code. Isn't it better to have the following methods: `blockSerialization()`, `txSerialization()`, `createTokenTxSerialization()`? They could call helpers to prevent code duplication.

src/models/transaction.ts
```diff
@@ -251,9 +251,7 @@ class Transaction {
     // Weight is a float with 8 bytes
     arr.push(helpers.floatToBytes(this.weight, 8));
     // Timestamp
-    if (this.timestamp) {
-      arr.push(helpers.intToBytes(this.timestamp, 4))
-    }
+    arr.push(helpers.intToBytes(this.timestamp!, 4))
```
msbrogli, line 254: `this.timetamp!`? What's the exclamation mark at the end?

src/models/transaction.ts
```diff
@@ -368,6 +366,17 @@ class Transaction {
...
   }
+
+  /**
+   * Set tx timestamp and weight
+   *
+   * @memberof Transaction
+   * @inner
+   */
+  prepareToSend() {
+    this.timestamp = Math.floor(Date.now() / 1000);
```
msbrogli, line 377: Maybe move to a separate method (e.g. `updateTimestamp()`) and call it from here?

src/utils/wallet.ts
```diff
@@ -176,6 +176,29 @@ const wallet = {
...
+  getXPubKeyFromSeed(seed: string, options: { passphrase?: string, networkName?: string, accountDerivationIndex?: string } = {}): string {
+    const methodOptions = Object.assign({passphrase: '', networkName: 'mainnet', accountDerivationIndex: '0\''}, options);
+    const { passphrase, networkName, accountDerivationIndex } = methodOptions;
+
+    const network = new Network(networkName);
+    const code = new Mnemonic(seed);
+    const xpriv = code.toHDPrivateKey(passphrase, network.bitcoreNetwork);
+    // We have a fixed derivation until the coin index
+    // after that we can receive a different account index, which the default is 0'
+    const privkey = xpriv.derive(`m/44'/${HATHOR_BIP44_CODE}'/${accountDerivationIndex}`);
```
msbrogli, line 198: I think we are doing two things here. Maybe we should split this method into `getXPrivKeyFromSeed` and call this new method from here. What do you think?

src/wallet/api/walletApi.ts
```diff
@@ -0,0 +1,62 @@
...
+import Input from '../../models/input';
+
+/**
+ * Api calls for wallet
+ *
+ * @namespace ApiWallet
+ */
+
+const walletApi = {
+  getWalletStatus(id: string) {
```
msbrogli, line 19: Shouldn't we put the type of the expected return? My comment relates to all methods in this file.

src/wallet/mineTransaction.ts
```diff
@@ -0,0 +1,161 @@
...
+ * - Get back mining response;
+ *
+ * It emits the following events:
+ * 'job-submitted': after job was submitted;
+ * 'estimation-updated': after getting the job status;
+ * 'job-done': after job is finished;
+ * 'error': if an error happens;
+ * 'unexpected-error': if an unexpected error happens;
+ **/
+class MineTransaction extends EventEmitter {
```
msbrogli, line 43: Duplicate code?

src/wallet/wallet.ts
```diff
@@ -0,0 +1,398 @@
...
+// if it ended loading and became ready
+const WALLET_STATUS_POLLING_TIMEOUT = 3000;
+
+enum walletState {
+  NOT_STARTED = 'Not started',
+  LOADING = 'Loading',
+  READY = 'Ready',
+}
+
+class HathorWallet extends EventEmitter {
```
msbrogli, line 32: Maybe rename to `HathorWalletServiceWallet`?

msbrogli (changes requested):

src/wallet/mineTransaction.ts
pedroferreira1: The idea was to migrate the `SendTransaction` class to typescript, then I adapted the code a bit for the new wallet service requirements.

src/models/transaction.ts
pedroferreira1: The attribute `timestamp` is optional in the constructor (`null` by default). And `intToBytes` can't receive `null`, then I need to use the `!` to tell the linter that timestamp won't be null in this moment.

src/wallet/api/walletApi.ts
pedroferreira1: I've added `Promise<any>` to all of the methods but in the new branch of the signature methods I've created new types for each API return and will add to this file correctly when I continue that branch

src/utils/wallet.ts
pedroferreira1: I agree. I created a method to get the xpriv of the root, and then in this method I derive. What do you think?

src/models/transaction.ts
pedroferreira1: Block and transactions can be handled the same way but I agree we have some custom serializations for the create token tx. Should I create a class `CreateTokenTransaction` that extends this `Transaction` class instead of having three different methods in the same class?

msbrogli: I guess so. Even blocks and transactions might differ significantly if we serialize MergeMinedBlocks for instance.

src/models/transaction.ts
msbrogli: Ok. Thanks!

src/utils/wallet.ts
msbrogli: Great. Thanks!

src/wallet/api/walletApi.ts
msbrogli: Ok. Thanks!

src/wallet/mineTransaction.ts
msbrogli: Can't we make one method call the other and remove duplication?

pedroferreira1: The idea is that the old class will be deleted eventually.

src/constants.js
```diff
@@ -222,3 +222,8 @@ export const TX_MINING_TESTNET_URL = 'https://txmining.testnet.hathor.network/';
  * Timeout in milliseconds to call the method to set all selected outputs of a tx as 'selected': false
  */
 export const SELECT_OUTPUTS_TIMEOUT = 1000 * 60;
+
+/**
+ * Wallet service URL
+ */
+export const WALLET_SERVICE_BASE_URL  = 'https://dol32y58xi.execute-api.us-east-1.amazonaws.com/production/';
```
andreabadesso, line 229: Maybe load from env variables?

andreabadesso (changes requested):

src/wallet/mineTransaction.ts
pedroferreira1: Done here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/184/commits/552aa00899aab7cc45b80bfac8254bcdbac7c335

src/models/transaction.ts
pedroferreira1: Done here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/184/commits/552aa00899aab7cc45b80bfac8254bcdbac7c335

src/constants.js
pedroferreira1: We currently have an issue (https://github.com/HathorNetwork/hathor-wallet-lib/issues/161) to improve the way we handle configurations in the wallet. The idea is that we could customize many constants that may depend on the network and we could use env vars to select a custom config file, for example.

I think this comment is valid but it's more complex than just that, so I believe we could keep the current implementation (like we have for the other configurations) for now and focus on that when we implement this issue, what do you think?

src/models/transaction.ts
msbrogli: Great! Thanks!

src/constants.js
msbrogli: I just feel that we should use a custom DNS instead of AWS's random DNS.

andreabadesso: I agree, a custom DNS should be enough

pedroferreira1: Added a TODO in the constants to change this and created an issue in the wallet service repository to create the DNS.

src/wallet/mineTransaction.ts
pedroferreira1: Can you see these two commits? 
https://github.com/HathorNetwork/hathor-wallet-lib/pull/184/commits/8b2f4cd5ffdbccfc5924834ba4f3590cc1ba96dd
https://github.com/HathorNetwork/hathor-wallet-lib/pull/184/commits/faa4e9c82833580f95218a1aa9d7d85bedbeb2b1

The first is the refactor to remove duplicate code and the second is because the first one was not working for create token tx.

msbrogli: Awesome! Thanks!

msbrogli (approved):

src/constants.js
```diff
@@ -226,5 +226,4 @@ export const SELECT_OUTPUTS_TIMEOUT = 1000 * 60;
 /**
  * Wallet service URL
  */
-// TODO Add a DNS for this domain (we should have one DNS for the testnet and another one for the mainnet)
-export const WALLET_SERVICE_BASE_URL  = 'https://dol32y58xi.execute-api.us-east-1.amazonaws.com/production/';
+export const WALLET_SERVICE_BASE_URL  = 'https://wallet-service.testnet.hathor.network/';
```
andreabadesso, line 229: LGTM

andreabadesso (approved):

PR #186, opened 2021-04-23
feat: changes to support wallet desktop integration with lib facade

pedroferreira1:


src/new/wallet.js
```diff
@@ -875,6 +962,20 @@ class HathorWallet extends EventEmitter {
     return promise;
   }
 
+  changeConnection(connection) {
```
msbrogli, line 965: What is it used for?

src/wallet.js
```diff
@@ -140,27 +140,37 @@ const wallet = {
...
    */
-  executeGenerateWalletFromXPriv(xpriv_root, pin, loadHistory) {
-    let xpriv = HDPrivateKey(xpriv_root);
-    let privkey = xpriv.derive(`m/44'/${HATHOR_BIP44_CODE}'/0'/0`);
+  executeGenerateWalletFromXPriv(xprivkey, pin, loadHistory) {
+    const xpriv = HDPrivateKey(xprivkey);
+    let initialAccessData;
+    let privkey;
+    if (xpriv.depth === 0) {
+      // TODO test if it's working for root xpriv
```
msbrogli, line 156: TODO?!

src/wallet.js
```diff
@@ -1192,8 +1202,11 @@ const wallet = {
    * @memberof Wallet
    * @inner
    */
-  cleanLoadedData() {
-    storage.removeItem('wallet:accessData');
+  cleanLoadedData({ cleanAccessData = true } = {}) {
+    // TODO Test with and without sending
```
msbrogli, line 1206: TODO?!

msbrogli (changes requested):

src/new/wallet.js
pedroferreira1: This is used when we change the server in the wallet desktop. In this case I can keep the same `HathorWallet` object but change the `Connection` object, stopping the old conn and starting the new one.

src/wallet.js
pedroferreira1: Created a unit test for this and removed the TODO

src/wallet.js
pedroferreira1: This already had a unit test. Removed the TODO

src/wallet.js
msbrogli: Great. Thanks!

src/new/wallet.js
msbrogli: I'm wondering why wouldn't we create a new object? I mean, so many things might go wrong that I feel we should just create a new `HathorWallet` instance.

pedroferreira1: I think you are right, there is no reason not to create a new object. Will do that.

pedroferreira1: Removed the method, is not needed anymore.

src/new/wallet.js
```diff
@@ -953,16 +1048,43 @@ class HathorWallet extends EventEmitter {
...
         }
 
-        // If output was already used, we can't use it
+        const ret = {tx_id, index, address: output.decoded.address};
+        // If output was already used, we can't use it, unless requested in options
         if (output.spent_by) {
-          continue;
+          if (skipSpent) {
+            continue;
+          } else {
```
msbrogli, line 1056: You don't need the else block here.

src/new/wallet.js
```diff
@@ -930,10 +1021,14 @@ class HathorWallet extends EventEmitter {
    *
    * @param {String} tokenUid UID of the token to select the authority utxo
    * @param {function} filterUTXOs Callback to check if the output is the authority I want (isMeltOutput or isMintOutput)
+   * @param {Object} options Object with custom options. 'many' if should return many utxos or just one (default false) and 'skipSpent' if should not include spent utxos (default true)
```
msbrogli, line 1024: Too long. Maybe put each option in a separate line?

src/new/wallet.js
```diff
@@ -953,16 +1048,43 @@ class HathorWallet extends EventEmitter {
...
         }
       }
     }
+
+    if (many) {
+      return utxos;
+    }
+  }
+
+  getMintAuthority(tokenUid, options = { many: false, skipSpent: true }) {
```
msbrogli, line 1082: Missing docstring.

src/new/wallet.js
```diff
@@ -953,16 +1048,43 @@ class HathorWallet extends EventEmitter {
...
+    return this.selectAuthorityUtxo(tokenUid, wallet.isMintOutput.bind(wallet), options);
+  }
+
+  getMeltAuthority(tokenUid, options = { many: false, skipSpent: true }) {
```
msbrogli, line 1086: Missing docstring.

src/new/wallet.js
msbrogli, line 1082: I fell you should let options be undefined. I see that the defaults here are the same as in `this.selectAuthorityUtxo` but weird behaviors might happen if the defaults are different.

src/new/wallet.js
msbrogli, line 1086: I fell you should let options be undefined. I see that the defaults here are the same as in `this.selectAuthorityUtxo` but weird behaviors might happen if the defaults are different.

src/new/wallet.js
```diff
@@ -1008,30 +1140,165 @@ class HathorWallet extends EventEmitter {
...
+
+    if (destroyInputs.length < count) {
+      return {success: false, message: `Don't have enough authority output available to destroy. Have ${destroyInputs.length} and requested ${count}.`}
+    }
+
+    const data = [];
+    for (const utxo of destroyInputs) {
+      const { tx_id, address, index } = utxo;
+      data.push({ tx_id, address, index, token: tokenUid });
+      // Even though count is expected as a number, I am using ==
```
msbrogli, line 1261: I think it is best to use `>=` and avoid infinite loops. We can do a type check at the beginning of the method to assert it has the right type.

src/new/wallet.js
```diff
@@ -1093,12 +1360,17 @@ class HathorWallet extends EventEmitter {
    *
    * @return {Object} Object with each token and it's balance in this tx for this wallet
    **/
-  getTxBalance(tx) {
+  getTxBalance(tx, options = { includeAuthorities: false }) {
```
msbrogli, line 1363: I think it is safer to always use `Object.assign(default, options)`.

src/new/wallet.js
```diff
@@ -1125,6 +1402,34 @@ class HathorWallet extends EventEmitter {
...
+   * The address might be in the input or output
+   * Removes duplicates
+   *
+   * @param {Object} tx Transaction data with array of inputs and outputs
+   *
+   * @return {Set} Set of strings with addresses
+   **/
+  getTxAddresses(tx) {
+    storage.setStore(this.store);
+    const walletAddresses = this.getAllAddresses();
```
msbrogli, line 1417: What's the type of `walletAddresses`? If it is a list, it might get really long and the searches inside the loops might be too expensive.

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
+      } else {
+        if (tokens.length === 0) {
+          // HTR
+          token = HATHOR_TOKEN_CONFIG.uid;
+        } else {
+          token = input.token;
+        }
+      }
+
+      tokensData[input.token].inputs = [...tokensData[input.token].inputs, {
```
msbrogli, line 844: Why do we recreate the list when we could just push a new item?

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
+    const historyTxs = 'historyTransactions' in walletData ? walletData.historyTransactions : {};
+    for (const tokenUid in tokensData) {
+      // For each token key in tokensData we prepare the data
+      const partialData = tokensData[tokenUid];
+      let chooseInputs = true;
+      if (partialData.inputs.length > 0) {
+        chooseInputs = false;
+      }
+
+      // Warning: prepareSendTokensData(...) might modify `partialData`.
```
msbrogli, line 863: What changes might occur?

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
    *  }
    *
    * @return {Promise} Promise that resolves when transaction is sent
    **/
-  sendManyOutputsTransaction(outputs, inputs = [], token = null, options = { changeAddress: null }) {
-    // XXX To accept here multi tokens in the same tx would be a bit more complicated because
-    // the method prepareSendTokensData would need a bigger refactor.
-    // I believe we should refactor all of that code (and it will be done on wallet-service)
-    // so I decided to change this method to accept only one token, which is enough for our needs right now
+  sendManyOutputsTransaction(outputs, inputs = [], token = null, options = {}) {
```
msbrogli, line 779: Maybe rename to `defaultToken`?

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
+    // then the first step is to separate the inputs/outputs for each token
+    const defaultData = { outputs: [], inputs: [] };
+    if (txToken && txToken.uid !== HATHOR_TOKEN_CONFIG.uid) {
+      tokensData[txToken.uid] = Object.assign({}, defaultData);
+      tokens.push(txToken.uid);
+    } else {
+      for (const output of outputs) {
+        if (output.token && output.token !== HATHOR_TOKEN_CONFIG.uid) {
+          if (!(output.token in tokensData)) {
+            tokens.push(output.token);
```
msbrogli, line 796: Isn't it better to remove the variable `tokens` and, after this loop, generate it from the keys of `tokensData`?

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
+            tokensData[output.token] = Object.assign({}, defaultData);
+          }
+        }
+      }
+
+      if (tokens.length === 0) {
+        // So it's HTR
+        tokensData[HATHOR_TOKEN_CONFIG.uid] = Object.assign({}, defaultData);
```
msbrogli, line 804: Maybe we should always have the HTR token in `tokensData`. It might avoid many if blocks in this method.

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
+      }
+    }
 
     for (const output of outputs) {
-      data.outputs.push({address: output.address, value: output.value, tokenData: isHathorToken ? 0 : 1})
+      let tokenData;
+      if (txToken && txToken.uid !== HATHOR_TOKEN_CONFIG.uid) {
```
msbrogli, line 810: Why does it override the `output.token`? I feel that, if `output.token` exists, it should be used no matter what.

src/new/wallet.js
```diff
@@ -765,44 +765,128 @@ class HathorWallet extends EventEmitter {
...
+        value: output.value,
+        timelock: output.timelock ? output.timelock : null,
+        tokenData,
+      }];
     }
 
     for (const input of inputs) {
-      data.inputs.push({tx_id: input.hash, index: input.index, token: HATHOR_TOKEN_CONFIG.uid });
+      let token;
+      if (txToken && txToken.uid !== HATHOR_TOKEN_CONFIG.uid) {
```
msbrogli, line 832: Why does it override `input.token`? If it exists, shouldn't it be always used?

msbrogli (changes requested):

src/new/wallet.js
pedroferreira1: I'm not sure I understand what do you mean by "let options be undefined".

I changed 
```
getMintAuthority(tokenUid, options = { many: false, skipSpent: true }) {
```
To
```
  getMintAuthority(tokenUid, options = {}) {
    const newOptions = Object.assign({many: false, skipSpent: true}, options);

```

That's what you meant?

src/new/wallet.js
pedroferreira1: I've changed to use another method `isAddressMine`, which accesses an object (where the keys are the addresses), then we can check if the address is from the wallet in O(1).

src/new/wallet.js
pedroferreira1: This will change in the branch (already started) to make the facades method signatures uniform. I don't think we should change this here also.

What do you think?

src/new/wallet.js
pedroferreira1: I use `tokensData` to create the outputs and we might have a tx without HTR. So I don't see any benefits in always having HTR and sometimes remove it. 

We avoid only one if block, as far as I can see.

src/new/wallet.js
pedroferreira1: This method may add inputs in the inputs array, if `chooseInputs=true` and also add the change output.

src/new/wallet.js
pedroferreira1: Makes sense, using `input.token` when exists

src/new/wallet.js
pedroferreira1: Makes sense, using `output.token` when exists

src/new/wallet.js
msbrogli: Exactly. Thanks!

src/new/wallet.js
msbrogli: Great. Thanks!

src/new/wallet.js

src/new/wallet.js
pedroferreira1: Created a method to return the defaultData and with it there is no need to use `cloneDeep`
```
    const getDefaultData = () => {
      return { outputs: [], inputs: [] };
    };
```

msbrogli: Great! Thanks!

msbrogli (approved):

src/new/wallet.js
```diff
@@ -1008,30 +1178,165 @@ class HathorWallet extends EventEmitter {
...
+   *  {
+   *   'createAnother': if should create another authority for the wallet. Default to true
+   *   'startMiningTx': boolean to trigger start mining (default true)
+   *  }
+   *
+   * @return {Object} Object with {success: true, sendTransaction, promise}, where sendTransaction is a
+   * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
+   **/
+  delegateAuthority(tokenUid, type, destinationAddress, options = {}) {
+    storage.setStore(this.store);
```
andreabadesso, line 1231: Why do we have to set the store on every method?

andreabadesso (approved):

src/new/wallet.js
pedroferreira1: The wallet headless might have many wallets handled by the same lib. Even though the wallets have separated storages and the class is prepared for that, the lib itself is not, thus we access the libStore many times.

For that to work with async calls we need to call `setStore` every time to indicates which store we want to use in each call.

src/new/wallet.js
```diff
@@ -737,7 +737,8 @@ class HathorWallet extends EventEmitter {
       chooseInputs = false;
     }
 
-    // Warning: prepareSendTokensData(...) might modify `partialData`.
+    // Warning: prepareSendTokensData(...) might modify `partialData`. It might add inputs in the inputs array
```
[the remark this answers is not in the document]
pedroferreira1: I agree with you. This is an old method used in many places right now and that will be deprecated soon. I've added a more explicit comment on it but I'd rather let the deprecation happen instead of changing it right now and update all the places using it (there are many places).

PR #188, opened 2021-05-13
feat: pin is required to send tx when it was not stored

pedroferreira1:


src/new/wallet.js
```diff
@@ -558,105 +566,8 @@ class HathorWallet extends EventEmitter {
...
-      const totalBalance = getBalance(tokenUid);
-      updatedBalanceMap[tokenUid] = totalBalance;
-    }
-    const newTokensHistory = Object.assign({}, state.tokensHistory, updatedHistoryMap);
-    const newTokensBalance = Object.assign({}, state.tokensBalance, updatedBalanceMap);
-
-    storage.setItem('local:tokens:history', newTokensHistory);
-    storage.setItem('local:tokens:balance', newTokensBalance);
-    */
   };
```
andreabadesso, line 569: Why was `sendMultiTokenTransaction` removed?

src/new/wallet.js
```diff
@@ -953,6 +881,10 @@ class HathorWallet extends EventEmitter {
         if (info.network.indexOf(this.conn.network) >= 0) {
           this.serverInfo = info;
           this.conn.start();
+          if (!this.storeSensitiveData) {
```
andreabadesso, line 884: I don't get it.

Wasn't it supposed to be:

```
if (this.storeSensitiveData) {
  this.pinCode = null;
  this.password = null;
}
```

To clear the `pinCode` and `password` from memory after a reconnect?

src/new/wallet.js
pedroferreira1: This method was not used. This was actually created to be used in an event a long time ago and it does not work. Maybe I should've removed it in another PR but it also used PIN, that's why I decided to remove it instead of refactoring it.

src/new/wallet.js
pedroferreira1: The idea here is that we always need this info to create the wallet and encrypt data in storage. After the start succeeds we remove the pin/password from memory if I don't want to store sensitive data.

msbrogli: Maybe we should optionally pass the pin and password in the `start(...)` method. So, if the user has set `this.pinCode` and `this.password`, they will be used. But, if they are empty, the user must pass both the pin and the password to the `start()` method.

What do you think? I feel that using the strategy we can remove the `this.storeSensitiveData` flag.

src/new/wallet.js
```diff
@@ -19,6 +19,8 @@ import Connection from './connection';
 import SendTransaction from './sendTransaction';
 import { AddressError } from '../../lib/errors';
 
+const ERROR_MESSAGE_PIN_REQUIRED = 'Pin is required.';
```
msbrogli, line 22: Should we put an error code as well?

msbrogli (changes requested):

src/new/wallet.js
pedroferreira1: Added

src/new/wallet.js
```diff
@@ -922,8 +843,20 @@ class HathorWallet extends EventEmitter {
...
+   *  {
+   *   'pinCode': pin to decrypt xpriv information. Required if not set in object.
+   *   'password': password to decrypt xpriv information. Required if not set in object.
+   *  }
    **/
-  start() {
+  start(optionsParams = {}) {
+    const options = Object.assign({ pinCode: null, password: null }, optionsParams);
+    const pinCode = options.pinCode || this.pinCode;
+    const password = options.password || this.password;
```
msbrogli, line 856: Isn't password required here?

msbrogli (approved):

src/new/wallet.js
pedroferreira1: Actually is not required if starting using xpriv. But I should add a verification `if (!xpriv && !password) => error`

pedroferreira1: Done

src/new/wallet.js
pedroferreira1: Added as optional in `start` and `constructor`

andreabadesso (approved):

PR #192, opened 2021-05-24
ECDSA message signatures with secp256k1 keys

r4mmer:
It's a port of bitcore-lib's [Message](https://github.com/bitpay/bitcore/blob/master/packages/bitcore-lib/lib/message.js).


r4mmer:
This PR will not be relevant as we will be using the more recent version of the lib it ports.

PR #194, opened 2021-05-27
Upgrade bitcore dependency

r4mmer:


r4mmer:
- Removed all uses of "new Buffer" to use "Buffer.alloc/Buffer.from"
    - This breaks for nodejs version 4 and lower, but our lowest target is 8
    - `new Buffer(value)` when value is a number we should use `alloc`, when it's a string `from` (preferably with encoding)
- `hexToBuffer` is not on bitcore-lib anymore, so i created our own `hexToBuffer`
    - `hexToBuffer` raises when value is not hex, `Buffer.from(val, 'hex')` ignore non-hex values
    - Only using hexToBuffer when value is not a hard-coded hex
- Upgrading bitcore-mnemonic makes bitcore-lib needed as a dependency, so it's included on package.json

r4mmer:
bitcore Networks data structure changed on a "recent" version to a dictionary of a list of network, this means that adding some network the mapped values (for Networks.get) will be a list and the 'get' method always return the first item on the list.

This means that using `Networks.get('mainnet')` always returns bitcoin's mainnet and not hathor's mainnet. I did not find an instance where we use this method with `'mainnet'` so i don't think it's a problem.
The solution would be removing bitcoin networks before adding hathor's, but this won't be possible since `Networks.remove` is not working correctly.

This could be a problem depending on how we use bitcore-lib's Networks, but we have our own method to get our network, so if we do not use `Networks.get` (and do not use the same version bytes as bitcoin) we won't encounter this problem.

__tests__/models/transaction.test.js
```diff
@@ -22,7 +22,7 @@ test('New tx', () => {
   const output1 = new Output(1000, address1);
   const output2 = new Output(1000, address2, {timelock: 1550249803});
   const inputDataHex = '4630440220317cd233801c1986c2de900bf8d344c6335d3c385e69d19d65e1fae7a0afd0af02207acddb824debf855798d79c45701cbe3a19aea00baad94bff5290c6f0b0acf8e210346cddff43dffab8e13398633ab7a7caf0d634551e89ae6fd563e282f6744b983';
-  const input1 = new Input('00034a15973117852c45520af9e4296c68adb9d39dc99a0342e23cd6686b295e', 0, {data: util.buffer.hexToBuffer(inputDataHex)})
+  const input1 = new Input('00034a15973117852c45520af9e4296c68adb9d39dc99a0342e23cd6686b295e', 0, {data: Buffer.from(inputDataHex, 'hex')})
```
pedroferreira1, line 25: Can we use our `hexToBuffer` utils method here?

__tests__/utils/wallet.test.js
```diff
@@ -87,12 +87,12 @@ test('Xpriv and xpub', () => {
 
   // To pubkey compressed
   const uncompressedPubKeyHex = '044f355bdcb7cc0af728ef3cceb9615d90684bb5b2ca5f859ab0f0b704075871aa385b6b1b8ead809ca67454d9683fcf2ba03456d6fe2c4abe2b07f0fbdbb2f1c1';
-  const compressedPubKey = wallet.toPubkeyCompressed(util.buffer.hexToBuffer(uncompressedPubKeyHex));
+  const compressedPubKey = wallet.toPubkeyCompressed(Buffer.from(uncompressedPubKeyHex, 'hex'));
```
pedroferreira1, line 90: Can we use our `hexToBuffer` utils method here?

__tests__/utils/wallet.test.js
```diff
@@ -87,12 +87,12 @@ test('Xpriv and xpub', () => {
...
   const expectedCompressedPubKeyHex = '034f355bdcb7cc0af728ef3cceb9615d90684bb5b2ca5f859ab0f0b704075871aa';
   expect(util.buffer.bufferToHex(compressedPubKey)).toBe(expectedCompressedPubKeyHex);
 
   // Invalid uncompressed public key must throw error
-  expect(() => wallet.toPubkeyCompressed(util.buffer.hexToBuffer(uncompressedPubKeyHex + 'ab'))).toThrowError(UncompressedPubKeyError);
+  expect(() => wallet.toPubkeyCompressed(Buffer.from(uncompressedPubKeyHex + 'ab', 'hex'))).toThrowError(UncompressedPubKeyError);
```
pedroferreira1, line 95: Can we use our `hexToBuffer` utils method here?

package.json
```diff
@@ -35,7 +35,8 @@
   ],
   "dependencies": {
     "axios": "^0.18.0",
-    "bitcore-mnemonic": "^1.7.0",
+    "bitcore-lib": "^8.25.10",
```
pedroferreira1, line 38: We have to update our `package-lock.json` as well, right?

pedroferreira1 (changes requested):

pedroferreira1:
About the methods `derive` -> `deriveChild`, when did it change? Do you know the version?

__tests__/models/transaction.test.js
r4mmer: We can, but since the input is a hard-coded hex, there is no difference between `hexToBuffer` and `Buffer.from`.

r4mmer:
> About the methods `derive` -> `deriveChild`, when did it change? Do you know the version?

Even on version 0.16.0 that we were using there was a [deprecation notice](https://github.com/bitpay/bitcore-lib/blob/v0.16.0/lib/hdpublickey.js#L89). For public keys the `derive` method uses the `deriveChild` method, but for private keys it uses `deriveNonCompliantChild` method ([here](https://github.com/bitpay/bitcore-lib/blob/v0.16.0/lib/hdprivatekey.js#L156)).

The non compliant method has a very important [warning](https://github.com/bitpay/bitcore-lib/blob/v0.16.0/lib/hdprivatekey.js#L205). We should evaluate the risks of changing this option carefully.

There has been no change in these files except the BufferUtils.hexToBuffer to Buffer.from, so not changing to deriveChild has no risks.

src/utils/buffer.ts
```diff
@@ -0,0 +1,12 @@
+import assert from 'assert';
+import buffer from 'buffer';
+
+const isHexa = (value: string): boolean => {
+  // test if value is string?
+  return /^[0-9a-fA-F]*$/.test(value);
+};
+
+export const hexToBuffer = (value: string): Buffer => {
+  assert(isHexa(value));
```
msbrogli, line 10: Does it work in the browser? I'm wondering whether it will work in hathor-wallet-desktop because it runs through Electron. Maybe we should use just raise an exception to avoid compatibility issues.

msbrogli (approved):
It will be great to update this dependency. Nice job!

src/wallet.js
```diff
@@ -153,7 +153,7 @@ const wallet = {
     let initialAccessData;
     let privkey;
     if (xpriv.depth === 0) {
-      privkey = xpriv.derive(`m/44'/${HATHOR_BIP44_CODE}'/0'/0`);
+      privkey = xpriv.deriveChild(`m/44'/${HATHOR_BIP44_CODE}'/0'/0`);
```
msbrogli, line 156: In a separate PR, add a test that create a wallet with given seeds and check their first 5 addresses. This test must pass without these changes and with these changes. It will guarantee that the address generation remains the same.

msbrogli (changes requested):

r4mmer:
# bitcore-lib upgrade concerns

## Overall changes that affect our lib

The investigation on the changes from version 0.16.0 to 8.25.10 showed a few possibly problematic changes.
- removal of method `hexToBuffer`
    - We use this method on our lib, so we need to update these calls.
- `new Buffer` to `Buffer.from/Buffer.alloc`
    - This is from an update on the version of NodeJs V4, so the changes are internal to bitcore-lib and do not affect our lib.
    - We should stop using `new Buffer` since it's deprecated either way.
- bitcore-lib (and mnemonic) were migrated to a joint repository (together with many other bitpay libs)
    - This means that any update on any package increases the version of bitcore-lib and bitcore-mnemonic
    - This is why in a short ammount of time we are on major version 8
- bitcore-lib introduced segwit and bech32 addresses
    - It doesn't affect our lib directly, but the network package changes some parameters, so we need to update our networks (they should use the same type).

## .derive to .deriveChild

This methods exist on `hdpublickey.js` and `hdprivatekey.js`, but on the `hdpublickey.js` derive just calls deriveChild internally, so they are interchangeable.
On `hdprivatekey.js` however there are some differences. The next sections will explain some of the tecnicals and implementations of bitcore-lib.

### Hardened childs

From bip32 [extended keys](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys):

> Each extended key has 2<sup>31</sup> normal child keys, and 2<sup>31</sup> hardened child keys. Each of these child keys has an index. The normal child keys use indices 0 through 2<sup>31</sup>-1. The hardened child keys use indices 2<sup>31</sup> through 2<sup>32-1</sup>.

The notation for hardened keys when using string is the index minus 2<sup>31</sup> with an apostrophe in the end (2<sup>31</sup> would be 0', and so on).

### deriveNonCompliantChild

The "non-compliant" part is because the private key should be padded with zeroes to be 33 bytes (total) on hardened derivation (see [ref](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--private-child-key), the note on the first step)
So "non-compliant" is a derivation that does not pad with zeroes, while deriveChild does.

### refs from bitcore-lib

- [`.deriveChild`](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L188)
    - Accepts a number or a string and calls `._deriveWithNumber` or `._deriveFromString` accordingly.
- [`.deriveNonCompliantChild`](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L214)
    - uses the same internal logic as .deriveChild, but passing a `true` for the argument `nonCompliant`
- [`._deriveFromString`](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L278)
    - parses the string path to get each derivation index and uses `._deriveWithNumber` on each of them
    - uses the `_getDerivationIndexes` method.
- [`._getDerivationIndexes`](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L98)
    - splits the path on each index
    - for each index if it ends with apostrophe, assumes it's hardened
    - on hardened indices adds 2<sup>31</sup>
    - return the list of indices for the path
- [`._deriveWithNumber`](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L224)
    - The function assumes it's a hardened derivation if the argument is greater than 2<sup>31</sup> or if the hardened argument is true.
    - Uses the nonCompliant argument to process the private key for derivation, padding with zeroes or not.

### Our wallet-lib

If we used derivations only on public keys or only with numbers, we would not be affected (since we never use the hardened argument), but that is not the case.
We derive private keys with paths (see [ref1](https://github.com/HathorNetwork/hathor-wallet-lib/blob/dev/src/wallet.js#L194) and [ref2](https://github.com/HathorNetwork/hathor-wallet-lib/blob/dev/src/wallet.js#L156)) and use hardened indexes on them.
So since whatever changes this padding brings will affect us, we still need to answer the question, how?

The padding will not bring any changes on private keys which themselves are represented as 32 bytes long buffers (the first byte is always 0x00).
So I made a script which gets the HDPrivateKey from the seed, converts it to the buffer (just like bitcore [does](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L241)) and test it's length to see if it's 32 bytes.
The initial tests did not gave any problems, so i generated 10.000 random seeds with 256 entropy (like our lib [does](https://github.com/HathorNetwork/hathor-wallet-lib/blob/dev/src/constants.js#L112)) and tested all of them.
Unfortunately, there were about 40 seeds which generated a private key with 31 bytes. I then tested the derivation paths on these 31 byte keys and they generated different addresses when using derive vs deriveChild (the private keys with 32 bytes generated the same addresses as expected).

### The scripts

It takes a couple minutes to run this for 10.000 seeds, but the chance to generate a seed with 31 bytes is approx. 0.4%, so testing with 100 may not generate any.

This script was executed with version 0.16.0 of bitcore-lib and 0.19.0 of `@hathor/wallet-lib`.

```js
const bitcore = require('bitcore-lib');
const Mnemonic = require('bitcore-mnemonic');
const hathor = require('@hathor/wallet-lib');

const genSeeds = () => {
    const seeds = [];
    for (let i = 0; i < 10000; i++) {
        const code = new Mnemonic(256);
        seeds.push(code.phrase);
    }
    return seeds;
};

// generate 10000 seed phrases with 256 entropy
const seeds = genSeeds();

console.log('[*] Testing for non 32 byte private keys');
seeds.forEach((seed) => {
  const xpriv = hathor.walletUtils.getXPrivKeyFromSeed(seed);
  if (xpriv.privateKey.bn.toBuffer().length !== 32) {
    console.log('[-] ', seed);
  }
});
```

Some functions to generate addresses from deriveChild and deriveNonCompliantChild from a seed:

```js
const baseDerivePath = "m/44'/280'/0'/0";

const deriveAddressesNonCompliant = (seed) => {
  const xpriv = hathor.walletUtils.getXPrivKeyFromSeed(seed);
  const dxpriv = xpriv.deriveNonCompliantChild(baseDerivePath);
  const ret = [];
  for (let i = 0; i <= 10; i++) {
    ret.push((i, dxpriv.deriveNonCompliantChild(i).publicKey.toAddress().toString()));
  }
  return ret;
};

const deriveAddresses = (seed) => {
  const xpriv = hathor.walletUtils.getXPrivKeyFromSeed(seed);
  const dxpriv = xpriv.deriveChild(baseDerivePath);
  const ret = [];
  for (let i = 0; i <= 10; i++) {
    ret.push((i, dxpriv.deriveChild(i).publicKey.toAddress().toString()));
  }
  return ret;
};
```

### Final considerations

Changing from .derive to .deriveChild may affect the users. We should explicitly call .deriveNonCompliantChild to denote the knowlegde of this api change and discuss if the change to .deriveChild is something we need.

msbrogli:
> ### Final considerations
> Changing from .derive to .deriveChild may affect the users. We should explicitly call .deriveNonCompliantChild to denote the knowlegde of this api change and discuss if the change to .deriveChild is something we need.

Great analysis. I feel we shouldn't change it in this PR. But I think we must include tests that generate addresses for a list of seeds and check whether the generated addresses are correct or not. @pedroferreira1 @r4mmer What do you think?

pedroferreira1:
> @pedroferreira1 @r4mmer What do you think?

I agree that we can move with `deriveNonCompliantChild`.

package.json
r4mmer: Done

src/models/network.ts
```diff
@@ -39,19 +40,24 @@ const mainnet = Networks.add({
   pubkeyhash: versionBytes['mainnet']['p2pkh'],
   privatekey: 0x80,
   scripthash: versionBytes['mainnet']['p2sh'],
+  bech32prefix: 'bc',
```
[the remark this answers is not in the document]
r4mmer: Done, i also changed the testnet to `tn`.

src/helpers.js
```diff
@@ -407,7 +407,7 @@ const helpers = {
 
     const addrMap = {};
     for (let index = startIndex; index < startIndex + quantity; index++) {
-      const key = xpub.derive(index);
+      const key = xpub.deriveChild(index);
```
pedroferreira1, line 410: Are you using `deriveChild` because it's an xpub? I know that for xpubs `deriveChild` and `deriveNonCompliantChild` but maybe we should keep using `deriveNonCompliantChild` everywhere to avoid any confusion, what do you think?

src/utils/wallet.ts
```diff
@@ -158,7 +158,7 @@ const wallet = {
     } catch (error) {
       throw new XPubError(error.message);
     }
-    const key = xpub.derive(index);
+    const key = xpub.deriveChild(index);
```
pedroferreira1, line 161: Same comment about the `deriveChild`. If we decide to continue using it I think we should have a comment explaining why we can safely use it with xpub and must use `deriveNonCompliantChild` with xpriv.

msbrogli: Is there a `deriveNonCompliantChild()` for pubkeys? If yes, I feel we should use it as well.

msbrogli (approved):

src/utils/wallet.ts
r4mmer: There isn't, the tests on commit 2abe487 failed because of this, i chose to use deriveChild but derive and deriveChild are interchangeable (derive calls deriveChild internally, [here](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdpublickey.js#L114)). The only difference is that maybe derive can be fully deprecated on a future version.

pedroferreira1: As we talked in the daily meeting yesterday, I think you should have a comment in all `deriveChild` uses to explain that `deriveChild` must be used ONLY for xpubs. Xprivs must always use `deriveNonCompliantChild` for now and to make sure in the future nobody will 'copy' xpub derivation method and use to derive xpriv as well

r4mmer:
## Network Error

When building an HDPrivateKey from seed the new version is not using our network.

### Explanation

When converting the seed to a `HDPrivateKey` the network object is passed as an argument, this will carry till the `._buildFromObject` method, it will then only use the xprivkey to build the buffers he pass on to `._buildFromBuffers`.
The `._buildFromBuffers` will get the Network from the version (the xprivkey from out network) which is the same as bitcore's livenet xprivkey, but on the new versions of bitcore the original bitcoin livenet is returned instead of our mainnet.

#### References

- `.toHDPrivateKey` [mnemonic.js#L211](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-mnemonic/lib/mnemonic.js#L211)
- `.fromSeed` [hdprivatekey.js#L422](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L422)
- HDPrivateKey constructor [hdprivatekey.js#L61](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L61)
- `._buildFromObject` [hdprivatekey.js#L368](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L368)
- `._buildFromBuffers` [hdprivatekey.js#L480](https://github.com/bitpay/bitcore/blob/v8.25.10/packages/bitcore-lib/lib/hdprivatekey.js#L480)

### Why

The internal `Object` that maps options (string and number) to Networks changed to map options to a list of Networks and the `.get` method only returns the first Network on the list.
This means that on version `0.16.0` we were replacing bitcoin livenet with our mainnet when registering our mainnet, but on the new version, bitcoin mainnet is not overwritten and is returned when using `.get` with any parameter other than "production" and our version bytes.

### Proposed solutions

- Change the version bytes we have in common with bitcoin
    - This will change the addresses and keys generated with seeds, so it's not an option for now.
- Make a PR with bitcore to change this behavior
    - We would need to wait for the next version that has this change, if it's approved
    - The bitpay team is not very responsive, so it may take a while.
    - This will lock the PR, so any issues waiting for this PR will also be locked.

__tests__/utils/wallet.test.js
```diff
@@ -174,43 +176,175 @@ test('hd derivation', () => {
...
-  const xpub32B = wallet.getXPubKeyFromSeed(Words32B, {accountDerivationIndex: '0\'/0'});
+  const xpub32B = wallet.getXPubKeyFromSeed(Words32B, {networkName: 'testnet', accountDerivationIndex: '0\'/0'});
   const addressObj32B = wallet.getAddresses(xpub32B, 0, 10, 'testnet');
   const address32B = Object.keys(addressObj32B);
   for (let i = 0; i < 10; i++) {
     expect(address32B[i]).toStrictEqual(Addr32B[i]);
   }
 });
+
+test('testnet: hd derivation with password', () => {
```
msbrogli, line 220: Change to passphrase. I know that it is confusing but, in our wallets, password, PIN, and passphrase have very different meanings.

__tests__/utils/wallet.test.js
```diff
@@ -174,43 +176,175 @@ test('hd derivation', () => {
...
+      },
+  );
+  const addressObj = wallet.getAddresses(xpub, 0, 10, 'testnet');
+  const address = Object.keys(addressObj);
+  for (let i = 0; i < 10; i++) {
+    expect(address[i]).toStrictEqual(Addr[i]);
+  }
+});
+
+test('mainnet: hd derivation', () => {
```
msbrogli, line 273: No passphrase for mainnet?

src/models/network.ts
```diff
@@ -39,8 +40,9 @@ const mainnet = Networks.add({
   pubkeyhash: versionBytes['mainnet']['p2pkh'],
   privatekey: 0x80,
   scripthash: versionBytes['mainnet']['p2sh'],
+  bech32prefix: 'ht',
   xpubkey: 0x0488b21e,
-  xprivkey: 0x0488ade4,
+  xprivkey: 0x0488ade5,
```
msbrogli, line 45: What is this parameter for? And why are we just increasing it by one? Should we create a value for Hathor? How? Randomly?

src/models/network.ts
```diff
@@ -52,8 +54,9 @@ const testnet = Networks.add({
   pubkeyhash: versionBytes['testnet']['p2pkh'],
   privatekey: 0x80,
   scripthash: versionBytes['testnet']['p2sh'],
+  bech32prefix: 'tn',
   xpubkey: 0x0488b21e,
-  xprivkey: 0x0488ade4,
+  xprivkey: 0x0488ade6,
```
msbrogli, line 59: Same thing here.

src/models/network.ts
```diff
@@ -39,8 +40,9 @@ const mainnet = Networks.add({
   pubkeyhash: versionBytes['mainnet']['p2pkh'],
   privatekey: 0x80,
   scripthash: versionBytes['mainnet']['p2sh'],
+  bech32prefix: 'ht',
```
msbrogli, line 43: What is it for? We don't have bech32 addresses in Hathor.

src/models/network.ts
```diff
@@ -52,8 +54,9 @@ const testnet = Networks.add({
   pubkeyhash: versionBytes['testnet']['p2pkh'],
   privatekey: 0x80,
   scripthash: versionBytes['testnet']['p2sh'],
+  bech32prefix: 'tn',
```
msbrogli, line 57: What is it for? We don't have bech32 addresses in Hathor.

src/models/network.ts
```diff
@@ -39,8 +40,9 @@ const mainnet = Networks.add({
   pubkeyhash: versionBytes['mainnet']['p2pkh'],
```
msbrogli, line 40: This comment is in the wrong line. I thought we would rename it to `htr-mainnet` and also we would remove all non-Hathor networks from bitcore. Why aren't we doing that?

msbrogli (changes requested):

src/models/network.ts
r4mmer: We cannot remove non-Hathor networks from bitcore since the `remove` method of bitcore is not working properly.

src/models/network.ts
r4mmer: We don't, but since the new version of bitcore has this new parameter i thought we should include it.

src/models/network.ts
r4mmer: This parameter only exists to signal which network the serialized xprivkey belongs to. The onty constraint i could find is that it must be 4 bytes long and must not be the same as the xpubkey (from the network object). This will **NOT** change the generated addresses, it only affects the metadata used internally by bitcore.

src/models/network.ts
pedroferreira1: I think the idea is to run the remove method even if it's not working properly. It's better for us to get an empty network and raise an error instead of getting a bitcoin network.

src/models/network.ts
```diff
@@ -28,32 +28,35 @@ const versionBytes = {
...
-  name: 'mainnet',
+  name: 'htr-mainnet',
   alias: 'production',
   pubkeyhash: versionBytes['mainnet']['p2pkh'],
   privatekey: 0x80,
   scripthash: versionBytes['mainnet']['p2sh'],
+  bech32prefix: 'ht',
   xpubkey: 0x0488b21e,
-  xprivkey: 0x0488ade4,
+  xprivkey: 0x0488ade5,
```
pedroferreira1, line 45: I just think we should have a more detailed explanation of this parameter and `xpubkey` in the code. To prevent any confusion in the future

pedroferreira1 (changes requested):

__tests__/models/network.test.js
```diff
@@ -13,28 +13,32 @@ test('Get and set network', () => {
     'mainnet': {
       'p2pkh': 0x28,
       'p2sh': 0x64,
+      'xpriv': 0x03523b05,
```
msbrogli, line 16: I think we should use:

- Mainnet, xpriv: `Hatr` in ASCII (`0x48617468`)
- Mainnet, xpub: (same as xpriv if possible)
- Testnet, xpriv: `htn!` in ASCII (`0x68746e21`)

What do you think? We should document it properly. Maybe an RFC?

msbrogli (approved):

src/models/network.ts
r4mmer: This wouldn't work, many places in bitcore use `Network.get(arg) || Network.defaultNetwork` which returns the livenet (even if we remove it). Later on the code it will search for this network and it will not find it generating an error (usually `"version argument is not a buffer, it's undefined"`). It broke 31 out of 137 of our tests when i tried.

__tests__/models/network.test.js
r4mmer: I think having the same identifier for xpriv and xpub would generate some confusion, but we could use these for now and discuss on another issue.

pedroferreira1 (approved):

PR #214, opened 2021-07-16
feat: integrate token actions in wallet service with new HathorWallet facade

pedroferreira1:
### Acceptance criteria

1. Authorization in the wallet service: creating signature and getting jwt token to be added on API header.
2. Token actions with new API from wallet service: send tx, create new token, mint, melt, delegate authority, destroy authority.

### Not done in this PR

1. Real time update
2. Xpriv management encrypted on storage

### DIFF
--- a/src/utils/transaction.ts
+++ b/src/utils/transaction.ts
@@ -1,0 +1,86 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+
+
+const transaction = {
+  /**
+   * Select best utxos with the algorithm described below. This method expects the utxos to be sorted by greatest value
+   *
+   * 1. If we have a single utxo capable of handle the full amount requested,
+   * we return the utxo with smaller amount among the ones that have an amount bigger than the requested
+   * 2. Otherwise we reverse sort the utxos by amount and select the utxos in order until the full amount is fulfilled.
+   *
+   * // TODO Add params/returns docstring and create Utxos typedef
+   *
+   * @memberof transaction
+   * @inner
+   */
+  selectUtxos(utxos, totalAmount: number): {utxos, changeAmount: number} {
+    if (totalAmount <= 0) {
+      throw new Error('Total amount must be a positive integer.');
+    }
+
+    if (utxos.length === 0) {
+      throw new Error('Don\'t have enough utxos to fill total amount.');
+    }
+
+    if (utxos[0].value < totalAmount) {
+      // We can't fill the total amount with a single utxo
+      // so we start pushing utxos to an array until we fill the total amount
+      const utxosToUse: any[] = [];
+      let filledAmount = 0;
+      for (const utxo of utxos) {
+        utxosToUse.push(utxo);
+        filledAmount += utxo.value;
+
+        if (filledAmount >= totalAmount) {
+          break;
+        }
+      }
+      return {
+        utxos: utxosToUse,
+        changeAmount: filledAmount - totalAmount
+      };
+    } else {
+      // We can fill the total amount with a single utxo
+      // we will find the smaller utxo that can fill the total amount
+      let lastUtxo = null;
+
+      for (const utxo of utxos) {
+        if (utxo.value === totalAmount) {
+          return {
+            utxos: [utxo],
+            changeAmount: 0
+          };
+        }
+
+        if (utxo.value < totalAmount) {
+          // The last one is the smallest single utxo that can fill the amount
+          // it's safe to use lastUtxo because it will never be null
+          // this is inside the else that checked that at least one utxo fills the total amount
+          return {
+            utxos: [lastUtxo],
+            // @ts-ignore
+            changeAmount: lastUtxo!.value - totalAmount
+          };
+        }
+
+        lastUtxo = utxo;
+      }
+      // If I get here, it means that all utxos in the array are bigger than the expected amount
+      // then I must use the last utxo, which has the smallest value
+      const smallestUtxo = utxos[utxos.length - 1];
+      return {
+        utxos: [smallestUtxo],
+        changeAmount: smallestUtxo.value - totalAmount
+      };
+    }
+  },
+}
+
+export default transaction;
--- a/src/utils/transaction.ts
+++ b/src/utils/transaction.ts
@@ -29,62 +29,27 @@ const transaction = {
       throw new UtxoError('Don\'t have enough utxos to fill total amount.');
     }
 
-    if (utxos[0].value < totalAmount) {
-      // We can't fill the total amount with a single utxo
-      // so we start pushing utxos to an array until we fill the total amount
-      const utxosToUse: any[] = [];
-      let filledAmount = 0;
-      for (const utxo of utxos) {
-        utxosToUse.push(utxo);
-        filledAmount += utxo.value;
-
+    let utxosToUse: Utxo[] = [];
+    let filledAmount = 0;
+    for (const utxo of utxos) {
+      if (utxo.value >= totalAmount) {
+        utxosToUse = [utxo];
+        filledAmount = utxo.value;
+      } else {
         if (filledAmount >= totalAmount) {
-          break;
+          break
         }
-      }
+        filledAmount += utxo.value;
+        utxosToUse.push(utxo);
+     }
+    }
+    if (filledAmount < totalAmount) {
+      throw new UtxoError('Don\'t have enough utxos to fill total amount.');
+    }
 
-      if (filledAmount < totalAmount) {
-        // It means that all utxos combined are not enough to fill the requested amount
-        throw new UtxoError('Don\'t have enough utxos to fill total amount.');
-      }
-
-      return {
-        utxos: utxosToUse,
-        changeAmount: filledAmount - totalAmount
-      };
-    } else {
-      // We can fill the total amount with a single utxo
-      // we will find the smallest utxo that can fill the total amount
-      let lastUtxo: Utxo | null = null;
-
-      for (const utxo of utxos) {
-        if (utxo.value === totalAmount) {
-          return {
-            utxos: [utxo],
-            changeAmount: 0
-          };
-        }
-
-        if (utxo.value < totalAmount) {
-          // The last one is the smallest single utxo that can fill the amount
-          // it's safe to use lastUtxo because it will never be null
-          // this is inside the else that checked that at least one utxo fills the total amount
-          return {
-            utxos: [lastUtxo!],
-            changeAmount: lastUtxo!.value - totalAmount
-          };
-        }
-
-        lastUtxo = utxo;
-      }
-
-      // If I got here, it means that all utxos in the array are bigger than the expected amount
-      // then I must use the last utxo, which has the smallest value
-      const smallestUtxo = utxos[utxos.length - 1];
-      return {
-        utxos: [smallestUtxo],
-        changeAmount: smallestUtxo.value - totalAmount
-      };
+    return {
+      utxos: utxosToUse,
+      changeAmount: filledAmount - totalAmount,
     }
   },
 }
--- a/src/utils/transaction.ts
+++ b/src/utils/transaction.ts
@@ -6,6 +6,8 @@
  */
 
 
+import { Utxo } from '../wallet/types';
+import { UtxoError } from '../errors';
 
 const transaction = {
   /**
@@ -20,13 +22,13 @@ const transaction = {
    * @memberof transaction
    * @inner
    */
-  selectUtxos(utxos, totalAmount: number): {utxos, changeAmount: number} {
+  selectUtxos(utxos: Utxo[], totalAmount: number): {utxos: Utxo[], changeAmount: number} {
     if (totalAmount <= 0) {
-      throw new Error('Total amount must be a positive integer.');
+      throw new UtxoError('Total amount must be a positive integer.');
     }
 
     if (utxos.length === 0) {
-      throw new Error('Don\'t have enough utxos to fill total amount.');
+      throw new UtxoError('Don\'t have enough utxos to fill total amount.');
     }
 
     if (utxos[0].value < totalAmount) {
@@ -42,14 +44,20 @@ const transaction = {
           break;
         }
       }
+
+      if (filledAmount < totalAmount) {
+        // It means that all utxos combined are not enough to fill the requested amount
+        throw new UtxoError('Don\'t have enough utxos to fill total amount.');
+      }
+
       return {
         utxos: utxosToUse,
         changeAmount: filledAmount - totalAmount
       };
     } else {
       // We can fill the total amount with a single utxo
-      // we will find the smaller utxo that can fill the total amount
-      let lastUtxo = null;
+      // we will find the smallest utxo that can fill the total amount
+      let lastUtxo: Utxo | null = null;
 
       for (const utxo of utxos) {
         if (utxo.value === totalAmount) {
@@ -64,15 +72,15 @@ const transaction = {
           // it's safe to use lastUtxo because it will never be null
           // this is inside the else that checked that at least one utxo fills the total amount
           return {
-            utxos: [lastUtxo],
-            // @ts-ignore
+            utxos: [lastUtxo!],
             changeAmount: lastUtxo!.value - totalAmount
           };
         }
 
         lastUtxo = utxo;
       }
-      // If I get here, it means that all utxos in the array are bigger than the expected amount
+
+      // If I got here, it means that all utxos in the array are bigger than the expected amount
       // then I must use the last utxo, which has the smallest value
       const smallestUtxo = utxos[utxos.length - 1];
       return {
--- a/src/utils/transaction.ts
+++ b/src/utils/transaction.ts
@@ -17,8 +17,6 @@ const transaction = {
    * we return the utxo with smaller amount among the ones that have an amount bigger than the requested
    * 2. Otherwise we reverse sort the utxos by amount and select the utxos in order until the full amount is fulfilled.
    *
-   * // TODO Add params/returns docstring and create Utxos typedef
-   *
    * @memberof transaction
    * @inner
    */
--- a/src/wallet/api/walletApi.ts
+++ b/src/wallet/api/walletApi.ts
@@ -14,6 +14,7 @@ import {
   NewAddressesResponseData,
   BalanceResponseData,
   HistoryResponseData,
+  TokensResponseData,
   TxProposalCreateResponseData,
   TxProposalUpdateResponseData,
   UtxoResponseData,
@@ -46,27 +47,14 @@ const walletApi = {
       data['firstAddress'] = firstAddress;
     }
     const axios = await axiosInstance();
-    try {
-      const response = await axios.post('wallet/init', data);
-      if (response.status === 200 && response.data.success) {
-        return response.data;
-      } else {
-        throw new WalletRequestError('Error creating wallet.');
-      }
-    } catch(error) {
-      if (error.response.status === 400) {
-        // Bad request
-        // Check if error is 'wallet-already-loaded'
-        const errorData = error.response.data;
-        if (errorData && errorData.error === 'wallet-already-loaded') {
-          // If it was already loaded, we have to check if it's ready
-          return errorData;
-        } else {
-          throw new WalletRequestError('Error creating wallet.');
-        }
-      } else {
-        throw new WalletRequestError('Error creating wallet.');
-      }
+    const response = await axios.post('wallet/init', data);
+    if (response.status === 200 && response.data.success) {
+      return response.data;
+    } else if (response.status === 400 && response.data.error === 'wallet-already-loaded') {
+      // If it was already loaded, we have to check if it's ready
+      return response.data;
+    } else {
+      throw new WalletRequestError('Error creating wallet.');
     }
   },
 
@@ -104,12 +92,18 @@ const walletApi = {
     }
   },
 
-  async getHistory(wallet: HathorWalletServiceWallet, token: string | null = null): Promise<HistoryResponseData> {
-    // TODO add pagination parameters
-    const data = { params: {} };
-    if (token) {
-      data['params']['token_id'] = token;
+  async getTokens(wallet: HathorWalletServiceWallet): Promise<TokensResponseData> {
+    const axios = await axiosInstance(wallet);
+    const response = await axios.get('wallet/tokens');
+    if (response.status === 200 && response.data.success === true) {
+      return response.data;
+    } else {
+      throw new WalletRequestError('Error getting list of tokens.');
     }
+  },
+
+  async getHistory(wallet: HathorWalletServiceWallet, options = {}): Promise<HistoryResponseData> {
+    const data = { params: options };
     const axios = await axiosInstance(wallet);
     const response = await axios.get('wallet/history', data);
     if (response.status === 200 && response.data.success === true) {
--- a/src/wallet/api/walletApi.ts
+++ b/src/wallet/api/walletApi.ts
@@ -40,8 +40,11 @@ const walletApi = {
     }
   },
 
-  async createWallet(xpubkey: string): Promise<WalletStatusResponseData> {
+  async createWallet(xpubkey: string, firstAddress: string | null = null): Promise<WalletStatusResponseData> {
     const data = { xpubkey };
+    if (firstAddress) {
+      data['firstAddress'] = firstAddress;
+    }
     const axios = await axiosInstance();
     try {
       const response = await axios.post('wallet/init', data);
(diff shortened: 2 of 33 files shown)


src/utils/transaction.ts
```diff
@@ -0,0 +1,92 @@
...
+  selectUtxos(utxos: Utxo[], totalAmount: number): {utxos: Utxo[], changeAmount: number} {
+    if (totalAmount <= 0) {
+      throw new UtxoError('Total amount must be a positive integer.');
+    }
+
+    if (utxos.length === 0) {
+      throw new UtxoError('Don\'t have enough utxos to fill total amount.');
+    }
+
+    if (utxos[0].value < totalAmount) {
```
msbrogli, line 32: I feel this algorithm can be simplified (but I haven't tested my code). What do you think?

```js
let utxosToUse = [];
let filledAmount = 0;
for (const utxo of utxos) {
  if (utxo.value >= totalAmount) {
    utxosToUse = [utxo];
    filledAmount = utxo.value;
  } else {
    if (filledAmount >= totalAmount) {
      break
    }
    filledAmount += utxo.value;
    utxosToUse.push(utxo);
 }
}
if (filledAmount < totalAmount) {
  throw new UtxoError('Don\'t have enough utxos to fill total amount.');
}
return {
  utxos: utxosToUse,
  changeAmount: filledAmount - totalAmount,
}
```

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
    * @inner
    */
   async startPollingStatus() {
     try {
-      const res = await walletApi.getWalletStatus(this.walletId!);
+      const res = await walletApi.getWalletStatus(this);
       const data = res.data;
       if (res.status === 200 && data.success) {
         if (data.status.status === 'creating') {
           setTimeout(async () => {
```
msbrogli, line 177: I feel this structure is too fragile because the pooling won't be called again if an exception occurs. I feel we should use a `setInterval(...)` that must be cancelled after the wallet is ready.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
    *
-   * @memberof HathorWallet
+   * @memberof HathorWalletServiceWallet
    * @inner
    */
   async startPollingStatus() {
     try {
-      const res = await walletApi.getWalletStatus(this.walletId!);
+      const res = await walletApi.getWalletStatus(this);
       const data = res.data;
```
msbrogli, line 174: Shouldn't we move this logic to the `walletApi`? I think this code must be way simpler to read because it would have to deal with a promise that might succeed or fail.

src/wallet/wallet.ts
```diff
@@ -65,35 +92,47 @@ class HathorWalletServiceWallet extends EventEmitter {
...
       } else {
-        this.setState(walletState.READY);
+        await this.onWalletReady();
       }
     }
+
     try {
-      const res = await walletApi.createWallet(xpub);
+      const res = await walletApi.createWallet(this.xpub);
       const data = res.data;
```
msbrogli, line 131: Shouldn't we move this logic to the walletApi? I think this code must be way simpler to read because it would have to deal with a promise that might succeed or fail.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
    * @inner
    */
-  async getAllAddresses(): Promise<string[]> {
-    const response = await walletApi.getAddresses(this.walletId!);
-    let addresses = [];
+  async getAllAddresses(): Promise<GetAddressesObject[]> {
+    this.checkWalletReady();
+    const response = await walletApi.getAddresses(this);
+    let addresses: GetAddressesObject[] = [];
     if (response.status === 200 && response.data.success === true) {
```
msbrogli, line 227: Shouldn't we move this logic to the walletApi? I think this code must be way simpler to read because it would have to deal with a promise that might succeed or fail.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
       addresses = response.data.addresses;
     } else {
-      // TODO What error should be handled here?
+      throw new WalletRequestError('Error getting wallet addresses.');
     }
     return addresses;
   }
 
+  private async getNewAddresses() {
```
msbrogli, line 235: Missing docstring.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+    this.checkWalletReady();
+    const response = await walletApi.getNewAddresses(this);
+    let addresses: AddressInfoObject[] = [];
+    if (response.status === 200 && response.data.success === true) {
```
msbrogli, line 239: Shouldn't we move this logic to the walletApi? I think this code must be way simpler to read because it would have to deal with a promise that might succeed or fail.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
    * @inner
    */
-  async getBalance(token: string | null = null) {
-    const response = await walletApi.getBalances(this.walletId!, token);
-    let balance = null;
+  async getBalance(token: string | null = null): Promise<GetBalanceObject[]> {
+    this.checkWalletReady();
+    const response = await walletApi.getBalances(this, token);
+    let balance: GetBalanceObject[] = [];
     if (response.status === 200 && response.data.success === true) {
```
msbrogli, line 258: Shouldn't we move this logic to the walletApi? I think this code must be way simpler to read because it would have to deal with a promise that might succeed or fail.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
   /**
    * Get the history of the wallet for a specific token
    *
-   * @memberof HathorWallet
+   * @memberof HathorWalletServiceWallet
    * @inner
    */
-  async getTxHistory(options: { token?: string } = {}) {
+  async getTxHistory(options: { token?: string } = {}): Promise<GetHistoryObject[]> {
     // TODO Add pagination parameters
```
msbrogli, line 273: TODO?

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+   */
+  async getUtxoFromId(txId: string, index: number): Promise<Utxo | null> {
+    const response = await walletApi.getUtxos(this, { txId, index });
+    if (response.status === 200 && response.data.success === true) {
+      const utxos = response.data.utxos;
+      if (utxos.length === 0) {
+        // No utxo for this txId/index or is not from the requested wallet
+        return null;
+      } else {
+        return utxos[0];
```
msbrogli, line 301: Why don't we return all utxos?

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+  }
+
+  /**
+   * Calculate sign message for auth token
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  getAuthSign(timestamp: number): string {
+    const xpriv = wallet.getXPrivKeyFromSeed(this.seed, {passphrase: this.passphrase, networkName: this.network.name});
```
msbrogli, line 368: Why don't we get the xprivkey directly from the storage? I mean, do we need really to access the seed and passphrase?

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+        utxos = ret.utxos;
+      }
+    } else {
+      throw new WalletRequestError('Error requesting utxo.');
+    }
+    return { utxos, changeAmount };
+  }
+
+  /**
+   * Calculate sign message for auth token
```
msbrogli, line 362: Suggestion: "Signs a message using derivation path XXX".

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  getAuthSign(timestamp: number): string {
```
msbrogli, line 367: Rename to `signMessage`? I understand that it is used by the authentication part but the signature is generic. Or split in two methods, one that signs and another that sends the auth message to be signed.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
    * Send a transaction from an array of outputs and inputs
    *
-   * @memberof HathorWallet
+   * @memberof HathorWalletServiceWallet
    * @inner
    */
-  async sendManyOutputsTransaction(outputs, options = { inputs: [], changeAddress: null }) {
-    return await this.sendTxProposal(outputs, options);
+  async sendManyOutputsTransaction(outputs: OutputRequestObj[], options: { inputs?: InputRequestObj[], changeAddress?: string } = {}): Promise<TxProposalUpdateResponseData> {
+    this.checkWalletReady();
```
msbrogli, line 418: Why do we need this call? It was added as first step in many methods of this file.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
       changeAddress: null
     }, options);
-    const { inputs, changeAddress } = newOptions;
-    const response = await walletApi.createTxProposal(this.walletId!, outputs, inputs);
-    if (response.status === 201) {
-      const responseData = response.data;
-      this.txProposalId = responseData.txProposalId;
+    const { inputs } = newOptions;
+
+
```
msbrogli, line 460: Blank line.

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+        } else {
+          amountInputMap[utxo.tokenId] = utxo.value;
+        }
       }
 
-      const outputsObj: Output[] = [];
-      for (const o of responseData.outputs) {
-        // TODO handle custom tokens
-        outputsObj.push(new Output(o.value, new Address(o.address)));
+      for (const t in amountOutputMap) {
```
msbrogli, line 518: Rename `t` to `tokenId`?

src/wallet/wallet.ts
```diff
@@ -103,170 +142,442 @@ class HathorWalletServiceWallet extends EventEmitter {
...
-   * Send a transaction proposal
+   * Send transaction
    *
-   * @memberof HathorWallet
+   * @memberof HathorWalletServiceWallet
    * @inner
    */
-  async sendTxProposal(outputs, options = { inputs: [], changeAddress: null }) {
-    const newOptions = Object.assign({
+  async createAndSendTx(outputs: OutputRequestObj[], options: { inputs?: InputRequestObj[], changeAddress?: string } = {}): Promise<TxProposalUpdateResponseData> {
```
msbrogli, line 449: Giant method! Should we split some parts in separate methods for clarity and simpler unittests?

src/wallet/wallet.ts
```diff
@@ -354,44 +679,434 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+      // d. HTR change output
+      const changeAddressStr = newOptions.changeAddress || this.getCurrentAddress({ markAsUsed: true }).address;
+      const changeAddress = new Address(changeAddressStr, {network: this.network});
+      if (!changeAddress.isValid()) {
+        throw new SendTxError(`Address ${newOptions.changeAddress} is not valid.`);
+      }
+      outputsObj.push(new Output(changeAmount, changeAddress));
+    }
+
+
```
msbrogli, line 800: Empty line.

src/wallet/wallet.ts
```diff
@@ -354,44 +679,434 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+      // c. HTR change output
+      const changeAddressStr = newOptions.changeAddress || this.getCurrentAddress({ markAsUsed: true }).address;
+      const changeAddress = new Address(changeAddressStr, {network: this.network});
+      if (!changeAddress.isValid()) {
+        throw new SendTxError(`Address ${newOptions.changeAddress} is not valid.`);
+      }
+      outputsObj.push(new Output(changeAmount, changeAddress));
+    }
+
+
```
msbrogli, line 885: Empty line.

src/utils/transaction.ts
```diff
@@ -0,0 +1,92 @@
...
+   */
+  selectUtxos(utxos: Utxo[], totalAmount: number): {utxos: Utxo[], changeAmount: number} {
+    if (totalAmount <= 0) {
+      throw new UtxoError('Total amount must be a positive integer.');
+    }
+
+    if (utxos.length === 0) {
+      throw new UtxoError('Don\'t have enough utxos to fill total amount.');
+    }
+
```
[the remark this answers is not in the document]
andreabadesso: The wallet service will return them sorted by value already

src/wallet/api/walletApi.ts
```diff
@@ -16,46 +28,68 @@ import Input from '../../models/input';
...
       data['params']['token_id'] = token;
     }
-    return axiosInstance().get('balances', data);
+    const axios = await axiosInstance(wallet);
+    return axios.get('wallet/balances', data);
   },
 
-  getHistory(id: string, token: string | null = null): Promise<any> {
+  async getHistory(wallet: HathorWalletServiceWallet, token: string | null = null): Promise<HistoryResponse> {
     // TODO add pagination parameters
```
andreabadesso, line 62: Why is this a TODO? We are not going to paginate it yet?

src/wallet/wallet.ts
```diff
@@ -354,44 +679,434 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+   * Get the current address to be used
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  getCurrentAddress({ markAsUsed = false } = {}): AddressInfoObject {
+    const newAddressesLen = this.newAddresses.length;
+    if (this.indexToUse > newAddressesLen - 1) {
+      const addressInfo = this.newAddresses[newAddressesLen - 1];
+      return {...addressInfo, error: 'GAP_LIMIT_REACHED'};
```
andreabadesso, line 692: Shouldn't this throw instead of returning an object?

src/wallet/wallet.ts
```diff
@@ -354,44 +679,434 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+    const deposit = tokens.getDepositAmount(amount);
+
+    // 2. Get utxos for HTR
+    const { utxos, changeAmount } = await this.getUtxos({ tokenId: HATHOR_TOKEN_CONFIG.uid, totalAmount: deposit });
+    if (utxos.length === 0) {
+      throw new UtxoError(`No utxos available to fill the request. Token: HTR - Amount: ${deposit}.`);
+    }
+
+    // 3. Get mint authority
+    const ret = await this.getUtxos({ tokenId: token, authority: 1 });
```
andreabadesso, line 843: Use constants? `hathorLib.constants.TOKEN_MINT_MASK`

src/wallet/wallet.ts
```diff
@@ -354,44 +679,434 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+    const withdraw = tokens.getWithdrawAmount(amount);
+
+    // 2. Get utxos for custom token to melt
+    const { utxos, changeAmount } = await this.getUtxos({ tokenId: token, totalAmount: amount });
+    if (utxos.length === 0) {
+      throw new UtxoError(`Not enough tokens to be melted. Token: ${token} - Amount: ${amount}.`);
+    }
+
+    // 3. Get mint authority
+    const ret = await this.getUtxos({ tokenId: token, authority: 2 });
```
andreabadesso, line 931: Use constants? `hathorLib.constants.TOKEN_MELT_MASK`

andreabadesso (changes requested):

src/utils/transaction.ts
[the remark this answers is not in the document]
pedroferreira1: Yeah and the docstring of the method says that it expects the utxos to be already sorted by greatest value.

src/wallet/wallet.ts
```diff
@@ -65,35 +92,47 @@ class HathorWalletServiceWallet extends EventEmitter {
     this.walletId = null;
     this.isSendingTx = false;
     this.txProposalId = null;
+    this.xpub = null;
 
     this.network = network;
+
+    this.authToken = null;
+
+    // TODO When we integrate the real time events from the wallet service
```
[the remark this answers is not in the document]
pedroferreira1: Yes, we are. This PR was supposed to handle  only the token actions integration. I had to also do the authorization methods, otherwise it would not work. So I decided not to add any more features here and left the real time update for another PR.

src/wallet/wallet.ts
pedroferreira1: Done

src/wallet/wallet.ts
pedroferreira1: Yes. This PR was supposed to handle only the token actions integration. I had to also do the authorization methods, otherwise it would not work. So I decided not to add any more features and this TODO is already removed and implemented in the PR that matches the faces signatures.

src/wallet/api/walletApi.ts
pedroferreira1: Yes, we are. This PR was supposed to handle only the token actions integration. I had to also do the authorization methods, otherwise it would not work. So I decided not to add any more features and this TODO is already removed and implemented in the PR that matches the faces signatures

src/wallet/wallet.ts
pedroferreira1: To get the private key from the storage we would need a bigger refactor in the wallets and/or in the auth methods. RIght now we depend on the xpriv to create the jwt auth and this expires in 30 minutes. We need this for every wallet service request and this refactor need to be discussed before being implemented. I believe this is not the best implementation about handling the xpriv but I think this should be handled in another PR

src/wallet/wallet.ts
pedroferreira1: The wallet service will return error in any request if the wallet is not ready (i.e. if it's still creating, or error). Then I just have this method to prevent useless lambda calls when the wallet is not ready.

src/wallet/wallet.ts
pedroferreira1: Done

src/wallet/wallet.ts
pedroferreira1: We had a great discussion in terms of interfaces of the new facades and I've done a bigger refactor in the sendTransaction code in the new branch (the one that matches the signatures). I prefer keeping like this here and just do the refactor there, what do you think?

src/wallet/wallet.ts
pedroferreira1: Done

src/wallet/wallet.ts
pedroferreira1: Done

src/wallet/wallet.ts
pedroferreira1: At first I agreed with you but I was changing this part of the code to throw the error and I had second thoughts. The user that calls the method `getCurrentAddress` wants an address to use. It doesn't matter which one, we just need to warn this user that the gap limit was reached and from now on the address will always be the same until one is used. This warning could come from the error thrown but how would I return the address to the user? We could assume that the user already has a previous address but he might not have it, so even when it reaches the gap limit, I need to return the address in this method, that's why I think throwing an error might not be the best approach here. What do you think?

src/wallet/wallet.ts
pedroferreira1: This is a method that get a utxo from txId and index, so it must return one utxo, or null.

src/wallet/wallet.ts
pedroferreira1: Updated.

src/wallet/wallet.ts
pedroferreira1: It makes sense to be `signMessage`

src/utils/transaction.ts
pedroferreira1: It's much simpler and it works. I have unit tests for this code and it still works

src/wallet/wallet.ts
pedroferreira1: Makes sense. Did the refactor

src/wallet/wallet.ts
pedroferreira1: Makes sense, done.

src/wallet/wallet.ts
pedroferreira1: Makes sense, done.

src/wallet/wallet.ts
pedroferreira1: Makes sense, done.

src/wallet/wallet.ts
pedroferreira1: Makes sense, done.

src/wallet/wallet.ts
andreabadesso: I see, so the main point here is that this is not an error, this is expected behavior -- we reached the gap limit

I would just change the attribute `error` to something like `info`. What do you think?

pedroferreira1: I agree! Changed to info

andreabadesso (approved):

src/utils/transaction.ts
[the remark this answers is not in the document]
msbrogli: Great. Thanks!

src/wallet/wallet.ts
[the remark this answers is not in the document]
msbrogli: Ok. Just to confirm: Do we have issues to track both future developments?

src/wallet/wallet.ts
msbrogli: I agree. Just make sure we have another issue to get it improved in the future.

src/wallet/wallet.ts
msbrogli: As it has already been done there, no need to do it here. Thanks!

msbrogli (approved):

src/wallet/wallet.ts
msbrogli: This refactor is great. It is way easier to read the code. Thanks!

src/wallet/wallet.ts
msbrogli: This refactor is great. It is way easier to read the code. Thanks!

src/wallet/wallet.ts
msbrogli: This refactor is great. It is way easier to read the code. Thanks!

src/wallet/wallet.ts
msbrogli: As far as I checked, you've already refactored it as well, right?

src/wallet/wallet.ts
msbrogli: Ok. Same case as our other conversation.

src/wallet/wallet.ts
msbrogli: Should we assert that `utxos.length == 1`?

src/wallet/wallet.ts
msbrogli: Great! Thanks!

src/wallet/wallet.ts
```diff
@@ -65,208 +94,435 @@ class HathorWalletServiceWallet extends EventEmitter {
...
    * @inner
    */
-  async getAllAddresses(): Promise<string[]> {
-    const response = await walletApi.getAddresses(this.walletId!);
-    let addresses = [];
-    if (response.status === 200 && response.data.success === true) {
-      addresses = response.data.addresses;
-    } else {
-      // TODO What error should be handled here?
+  private checkWalletReady() {
```
msbrogli, line 174: I'd rename it to `failIfWalletNotReady()`. What do you think?

src/wallet/wallet.ts
msbrogli: Ok. I'd just rename it to `failIfWalletNotReady()`. What do you think?

src/wallet/wallet.ts
pedroferreira1: Yes! I forgot to reply this one, sorry

src/wallet/wallet.ts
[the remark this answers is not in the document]
pedroferreira1: Yes

src/wallet/wallet.ts
pedroferreira1: We have

src/wallet/wallet.ts
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/214/commits/e2f5bb90b67e0466d052f0398baf969b7cb12b2c

src/wallet/wallet.ts
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/214/commits/2c8f734fab81b253460db4cf59e6742742b2d119

PR #217, opened 2021-07-20
fix: node versions does not exist on the web browser

pedroferreira1:


src/utils/buffer.ts
```diff
@@ -81,7 +80,10 @@ export const unpackToInt = (n: number, signed: boolean, buff: Buffer): [number,
     }
   } else if (n === 8) {
     // We have only signed ints here
-    if (parseInt(nodeMajorVersion) > 8) {
+    const nodeMajorVersion = get(process, 'versions.node');
```
giovanecosta, line 83: Will this return undefined in browser? You can use a default value as 3rd arg or use `_.has` to check if this property exists

src/utils/buffer.ts
```diff
@@ -81,7 +80,10 @@ export const unpackToInt = (n: number, signed: boolean, buff: Buffer): [number,
...
+    // Browsers don't have access to versions.node and they don't have readBigInt64BE method
+    // slicedBuff.readIntBE(0, 8) worked fine testing on electron and react native (desktop and mobile wallets)
+    if (nodeMajorVersion && parseInt(nodeMajorVersion.split('.')[0]) > 8) {
```
giovanecosta, line 86: Also here you can check only `parseInt(nodeMajorVersion) > 8` because:
```
parseInt('12.34.32') > 8
true
parseInt(undefined) > 8
false
parseInt(null) > 8
false
parseInt('randomstr') > 8
false
parseInt('12.32.beta-rc') > 8
true
```

msbrogli, line 86: Why don't we just check whether `readBigInt64BE` exists? For example:

```js
if (slicedBuff.readBigInt64BE) {
  retInt = Number(slicedBuff.readBigInt64BE());
} else {
  retInt = slicedBuff.readIntBE(0, 8);
}
```

msbrogli (changes requested):

andreabadesso (approved):

src/utils/buffer.ts
pedroferreira1: This code has changed, can you see again please?

src/utils/buffer.ts
pedroferreira1: This code has changed, can you see again please?

pedroferreira1: Done

giovanecosta (approved):

msbrogli (approved):

PR #220, opened 2021-07-29
feat: generate wallet from xpub

r4mmer:


src/new/wallet.js
```diff
@@ -854,7 +857,7 @@ class HathorWallet extends EventEmitter {
     const options = Object.assign({ pinCode: null, password: null }, optionsParams);
     const pinCode = options.pinCode || this.pinCode;
     const password = options.password || this.password;
-    if (!pinCode) {
+    if (!this.xpub && !pinCode) {
```
pedroferreira1, line 860: There are more methods in this wallet class where you must have the xpriv to run, e.g. send tx, create token, mint, ... all those methods should be protected and throw an error if you try to call them and you have only the xpub.

pedroferreira1 (changes requested):

src/wallet.js
```diff
@@ -630,13 +651,17 @@ const wallet = {
    * @inner
    */
   changePin(oldPin, newPin) {
+    const accessData = this.getWalletAccessData();
+
+    if (accessData['from_xpub']) {
```
msbrogli, line 656: Create a method `isFromXpub()` and use it in all these verifications.

src/wallet.js
```diff
@@ -1511,6 +1536,9 @@ const wallet = {
    */
   getWalletWords(password) {
     const accessData = this.getWalletAccessData();
+    if (accessData['from_xpub']) {
```
msbrogli, line 1539: Use `isFromXpub()` here.

src/wallet.js
```diff
@@ -2435,6 +2463,11 @@ const wallet = {
    */
   getXprivKey(pin) {
     const accessData = this.getWalletAccessData();
+
+    if (accessData['from_xpub']) {
```
msbrogli, line 2467: Use `isFromXpub()` here.

msbrogli (approved):

src/new/wallet.js
r4mmer: Done

src/wallet.js
r4mmer: Done

msbrogli (approved):

pedroferreira1 (approved):

PR #221, opened 2021-07-30
feat: match both facades method signatures and some improvements to integrate in the wallet service rollout

pedroferreira1:


src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+    // in case it's not sent and (ii) create the token array of the tx
+    const amountOutputMap = {};
+    for (const output of this.outputs) {
+      if (output.token in amountOutputMap) {
+        amountOutputMap[output.token] += output.value;
+      } else {
+        amountOutputMap[output.token] = output.value;
+      }
+    }
+
```
andreabadesso, line 83: **Comment unrelated to line number, but related to the current method**

I feel that this method is big, we should add more comments explaining each step starting from this line

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+      const responseData = await walletApi.createTxProposal(this.wallet, txHex);
+      const txProposalId = responseData.txProposalId;
+      const sendData = await walletApi.updateTxProposal(this.wallet, txProposalId, txHex);
+      this.transaction.updateHash();
+      this.emit('send-tx-success', this.transaction);
+    } catch (err) {
+      if (err instanceof WalletRequestError) {
+        this.emit('send-error', 'Error sending tx proposal.');
+      } else {
+        throw err;
```
andreabadesso, line 288: What will happen if this throws? Shouldn't we emit `send-error` here as well?

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+   * @inner
+   */
+  runFromMining(until = null) {
+    try {
+      this.mineTx();
+      if (until === 'mine-tx') {
+        return;
+      }
+
+      this.once('mine-tx-ended', (data) => {
```
andreabadesso, line 309: Lint suggestion:

```
this.once('mine-tx-ended', () => this.handleSendTxProposal());
```

andreabadesso (changes requested):

src/new/wallet.js
```diff
@@ -533,7 +715,96 @@ class HathorWallet extends EventEmitter {
...
+   * Get balance for a token (same as old method to be used for compatibility)
+   *
+   * @params {string} tokenUid Token uid
+   *
+   * @return {Object} Object with balance { available, locked }
+   *
+   * @memberof HathorWallet
+   * @inner
+   **/
+  getOldBalance(tokenUid) {
```
msbrogli, line 728: Should we mark it as deprecated? I still didn't understand what's changing in the storage so we need this compatibility methods.

src/new/wallet.js
```diff
@@ -533,7 +715,96 @@ class HathorWallet extends EventEmitter {
...
+
+  /**
+   * Get full wallet history (same as old method to be used for compatibility)
+   *
+   * @return {Object} Object with transaction data { tx_id: { full_transaction_data }}
+   *
+   * @memberof HathorWallet
+   * @inner
+   **/
+  getOldHistory() {
```
msbrogli, line 743: Same here. What is changing that we need a compatibility method?

src/new/wallet.js
```diff
@@ -533,7 +715,96 @@ class HathorWallet extends EventEmitter {
...
+      // update token total balance
+      tokensBalance[tokenUid] = totalBalance;
+    }
+
+    // in the end, sort (in place) all tx lists in descending order by timestamp
+    for (const txList of Object.values(tokensHistory)) {
+      txList.sort((elem1, elem2) => elem2.timestamp - elem1.timestamp);
+    }
+
+    storage.setItem('old-facade:tokens', Object.keys(tokensHistory));
```
msbrogli, line 796: I don't like this `old-facade` prefix. What does it mean? I don't think we are planning to remove this facade so I don't get why we are using the term "old" here as well.

src/new/wallet.js
```diff
@@ -692,153 +869,29 @@ class HathorWallet extends EventEmitter {
...
-    }
-    const ret = {success: true, promise: sendTransaction.promise, sendTransaction};
-    if (this.debug) {
-      ret.debug = {
-        balanceHTR: this.getBalance(),
-        data: data,
-      };
-    }
-    return ret;
+    sendTransaction.run();
```
msbrogli, line 893: Should we return a promise in the `run()` method? So we can return it (instead of creating a new one).

src/new/wallet.js
```diff
@@ -692,153 +869,29 @@ class HathorWallet extends EventEmitter {
    *   'pinCode': pin to decrypt xpriv information. Optional but required if not set in this
    *  }
    *
-   * @return {Promise} Promise that resolves when transaction is sent
+   * @return {Promise<Transaction>} Promise that resolves when transaction is sent
    **/
-  sendManyOutputsTransaction(outputs, inputs = [], token = null, options = {}) {
+  sendManyOutputsTransaction(outputs, options = {}) {
```
msbrogli, line 874: `async`?

src/new/wallet.js
```diff
@@ -925,42 +978,96 @@ class HathorWallet extends EventEmitter {
...
+      sendTransaction.on('send-tx-success', (transaction) => {
+        resolve(transaction);
+      });
+
+      sendTransaction.on('send-error', (err) => {
+        reject(err);
+      });
+    });
+
+    sendTransaction.runFromMining();
```
msbrogli, line 1004: Same here about returning a promise.

src/new/wallet.js
```diff
@@ -1081,12 +1188,16 @@ class HathorWallet extends EventEmitter {
...
-   * @return {Object} Object with {success: true, sendTransaction, promise}, where sendTransaction is a
-   * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
+   * @return {Promise} Promise that resolves with transaction object if succeeds
+   * or with error message if it fails
+   *
+   * @memberof HathorWallet
+   * @inner
    **/
-  mintTokens(tokenUid, amount, address, options = {}) {
+  prepareMintTokensData(tokenUid, amount, options = {}) {
```
msbrogli, line 1197: `async`?

src/new/wallet.js
```diff
@@ -1095,50 +1206,73 @@ class HathorWallet extends EventEmitter {
...
-   * @return {Object} Object with {success: true, sendTransaction, promise}, where sendTransaction is a
-   * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
+   * @return {Promise} Promise that resolves with transaction object if succeeds
+   * or with error message if it fails
+   *
+   * @memberof HathorWallet
+   * @inner
    **/
-  meltTokens(tokenUid, amount, options = {}) {
+  prepareMeltTokensData(tokenUid, amount, options = {}) {
```
msbrogli, line 1272: `async`?

src/new/wallet.js
```diff
@@ -1182,15 +1337,18 @@ class HathorWallet extends EventEmitter {
...
-   * @return {Object} Object with {success: true, sendTransaction, promise}, where sendTransaction is a
-   * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
+   * @return {Promise} Promise that resolves with transaction object if succeeds
+   * or with error message if it fails
+   *
+   * @memberof HathorWallet
+   * @inner
    **/
-  delegateAuthority(tokenUid, type, destinationAddress, options = {}) {
+  prepareDelegateAuthorityData(tokenUid, type, destinationAddress, options = {}) {
```
msbrogli, line 1346: `async`?

src/new/wallet.js
```diff
@@ -1235,15 +1413,18 @@ class HathorWallet extends EventEmitter {
...
-   * @return {Object} Object with {success: true, sendTransaction, promise}, where sendTransaction is a
-   * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
+   * @return {Promise} Promise that resolves with transaction object if succeeds
+   * or with error message if it fails
+   *
+   * @memberof HathorWallet
+   * @inner
    **/
-  destroyAuthority(tokenUid, type, count, options = {}) {
+  prepareDestroyAuthorityData(tokenUid, type, count, options = {}) {
```
msbrogli, line 1422: `async`?

src/new/wallet.js
```diff
@@ -195,22 +195,118 @@ class HathorWallet extends EventEmitter {
...
+  /**
+   * Old getAllAddresses method used to keep compatibility
+   * with some methods that used to need it
+   *
+   * @return {Array} Array of addresses (string)
+   *
+   * @memberof HathorWallet
+   * @inner
+   **/
+  getAllAddressesSimple() {
```
msbrogli, line 207: What does simple mean?

src/wallet/sendTransaction.ts
msbrogli: I agree. Instead of comments, I feel we should split it in several methods that can be independently tested.

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    try {
+      const responseData = await walletApi.createTxProposal(this.wallet, txHex);
+      const txProposalId = responseData.txProposalId;
+      const sendData = await walletApi.updateTxProposal(this.wallet, txProposalId, txHex);
+      this.transaction.updateHash();
+      this.emit('send-tx-success', this.transaction);
+    } catch (err) {
+      if (err instanceof WalletRequestError) {
+        this.emit('send-error', 'Error sending tx proposal.');
```
msbrogli, line 286: Should we emit the event and omit the exception? I feel that we should let the exception be thrown and handled by the caller.

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+      if (until === 'mine-tx') {
+        return;
+      }
+
+      this.once('mine-tx-ended', (data) => {
+        this.handleSendTxProposal();
+      });
+    } catch (err) {
+      if (err instanceof WalletError) {
+        this.emit('send-error', err);
```
msbrogli, line 314: Should we omit the exception?

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+  /**
+   * Run sendTransaction from mining, i.e. expects this.transaction to be prepared and signed
+   * then it will mine and handle tx proposal
+   *
+   * 'until' parameter can be 'mine-tx', in order to only mine the transaction without propagating
+   *
+   * @memberof SendTransactionWalletService
+   * @inner
+   */
+  runFromMining(until = null) {
```
msbrogli, line 302: I feel it could return a promise that will be resolved or rejected accordingly. What do you think?

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+  /**
+   * Run sendTransaction from preparing, i.e. prepare, sign, mine and send the tx
+   *
+   * 'until' parameter can be 'prepare-tx' (it will stop before signing the tx), 'sign-tx' (it will stop before mining the tx),
+   * or 'mine-tx' (it will stop before send tx proposal, i.e. propagating the tx)
+   *
+   * @memberof SendTransactionWalletService
+   * @inner
+   */
+  async run(until = null) {
```
msbrogli, line 330: I feel it could return a promise that will be resolved or rejected accordingly. What do you think?

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+
+      this.signTx(preparedData.utxosAddressPath);
+      if (until === 'sign-tx') {
+        return;
+      }
+
+      this.runFromMining(until);
+    } catch (err) {
+      if (err instanceof WalletError) {
+        this.emit('send-error', err);
```
msbrogli, line 345: Why are we omitting this exception?

src/new/sendTransaction.js
```diff
@@ -6,18 +6,15 @@
...
+import oldHelpers from '../helpers';
 import storage from '../storage';
-import Input from '../models/input';
-import Output from '../models/output';
-import Transaction from '../models/transaction';
-import CreateTokenTransaction from '../models/create_token_transaction';
-import Address from '../models/address';
 import MineTransaction from '../wallet/mineTransaction';
 
 /**
```
msbrogli, line 20: Comment not related to this line.

Should we create an interface that is implemented by both `SendTransaction` classes?

src/new/sendTransaction.js
```diff
@@ -172,15 +276,72 @@ class SendTransaction extends EventEmitter {
...
-   * Start object (submit job)
+   * Run sendTransaction from mining, i.e. expects this.transaction to be prepared and signed
+   * then it will mine and push tx
+   *
+   * 'until' parameter can be 'mine-tx', in order to only mine the transaction without propagating
+   *
+   * @memberof SendTransaction
+   * @inner
+   */
+  runFromMining(until = null) {
```
msbrogli, line 291: Shouldn't it return a promise that will be resolved or rejected accordingly?

src/new/sendTransaction.js
```diff
@@ -172,15 +276,72 @@ class SendTransaction extends EventEmitter {
...
+  /**
+   * Run sendTransaction from preparing, i.e. prepare, sign, mine and push the tx
+   *
+   * 'until' parameter can be 'prepare-tx' (it will stop before signing the tx),
+   * or 'mine-tx' (it will stop before send tx proposal, i.e. propagating the tx)
+   *
+   * @memberof SendTransaction
+   * @inner
+   */
+  run(until = null) {
```
msbrogli, line 330: Shouldn't it return a promise that will be resolved or rejected accordingly?

src/wallet/sendTransaction.ts
```diff
@@ -0,0 +1,353 @@
...
+import { OutputRequestObj, InputRequestObj } from './types';
+
+type optionsType = {
+  outputs?: OutputRequestObj[],
+  inputs?: InputRequestObj[],
+  changeAddress?: string | null,
+  transaction?: Transaction | null,
+};
+
+class SendTransactionWalletService extends EventEmitter {
```
msbrogli, line 28: Maybe we should rename the file to `sendTransactionWalletService.ts` or move it to `wallet-service/sendTransaction.ts`. What do you think?

msbrogli (changes requested):

src/new/sendTransaction.js
```diff
@@ -41,63 +38,179 @@ class SendTransaction extends EventEmitter {
...
+        reject(message);
+      });
+
+      this.on('unexpected-error', (message) => {
+        reject(message);
+      });
+    });
+
+    // Error to be shown in case of an unexpected error when executing push tx
+    this.unexpectedPushTxError = 'An unexpected error happened. Check if the transaction has been sent looking into the history and try again if it hasn\'t.';
```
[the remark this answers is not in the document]
pedroferreira1: Done [here](https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/25fbb32c518a7f854f8354487cd12c3932b27b72)

src/new/sendTransaction.js
```diff
@@ -41,63 +38,179 @@ class SendTransaction extends EventEmitter {
...
+   * @params {Object} options Optional object with {'startMiningTx', 'maxTxMiningRetries'}
+   *
+   * @throws WalletError
+   *
+   * @memberof SendTransaction
+   * @inner
+   */
+  mineTx(options = {}) {
+    if (this.transaction === null) {
+      throw new WalletError('Can\'t mine transaction if it\'s null.');
```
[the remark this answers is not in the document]
pedroferreira1: Done [here](https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/25fbb32c518a7f854f8354487cd12c3932b27b72)

src/utils/helpers.ts
```diff
@@ -296,6 +298,70 @@ const helpers = {
...
+        options
+      );
+    } else if (data.version === DEFAULT_TX_VERSION) {
+      return new Transaction(
+        inputs,
+        outputs,
+        options
+      );
+    } else {
+        throw new ParseError('We currently support only the Transaction and CreateTokenTransaction types. Other types will be supported in the future.');
```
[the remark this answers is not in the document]
pedroferreira1: Done [here](https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/25fbb32c518a7f854f8354487cd12c3932b27b72)

src/wallet/sendTransaction.ts
pedroferreira1: I don't think we should emit. I have some expected errors in the `try..catch` block. I feel we should capture the specific error to emit and just throw if something unexpected happens.

src/new/sendTransaction.js
pedroferreira1: When I started doing this task I created the interface but javascript code does not have any interface implementation. Because of that, we would not be able to use it to enforce the same method returns/signatures on it, then I decided not to do it. 

I can create it and the classes on typescript may implement it but the ones still using javascript wouldn't benefit from it.

(This comment is valid for both `SendTransaction` and `HathorWallet` classes)

src/wallet/sendTransaction.ts
pedroferreira1: I agree with that if we decide to move the files in `src/new/` -> `src/wallet/`, otherwise I see no benefits on doing that. Should we create an issue to migrate the old facade to this new folder and migrate the code to typescript? I believe we will need to continue having both facades running, right?

src/wallet/wallet.ts
```diff
@@ -21,20 +21,22 @@ import Input from '../models/input';
...
-  TxProposalUpdateResponseData,
   SendManyTxOptionsParam,
   SendTxOptionsParam,
   WalletStatus,
   Utxo,
   OutputRequestObj,
-  InputRequestObj
+  InputRequestObj,
+  SendTransactionEvents,
+  SendTransactionResponse
```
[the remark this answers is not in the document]
pedroferreira1: When I started doing this task I created the interface but javascript code does not have any interface implementation. Because of that, we would not be able to use it to enforce the same method returns/signatures on it, then I decided not to do it.

I can create it and the classes on typescript may implement it but the ones still using javascript wouldn't benefit from it.

(This comment is valid for both SendTransaction and HathorWallet classes)

src/wallet/api/walletApi.ts
```diff
@@ -104,12 +92,18 @@ const walletApi = {
...
 
-  async getHistory(wallet: HathorWalletServiceWallet, token: string | null = null): Promise<HistoryResponseData> {
-    // TODO add pagination parameters
-    const data = { params: {} };
-    if (token) {
-      data['params']['token_id'] = token;
+  async getTokens(wallet: HathorWalletServiceWallet): Promise<TokensResponseData> {
+    const axios = await axiosInstance(wallet);
+    const response = await axios.get('wallet/tokens');
+    if (response.status === 200 && response.data.success === true) {
```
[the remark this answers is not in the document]
pedroferreira1: Actually not. If we decide to resolve promises only for 2xx the code will be like that:

```
try{
    const response = await axios.get('wallet/addresses');
    if (response.data.success === true) {
      return response.data;
    } else {
      throw new WalletRequestError('Error getting wallet addresses.');
    }
catch (e) {
  throw new WalletRequestError(e.message);
}
```

The `if...else` would still be needed because of the `response.data.success`

src/new/wallet.js
```diff
@@ -195,22 +195,118 @@ class HathorWallet extends EventEmitter {
...
 
+  /**
+   * Return all addresses of the wallet with info of each of them
+   *
+   * @return {Promise<Array>} Array of objects { address, index, transactions } where transactions is the count of txs for this address
+   *
+   * @memberof HathorWallet
+   * @inner
+   **/
+  async getAllAddresses() {
```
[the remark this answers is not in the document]
pedroferreira1: The idea is to have both methods (from both facades) working the same way. The one from wallet service just returns the array of addresses. I could change the other one as well but not sure if we should, the wallet service API already returns the full array anyway

src/new/wallet.js
```diff
@@ -231,18 +326,104 @@ class HathorWallet extends EventEmitter {
...
+   *   balance: {unlocked. locked},
+   *   transactions: number,
+   *   lockExpires: number | null,
+   *   tokenAuthorities: {unlocked: {mint, melt}. locked: {mint, melt}}
+   * }
+   *
+   * @memberof HathorWallet
+   * @inner
+   **/
+  getTxHistory(options = {}) {
```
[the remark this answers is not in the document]
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
```diff
@@ -231,18 +326,104 @@ class HathorWallet extends EventEmitter {
...
+
+  /**
+   * Get tokens that this wallet has transactions
+   *
+   * @return {Promise<Array>} Array of strings (token uid)
+   *
+   * @memberof HathorWallet
+   * @inner
+   **/
+  getTokens() {
```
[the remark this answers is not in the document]
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
```diff
@@ -925,42 +977,96 @@ class HathorWallet extends EventEmitter {
...
+   *
+   * @param {Transaction} transaction Transaction object to be mined and pushed to the network
+   *
+   * @return {Promise} Promise that resolves with transaction object if succeeds
+   * or with error message if it fails
+   *
+   * @memberof HathorWallet
+   * @inner
+   */
+  handleSendPreparedTransaction(transaction) {
```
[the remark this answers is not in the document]
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/2f4b0486a70bf36ddec081d60a3202f22f2393a6

src/new/wallet.js
```diff
@@ -533,7 +714,96 @@ class HathorWallet extends EventEmitter {
...
+    }
+
+    const tokensBalance = {};
+    for (const tokenUid of Object.keys(tokensHistory)) {
+      const totalBalance = this.getOldBalance(tokenUid);
+      // update token total balance
+      tokensBalance[tokenUid] = totalBalance;
+    }
+
+    // in the end, sort (in place) all tx lists in descending order by timestamp
```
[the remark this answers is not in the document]
pedroferreira1: The other one I added but this one I don't think should be on the docstring. I think it's a comment about a line in the code only. Why do you think it should be in a remark?

src/wallet/sendTransaction.ts
pedroferreira1: I added more comments and split the method into some smaller ones. Can you check please?
https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/d7e7446efd4b3b7ab2f1e026a12f8f9427baadf0

src/wallet/sendTransaction.ts
pedroferreira1: I am omitting the exception if we use the event handler object. The normal use in the wallet would be to call

```
try {
  wallet.sendTransaction();
} catch (e) {
  // handle send transaction error
}
```

I thought about continue throwing the exception in the event emitter class but I feel it's not bad to handle exceptions listening to the `send-error` event

src/wallet/sendTransaction.ts
pedroferreira1: This is exactly what's being done in the wallet send transaction method.

```
  async sendManyOutputsTransaction(outputs: OutputRequestObj[], options: { inputs?: InputRequestObj[], changeAddress?: string } = {}): Promise<Transaction | string> {
    this.failIfWalletNotReady();
    const newOptions = Object.assign({
      inputs: [],
      changeAddress: null
    }, options);
    const { inputs, changeAddress } = newOptions;
    const sendTransaction = new SendTransactionWalletService(this, { outputs, inputs, changeAddress });
    const promise: Promise<Transaction | string> = new Promise((resolve, reject) => {
      sendTransaction.on('send-tx-success', (transaction) => {
        resolve(transaction);
      });

      sendTransaction.on('send-error', (err) => {
        reject(err);
      });
    });

    sendTransaction.run();
    return promise;
  }
```

At first I thought about having the sendTransaction only with the events. Should we add the promise also here?

src/new/wallet.js
```diff
@@ -231,18 +327,104 @@ class HathorWallet extends EventEmitter {
...
+   **/
+  getBalance(token = null) {
     storage.setStore(this.store);
-    const uid = tokenUid || this.token.uid;
-    const historyTransactions = this.getTxHistory();
-    return wallet.calculateBalance(Object.values(historyTransactions), uid);
+    // TODO if token is null we should get the balance for each token I have
+    // but we don't use it in the wallets, so I won't implement it
+    const uid = token || this.token.uid;
+    const balanceByToken = storage.getItem('old-facade:balanceByToken');
```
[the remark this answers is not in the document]
pedroferreira1: As we discussed in the call earlier today, I've move the pre processed data from localStorage to memory. Refactor done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/075b763e73f445bf23624eeb5d9b3600df061def

src/new/wallet.js
```diff
@@ -533,7 +714,96 @@ class HathorWallet extends EventEmitter {
...
+      // update token total balance
+      tokensBalance[tokenUid] = totalBalance;
+    }
+
+    // in the end, sort (in place) all tx lists in descending order by timestamp
+    for (const txList of Object.values(tokensHistory)) {
+      txList.sort((elem1, elem2) => elem2.timestamp - elem1.timestamp);
+    }
+
+    storage.setItem('old-facade:tokens', Object.keys(tokensHistory));
```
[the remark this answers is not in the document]
pedroferreira1: As we discussed in the call earlier today, I've move the pre processed data from localStorage to memory. Refactor done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/075b763e73f445bf23624eeb5d9b3600df061def

src/new/wallet.js
pedroferreira1: This method is still used as an auxiliar method in the whole facade and it's needed because the return is different from the wallet service return, it's much simpler. I agree that `getOldHistory` is a bad name, should I call it `getSimpleHistory`?

src/wallet/sendTransaction.ts
andreabadesso: Looking great to me :+1:

src/new/wallet.js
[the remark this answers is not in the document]
andreabadesso: You are right, this will just sort the `tokensHistory` in place, it is not being returned by the function

andreabadesso (approved):

src/wallet/wallet.ts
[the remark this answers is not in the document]
msbrogli: I see. I feel that we should do it to explicitly define the interface and understand the impacts of future changes. Regarding the javascript file, I hope it will be migrated to type script and we might benefit from the linter in the future.

src/wallet/wallet.ts
```diff
@@ -359,166 +364,48 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+      sendTransaction.on('send-tx-success', (transaction) => {
+        resolve(transaction);
+      });
+
+      sendTransaction.on('send-error', (err) => {
+        reject(err);
+      });
+    });
+
+    sendTransaction.run();
```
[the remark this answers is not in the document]
msbrogli: Did you see this conversation?

src/wallet/wallet.ts
```diff
@@ -671,12 +534,35 @@ class HathorWalletServiceWallet extends EventEmitter {
...
+      sendTransaction.on('send-tx-success', (tx) => {
+        resolve(tx);
+      });
+
+      sendTransaction.on('send-error', (err) => {
+        reject(err);
+      });
+    });
+
+    sendTransaction.runFromMining();
```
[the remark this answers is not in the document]
msbrogli: Did you see this conversation?

src/wallet/sendTransaction.ts
msbrogli: I guess we will need both facades from a long time. So I agree that we should refactor these files in the future. This change won't be compatible and will require update from use cases.

My points about renaming the file is that one might think that `SendTransactionWalletService` is the only one available since it is looking into the generic file `sendTransaction.ts`. Maybe we should move the code to `sendTransactionWalletService.ts` and keep just the export in the `sendTransaction.ts`. So, in the future, we can have both implementation available from the same file (e.g. `import { sendTransactionFullNode } from '@hathor/wallet-lib/wallet/sendTransaction';`).

src/wallet/sendTransaction.ts
msbrogli: Well done. Thanks!

src/wallet/sendTransaction.ts
msbrogli: Make sense. Thanks!

src/wallet/sendTransaction.ts
msbrogli: I feel it could return the promise and it would be used by `sendManyOutputsTransaction()` and others. I mean, the user can decide what's the best approach for their case and use the promise or the events accordingly.

src/new/wallet.js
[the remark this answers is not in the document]
msbrogli: I feel that an iterator would make the use simpler because we might need to paginate large returns, and we wouldn't have to request all pages before returning. Besides that, it might use less memory in cases with a large number of addresses.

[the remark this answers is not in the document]
msbrogli: I think we should do it now because it seems quick to be develop and it would barely affect the current code. It just keeps things more flexible for future features.

src/new/wallet.js
msbrogli: I'd say `_getBalance()` or `_getRawBalance()`. What do you think?

src/new/wallet.js
msbrogli: Maybe `_getRawHistory()`?

src/new/sendTransaction.js
msbrogli: I got it. I still think it is valuable to have an explicit interface implemented by all classes. I have the same considerations that I told in the other conversation.

msbrogli (approved):

src/new/wallet.js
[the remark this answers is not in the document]
msbrogli: Well done! Great!

src/new/wallet.js
msbrogli: Well done! Great!

src/wallet/api/walletApi.ts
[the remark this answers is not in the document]
msbrogli: You're right. Thanks!

src/wallet/api/walletServiceAxios.ts
```diff
@@ -26,6 +26,12 @@ export const axiosInstance = async (wallet: HathorWalletServiceWallet | null = n
   const defaultOptions = {
     baseURL: WALLET_SERVICE_BASE_URL,
     timeout: timeout,
+    // `validateStatus` defines whether to resolve or reject the promise for a given
+    // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
+    // or `undefined`), the promise will be resolved; otherwise, the promise will be
+    // rejected. The default behaviour of axios is to reject anything different than 2xx
+    // We need to handle some 400 manually (e.g. create wallet might already be loaded)
+    validateStatus: (status) => status >= 200 && status < 500,
```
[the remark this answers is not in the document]
pedroferreira1: I'm not sure I agree. We want to be able to handle some errors accordingly with what we have in the wallet service, e.g. the 400 error in the load wallet that we want to manually handle it. Why should we leave axios to reject it if we need to handle them?

src/new/wallet.js
pedroferreira1: Changed to `_getBalanceRaw`

src/new/wallet.js
pedroferreira1: Changed to `_getHistoryRaw`

src/wallet/sendTransaction.ts
pedroferreira1: The same as in this comment here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/221#discussion_r687262959

src/wallet/sendTransaction.ts
pedroferreira1: The same as in this comment here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/221#discussion_r687262959

src/wallet/sendTransaction.ts
pedroferreira1: Yeah, I think I agree with you. Refactor done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/new/sendTransaction.js
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/new/sendTransaction.js
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/wallet/sendTransaction.ts
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/new/wallet.js
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/new/wallet.js
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/wallet/wallet.ts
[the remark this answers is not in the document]
pedroferreira1: Yes, I saw it but it was a comment similar to another and I was waiting for the other conclusion. I did the refactor here 
 https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/wallet/wallet.ts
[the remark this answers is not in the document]
pedroferreira1: Yes, I saw it but it was a comment similar to another and I was waiting for the other conclusion. I did the refactor here 
 https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/43cdd56eebf3e443132890aaddd495d24aaf24f5

src/wallet/sendTransaction.ts
pedroferreira1: Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/fa74421d218315e450eb1f4a542f127eae22ac9c

src/new/wallet.js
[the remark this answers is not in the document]
pedroferreira1: I agree. Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/61783cc94fbfb8328609a96417c0fffa6af8da2c

src/new/sendTransaction.js
pedroferreira1: Makes sense, done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/9dd1b760fdfdce10f2f4022fad206998e1777527

src/wallet/wallet.ts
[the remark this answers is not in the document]
pedroferreira1: Makes sense, done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/9dd1b760fdfdce10f2f4022fad206998e1777527

src/wallet/api/walletServiceAxios.ts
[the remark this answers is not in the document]
msbrogli: Can't you use something like this? I mean, I don't think we should change the default behavior because just one API needs a different behavior. https://github.com/axios/axios#handling-errors

src/wallet/wallet.ts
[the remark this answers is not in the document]
msbrogli: That's great. Thanks!

src/new/wallet.js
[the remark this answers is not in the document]
msbrogli: That's great! Thanks!

src/wallet/wallet.ts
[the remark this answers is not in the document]
msbrogli: That's great! Thanks!

src/new/wallet.js
msbrogli: That's great! Thanks!

src/new/wallet.js
msbrogli: That's great! Thanks!

src/wallet/wallet.ts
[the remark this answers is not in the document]
msbrogli: That's great! Thanks!

src/wallet/api/walletApi.ts
```diff
@@ -46,27 +47,14 @@ const walletApi = {
...
-          throw new WalletRequestError('Error creating wallet.');
-        }
-      } else {
-        throw new WalletRequestError('Error creating wallet.');
-      }
+    const response = await axios.post('wallet/init', data);
+    if (response.status === 200 && response.data.success) {
+      return response.data;
+    } else if (response.status === 400 && response.data.error === 'wallet-already-loaded') {
+      // If it was already loaded, we have to check if it's ready
```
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

src/wallet/sendTransaction.ts
msbrogli: That's great! Thanks!

src/wallet/sendTransaction.ts
msbrogli: That's great! Thanks!

src/new/sendTransaction.js
msbrogli: Well done! Thanks!

src/new/sendTransaction.js
msbrogli: That's great! Thanks!

src/new/sendTransaction.js
msbrogli: That's great! Thanks!

msbrogli (approved):

src/wallet/api/walletServiceAxios.ts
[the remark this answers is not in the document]
pedroferreira1: Yeah, makes sense. Done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/221/commits/ac0b4b4d5f9f57d0aedff47e5d4f18baf3b7a00b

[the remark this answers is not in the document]
msbrogli: Great! Thanks!

PR #229, opened 2021-08-08
feat: auto retry when there is a client timeout when loading wallet history

pedroferreira1:


src/wallet.js
```diff
@@ -435,7 +435,19 @@ const wallet = {
...
       while (hasMore === true) {
         let response;
-        response = await walletApi.getAddressHistoryForAwait(addressesToSearch, firstHash);
+        try {
+          response = await walletApi.getAddressHistoryForAwait(addressesToSearch, firstHash);
+        } catch (e) {
+          // We will retry the request that fails with client timeout
+          // in this request error we don't have the response because
+          // the client closed the connection
+          if (e.code === 'ECONNABORTED' && e.response === undefined && e.message.toLowerCase().includes('timeout')) {
```
msbrogli, line 444: Should we always keep trying? Any reason to give up if there's another error from the server? I mean, I feel we can safely remove the `timeout` substring in the error message. And why does the response must be undefined?

msbrogli (changes requested):

src/wallet.js
pedroferreira1: I am not sure we should keep trying if we receive a 400, 500, or even 429 (rate limit, in that case maybe we should). That's why I tried to identify when we have a client timeout, that's why I check if the response is undefined, because the client closed the connection.

Do you think we should retry forever any request error?

pedroferreira1: As we've discussed in a call, I tried to make it easier to identify a client timeout error but it's not easy with what axios have. (I've added more comments explaining the problems I had and some github issues about the problems)

Besides that, I've added a retry (limited) for any other error different from client timeout.

andreabadesso: I don't like the idea of the **lib** retrying forever, without giving feedback to the consumer

I think we should fail with a reason and let the client decide if he wants to try again

andreabadesso (changes requested):

src/wallet.js
pedroferreira1: I think I agree with @andreabadesso that the best approach would be to have a way for the client to handle the errors as needed (with a retry or throwing an error). However I don't see an easy way to do this with the current code, a bigger refactor would need to be implemented in order to achieve this. I see 3 options for us:

1. Dedicate more time to this task to implement the bigger refactor and do as @andreabadesso suggested.
2. Given that the current solution is not the best and there are other task with more priority, stop this task until we have more time and this becomes a priority.
3. Accept the current implemented solution as a first step in order to prevent any other problems while loading the wallets and create an issue to implement the better solution later.

In my opinion we should go with 3. What do you think @andreabadesso @msbrogli ?

msbrogli: I agree that we can move forward with proposal 3. So, let's create a new issue to refactor and add this feature later.

msbrogli: My point is that this would be a new feature even without this change. If an error occurs, users must restart the load anyway. In case of an "infinity loop" during the load, users will have the feedback from the screen because the number of loaded addresses will freeze.

msbrogli (approved):

src/wallet.js
andreabadesso: I agree with creating an issue to refactor this later and moving forward for now

My point is that eventually the wallet-lib will be used by services other than our wallets, and having a method that loops indefinitely when timeouts occur (maybe because of a proxy misconfiguration) looks like a bad idea to me

andreabadesso (approved):

src/wallet.js
pedroferreira1: Issue created: https://github.com/HathorNetwork/hathor-wallet-lib/issues/236

PR #232, opened 2021-08-09
refactor: output model now does not have address anymore, it has a buffer script that can be decoded

pedroferreira1:


src/wallet/sendTransaction.ts
```diff
@@ -166,8 +167,10 @@ class SendTransactionWalletService extends EventEmitter {
         throw new SendTxError(`Address ${o.address} is not valid.`);
       }
       const tokenData = (o.token in tokens) ? tokens.indexOf(o.token) + 1 : 0;
-      const outputOptions = { tokenData, timelock: o.timelock || null };
-      outputsObj.push(new Output(o.value, address, outputOptions));
+      const outputOptions = { tokenData };
+      const p2pkh = new P2PKH(address, { timelock: o.timelock || null });
+      const p2pkhScript = p2pkh.createScript()
+      outputsObj.push(new Output(o.value, p2pkhScript, outputOptions));
```
msbrogli, line 173: Maybe `p2pkh.createOutput()`?

msbrogli (approved):
Well done! \o/

src/wallet/sendTransaction.ts
pedroferreira1: Do you mean to create a method in P2PKH class that receives value and tokenData and creates an output from a P2PKH object?

pedroferreira1: As we've discussed in a call there is no need to create this method, it wouldn't help much.

andreabadesso (approved):

PR #243, opened 2021-08-19
fix: small fixes when sending transactions (mint/melt as well) with new signatures

pedroferreira1:
After the merge of the new facade signatures to dev and the refactors that were done, some specific problems were found:

1. Add handling of error when the input sent was from a token that was not in the outputs
2. Mint and melt manual address selection had wrong signature (comparing to new wallet service facade and IHathorWallet interface)
3. Utxo consolidation method was not updated to new sendManyOutputsTransaction method signature and return.

### Acceptance criteria

All APIs of the wallet headless must work fine will all parameter options possible.

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -662,7 +662,7 @@ class HathorWallet extends EventEmitter {
       }
       const utxo = utxoDetails.utxos[i];
       inputs.push({
-        tx_id: utxo.tx_id,
+        txId: utxo.tx_id,
         index: utxo.index,
       });
       utxos.push(utxo);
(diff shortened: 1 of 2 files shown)


src/new/wallet.js
```diff
@@ -662,7 +662,7 @@ class HathorWallet extends EventEmitter {
       }
       const utxo = utxoDetails.utxos[i];
       inputs.push({
-        tx_id: utxo.tx_id,
+        txId: utxo.tx_id,
```
msbrogli, line 665: I just saw `tx_id` below. Should we change all to the same pattern? It seems a good opportunity to do it since we will have to update all wallets to use the new signatures.

msbrogli (approved):

src/new/wallet.js
pedroferreira1: Yeah, I guess you are right. Done.

andreabadesso (approved):

PR #251, opened 2021-08-30
refactor: passing network and deciding what url to use on wallet-service

andreabadesso:
# Acceptance criteria

The app must connect to the testnet wallet service instance when the network is `testnet`

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -226,7 +226,8 @@ export const SELECT_OUTPUTS_TIMEOUT = 1000 * 60;
 /**
  * Wallet service URL
  */
-export const WALLET_SERVICE_BASE_URL  = 'https://wallet-service.testnet.hathor.network/';
+export const WALLET_SERVICE_BASE_URL  = 'https://wallet-service.hathor.network/';
+export const WALLET_SERVICE_TESTNET_BASE_URL  = 'https://wallet-service.testnet.hathor.network/';
 
 /**
  * Size in bytes of a transaction hash (32 bytes)
--- a/src/wallet/api/walletApi.ts
+++ b/src/wallet/api/walletApi.ts
@@ -32,7 +32,7 @@ import { WalletRequestError } from '../../errors';
 
 const walletApi = {
   async getWalletStatus(wallet: HathorWalletServiceWallet): Promise<WalletStatusResponseData> {
-    const axios = await axiosInstance(wallet.network, wallet);
+    const axios = await axiosInstance(wallet, true);
     const response = await axios.get('wallet/status');
     const data = response.data;
     if (response.status === 200 && data.success) {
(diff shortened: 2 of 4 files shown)


src/wallet/api/walletServiceAxios.ts
```diff
@@ -46,4 +51,17 @@ export const axiosInstance = async (wallet: HathorWalletServiceWallet | null = n
   return axios.create(defaultOptions);
 }
 
-export default axiosInstance;
\ No newline at end of file
+/**
+ * Returns the correct base url constant for wallet service based on the network
+ *
+ * @param {Network} network The network, can be either mainnet or testnet but will default to testnet
```
pedroferreira1, line 57: There is no default, the parameter is required, right?

src/wallet/api/walletServiceAxios.ts
```diff
@@ -20,11 +25,11 @@ import HathorWalletServiceWallet from '../wallet';
  *
  * @param {number} timeout Timeout in milliseconds for the request
  */
-export const axiosInstance = async (wallet: HathorWalletServiceWallet | null = null, timeout: number = TIMEOUT) => {
+export const axiosInstance = async (wallet: HathorWalletServiceWallet | null = null, network: Network, timeout: number = TIMEOUT) => {
```
pedroferreira1, line 28: Should we refactor the parameters and set `network` as the first and only required parameter and add `wallet` and `timeout` as an optional options parameter just like we've been doing with optional parameters in the lib?

pedroferreira1 (changes requested):

src/wallet/api/walletServiceAxios.ts
andreabadesso: Yes.

What do you think about it?

src/wallet/api/walletServiceAxios.ts
andreabadesso: I agree, done in 2b50b0be879e9da40172f4463b9e63dcd783a899

src/wallet/api/walletServiceAxios.ts
pedroferreira1: I like it.

src/wallet/api/walletApi.ts
```diff
@@ -8,6 +8,7 @@
 import { axiosInstance } from './walletServiceAxios';
 import Output from '../../models/output';
 import Input from '../../models/input';
+import Network from '../../models/network';
```
pedroferreira1, line 11: Are we using this?

pedroferreira1 (approved):

src/constants.js
```diff
@@ -226,7 +226,8 @@ export const SELECT_OUTPUTS_TIMEOUT = 1000 * 60;
 /**
  * Wallet service URL
  */
-export const WALLET_SERVICE_BASE_URL  = 'https://wallet-service.testnet.hathor.network/';
+export const WALLET_SERVICE_BASE_URL  = 'https://wallet-service.hathor.network/';
```
msbrogli, line 229: Shouldn't we use `wallet-service.mainnet.hathor.network`? I'm not sure what strategy we've been using.

msbrogli (approved):

src/constants.js
andreabadesso: I think this is a valid discussion, @luislhl probably followed the explorer URLs pattern:

https://explorer.hathor.network/ for `mainnet`
https://explorer.testnet.hathor.network/ for `testnet`

I like the current URL, without `.mainnet` but we can discuss this in another issue. What do you think?

PR #252, opened 2021-08-31
feat: add support for create NFT in the wallet service facade

pedroferreira1:


src/wallet/wallet.ts
```diff
@@ -994,6 +1007,31 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+
+  /**
+   * Create an NFT in the network
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  async createNFT(name: string, symbol: string, amount: number, data: string, options = {}): Promise<Transaction>  {
+    this.failIfWalletNotReady();
+    type optionsType = {
```
andreabadesso, line 1019: What is this for?

Typescript will only perform static type checking at compile time, so I guess this is pointless unless you type the method parameter

andreabadesso (changes requested):

src/wallet/wallet.ts
pedroferreira1: As we discussed in a call, I've been using this instead of typing the parameter to prevent using `| undefined` in all parameters because they are optional inside options. I understand it won't execute type checking on it but I feel it's better than not to have because anyone reading the code can clearly understand what are the types. I will create an issue to study further what would be the best approach for this situation and we can refactor all methods that are using this.

msbrogli (approved):

andreabadesso (approved):

PR #255, opened 2021-09-11
fix: we should calculate tx weight after adding input data

pedroferreira1:
### Motivation

We are executing `calculateWeight` method before setting the input data (with signatures), then the tx weight is smaller than it should be. This error happens only on the wallet service facade and was not happening before because the tests were made only on the testnet and the min_weight is 8 there.

### Acceptance criteria

The weight should be calculated fine for the transaction and we must be able to send a transaction using the wallet service on the mainnet.


luislhl (approved):
Should we create an issue to add unit tests on the methods involved?

pedroferreira1:
> Should we create an issue to add unit tests on the methods involved?

Unit tests can be created with mocks but the best tests would be integration tests with wallet service and full node. We have an issue to handle this soon and I agree it's really important.

andreabadesso (approved):

PR #261, opened 2021-09-22
feat: fetch wallet tokens metadata on load wallet (old facade)

pedroferreira1:
### Acceptance criteria

- We must have methods to load token metadata from explorer service (working for testnet or mainnet)

### DIFF
--- a/src/api/explorerServiceAxios.ts
+++ b/src/api/explorerServiceAxios.ts
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import axios from 'axios';
+import {
+    EXPLORER_SERVICE_BASE_URL,
+    EXPLORER_SERVICE_TESTNET_BASE_URL,
+    TIMEOUT,
+} from '../constants';
+
+/**
+ * Method that creates an axios instance
+ *
+ * @module Axios
+ */
+
+/**
+ * Create an axios instance to be used when sending requests to the explorer service
+ *
+ * @param {string} network The network to access the explorer service
+ * @param {number} timeout Timeout in milliseconds for the request
+ */
+export const axiosInstance = async (network: string, timeout: number = TIMEOUT) => {
+  const defaultOptions = {
+    baseURL: getBaseUrl(network),
+    timeout: timeout,
+    headers: {
+      'Content-Type': 'application/json',
+    },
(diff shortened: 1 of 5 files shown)


andreabadesso (approved):

andreabadesso (changes requested):
> The wallet object must load the wallet tokens metadata and emit an event with it

The code looks good, but where is this event being emitted? Is this PR still a draft?

andreabadesso (approved):

src/api/metadataApi.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import explorerServiceAxios from './explorerServiceAxios';
+
+const metadataApi = {
+  async getDag(id: string, network: string) {
```
msbrogli, line 11: I'd rename it to `getDagMetadata(...)`.

src/api/explorerServiceAxios.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Returns the correct base url constant for wallet service based on the network
+ *
+ * @param {string} network The network, can be either mainnet or testnet
+ */
+const getBaseUrl = (network: string): string => {
+  if (network === 'mainnet') {
+    return EXPLORER_SERVICE_BASE_URL;
+  } else {
+    return EXPLORER_SERVICE_TESTNET_BASE_URL;
```
msbrogli, line 48: Shouldn't we allow the client to set a different URL? This seems relevant for testing private networks, for example.

msbrogli (changes requested):

msbrogli (approved):

src/api/explorerServiceAxios.ts
pedroferreira1: I agree with you but we already have another issue more generic to handle this kind of configuration (https://github.com/HathorNetwork/hathor-wallet-lib/issues/161)

I think we should leave this to another PR. What do you think?

src/api/metadataApi.ts
msbrogli: Thanks!

src/api/explorerServiceAxios.ts
msbrogli: I agree. Luis should work on it this week, right?

pedroferreira1: Yes.

PR #262, opened 2021-09-28
fix: clear sensitive data after wallet load

andreabadesso:
# Acceptance criteria

We must avoid keeping the seed/privkey on memory after it is no longer necessary to load the wallet.

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1045,6 +1045,7 @@ class HathorWallet extends EventEmitter {
       throw "This should never happen";
     }
 
+    this.clearSensitiveData();
     this.getTokenData();
     this.serverInfo = null;
     this.walletStopped = false;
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):

src/new/wallet.js
```diff
@@ -1611,6 +1612,14 @@ class HathorWallet extends EventEmitter {
     return this.handleSendPreparedTransaction(tx);
   }
 
+  /**
+   * Remove sensitive data from memory
```
luislhl, line 1616: Do we really remove the data from memory when doing this?

According to a quick research I did here, we don't. They can still persist for some time until garbage collection erases them.

Apparently the exact behavior depends on the JS engine in use, so how do we guarantee all engines will remove things from memory when doing this?

andreabadesso: We don't remove it from memory immediately, but JavaScript does not provide an way to force garbage collection. So yes, it will be in memory for a while

andreabadesso: Since this is a lib, meant to be used by multiple javascript engines, we can't guarantee that this will be garbage collected immediately, so I think that adding a comment explaining this should be enough for now.

Added in 5c63472e7ffeb96a099cb72c4f4f654c935a1d56

luislhl (approved):

PR #265, opened 2021-09-29
feat: Create an easy way to init/edit configuration variables

luislhl:


__tests__/api/txMiningAxios.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+test('use testnet tx mining when network is testnet', () => {
+    networkIntance.setNetwork("testnet");
+    const client = txMiningRequestClient(null, null);
+
+    expect(client.defaults.baseURL).toEqual("https://txmining.testnet.hathor.network/");
+});
+
+test('use explicitly configured tx mining', () => {
+    config.setTxMiningUrl("txmining.url");
```
luislhl, line 36: Maybe we should also clear the side effects generated by setting this

src/config.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+    * Returns the correct base url constant for explorer service.
+    * If the url was explicitly set using the config object, it is always returned.
+    * Otherwise, we return it based on the provided network object.
+    *
+    * If the url was not set in the config, and no network is provided, we throw an Error.
+    *
+    * @summary If the description is long, write your summary here. Otherwise, feel free to remove this.
```
pedroferreira1, line 110: Should we remove this?

src/constants.js
```diff
@@ -56,19 +41,6 @@ export const LIMIT_ADDRESS_GENERATION = true;
  */
 export const HATHOR_BIP44_CODE = 280;
 
-/**
- * Server options for the user to choose which one to connect
- */
-export const DEFAULT_SERVERS = [
```
pedroferreira1, line 62: I'm thinking if we should leave those constants here for now and mark them as deprecated.

Even though our wallets don't use them directly, maybe someone is using.

What do you think?

src/helpers.js
```diff
@@ -169,28 +169,6 @@ const helpers = {
...
-   * Get the server URL that the wallet is connected
-   *
-   * If a server was not selected, returns the default one
-   *
-   * @return {string} Server URL
-   *
-   * @memberof Helpers
-   * @inner
-   */
-  getServerURL() {
```
pedroferreira1, line 182: This method is used by the wallets. I think every method and constant we are removing should be checked in all the wallets

src/new/connection.js
```diff
@@ -49,8 +49,9 @@ class Connection extends EventEmitter {
     this.onConnectionChange = this.onConnectionChange.bind(this);
     this.handleWalletMessage = this.handleWalletMessage.bind(this);
 
-    this.servers = servers || [...DEFAULT_SERVERS];
-    this.currentServer = this.servers[0];
+    // TODO Should we use `config.getServerUrl` here, instead of DEFAULT_SERVER?
+    //  By using DEFAULT_SERVER, we are ignoring anything set in storage in the 'wallet:server' key
+    this.currentServer = servers[0] || DEFAULT_SERVER;
```
pedroferreira1, line 54: I agree to use `config.getServerUrl`

src/constants.js
luislhl: I think it's a good idea, will do this

src/helpers.js
luislhl: I'll make it a proxy for the new method

src/constants.js
```diff
@@ -17,13 +18,19 @@ export const DECIMAL_PLACES = 2;
 
 /**
  * ID of the genesis block
+ *
+ * @deprecated since version 0.24.0.
+ * Will be deleted in version 1.0.0
```
pedroferreira1, line 23: I like the deprecate tag, I just don't think we should set a version to deprecate it.

We can write "Will be deleted in the near future", or something like that. We don't know when we are going to release a v1.0.0

What do you think?

pedroferreira1 (approved):

andreabadesso (approved):

andreabadesso:
Approved, but I think you need a rebase from dev

src/constants.js
luislhl: The logic behind putting 1.0.0 as the deprecation version is that when we remove those constants, this should be considered a breaking change. So they will be removed in the next major version, which is 1.0.0.

It wouldn't make sense to remove them in some `v0.*`

And it wouldn't also make sense to release a v1.0.0 without removing those constants.

So, by this logic, it will be surely on v1.0.0.

Does it make sense?
But

luislhl: Unless we are currently in a period of not guaranteeing the backwards compatibility of minor versions.
Are we?

In this case, we could remove them in `v0.*`.

But this fact of not guaranteeing backwards compatibility should be stated somewhere in the docs.

andreabadesso (approved):

pedroferreira1 (approved):

PR #266, opened 2021-10-01
feat: add error handling and retry mechanism in getDagMetadata API method

pedroferreira1:


src/api/metadataApi.ts
```diff
@@ -6,11 +6,48 @@
...
+          // throw Error and the catch will handle it
+          throw Error;
+        }
+      } catch (e) {
+        if (e.response && e.response.status === 404) {
+          // No need to do anything, the metadata for this token was not found
+          // There is no error here, we just return null
+          return null;
+        } else {
+          // Error downloading metadata, then we should wait a few seconds and retry if still didn't reached retry limit
```
andreabadesso, line 41: I suggest changing the position of this comment to the line just above

`await helpers.sleep(retryInterval);`

And add another comment explaining that when `retries === 0` the method will propagate the error

andreabadesso (changes requested):

src/api/metadataApi.ts
pedroferreira1: Improved the comments here: https://github.com/HathorNetwork/hathor-wallet-lib/pull/266/commits/0085f4c46108282fde98a7bcb9e536d48d0a8e2d

What do you think?

andreabadesso (approved):

src/api/metadataApi.ts
```diff
@@ -6,11 +6,50 @@
...
+    while (retries >= 0) {
+      const axios = await explorerServiceAxios(network);
+      try {
+        const response = await axios.get(`metadata/dag`, { params: { id }});
+        if (response.data) {
+          return response.data;
+        } else {
+          // Error downloading metadata
+          // throw Error and the catch will handle it
+          throw Error;
```
msbrogli, line 33: Should we customize the error? We can also add a message to it but I feel a customized error is better.

msbrogli (approved):

src/api/metadataApi.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/266/commits/1c957a38e489e17807245a45df8f97105841e99d

PR #276, opened 2021-10-07
fix: error handling when sending transaction

pedroferreira1:
### Acceptance criteria

I've done 3 different things in this PR

1. fix: error handling in `SendTransaction` class is now working. If we have an error when mining or pushing tx, it will be handled by the promise.
2. feat: add validation for name/symbol size in the CreateTokenTransaction object
3. feat: add address validation in some places we hadn't added it (especially this one will require some changes in the wallets)

### DIFF
--- a/src/new/sendTransaction.js
+++ b/src/new/sendTransaction.js
@@ -272,10 +272,10 @@ class SendTransaction extends EventEmitter {
           const err = new SendTxError(response.message);
           reject(err);
         }
-      }).catch(() => {
+      }).catch((e) => {
         this.updateOutputSelected(false);
-        const err = new SendTxError(this.unexpectedPushTxError);
-        reject(err);
+        this.emit('send-error', e.message);
+        reject(e);
       });
     });
 
--- a/src/new/sendTransaction.js
+++ b/src/new/sendTransaction.js
@@ -232,10 +232,6 @@ class SendTransaction extends EventEmitter {
     })
 
     this.mineTransaction.on('success', (data) => {
-      this.transaction.parents = data.parents;
-      this.transaction.timestamp = data.timestamp;
-      this.transaction.nonce = data.nonce;
-      this.transaction.weight = data.weight;
       this.emit('mine-tx-ended', data);
     })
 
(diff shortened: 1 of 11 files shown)


src/new/sendTransaction.js
```diff
@@ -256,25 +254,32 @@ class SendTransaction extends EventEmitter {
...
-    const txHex = this.transaction.toHex();
-    txApi.pushTx(txHex, false, (response) => {
-      if (response.success) {
-        this.transaction.updateHash();
-        this.emit('send-tx-success', this.transaction);
-        if (this._unmarkAsSelectedTimer !== null) {
-          // After finishing the push_tx we can clearTimeout to unmark
-          clearTimeout(this._unmarkAsSelectedTimer);
-          this._unmarkAsSelectedTimer = null;
+    const promise = new Promise((resolve, reject) => {
```
luislhl, line 257: I feel like if we try to use more `async/await` instead of the Promise class directly, we could improve the code readability.

It is a little unfriendly to the reader.

But this is just a code style suggestion.

src/new/sendTransaction.js
```diff
@@ -287,21 +292,24 @@ class SendTransaction extends EventEmitter {
    * @inner
    */
   async runFromMining(until = null) {
-    const promise = new Promise((resolve, reject) => {
+    const promise = new Promise(async (resolve, reject) => {
```
luislhl, line 295: Do we really need a Promise wrapping this code here?

Why not just this (which I think works the same, but didn't test to confirm)?

```ts
 async runFromMining(until = null) {
    try {
      // This will await until mine tx is fully completed
      // mineTx method returns a promise that resolves when
      // mining succeeds or rejects when there is an error
      const mineData = await this.mineTx();
      this.transaction.parents = mineData.parents;
      this.transaction.timestamp = mineData.timestamp;
      this.transaction.nonce = mineData.nonce;
      this.transaction.weight = mineData.weight;

      if (until === 'mine-tx') {
        return this.transaction;
      }

      const pushedTx = await this.handlePushTx();
      return this.transaction;
    } catch (err) {
      if (err instanceof WalletError) {
        this.emit('send-error', err.message);
      }
      throw err;
    }
  }
```

luislhl (approved):

src/new/sendTransaction.js
pedroferreira1: I agree about being more friendly to read, however the way we have our API calls from the lib we expect one parameter as the callback from the promise resolve. This is not good and should be refactored in the future but right now we can't use only async await with it.

For the new APIs (you can see in the wallet service facade) we are using only async/await.

src/new/sendTransaction.js
pedroferreira1: I agree it's much better. Refactor done here https://github.com/HathorNetwork/hathor-wallet-lib/pull/276/commits/7b6165efb359a999abd7869a5ca3173d3d99a232. Can you see again please?

luislhl (approved):

PR #277, opened 2021-10-07
feat: allow private networks

luislhl:
### Acceptance Criteria
- [ ] It should be possible to use this version of the lib in our Explorer and make it work with private-networks
- [ ] Allow dynamically adding more network configuration options

### Notes
- We are using the same version bytes as the Testnet for the private networks, because we would have problems with TxMiningService if using different ones, since it validates the addresses.

### TODO
- [ ] Release new version of the Explorer once a new version of this lib is available, which should close https://github.com/HathorNetwork/hathor-explorer/issues/125
- [ ] Also release a new version of the Wallet Headless using the new lib version

### DIFF
--- a/src/models/network.ts
+++ b/src/models/network.ts
@@ -24,8 +24,8 @@ const versionBytes = {
     'xpub': 0x0488b21e,  // xpub // 0x0434c85b -> tnpb
   },
   'privatenet': {
-    'p2pkh': 0x70,
-    'p2sh': 0x93,
+    'p2pkh': 0x49,
+    'p2sh': 0x87,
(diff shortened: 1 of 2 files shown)


src/models/network.ts
```diff
@@ -114,19 +171,50 @@ class Network {
     this.bitcoreNetwork = networkOptions[name];
   }
 
+  /**
+   * Registers a new network configuration.
+   *
+   * We validate the configuration before registering.
+   */
+  static registerNetwork(networkConfig: networkConfigType) {
```
luislhl, line 179: Move this to a different PR, because we should test it further

msbrogli (approved):

PR #291, opened 2021-11-24
feat: change password

andreabadesso:
### Acceptance Criteria
- Added a method to change the password
- It should also change everything that is encrypted on storage
- We should also store its hash on storage

### Security Checklist

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -260,7 +260,7 @@ const wallet = {
     const initialAccessData = this.getWalletAccessData() || {};
 
     const encryptedDataWords = this.encryptData(words, password);
-    initialAccessData['words'] = encryptedDataWords.encrypted.toString(),
+    initialAccessData['words'] = encryptedDataWords.encrypted.toString();
 
     this.setWalletAccessData(initialAccessData);
   },
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -724,6 +724,44 @@ const wallet = {
   },
 
   /**
+   * Validate old password and change it for the new one
+   *
+   * @param {string} oldPassword
+   * @param {string} newPassword
+   *
+   * @return {boolean} true if the PIN was successfully changed
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  changePassword(oldPassword, newPassword) {
+    if (this.isFromXPub()) {
+        throw WalletFromXPubGuard('changePassword');
+    }
+
+    const isCorrect = this.isPasswordCorrect(oldPassword);
+    if (!isCorrect) {
+      return false;
+    }
+
+    // Get new password hash
+    const newHash = this.hashPassword(newPassword);
+    // Update new Password data in storage
+    const accessData = this.getWalletAccessData();
+    accessData['hashPasswd'] = newHash.key.toString();
+    accessData['saltPasswd'] = newHash.salt;
+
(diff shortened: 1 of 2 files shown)


src/wallet.js
```diff
@@ -723,6 +723,84 @@ const wallet = {
...
+          break;
+        }
+      } catch(e) {}
+    }
+
+    if (pinFound !== -1) {
+      return [true, pinFound];
+    }
+
+    return [false];
```
msbrogli, line 801: Should we return `[false, '']` for consistency in the return type?

src/wallet.js
```diff
@@ -723,6 +723,84 @@ const wallet = {
...
+    return true;
+  },
+
+  /**
+   * @return {[boolean, string]} Tuple <success:boolean, pin:string> with the result
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  guessPin() {
```
msbrogli, line 770: I'm not sure we should place it in hathor-wallet-lib. It seems a temporary fix needed only in hathor-wallet-mobile. I feel we should keep it there and remove it in a later version.

src/wallet.js
```diff
@@ -723,6 +723,84 @@ const wallet = {
...
+    const accessData = this.getWalletAccessData();
+    const data = accessData.words;
+
+    const decrypt = (data, pin) => CryptoJS
+      .AES
+      .decrypt(data, pin)
+      .toString(CryptoJS.enc.Utf8);
+
+    let pinFound = -1;
+    for (let i = 0; i <= 999999; i++) {
```
msbrogli, line 781: Should we block the main loop? I think we should do something like this:

```js
guessPartial(begin, end) {
  :
  for (let i = begin; i <= end; i++) {
    :
  }
  :
}

guessScheduler(begin, step, cb) {
  const end = begin + step;
  ret = guessPartial(begin, end);
  if (ret[0] == false) {
    if (end < 999999) {
      setTimeout(guessScheduler, 0, end + 1, step, cb);
    }
  } else {
    cb(pin);
  }
}

guessPin(cb) {
  guessScheduler(0, 1000, cb);
}

**Review by msbrogli — CHANGES_REQUESTED** [review]

File: src/wallet.js (hunk 3 above)
Comment (msbrogli) [reply src/wallet.js:781]: Maybe we should call `cb(...)` multiple times so the caller can update the interface with progress.

Comment (andreabadesso) [reply src/wallet.js]: @msbrogli This was moved to the hathor-wallet-mobile repo, can you check it there?

File: src/wallet.js (hunk 2 above)
Comment (andreabadesso) [reply src/wallet.js]: I agree, moved to the hathor-wallet-mobile repo and removed the latest commit

File: src/wallet.js (hunk 1 above)
Comment (andreabadesso) [reply src/wallet.js]: I agree, done in the hathor-wallet-mobile repo

File: src/wallet.js [hunk 4]
```diff
@@ -723,6 +723,84 @@ const wallet = {
...
+      const pin = `${i}`.padStart(6, '0');
+
+      try {
+        const decrypted = decrypt(data, pin);
+
+        if (decrypted.length > 250) continue;
```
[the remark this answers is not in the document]
andreabadesso: Done in the hathor-wallet-mobile repo

pedroferreira1 (approved):

msbrogli (approved):

PR #295, opened 2021-11-30
feat: create parse script data method

pedroferreira1:
### Acceptance Criteria
- We must have a method to parse an output script that is a script data.

### Security Checklist

### DIFF
--- a/__tests__/models/output.test.js
+++ b/__tests__/models/output.test.js
@@ -99,23 +99,3 @@ test('Script', () => {
     parseP2PKH(o1.script.slice(10), network);
   }).toThrowError(ParseScriptError);
 });
-
-test('Script data', () => {
-  const data = 'test';
-  const scriptDataObj = new ScriptData(data);
-  const scriptData = scriptDataObj.createScript();
-
-  const parsedData = parseScriptData(scriptData);
-  expect(parsedData.data).toBe(data);
-
-  const wrongData = buffer.Buffer.from('a', 'utf-8');
-  expect(() => {
-    parseScriptData(wrongData);
-  }).toThrowError(ParseScriptError);
-
-  // Remove last element from scriptData (OP_CHECKSIG), then should fail
-  const wrongData2 = scriptData.slice(scriptData.length - 1)
-  expect(() => {
-    parseScriptData(wrongData2);
-  }).toThrowError(ParseScriptError);
-});
--- a/__tests__/models/output.test.js
+++ b/__tests__/models/output.test.js
@@ -8,11 +8,10 @@
 import Output from '../../src/models/output';
 import Address from '../../src/models/address';
 import P2PKH from '../../src/models/p2pkh';
-import ScriptData from '../../src/models/script_data';
 import Network from '../../src/models/network';
 import { OutputValueError, ParseScriptError } from '../../src/errors';
 import buffer from 'buffer';
-import { parseP2PKH, parseScriptData } from '../../src/utils/scripts';
+import { parseP2PKH } from '../../src/utils/scripts';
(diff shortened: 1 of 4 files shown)


src/utils/scripts.ts
```diff
@@ -37,3 +39,87 @@ export const parseP2PKH = (buff: Buffer, network: Network): P2PKH => {
...
+  expectedLen += dataBytesLen;
+
+  if (expectedLen !== scriptBuf.length) {
+    // The script has different qty of bytes than expected
+    throw new ParseScriptError('Invalid output script.');
+  }
+
+  if (scriptBuf[expectedLen - 1] !== OP_CHECKSIG[0]) {
+    // Last byte must be an OP_CHECKSIG
+    throw new ParseScriptError('Invalid output script.');
```
luislhl, line 83: Should the error message better describe what is happening in each case?
Like: `Invalid output script. Last byte must be an OP_CHECKSIG`.

andreabadesso (approved):

__tests__/models/output.test.js
```diff
@@ -97,4 +98,24 @@ test('Script', () => {
   expect(() => {
     parseP2PKH(o1.script.slice(10), network);
   }).toThrowError(ParseScriptError);
+});
+
+test('Script data', () => {
```
luislhl, line 103: Shouldn't this test be included in a new file `__tests__/models/script_data.test.js`?

If I understand the naming pattern correctly, each model should have its own test file, right?

src/utils/scripts.ts
```diff
@@ -37,3 +39,87 @@ export const parseP2PKH = (buff: Buffer, network: Network): P2PKH => {
...
+  let scriptBuf = _.clone(buff);
+
+  if (scriptBuf.length === 0) {
+    throw new ParseError('Invalid buffer.');
+  }
+
+  let lenData, start;
+
+  if (scriptBuf[0] > 75) {
+    lenData = scriptBuf[1];
```
luislhl, line 118: Should we have a test for this case?

src/utils/scripts.ts
```diff
@@ -37,3 +39,87 @@ export const parseP2PKH = (buff: Buffer, network: Network): P2PKH => {
...
+  // The expected len will be at least 2 bytes
+  // 1 for the script len and 1 for the OP_CHECKSIG in the end
+  let expectedLen = 2;
+  let dataBytesLen;
+
+  // If we have OP_PUSHDATA1 as first byte, the second byte has the length of data
+  // otherwise, the first byte already has the length of data
+  if (scriptBuf[0] === OP_PUSHDATA1[0]) {
+    expectedLen += 1;
+    dataBytesLen = scriptBuf[1];
```
luislhl, line 68: Should we have a test for this case?

pedroferreira1: Agreed. Added here https://github.com/HathorNetwork/hathor-wallet-lib/pull/295/commits/b7bf56105ac7290106987153ba64e1e302821ee5

src/utils/scripts.ts
pedroferreira1: Agreed. Added here https://github.com/HathorNetwork/hathor-wallet-lib/pull/295/commits/b7bf56105ac7290106987153ba64e1e302821ee5

src/utils/scripts.ts
pedroferreira1: Makes sense. Added more details in all error messages here https://github.com/HathorNetwork/hathor-wallet-lib/pull/295/commits/8d6c46cfdf5d96b377f4a33118805f191095380f

luislhl (approved):

PR #296, opened 2021-12-02
feat: added a changePinAndPassword method

andreabadesso:
### Acceptance Criteria
- There must be a method that updates the pin and password simultaneously so we can guarantee that they will be updated (or fail) at the same time

### Security Checklist

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -724,6 +724,38 @@ const wallet = {
   },
 
   /**
+   * Validate old password and returns a new accessData object
+   * with updated data using it.
+   *
+   * @param {Object} oldAccessData - The old access data to update
+   * @param {string} oldPassword
+   * @param {string} newPassword
+   *
+   * @return {Object} The new access data
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  _getChangePasswordAccessData (oldAccessData, oldPassword, newPassword) {
+    // Get new password hash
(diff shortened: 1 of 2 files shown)


msbrogli (approved):

src/wallet.js
```diff
@@ -778,23 +833,67 @@ const wallet = {
...
+   * @param {string} newPin
+   * @param {string} oldPassword
+   * @param {string} newPassword
+   *
+   * @return {boolean} true if PIN and password were successfully changed
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  changePinAndPassword(oldPin, newPin, oldPassword, newPassword) {
```
pedroferreira1, line 869: I don't think this method should be in the lib. Changing PIN and password at the same time is something really unique for the mobile app and if the idea is to prevent success when changing pin and error when changing password, we can add `if (isPinCorrect(oldPin) && isPasswordCorrect(oldPassword)) do things` before changing both.

I'm not against the code, I think the PR is fine, so if you think it makes sense, we can go with it, I just think this shouldn't be in the lib.

If we move with this in the lib I think we should refactor this method a bit to call `changePin` and `changePassword`, so we don't repeat all the newAccessData code in the end.

pedroferreira1 (approved):

msbrogli (approved):

PR #302, opened 2021-12-12
feat: wallet service facade realtime

andreabadesso:
### Acceptance Criteria
- We should receive real-time events from the wallet service for new txs
- We should receive real-time events from the wallet service for updates in txs
- We should correctly display the balance and transaction history for tokens updated in received transactions
- We should point to the correct wallet-service websocket URLs

### Missing on this PR
- We are not yet receiving events for height unlocks on the wallet-service websocket
- We are not yet receiving events for time unlocks on the wallet-service websocket
- Transactions are taking up to 10 seconds to show up, this is yet to be debugged

### Breaking changes
To keep compatibility with the new facade, we had to change a few methods to match the signatures. The main change was the `getTxBalance` method, which is now `async`. But I also had to change a few other methods that called the `getTxBalance`, so this is the list of methods that had their signature changed (from sync to async):

- `getTxBalance`
- `preProcessWalletData`
- `onTxArrived`
- `getPreProcessedData`
- `setState`

They are mostly internal methods, but the `getTxBalance` signature change will have to be refactored on the wallet mobile and wallet desktop (the headless wallet does not use it).

### Security Checklist


pedroferreira1 (changes requested):

src/config.ts
```diff
@@ -91,6 +96,33 @@ class Config {
...
 
+    /**
+     * Returns the correct websocket base url constant for wallet service.
+     * If the url was explicitly set using the config object, it is always returned.
+     * Otherwise, we return it based on the provided network object.
+     *
+     * @param {Network} network The network, can be either mainnet or testnet
+     * @return {string} The wallet service websocket url
+     */
+    getWalletServiceBaseWsUrl(network?: Network) {
```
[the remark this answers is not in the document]
andreabadesso: Yes! Fixed in 70fbe80bcd021716d3064e78e71064ad871cfdd6

src/config.ts
```diff
@@ -91,6 +96,33 @@ class Config {
...
+        if (this.WALLET_SERVICE_WS_BASE_URL) {
+            return this.WALLET_SERVICE_WS_BASE_URL;
+        }
+
+        if (!network) {
+            throw new Error('You should either provide a network or call setWalletServiceBaseUrl before calling this.');
```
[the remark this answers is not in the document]
andreabadesso: Yes, fixed in e9bc90831070f1524535aba7ecbf604dfd4034ba alongside with the refactor from your last comment

src/config.ts
```diff
@@ -91,6 +96,33 @@ class Config {
...
+        }
+
+        // Keeps the old behavior for cases that don't explicitly set a WALLET_SERVICE_BASE_WS_URL
```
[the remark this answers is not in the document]
andreabadesso: Makes sense, done in e9bc90831070f1524535aba7ecbf604dfd4034ba and updated in the mobile PR

src/wallet/wallet.ts
```diff
@@ -147,6 +160,25 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     await handleCreate(data.status);
   }
 
+  async onNewTx(newTx: any) {
```
pedroferreira1, line 163: Missing docstring

pedroferreira1, line 163: Why type `any`? I think it's a defined type that you will receive from the conn object, no?

src/wallet/wallet.ts
```diff
@@ -147,6 +160,25 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    const outputs: any[] = newTx.outputs;
```
pedroferreira1, line 164: Why type `any` here?

src/wallet/wallet.ts
```diff
@@ -147,6 +160,25 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    for (const output of outputs) {
+      if (this.newAddresses.find((newAddress) => newAddress.address === output.decoded.address)) {
+        // break early
+        shouldGetNewAddresses = true;
+        break;
+      }
+    }
+
+    if (shouldGetNewAddresses) {
+      await this.getNewAddresses();
```
pedroferreira1, line 176: I think it's important leave a comment here explaining why you are doing this. This is related to this TODO in the constructor, right? If so, we need to explain here the reason we are doing this and remove the TODO

```
    // TODO When we integrate the real time events from the wallet service
    // we will need to have a trigger to update this array every new transaction
    // because the new tx might have used one of those addresses
    // so we just need to call await this.getNewAddresses(); on the new tx event
```

src/wallet/wallet.ts
```diff
@@ -157,6 +189,66 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+  /**
+   * Returns the balance for each token in tx, if the input/output belongs to this wallet
+   *
+   * This method is meant to keep compatibility with the old facade
+   *
+   * @param {Object} tx Transaction data with array of inputs and outputs
+   *
+   * @return {Object} Object with each token and it's balance in this tx for this wallet
+   **/
+  async getTxBalance(tx, optionsParam = {}) {
```
pedroferreira1, line 201: Missing parameters types and return type

pedroferreira1, line 201: Should we add this method to the `IHathorWallet` interface, since it's now a method that must be implemented?

src/wallet/connection.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.websocket.on('update-tx', (payload) => this.emit('update-tx', payload.data));
+
+    this.setState(ConnectionState.CONNECTING);
+    this.websocket.setup();
+  }
+
+  /**
+   * Close the connections and stop emitting events.
+   **/
+  stop() {
```
pedroferreira1, line 116: Why you don't end websocket connection on stop?

Maybe we should have a parent class that both Connection classes extend and change

pedroferreira1 (changes requested):

src/websocket.js
```diff
@@ -184,59 +55,7 @@ class WS extends EventEmitter {
...
-      }, 1000);
-    }
-  }
-
-  /**
-   * Ping method to check if server is still alive
-   *
+   * Ping method to check if server is still alive.
    */
   sendPing() {
```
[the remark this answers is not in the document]
andreabadesso: The signature, the wallet service `sendPing` uses `action`, the old facade uses `type`

src/wallet/wallet.ts
andreabadesso: Done in 1e3e15236ba33f9bc5fd3d8ca94cdb32c12ccb56

andreabadesso: Yea, done in 1e3e15236ba33f9bc5fd3d8ca94cdb32c12ccb56

src/wallet/wallet.ts
andreabadesso: Done in 1e3e15236ba33f9bc5fd3d8ca94cdb32c12ccb56

src/wallet/wallet.ts
andreabadesso: You are absolutely right, done in 11ca6458aaa1643201732a1f74e54da6f6bb9348

src/wallet/wallet.ts
andreabadesso: Done in 9ffc52b249d1207bc156ec52fb4a486310447167

andreabadesso: a16aa65f6bd739dfb8df7e9388de1644e3032661

src/wallet/connection.ts
andreabadesso: This was probably a mistake, can you check again?

src/websocket/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import BaseWebSocket, { WsOptions } from './base';
+
+/**
+ * Handles websocket connections and message transmission
+ *
+ * @class
+ * @name WS
```
luislhl, line 14: Show we update these annotations?

src/wallet/websocket.ts
```diff
@@ -0,0 +1,147 @@
...
+} from '../websocket/base';
+
+export interface WalletServiceWebSocketOptions extends WsOptions {
+  walletId: string;
+};
+
+const JOIN_TIMEOUT = 5000;
+
+/**
+ * Handles websocket connections and message transmission
```
luislhl, line 20: We should improve the description in this class and in class `WalletWebSocket`. They currently have the same description, so it's hard to know the different between them, and in which cases each one should be used.

src/wallet/websocket.ts
```diff
@@ -0,0 +1,147 @@
+/**
```
luislhl, line 1: The intent of this `src/wallet` folder is to aggregate everything related to wallet-service, right?

Why don't we name it `src/wallet-service` instead?

src/wallet/websocket.ts
```diff
@@ -0,0 +1,147 @@
...
+import BaseWebSocket, {
+  WsOptions,
+  DEFAULT_WS_OPTIONS,
+} from '../websocket/base';
+
+export interface WalletServiceWebSocketOptions extends WsOptions {
+  walletId: string;
+};
+
+const JOIN_TIMEOUT = 5000;
```
luislhl, line 17: Should we include this in the `WalletServiceWebSocketOptions`?

Other timeouts are configurable through `WsOptions`

src/wallet/websocket.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.walletId = walletId;
+    this.joinTimeoutTimer = null;
+  }
+
+  /**
+   * Handle message receiving from websocket
+   *
+   * @param {Object} evt Event that has data (evt.data) sent in the websocket
+   */
+  onMessage(evt) {
```
luislhl, line 61: Would it make sense to make this class `WalletServiceWebSocket` be just a wrapper around an instance of `WalletWebSocket`, instead of a sibling of it?

This way I think we would not need to re-implement this method `onMessage` here.

We would just have an instance of `WalletWebSocket` on `this.websocket`, for instance, then we would do:
```
    this.websocket.on('join-success', () => this.onJoinSuccess());
```

Does it make sense? I didn't really checked if this change change could bring difficulties in some other part of the code.

pedroferreira1 (changes requested):

src/config.ts
```diff
@@ -78,17 +75,35 @@ class Config {
...
-        if (network.name == 'mainnet') {
-            return WALLET_SERVICE_MAINNET_BASE_URL;
-        } else if (network.name == 'testnet'){
-            return WALLET_SERVICE_TESTNET_BASE_URL;
-        } else {
-            throw new Error(`Network ${network.name} doesn't have a correspondent wallet service url. You should set it explicitly.`);
+    /**
+     * Returns the correct websocket base url constant for wallet service.
+     * If the url was explicitly set using the config object, it is always returned.
+     * Otherwise, we return it based on the provided network object.
```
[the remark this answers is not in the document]
andreabadesso: Fixed in df7df40b6ca069fcc7cb8234a826d6e4593762f9

src/config.ts
```diff
@@ -78,17 +75,35 @@ class Config {
...
+     *
+     * @param {Network} network The network, can be either mainnet or testnet
+     * @return {string} The wallet service websocket url
+     */
+    getWalletServiceBaseWsUrl(network?: Network) {
```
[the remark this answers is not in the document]
andreabadesso: Added it explicitly on df7df40b6ca069fcc7cb8234a826d6e4593762f9

src/config.ts
```diff
@@ -78,17 +75,35 @@ class Config {
...
+        if (this.WALLET_SERVICE_BASE_WS_URL) {
```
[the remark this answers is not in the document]
andreabadesso: Added a throw on df7df40b6ca069fcc7cb8234a826d6e4593762f9

src/config.ts
```diff
@@ -78,17 +75,35 @@ class Config {
         }
 
         if (!network) {
-            throw new Error("You should either provide a network or call setWalletServiceBaseUrl before calling this.")
+            throw new Error('You should either provide a network or call setWalletServiceBaseUrl before calling this.');
         }
+    }
 
-        // Keeps the old behavior for cases that don't explicitly set a WALLET_SERVICE_BASE_URL
```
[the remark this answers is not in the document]
andreabadesso: I think it is safe to assume that nobody is using this default URL and that we should force the user to set it before using.

Added throws for the case when it is not set yet on df7df40b6ca069fcc7cb8234a826d6e4593762f9

src/connection.ts
```diff
@@ -25,48 +44,50 @@ import WS from '../websocket';
  * - state: Fired when the state of the Wallet changes.
  * - wallet-update: Fired when a new wallet message arrive from the websocket.
  **/
-class Connection extends EventEmitter {
+abstract class Connection extends EventEmitter {
+  // network: 'testnet' or 'mainnet'
+  public network: string;
```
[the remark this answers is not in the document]
andreabadesso: Because otherwise the `WalletServiceConnection` and `WalletConnection` classes wouldn't be able to use it as the network would be private to the `Connection` (base) class

src/connection.ts
```diff
@@ -25,48 +44,50 @@ import WS from '../websocket';
...
-    network,
-    servers = [],
-    connectionTimeout = null,
-  } = {}) {
+  constructor(options: ConnectionParams) {
     super();
 
+    const {
+      network,
+      servers,
```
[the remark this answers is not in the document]
andreabadesso: Not here, I'm handling `connectionTimeout` on `WalletServiceConnection` and `WalletConnection` because they instantiate the `WebSocket` differently on their constructors

src/wallet/wallet.ts
```diff
@@ -138,6 +143,12 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
         }, WALLET_STATUS_POLLING_INTERVAL);
       } else if (data.status === 'ready') {
         await this.onWalletReady();
+        this.conn = new WalletServiceConnection({
+          walletId: this.walletId,
+        });
+        this.conn.start();
+        this.conn.on('new-tx', (newTx: WsTransaction) => this.onNewTx(newTx));
+        this.conn.on('update-tx', (updatedTx) => this.emit('update-tx', updatedTx));
```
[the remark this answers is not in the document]
andreabadesso: We sure can, done in 3e5338c0b25dfd9bc05aa6f2e2f2f16f404a75a0

src/wallet/wallet.ts
```diff
@@ -147,6 +158,34 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+
+    for (const output of outputs) {
+      if (this.newAddresses.find((newAddress) => newAddress.address === output.decoded.address)) {
+        // break early
+        shouldGetNewAddresses = true;
+        break;
+      }
+    }
+
+    // We need to have a to update the `newAddresses` array every new transaction
```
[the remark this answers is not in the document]
andreabadesso: Fixed in b5bba45051d3138a0e6d3bd85c9e99b59e0aa787

src/wallet/websocket.ts
andreabadesso: Because it is meant to replace the old implementations and be the default implementation for the `wallet`

I think we should keep it as `wallet` and replace the `new` folder which is the old facade to something else, maybe `old` 😂 

But I wouldn't do this refactor in this PR, I suggest opening an issue to do that. What do you think, @luislhl?

src/wallet/websocket.ts
andreabadesso: I like the idea, done in a0e5091ee509a7a0e0e3672b7887973e4f845b33

src/wallet/websocket.ts
andreabadesso: I agree, improved the documentation for both classes on 8e3c6f98808c1982f33ba1546e0e985e27e033b9

src/wallet/websocket.ts
andreabadesso: Both `WalletWebSocket` and `WalletServiceWebSocket` extends the base `WebSocket` class in `src/websocket/base.ts`

The idea of they being extensions of the abstract base class is that they will always share common methods (like keeping track of ping/pong and calculating uptime, etc...) but may have completely different implementations for specific methods, like the `onMessage` here (that is currently very similar)

src/websocket/base.ts
```diff
@@ -5,71 +5,97 @@
...
+  // Open connection timeout in milliseconds
+  private openConnectionTimeout: number;
+  // Date of connection.
+  private connectedDate: Date | null;
+  // Date of latest setup call. The setup is the way to open a new connection.
+  private latestSetupDate: Date | null;
+  // Latest round trip time measured by PING/PONG.
+  private latestRTT: number | null;
+  // Date of latest ping.
+  public latestPingDate: Date | null;
```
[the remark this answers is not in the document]
andreabadesso: Because otherwise the `WalletWebSocket` and `WalletServiceWebSocket` would not be able to mutate it

src/websocket/index.ts
andreabadesso: We should, done in 8e3c6f98808c1982f33ba1546e0e985e27e033b9

src/websocket/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        this.timeoutTimer = setTimeout(() => this.onConnectionDown(), this.connectionTimeout);
+      }
+    }
+    this.emit(_type, message)
+  }
+
+  /**
+   * Ping method to check if server is still alive
+   */
+  sendPing() {
```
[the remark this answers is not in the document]
andreabadesso: It doesn't feel right to me, maybe in another implementation of a `WebSocket` class we would just send a string `PING` instead of an object

But I can add the `key` parameter if you think we should

src/config.ts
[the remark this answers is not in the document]
pedroferreira1: Should we create an specific error for that type? We already have two of them and should have more as we grow this config class.

src/wallet/wallet.ts
```diff
@@ -147,6 +158,44 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     await handleCreate(data.status);
   }
 
+  /**
+   * onUpdateTx: Event called when a transaction is updated
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  async onUpdateTx(updatedTx) {
```
pedroferreira1, line 167: Why is it async?

src/websocket/index.ts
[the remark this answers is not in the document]
pedroferreira1: I think I agree with you, this doesn't seem right. The only thing that bothers me is that any change in this method if we find a bug will need to be replicated in all subclasses and this is even worse to me.

We could also have a new abstract method `getPingMessage`, which is the only different method in both classes, the `sendPing` would be unique and like this:

```
sendPing() {
  if (this.latestPingDate) {
      // Skipping sendPing. Still waiting for pong...
      return;
    }
    const msg = this.getPingMessage();
    this.latestPingDate = new Date();
    this.timeoutTimer = setTimeout(() => this.onConnectionDown(), this.connectionTimeout);
    this.sendMessage(msg);
  }
```

What do you think? I'm okay with whatever you choose.

src/websocket/base.ts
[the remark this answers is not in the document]
pedroferreira1: Should we set them as `protected`?

src/connection.ts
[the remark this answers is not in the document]
pedroferreira1: Should we use `protected` then?

src/wallet/websocket.ts
luislhl: Looks good. I like how the `new` folder is actually the `old` facade.

src/wallet/wallet.ts
```diff
@@ -157,6 +206,66 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+          }
+        }
+        continue;
+      }
+      if (txin.decoded && txin.decoded.address
+          && addresses.includes(txin.decoded.address)) {
+        if (!balance[txin.token]) {
+          balance[txin.token] = 0;
+        }
+        balance[txin.token] -= txin.value;
```
luislhl, line 262: Should we have tests for these new methods?

src/config.ts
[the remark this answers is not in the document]
andreabadesso: Sounds good!

Done in 62b498435866005c69603251a487a918ac488bf7

src/connection.ts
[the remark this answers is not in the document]
andreabadesso: Yes we should 😳

Done in 03477495831cfeb02ed5cc0779843bc151586ff9

src/websocket/index.ts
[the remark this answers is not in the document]
andreabadesso: I like it, done in f246cdd5911f141d5b344a89ba6389f0dfc7c07d

src/websocket/base.ts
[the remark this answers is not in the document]
andreabadesso: Yes we should 😳

Done in 03477495831cfeb02ed5cc0779843bc151586ff9

src/wallet/wallet.ts
andreabadesso: It shouldn't be! Fixed in a483314d5ded017d66166fa8a282c23e98260626

src/wallet/wallet.ts
andreabadesso: Added tests for the `getTxBalance` method in fc56280984abb512c6e7443a383d9e92c98a8396

src/config.ts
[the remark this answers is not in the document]
pedroferreira1: I thought about creating a generic error `ConfigNotSetError`. We could use it to all of them or create as base error and then extend it to child error classes. What do you think?

pedroferreira1 (approved):

src/wallet/wallet.ts
```diff
@@ -157,6 +206,67 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * @param {Object} tx Transaction data with array of inputs and outputs
+   *
+   * @return {Object} Object with each token and it's balance in this tx for this wallet
+   **/
+  async getTxBalance(tx: WsTransaction, optionsParam = {}): Promise<{[tokenId: string]: number}> {
+    const options = Object.assign({ includeAuthorities: false }, optionsParam);
+
+    const addresses: string[] = [];
+
+    const generator = this.getAllAddresses();
```
pedroferreira1, line 223: I think a comment here explaining why you are using this syntax is important.

src/new/wallet.js
```diff
@@ -874,6 +886,7 @@ class HathorWallet extends EventEmitter {
         tokenHistory.sort((elem1, elem2) => elem2.timestamp - elem1.timestamp);
       } else {
         const currentHistory = tokensHistory[tokenUid];
+        console.log('will try to find index');
```
pedroferreira1, line 889: Debug log

src/new/wallet.js
```diff
@@ -927,19 +940,19 @@ class HathorWallet extends EventEmitter {
...
+  async getPreProcessedData(key) {
     if (Object.keys(this.preProcessedData).length === 0) {
-      this.preProcessWalletData();
+      await this.preProcessWalletData();
     }
     return this.preProcessedData[key];
   }
 
-  setState(state) {
+  async setState(state) {
```
pedroferreira1, line 950: We have this method in the new facade as well and it's not `async`. Maybe we should keep this method not async and just call `this.preProcessWalletData();` without await. It should have the same behaviour and we keep the same method signatures in both facades.

src/new/wallet.js
andreabadesso: Sorry for that, removed in 5bd097eed520f8eafcaf9e559ad40b74f869a6d3

src/new/wallet.js
andreabadesso: I was worried about potential race conditions of setting the state to `READY` without the `preProcessWalletData`

But since the `getTxbalance` resolves immediately, we should be fine, I've tested it and noticed no issues so far.

Added a comment with a notice on d192e3d2df1485dc1641a742f3e2d8627487de1b

What do you think?

src/wallet/wallet.ts
andreabadesso: I agree, done in c60d101d7eb4e332429190099b6939c40c905a60

src/config.ts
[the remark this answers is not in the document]
andreabadesso: Done in 16018606fae8d7952e5aa3471150617394bd3100

Is that what you mean?

src/new/wallet.js
pedroferreira1: I like it.

PR #306, opened 2022-01-10
feat: split prepare and sign tx

r4mmer:
### Summary

The `sendTransaction` class prepares and signs the transaction on the same method.

To use this with the hardware wallet we implement the preparation on the wallet and instantiate the `sendTransaction` with the prepared transaction with the signatures from the hardware wallet.

This PR aims to unify this preparation on the wallet-lib, creating the `prepareTxData` that will be used to send the tx to the hardware wallet and `prepareTxFrom` to create the tx from the signatures from the hardware wallet while keeping the same functionality at the `prepareTx` (as to not disturb the functionality of the other wallets that use this method).

### Acceptance Criteria

- Create a method to prepare the tx data without signing so it can be used by a hardware wallet.
- Create a method to complete the tx with the signatures from the hardware wallet
- Update the `prepareTx` method to use the same method to prepare the tx data.
  - Keep the same usage to not create issues with the software wallets using this method

### Security Checklist

### DIFF
--- a/src/new/sendTransaction.js
+++ b/src/new/sendTransaction.js
@@ -64,16 +64,16 @@ class SendTransaction extends EventEmitter {
 
   /**
    * Prepare transaction data from inputs and outputs
-   * Fill the inputs if needed, create output change if needed and sign inputs
+   * Fill the inputs if needed, create output change if needed
    *
    * @throws SendTxError
    *
-   * @return {Transaction} Transaction object prepared to be mined
+   * @return {Object} fullTxData with tokens array, inputs and outputs
    *
    * @memberof SendTransaction
    * @inner
    */
-  prepareTx() {
+  prepareTxData() {
     const tokensData = {};
     const HTR_UID = HATHOR_TOKEN_CONFIG.uid;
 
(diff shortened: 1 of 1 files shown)


src/new/sendTransaction.js
```diff
@@ -168,9 +169,67 @@ class SendTransaction extends EventEmitter {
...
+   * Fill the inputs if needed, create output change if needed
+   *
+   * @throws SendTxError
+   *
+   * @return {Transaction} Transaction object prepared to be mined
+   *
+   * @memberof SendTransaction
+   * @inner
+   */
+  prepareTxFrom(signatures) {
```
andreabadesso, line 213: I'm confused

The method is called `prepareTxFrom` and it receives `signatures`, but the docstring says it prepares the transaction from inputs, outputs and signatures

Should we update the docstring?

andreabadesso (changes requested):

src/new/sendTransaction.js
r4mmer: Changed the docstring.

This method prepares the transaction to be mined from the signatures (and pre-prepared tx data).

andreabadesso (approved):

src/new/sendTransaction.js
```diff
@@ -168,9 +169,69 @@ class SendTransaction extends EventEmitter {
...
+    if (this.fullTxData === null) {
+      // This method can only be called with a prepared tx data
+      // because prepareTxData may modify the inputs and outputs
+      throw new SendTxError(ErrorMessages.TRANSACTION_IS_NULL);
+    }
+
+    // add each input data from signature
+    const keys = wallet.getWalletData().keys;
+    for (const [index, input] of this.fullTxData.inputs.entries()) {
+      const signature = Buffer.from(signatures[index]);
```
pedroferreira1, line 225: I'm a bit confused about this `signatures` parameter. Is is a string or an array of bytes, or an array of string? We should add the type in the docstring with an example and explain that this method is mainly used to sign transactions with signatures that come from ledger.

pedroferreira1 (approved):

src/new/sendTransaction.js
r4mmer: Changed the doctring to reflect what should be the parameter (an array of Buffer, each being a signature). This `Buffer.from` ends up being redundant so I removed it.

PR #316, opened 2022-02-03
refactor: using auth xpubkey to authenticate wallet service requests

andreabadesso:
## Summary
Implements https://github.com/HathorNetwork/internal-issues/issues/76

### Acceptance Criteria
- We should use a different derivation from the one used to start the wallet to authenticate requests on the wallet service
- We should prove we own both `xpubs` (the wallet's account level and the auth_xpubkey) on the start wallet API
- We should use the `auth_xpubkey` to request a JWT token from the wallet service

### Security Checklist

### DIFF
--- a/src/utils/wallet.ts
+++ b/src/utils/wallet.ts
@@ -198,6 +198,39 @@ const wallet = {
   },
 
   /**
+   * Get auth xpubkey in the 280' purpose path from seed
+   *
+   * @param {String} seed 24 words
+   * @param {Object} options Options with passphrase and networkName
+   *
+   * @return {String} Wallet xpubkey
+   * @memberof Wallet
+   * @inner
+   */
+  getAuthXPubKeyFromSeed(seed: string, options: { passphrase?: string, networkName?: string } = {}): string {
+    const methodOptions = Object.assign({passphrase: '', networkName: 'mainnet', accountDerivationIndex: '0\''}, options);
+    const { accountDerivationIndex } = methodOptions;
+
+    const xpriv = this.getXPrivKeyFromSeed(seed, methodOptions);
+    const privkey = this.deriveAuthXpriv(xpriv);
+
+    return privkey.xpubkey;
+  },
+
+  /**
+   * Derive xpriv from root to the auth specific purpose derivation path
+   *
+   * @param {string} accountDerivationIndex String with derivation index of account (can be hardened)
+   *
+   * @return {HDPrivateKey} Derived private key
+   * @memberof Wallet
+   * @inner
+   */
+  deriveAuthXpriv(xpriv: HDPrivateKey): HDPrivateKey {
+    return xpriv.deriveNonCompliantChild(`m/${HATHOR_BIP44_CODE}'/${HATHOR_BIP44_CODE}'`);
+  },
+
+  /**
    * Get root xpriv from seed
    *
    * @param {String} seed 24 words
--- a/src/utils/wallet.ts
+++ b/src/utils/wallet.ts
@@ -7,6 +7,7 @@
 
 
 import { crypto, HDPublicKey, HDPrivateKey, Address } from 'bitcore-lib';
+import { createHash, HexBase64Latin1Encoding } from 'crypto';
 import Mnemonic from 'bitcore-mnemonic';
 import { HD_WALLET_ENTROPY, HATHOR_BIP44_CODE } from '../constants';
 import { XPubError, InvalidWords, UncompressedPubKeyError } from '../errors';
@@ -198,39 +199,6 @@ const wallet = {
   },
 
   /**
-   * Get auth xpubkey in the 280' purpose path from seed
-   *
-   * @param {String} seed 24 words
-   * @param {Object} options Options with passphrase and networkName
-   *
-   * @return {String} Wallet xpubkey
-   * @memberof Wallet
-   * @inner
-   */
-  getAuthXPubKeyFromSeed(seed: string, options: { passphrase?: string, networkName?: string } = {}): string {
-    const methodOptions = Object.assign({passphrase: '', networkName: 'mainnet', accountDerivationIndex: '0\''}, options);
-    const { accountDerivationIndex } = methodOptions;
-
-    const xpriv = this.getXPrivKeyFromSeed(seed, methodOptions);
-    const privkey = this.deriveAuthXpriv(xpriv);
-
-    return privkey.xpubkey;
-  },
-
-  /**
-   * Derive xpriv from root to the auth specific purpose derivation path
-   *
-   * @param {string} accountDerivationIndex String with derivation index of account (can be hardened)
-   *
-   * @return {HDPrivateKey} Derived private key
-   * @memberof Wallet
-   * @inner
-   */
-  deriveAuthXpriv(xpriv: HDPrivateKey): HDPrivateKey {
-    return xpriv.deriveNonCompliantChild(`m/${HATHOR_BIP44_CODE}'/${HATHOR_BIP44_CODE}'`);
-  },
-
-  /**
    * Get root xpriv from seed
    *
    * @param {String} seed 24 words
@@ -365,7 +333,25 @@ const wallet = {
 
     const derivedXpub = xpub.deriveChild(derivationIndex);
     return derivedXpub.xpubkey;
-  }
+  },
+
+  /**
+ * Calculate the double sha256 hash of the data.
+ *
+ * @remarks
+ * If encoding is provided a string will be returned; otherwise a Buffer is returned.
+ *
+ * @param data - Data to be hashed
+ * @param encoding - The encoding of the returned object
+ * @returns The sha256d hash of the data
+ */
+  sha256d(data: string, encoding: HexBase64Latin1Encoding): string {
+    const hash1 = createHash('sha256');
+    hash1.update(data);
+    const hash2 = createHash('sha256');
+    hash2.update(hash1.digest());
+    return hash2.digest(encoding);
+  },
 }
 
 export default wallet;
--- a/src/utils/wallet.ts
+++ b/src/utils/wallet.ts
@@ -345,7 +345,7 @@ const wallet = {
  * @param encoding - The encoding of the returned object
  * @returns The sha256d hash of the data
  */
-  sha256d(data: string, encoding: HexBase64Latin1Encoding): string {
+  sha256d(data: string, encoding: HexBase64Latin1Encoding): string | Buffer {
     const hash1 = createHash('sha256');
     hash1.update(data);
     const hash2 = createHash('sha256');
--- a/src/utils/wallet.ts
+++ b/src/utils/wallet.ts
@@ -334,24 +334,6 @@ const wallet = {
     const derivedXpub = xpub.deriveChild(derivationIndex);
     return derivedXpub.xpubkey;
   },
-
-  /**
- * Calculate the double sha256 hash of the data.
- *
- * @remarks
- * If encoding is provided a string will be returned; otherwise a Buffer is returned.
- *
- * @param data - Data to be hashed
- * @param encoding - The encoding of the returned object
- * @returns The sha256d hash of the data
- */
-  sha256d(data: string, encoding: HexBase64Latin1Encoding): string | Buffer {
-    const hash1 = createHash('sha256');
-    hash1.update(data);
-    const hash2 = createHash('sha256');
-    hash2.update(hash1.digest());
-    return hash2.digest(encoding);
-  },
 }
 
 export default wallet;
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -837,7 +837,7 @@ const wallet = {
    */
   changePinAndPassword({ oldPin, newPin, oldPassword, newPassword }) {
     if (this.isFromXPub()) {
-        throw WalletFromXPubGuard('changePinAndPassword');
+      throw WalletFromXPubGuard('changePinAndPassword');
     }
 
     if (oldPassword && !newPassword) {
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -1774,9 +1774,9 @@ const wallet = {
   /**
    * Get the privKey of the auth derivation
    *
-   * @param {string} pin Pin to decrypt the words
+   * @param {string} pin Pin to decrypt the auth key
    *
-   * @return {string} The auth private key
+   * @return {HDPrivateKey} The auth private key
    *
    * @memberof Wallet
    * @inner
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -1749,6 +1749,21 @@ const wallet = {
     return this.decryptData(accessData.words, password);
   },
 
+  /**
+   * Get the privKey of the auth derivation
+   *
+   * @param {string} pin Pin to decrypt the words
+   *
+   * @return {string} The auth private key
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getAuthKey(pin) {
+    const accessData = this.getWalletAccessData();
+    return HDPrivateKey(this.decryptData(accessData.authKey, pin));
+  },
+
   /*
    * Save backup done in storage
    *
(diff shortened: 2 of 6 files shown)


src/wallet.js
```diff
@@ -1746,6 +1771,21 @@ const wallet = {
     return this.decryptData(accessData.words, password);
   },
 
+  /**
+   * Get the privKey of the auth derivation
+   *
+   * @param {string} pin Pin to decrypt the words
```
r4mmer, line 1777:
```suggestion
   * @param {string} pin Pin to decrypt the auth key
```

src/utils/wallet.ts
```diff
@@ -332,7 +333,25 @@ const wallet = {
...
+ * Calculate the double sha256 hash of the data.
+ *
+ * @remarks
+ * If encoding is provided a string will be returned; otherwise a Buffer is returned.
+ *
+ * @param data - Data to be hashed
+ * @param encoding - The encoding of the returned object
+ * @returns The sha256d hash of the data
+ */
+  sha256d(data: string, encoding: HexBase64Latin1Encoding): string {
```
r4mmer, line 348:
```suggestion
  sha256d(data: string, encoding?: HexBase64Latin1Encoding): string | Buffer {
```

src/wallet.js
```diff
@@ -1746,6 +1771,21 @@ const wallet = {
...
+   *
+   * @return {string} The auth private key
```
r4mmer, line 1779:
```suggestion
   * @return {HDPrivateKey} The auth xpriv
```

src/wallet/wallet.ts
```diff
@@ -135,6 +138,40 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    this.authXpriv = null;
+  }
+
+  /**
+   * Get auth xpubkey in the 280' purpose path from seed
+   *
+   * @param {String} seed 24 words
+   * @param {Object} options Options with passphrase and networkName
+   *
+   * @return {String} Wallet xpubkey
```
r4mmer, line 150:
```suggestion
   * @return {String} auth xpubkey
```

src/wallet/wallet.ts
```diff
@@ -135,6 +138,40 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  static getAuthXPubKeyFromSeed(seed: string, options: { passphrase?: string, networkName?: string } = {}): string {
+    const methodOptions = Object.assign({passphrase: '', networkName: 'mainnet', accountDerivationIndex: '0\''}, options);
+    const { accountDerivationIndex } = methodOptions;
```
r4mmer, line 156: I think we are not using `accountDerivationIndex`

src/wallet/wallet.ts
```diff
@@ -135,6 +138,40 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * Derive xpriv from root to the auth specific purpose derivation path
+   *
+   * @param {string} accountDerivationIndex String with derivation index of account (can be hardened)
+   *
+   * @return {HDPrivateKey} Derived private key
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  static deriveAuthXpriv(xpriv: bitcore.HDPrivateKey): bitcore.HDPrivateKey {
+    return xpriv.deriveNonCompliantChild(`m/${HATHOR_BIP44_CODE}'/${HATHOR_BIP44_CODE}'`);
```
r4mmer, line 174: Maybe have the auth path on `src/constants.js`

src/wallet/wallet.ts
```diff
@@ -143,16 +180,45 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    });
+
+    const walletId = walletUtils.sha256d(xpub, 'hex');
+    const now = new Date();
+    const timestampNow = Math.floor(now.getTime() / 1000);
+
+    // prove we own the xpubkey
+    const xpubAccountPath = walletUtils.deriveXpriv(xpriv, '0\'')
+    const address = xpubAccountPath.publicKey.toAddress(this.network.getNetwork()).toString();
+    const message = new bitcore.Message(String(timestampNow).concat(walletId).concat(address));
```
r4mmer, line 204: Should we make the message creation as a static method on `HathorWalletServiceWallet` that receive the timestamp, walletId and address?

r4mmer (changes requested):

src/utils/wallet.ts
andreabadesso: Done in c8600b30a1fb0f5b8da60993a30dd65205e23505

src/wallet/wallet.ts
andreabadesso: Good idea! Done in 32685e3fe69c536c4659ec42d46594fe5d816d8d

src/wallet.js
andreabadesso: Thanks! Done in 7be2bddfaa7280addf82144f2b3ef67767fac655

src/wallet.js
andreabadesso: Thanks! Fixed in 7be2bddfaa7280addf82144f2b3ef67767fac655

src/wallet/wallet.ts
andreabadesso: Thanks! Fixed in 7be2bddfaa7280addf82144f2b3ef67767fac655

src/wallet/wallet.ts
andreabadesso: We are not! Thanks

Fixed in 7be2bddfaa7280addf82144f2b3ef67767fac655

src/wallet/wallet.ts
```diff
@@ -143,16 +186,45 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+
+    // prove we own the xpubkey
+    const xpubAccountPath = walletUtils.deriveXpriv(xpriv, '0\'')
+    const address = xpubAccountPath.publicKey.toAddress(this.network.getNetwork()).toString();
+    const message = new bitcore.Message(String(timestampNow).concat(walletId as string).concat(address));
+    const xpubkeySignature = message.sign(xpubAccountPath.privateKey);
+
+    // prove we own the auth_xpubkey
+    const authDerivedPrivKey = HathorWalletServiceWallet.deriveAuthXpriv(xpriv);
+    const authAddress = authDerivedPrivKey.publicKey.toAddress(this.network.getNetwork());
```
luislhl, line 215: The `.toString()` call isn't necessary here as it is in the code above?

luislhl (approved):

src/wallet/wallet.ts
```diff
@@ -135,6 +145,39 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   *
+   * @param {String} seed 24 words
+   * @param {Object} options Options with passphrase and networkName
+   *
+   * @return {String} auth xpubkey
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  static getAuthXPubKeyFromSeed(seed: string, options: { passphrase?: string, networkName?: string } = {}): string {
+    const methodOptions = Object.assign({passphrase: '', networkName: 'mainnet', accountDerivationIndex: '0\''}, options);
```
pedroferreira1, line 162: Why do you need `accountDerivationIndex`?

src/wallet/wallet.ts
```diff
@@ -135,6 +145,39 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    const xpriv = walletUtils.getXPrivKeyFromSeed(seed, methodOptions);
+    const privkey = HathorWalletServiceWallet.deriveAuthXpriv(xpriv);
+
+    return privkey.xpubkey;
+  }
+
+  /**
+   * Derive xpriv from root to the auth specific purpose derivation path
+   *
+   * @param {string} accountDerivationIndex String with derivation index of account (can be hardened)
```
pedroferreira1, line 173: What's this parameter? As far as I understood this is not the `xpriv` param, right?

src/wallet/wallet.ts
```diff
@@ -143,16 +186,45 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @memberof HathorWalletServiceWallet
    * @inner
    */
-  async start() {
+  async start({ pinCode }) {
```
pedroferreira1, line 189: Missing new parameter on docstring

src/wallet/wallet.ts
```diff
@@ -464,14 +545,12 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @memberof HathorWalletServiceWallet
    * @inner
    */
-  signMessage(seed: string, timestamp: number): string {
-    const xpriv = walletUtils.getXPrivKeyFromSeed(seed, {passphrase: this.passphrase, networkName: this.network.name});
-    const derivedPrivKey = walletUtils.deriveXpriv(xpriv, '0\'');
-    const address = derivedPrivKey.publicKey.toAddress(this.network.getNetwork()).toString();
-
+  signMessage(xpriv: bitcore.HDPrivateKey, timestamp: number): string {
+    // const xpriv = walletUtils.getXPrivKeyFromSeed(seed, {passphrase: this.passphrase, networkName: this.network.name});
```
pedroferreira1, line 549: Can we remove this?

src/wallet/wallet.ts
```diff
@@ -143,16 +186,45 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      passphrase: this.passphrase,
+      networkName: this.network.name,
+    });
+
+    const walletId = walletUtils.sha256d(xpub, 'hex');
+    const now = new Date();
+    const timestampNow = Math.floor(now.getTime() / 1000);
+
+    // prove we own the xpubkey
+    const xpubAccountPath = walletUtils.deriveXpriv(xpriv, '0\'')
```
pedroferreira1, line 208: The following line above 

```
const xpub = walletUtils.getXPubKeyFromSeed(this.seed, {passphrase: this.passphrase, networkName: this.network.name});
```

Already derives this exactly xpub I think. We probably can use it instead of deriving again.

pedroferreira1: Actually I think you are deriving an xpriv because you are using it to sign the message below. If that's the case you can ignore my previous comment and would be great to change the variable name, it's misleading.

src/wallet/wallet.ts
```diff
@@ -143,16 +186,45 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    const address = xpubAccountPath.publicKey.toAddress(this.network.getNetwork()).toString();
+    const message = new bitcore.Message(String(timestampNow).concat(walletId as string).concat(address));
+    const xpubkeySignature = message.sign(xpubAccountPath.privateKey);
+
+    // prove we own the auth_xpubkey
+    const authDerivedPrivKey = HathorWalletServiceWallet.deriveAuthXpriv(xpriv);
+    const authAddress = authDerivedPrivKey.publicKey.toAddress(this.network.getNetwork());
+
+    const authMessage = new bitcore.Message(String(timestampNow).concat(walletId as string).concat(authAddress));
+    const authXpubkeySignature = authMessage.sign(authDerivedPrivKey.privateKey);
```
pedroferreira1, line 218: Maybe we can create a method to get the signature (I think we already has this method but we should only refactor it). The method would receive the `xpriv`, `timestamp` and `walletId`, then it can create the address, create the message and return the signature.

This would prevent differences in the code as the onde mentioned by @luislhl in the comment above.

pedroferreira1: I just checked and we can use the method `signMessage` that already exists. You don't even need to send `walletId`, given that it uses from `this`.

src/wallet/wallet.ts
```diff
@@ -143,16 +186,45 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+
     const xpubChangeDerivation = walletUtils.xpubDeriveChild(xpub, 0);
     const firstAddress = walletUtils.getAddressAtIndex(xpubChangeDerivation, 0, this.network.name);
+
     this.xpub = xpub;
+    this.authXpriv = authDerivedPrivKey;
```
pedroferreira1, line 224: You are setting to the `authXpriv` a private key and not an xpriv, is that correct?

pedroferreira1 (changes requested):

src/utils/wallet.ts
```diff
@@ -332,7 +333,25 @@ const wallet = {
...
+ * Calculate the double sha256 hash of the data.
+ *
+ * @remarks
+ * If encoding is provided a string will be returned; otherwise a Buffer is returned.
+ *
+ * @param data - Data to be hashed
+ * @param encoding - The encoding of the returned object
+ * @returns The sha256d hash of the data
+ */
+  sha256d(data: string, encoding: HexBase64Latin1Encoding): string | Buffer {
```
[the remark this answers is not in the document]
andreabadesso: We can, refactored it on 37b8cba0030328e82191a71ef0b66af6b6d692cb

src/wallet.js
```diff
@@ -258,13 +261,32 @@ const wallet = {
...
+   * Encrypt a auth xpriv to be used for auth on the Wallet Service facade with
+   * a password and save it on localStorage
+   *
+   * @param {string} key Key to be encrypted
+   * @param {string} password Password to encrypt
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  storeEncryptedAuthKey(key, password) {
```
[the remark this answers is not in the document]
andreabadesso: Done in 9196828eeec134bf5b4e68bb40b51fc88f196f1d

src/wallet/wallet.ts
andreabadesso: We don't need it

Removed in 504cf8c44a754aab33c85a3f99791de9e080afa3

src/wallet/wallet.ts
andreabadesso: Fixed in 35809f7e9114b604a02e97799a11921286c42ee2

src/wallet/wallet.ts
andreabadesso: Fixed in 35809f7e9114b604a02e97799a11921286c42ee2

src/wallet/wallet.ts
andreabadesso: You are absolutely right, sorry for that

Refactored the `signMessage` method to accept `walletId` and refactored the `start` method to use it instead of manually signing messages

src/wallet/wallet.ts
andreabadesso: Done in 3ee21d0a8c4a456520383a10647758af308ad7e7

src/wallet/wallet.ts
andreabadesso: Refactored in 3ee21d0a8c4a456520383a10647758af308ad7e7 along with other changes on `signMessage`

src/wallet/wallet.ts
andreabadesso: You are right, refactored methods at 76913ffaad9f3379fa19ba8bf1fa53ae3a3f05c1

src/wallet/wallet.ts
andreabadesso: I think we don't need it to be `static` as it is only being used on this class, I refactored this part of the code on 715e5f9abc085093f6180fd809d86959c1e7a10b

Can you please check again?

src/wallet/wallet.ts
andreabadesso: @luislhl I refactored this part of the code on this commit: 715e5f9abc085093f6180fd809d86959c1e7a10b

Can you please check again?

src/wallet.js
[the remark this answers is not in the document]
andreabadesso: It would make sense as it is exclusively being used by the wallet service facade

But I prefer keeping it on `src/wallet.js` so everything that is related to data storage is on this file, we don't have to find what is mutating the storage data anywhere else on future code changes

What do you think?

r4mmer (approved):

pedroferreira1 (approved):

PR #326, opened 2022-03-04
fix: waiting for wallet load (polling) before setting up websocket

andreabadesso:
### Acceptance Criteria
- Include here all things that this PR should solve

### Security Checklist

### DIFF
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -265,21 +265,14 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
 
     const handleCreate = async (data: WalletStatus) => {
       this.walletId = data.walletId;
+
       if (data.status === 'creating') {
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.ts
```diff
@@ -265,22 +266,19 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
 
     const handleCreate = async (data: WalletStatus) => {
       this.walletId = data.walletId;
+
       if (data.status === 'creating') {
-        this.walletStatusInterval = setInterval(async () => {
```
pedroferreira1, line 269: Should we remove this attribute from the class and constructor?

andreabadesso: Yes we should, thanks

Done in 82f4de533c16baad51c83356a3609c393b5175aa

pedroferreira1 (approved):

PR #337, opened 2022-03-23
feat: handle tx-mining rate limit

luislhl:
### Acceptance Criteria
The lib should be able no nicely handle Http Code 429 that it may get from tx-mining-service

### Notes
1. I don't know how I should go about translating the new error message into other languages in wallet-mobile.

2. In wallet-mobile, the message would be stripped:

mobile_rate_limit

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
       "<rootDir>/setupTests.js"
     ],
     "collectCoverageFrom": [
-      "<rootDir>/src/*.js",
+      "<rootDir>/src/**/*.js",
+      "<rootDir>/src/**/*.ts",
       "!<rootDir>/node_modules/"
     ],
     "modulePathIgnorePatterns": [
--- a/src/api/txMining.js
+++ b/src/api/txMining.js
@@ -33,8 +33,8 @@ const txMiningApi = {
     }
     return txMiningRequestClient(resolve).post('submit-job', postData).then((res) => {
       resolve(res.data)
-    }, (res) => {
-      return Promise.reject(res);
+    }, (error) => {
+      return Promise.reject(error);
     });
   },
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -7,7 +7,8 @@
       "<rootDir>/setupTests.js"
     ],
     "collectCoverageFrom": [
-      "<rootDir>/src/*.js",
+      "<rootDir>/src/**/*.js",
+      "<rootDir>/src/**/*.ts",
```
luislhl, line 11: I noticed we were not including some files in the cli coverage

src/api/txMining.js
```diff
@@ -33,8 +33,8 @@ const txMiningApi = {
     }
     return txMiningRequestClient(resolve).post('submit-job', postData).then((res) => {
       resolve(res.data)
-    }, (res) => {
-      return Promise.reject(res);
+    }, (error) => {
```
luislhl, line 36: I noticed this object is not a response, it's actually an error object that contains the response (among other things)

src/wallet/mineTransaction.ts
```diff
@@ -81,6 +84,20 @@ class MineTransaction extends EventEmitter {
...
+  /**
+   * Used to handle errors in requests to the tx mining API 
+   * 
+   * @param error: The error that was received from the axiosInstance
+   */
+  handleRequestError(error: any) {
+    if (error.response && error.response.status === 429) {
+      this.emit('error', rateLimitExceededError);
+    } else {
+      console.error(error);
```
luislhl, line 96: Logging the error will give more info to anyone running a wallet about eventual problems. Especially for people running wallet-headless, because the error will be logged its stdout.

pedroferreira1 (approved):

andreabadesso (approved):

PR #342, opened 2022-04-01
feat: wallet service facade should be able to start from a xprivkey on the account path derivation

andreabadesso:
## Motivation
Currently, the wallet service facade expects to always receive a seed to be able to derive the change level path private key it uses to identify each wallet (we use a double sha256 encoded hash of the xpubkey on that path as a `wallet-id`).

The problem is that the `wallet-desktop` expects to be able to start the wallet using an `xprivkey` when restoring an already loaded wallet and a `xpubkey` when using the Hardware Wallet and currently, the `mainKey` stored on the device's local storage is on the change level path, making it impossible for the wallet-service to get the account level path

The idea here is to store a account level path xprivkey on storage, encrypted using the PIN code, every time the wallet is started from the seed and use it when needed

### Acceptance Criteria
- We should be able to load a wallet from a `xprivkey` on the account path derivation
- We should store a `xprivkey` on the account path derivation when starting from the seed

### Security Checklist

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -22,6 +22,7 @@ import {
   LOAD_WALLET_MAX_RETRY,
   LOAD_WALLET_RETRY_SLEEP,
   WALLET_SERVICE_AUTH_DERIVATION_PATH,
+  P2PKH_ACCT_PATH,
   P2SH_ACCT_PATH,
 } from './constants';
 import Mnemonic from 'bitcore-mnemonic';
@@ -276,20 +277,25 @@ const wallet = {
   executeGenerateWallet(words, passphrase, pin, password, loadHistory, multisig) {
     let code = new Mnemonic(words);
     let xpriv = code.toHDPrivateKey(passphrase, network.getNetwork());
-    let authXpriv = xpriv.deriveNonCompliantChild(WALLET_SERVICE_AUTH_DERIVATION_PATH);
+    const authXpriv = xpriv.deriveNonCompliantChild(WALLET_SERVICE_AUTH_DERIVATION_PATH);
+    const accPrivKey = xpriv.deriveNonCompliantChild(P2PKH_ACCT_PATH);
+
     let privkey;
     if (multisig) {
       privkey = xpriv.deriveNonCompliantChild(`${P2SH_ACCT_PATH}/0`);
     } else {
       privkey = xpriv.deriveNonCompliantChild(`m/44'/${HATHOR_BIP44_CODE}'/0'/0`);
     }
+
     let encryptedData = this.encryptData(privkey.xprivkey, pin)
-    let encryptedAuthXpriv = this.encryptData(authXpriv.xprivkey, pin)
-    let encryptedDataWords = this.encryptData(words, password)
+    let encryptedAccountPathXpriv = this.encryptData(accPrivKey.xprivkey, pin);
+    let encryptedAuthXpriv = this.encryptData(authXpriv.xprivkey, pin);
+    let encryptedDataWords = this.encryptData(words, password);
 
     // Save in storage the encrypted private key and the hash of the pin and password
     let access = {
       mainKey: encryptedData.encrypted.toString(),
+      acctPathMainKey: encryptedAccountPathXpriv.encrypted.toString(),
       hash: encryptedData.hash.key.toString(),
       salt: encryptedData.hash.salt,
       words: encryptedDataWords.encrypted.toString(),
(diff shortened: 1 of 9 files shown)


src/wallet/wallet.ts
```diff
@@ -275,24 +307,44 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
 
-  generateCreateWalletAuthData(seed: string): CreateWalletAuthData {
-    const xpriv = walletUtils.getXPrivKeyFromSeed(seed, {passphrase: this.passphrase, networkName: this.network.name});
-    const xpub = walletUtils.getXPubKeyFromSeed(seed, {passphrase: this.passphrase, networkName: this.network.name});
-    const authXpub = HathorWalletServiceWallet.getAuthXPubKeyFromSeed(seed, {
-      passphrase: this.passphrase,
-      networkName: this.network.name,
-    });
+  generateCreateWalletAuthData(pinCode: string): CreateWalletAuthData {
+    let xpriv: bitcore.HDPrivateKey;
```
andreabadesso, line 311: Not a `xpriv`, we should change the variable name

src/wallet/wallet.ts
```diff
@@ -275,24 +307,44 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    let xpub: string;
+    let authXpub: string;
+    let xprivAccountPath: string;
```
andreabadesso, line 314: Not a string and not a `xpriv`, we should change the type and variable name

src/wallet/wallet.ts
```diff
@@ -275,24 +305,44 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    let authDerivedPrivKey: bitcore.HDPrivateKey;
 
     const now = new Date();
     const timestampNow = Math.floor(now.getTime() / 1000); // in seconds
-    const walletId = HathorWalletServiceWallet.getWalletIdFromXPub(xpub);
+
+    if (this.seed) {
+      xpriv = walletUtils.getXPrivKeyFromSeed(this.seed, {passphrase: this.passphrase, networkName: this.network.name});
+      xpub = walletUtils.getXPubKeyFromSeed(this.seed, {passphrase: this.passphrase, networkName: this.network.name});
```
pedroferreira1, line 320: I would just add a comment here and above.

`getXPubKeyFromSeed` gets the xpub in the account derivation path while `getXPrivKeyFromSeed` gets the privateKey in the root derivation path.

src/wallet/wallet.ts
pedroferreira1: I think this shouldn't be defined here. It's only used inside each `if..else` block and inside each one has different meanings. In the first, it's the root private key and in the second is the private key in the account path.

src/wallet/wallet.ts
andreabadesso: Done in 99fbab7ef1e29028e8ef2e1fcc8688643c14d9a4

src/wallet.js
```diff
@@ -276,20 +278,25 @@ const wallet = {
...
     if (multisig) {
       privkey = xpriv.deriveNonCompliantChild(`${P2SH_ACCT_PATH}/0`);
     } else {
       privkey = xpriv.deriveNonCompliantChild(`m/44'/${HATHOR_BIP44_CODE}'/0'/0`);
     }
+
     let encryptedData = this.encryptData(privkey.xprivkey, pin)
-    let encryptedAuthXpriv = this.encryptData(authXpriv.xprivkey, pin)
-    let encryptedDataWords = this.encryptData(words, password)
+    let encryptedAccountPathXpriv = this.encryptData(accPrivKey.xprivkey, pin);
```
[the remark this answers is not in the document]
andreabadesso: Refactored all of them to `const` as we are not mutating those variables

96a47b2f9975dd3fee9bf152e9e3f471edd72fe7

src/wallet/wallet.ts
```diff
@@ -80,8 +78,12 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
   network: Network;
   // Method to request the password from the client
   private requestPassword: Function;
+  // String with 24 words separated by space
+  private seed: string | null;
   // Xpub of the wallet
   private xpub: string | null;
+  // Xpriv of the wallet
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 99fbab7ef1e29028e8ef2e1fcc8688643c14d9a4

src/wallet/types.ts
```diff
@@ -237,6 +237,7 @@ export interface IHathorWallet {
   getFullHistory(): TransactionFullObject[];
   getTxBalance(tx: WsTransaction, optionsParams): Promise<{[tokenId: string]: number}>;
   onConnectionChangedState(newState: ConnectionState): void;
+  generateCreateWalletAuthData(pinCode: string): CreateWalletAuthData;
```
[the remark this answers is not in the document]
andreabadesso: We are not, removed it from the general `IHathorWallet` type as it should be used (as soon as the old facade gets converted to typescript) on both facades

r4mmer (approved):
Are we going to add tests for the new methods and features?

pedroferreira1 (approved):

andreabadesso:
@r4mmer Will create tests for the new methods in a new PR, created an issue for that [here](https://github.com/HathorNetwork/hathor-wallet-lib/issues/351)

PR #344, opened 2022-04-08
feat: get mint and melt authority method on both facades

andreabadesso:
### Acceptance Criteria
- We should include `getMintAuthority` and `getMeltAuthority` on the wallet service facade
- Both methods should be async and have the same signature

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1438,7 +1438,7 @@ class HathorWallet extends EventEmitter {
    *
    * @return {Array} Array of objects with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
    **/
-  getMintAuthority(tokenUid, options = {}) {
+  async getMintAuthority(tokenUid, options = {}) {
     const newOptions = Object.assign({many: false, skipSpent: true}, options);
     return this.selectAuthorityUtxo(tokenUid, wallet.isMintOutput.bind(wallet), newOptions);
   }
@@ -1456,7 +1456,7 @@ class HathorWallet extends EventEmitter {
    *
    * @return {Array} Array of objects with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
    **/
-  getMeltAuthority(tokenUid, options = {}) {
+  async getMeltAuthority(tokenUid, options = {}) {
     const newOptions = Object.assign({many: false, skipSpent: true}, options);
     return this.selectAuthorityUtxo(tokenUid, wallet.isMeltOutput.bind(wallet), newOptions);
   }
--- a/src/wallet/types.ts
+++ b/src/wallet/types.ts
@@ -212,14 +212,14 @@ export interface IStopWalletParams {
 };
 
 export interface IHathorWallet {
-  start(options: { pinCode: string, password: string });
+  start(options: { pinCode: string, password: string }): Promise<void>;
   getAllAddresses(): AsyncGenerator<GetAddressesObject>;
   getBalance(token: string | null): Promise<GetBalanceObject[]>;
   getTokens(): Promise<string[]>;
   getTxHistory(options: { token_id?: string, count?: number, skip?: number }): Promise<GetHistoryObject[]>;
   sendManyOutputsTransaction(outputs: OutputRequestObj[], options: { inputs?: InputRequestObj[], changeAddress?: string }): Promise<Transaction>;
   sendTransaction(address: string, value: number, options: { token?: string, changeAddress?: string }): Promise<Transaction>;
-  stop(params?: IStopWalletParams);
+  stop(params?: IStopWalletParams): void;
   getAddressAtIndex(index: number): string;
   getCurrentAddress({ markAsUsed: boolean }): AddressInfoObject;
   getNextAddress(): AddressInfoObject;
@@ -237,6 +237,7 @@ export interface IHathorWallet {
   getFullHistory(): TransactionFullObject[];
   getTxBalance(tx: WsTransaction, optionsParams): Promise<{[tokenId: string]: number}>;
   onConnectionChangedState(newState: ConnectionState): void;
+  generateCreateWalletAuthData(pinCode: string): CreateWalletAuthData;
 }
 
 export interface ISendTransaction {
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -260,7 +260,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
         password,
         false,
       );
-    } else if(this.xpriv) {
+    } else if (this.xpriv) {
       wallet.executeGenerateWalletFromXPriv(
         this.xpriv,
         pinCode,
@@ -512,10 +512,11 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @inner
    */
   private async onWalletReady() {
+    // We should wait for new addresses before setting wallet to ready
+    await this.getNewAddresses(true);
+
     this.setupConnection();
     this.setState(walletState.READY);
-
-    await this.getNewAddresses();
   }
 
   setupConnection() {
@@ -575,8 +576,10 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @memberof HathorWalletServiceWallet
    * @inner
    */
-  private async getNewAddresses() {
-    this.failIfWalletNotReady();
+  private async getNewAddresses(ignoreWalletReady = false) {
+    if (!ignoreWalletReady) {
+      this.failIfWalletNotReady();
+    }
     const data = await walletApi.getNewAddresses(this);
     this.newAddresses = data.addresses;
     this.indexToUse = 0;
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -70,8 +70,6 @@ enum walletState {
 }
 
 class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
-  // String with 24 words separated by space
-  seed: string | null;
   // String with wallet passphrase
   passphrase: string;
   // Wallet id from the wallet service
@@ -80,8 +78,12 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
   network: Network;
   // Method to request the password from the client
   private requestPassword: Function;
+  // String with 24 words separated by space
+  private seed: string | null;
   // Xpub of the wallet
   private xpub: string | null;
+  // Xpriv of the wallet
+  private xpriv: string | null;
   // Xpriv of the auth derivation path
   private authPrivKey: bitcore.HDPrivateKey | null;
   // State of the wallet. One of the walletState enum options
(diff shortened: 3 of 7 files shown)


r4mmer (approved):

src/wallet/wallet.ts
```diff
@@ -1099,6 +1099,57 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * @param tokenId of the token to select the authority utxo
+   * @param options Object with custom options.
+   *  {
+   *    'many': if should return many utxos or just one (default false),
+   *    'skipSpent': if should not include spent utxos (default true)
+   *  }
+   *
+   * @return Array of objects with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
+   **/
+  async getMintAuthority(tokenId: string, options: { many?: boolean, skipSpent?: boolean } = {}) {
```
pedroferreira1, line 1116: Missing return type

src/wallet/wallet.ts
```diff
@@ -1099,6 +1099,57 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      skipSpent: newOptions.skipSpent,
+      count: newOptions.many ? undefined : 1,
+    });
+
+    return utxos;
+  }
+
+  /**
+   * Get melt authorities
+   * Uses the getUtxos method to return one or many mint authorities
```
pedroferreira1, line 1130: melt*

src/wallet/wallet.ts
andreabadesso: Normalized the return of both facades on 3bc7cb238f85a5f9edb650477382f4e02c1f4eac, can you please check again?

src/wallet/wallet.ts
```diff
@@ -648,7 +648,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @memberof HathorWalletServiceWallet
    * @inner
    */
-  async getUtxos(options: { tokenId?: string, authority?: number, addresses?: string[], totalAmount?: number, count?: number } = {}): Promise<{ utxos: Utxo[], changeAmount: number }> {
+  async getUtxos(options: { tokenId?: string, authority?: number, addresses?: string[], totalAmount?: number, count?: number, skipSpent?: boolean } = {}): Promise<{ utxos: Utxo[], changeAmount: number }> {
```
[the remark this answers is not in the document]
andreabadesso: Done in 8a933d7a05619e05004b18b02ab707b6cf5bfdc1

src/wallet/wallet.ts
```diff
@@ -1099,6 +1099,57 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * @param tokenId of the token to select the authority utxo
+   * @param options Object with custom options.
+   *  {
+   *    'many': if should return many utxos or just one (default false),
+   *    'skipSpent': if should not include spent utxos (default true)
+   *  }
+   *
+   * @return Array of objects with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
+   **/
+  async getMeltAuthority(tokenId: string, options: { many?: boolean, skipSpent?: boolean } = {}) {
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 3bc7cb238f85a5f9edb650477382f4e02c1f4eac

Also refactored the return to be exactly equal to the old facade

src/wallet/wallet.ts
andreabadesso: Fixed in c427d6afea611a3f335cdec5dedc57a4f6947d11

andreabadesso:
@pedroferreira1 @r4mmer 
Updated this PR with the necessary changes to use the new `getTxOutputs` API method from [this PR](https://github.com/HathorNetwork/hathor-wallet-service/pull/200)

src/new/wallet.js
```diff
@@ -1454,11 +1464,21 @@ class HathorWallet extends EventEmitter {
...
-   * @return {Array} Array of objects with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
+   * @return {Array} Array of objects with {txId, index, address} of the authority output. Returns an empty array in case there are no tx_outupts for this type
    **/
-  getMeltAuthority(tokenUid, options = {}) {
+  async getMeltAuthority(tokenUid, options = {}) {
     const newOptions = Object.assign({many: false, skipSpent: true}, options);
-    return this.selectAuthorityUtxo(tokenUid, wallet.isMeltOutput.bind(wallet), newOptions);
+    const txOutputs = this.selectAuthorityUtxo(tokenUid, wallet.isMeltOutput.bind(wallet), newOptions);
+
+    if (!txOutputs) {
```
pedroferreira1, line 1473: Maybe we can create a method with the lines from here and below. So if we decide to add a new key in the return we don't need to add in both methods

src/wallet/wallet.ts
```diff
@@ -628,8 +629,8 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @inner
    */
   async getUtxoFromId(txId: string, index: number): Promise<Utxo | null> {
-    const data = await walletApi.getUtxos(this, { txId, index });
-    const utxos = data.utxos;
+    const data = await walletApi.getTxOutputs(this, { txId, index });
```
pedroferreira1, line 632: I think you should continue using the utxo API. For example, in this case here we don't want any return if it's spent, so the tx outputs API should never be used here, only the utxo.

We could force just the parameter `skipSpent: true` but for what I've seen in the wallet service implementation, if we send `txId` and `index`, you ignore the other parameters.

I think in this line here we should return only utxos with one of the solutions above, not sure which one is the best.

src/wallet/wallet.ts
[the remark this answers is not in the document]
pedroferreira1: Actually now you are not using this method for the mint/melt authorities, then I believe you should remove this parameter. Makes no sense to have a `skipSpent` parameter in the `getUtxos` method

src/wallet/wallet.ts
```diff
@@ -1099,6 +1102,65 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    return txOutputs.map((txOutput) => ({
+      txId: txOutput.txId,
+      index: txOutput.index,
+      address: txOutput.address,
+    }));
+  }
+
+  /**
+   * Get melt authorities
+   * Uses the getTxOutputs API method to return one or many mint authorities
```
pedroferreira1, line 1137: melt*

src/wallet/wallet.ts
```diff
@@ -1099,6 +1102,65 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * @param tokenId of the token to select the authority utxo
+   * @param options Object with custom options.
+   *  {
+   *    'many': if should return many utxos or just one (default false),
+   *    'skipSpent': if should not include spent utxos (default true)
+   *  }
+   *
+   * @return Array of objects with {txId, index, address} of the authority output. Returns an empty array in case there are no tx outputs for this type
+   **/
+  async getMeltAuthority(tokenId: string, options: { many?: boolean, skipSpent?: boolean } = {}): Promise<AuthorityTxOutput[]> {
```
pedroferreira1, line 1148: This method and the one above are pretty much the same, except for the authority mask. We should refactor to remove this code duplication.

pedroferreira1 (changes requested):

src/wallet/wallet.ts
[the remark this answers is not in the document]
andreabadesso: You are right, removed it on 6d99da5acbf43946e7a5017bdf98dd472903e392

src/new/wallet.js
andreabadesso: We absolutely can, done in 3629080e5253bdc4439d33d9ef70ca66671335be

src/wallet/wallet.ts
andreabadesso: As we've discussed on slack, we will use the `tx_output` API, but will always send the `skipSpent: true`. The tx_output API will only return the tx_output if it is an UTXO (spentBy is null). This has been done on the wallet-service API PR

a5300f801e24c0149aea4630e34ae5f9c8345b04

src/wallet/wallet.ts
andreabadesso: Fixed in b058f570f3e8fc3012f6ac6d3f112d8ff4cf5410

src/wallet/wallet.ts
andreabadesso: Refactored in d04aa7de5199f922b6f4fd140422d633c192bb0b

src/wallet/types.ts
```diff
@@ -161,6 +161,12 @@ export interface Utxo {
   addressPath: string; // path to generate output address
 }
 
+export interface AuthorityTxOutput {
```
r4mmer, line 164: Should we have which kind of authority this is?

r4mmer (approved):

src/new/wallet.js
```diff
@@ -1436,11 +1455,13 @@ class HathorWallet extends EventEmitter {
    *    'skipSpent': if should not include spent utxos (default true)
    *  }
    *
-   * @return {Array} Array of objects with {tx_id, index, address} of the authority output. Returns null in case there are no utxos for this type
+   * @return {Array} Array of objects with {txId, index, address} of the authority output. Returns an empty array in case there are no tx_outupts for this type
```
pedroferreira1, line 1458: Returns a promise now that the method is async

src/new/wallet.js
```diff
@@ -1454,11 +1475,13 @@ class HathorWallet extends EventEmitter {
...
```
pedroferreira1, line 1478: Returns a promise now that the method is async

src/wallet/wallet.ts
```diff
@@ -1100,6 +1106,68 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     return parseInt(parts[5], 10);
   }
 
+  /**
+   * Helper method to get authority tx_outputs
+   * Uses the getTxOutputs API method to return one or many mint authorities
```
pedroferreira1, line 1111: one or many authorities* not only mint

src/wallet/wallet.ts
```diff
@@ -1100,6 +1106,68 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   */
+  async _getAuthorityTxOutput(options: {tokenId: string, authority: string, skipSpent: boolean, maxOutputs?: number }): Promise<AuthorityTxOutput[]> {
```
pedroferreira1, line 1113: Shouldn't `authority` be a `number`?

andreabadesso: Yes, fixed in 5b788957fd1124ee7f4f2f51eda098737313bb7a

src/wallet/wallet.ts
andreabadesso: Yes, fixed in 5b788957fd1124ee7f4f2f51eda098737313bb7a

src/new/wallet.js
andreabadesso: Done in 5b788957fd1124ee7f4f2f51eda098737313bb7a

src/new/wallet.js
andreabadesso: Fixed in 5b788957fd1124ee7f4f2f51eda098737313bb7a

src/wallet/types.ts
```diff
@@ -255,6 +261,7 @@ export interface IHathorWallet {
   getTxBalance(tx: WsTransaction, optionsParams): Promise<{[tokenId: string]: number}>;
   onConnectionChangedState(newState: ConnectionState): void;
   getTokenDetails(tokenId: string): Promise<TokenDetailsObject>;
+  generateCreateWalletAuthData(pinCode: string): CreateWalletAuthData;
```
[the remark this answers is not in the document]
andreabadesso: We are not, removed it in d065e231f094d39f7e6374cba5e69831420c4793

Also added docstring to `generateCreateWalletAuthData`

src/wallet/types.ts
andreabadesso: That's a nice refactor, done in b033fd954ad7715f7dc82c775082ba3dbf5b53cc

pedroferreira1 (approved):

PR #345, opened 2022-04-11
feat: get tokens details from both facades

andreabadesso:
## Motivation

Currently, our wallet desktop is querying the fullnode for token details, we should use the same method on both facades to query this information from the wallet-service (if on the facade) or the full-node. This new method will utilize the token info API, [created on the wallet-service](https://github.com/HathorNetwork/hathor-wallet-service/pull/202)

### Acceptance Criteria
- We should have a method on both facades to query token details

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1847,13 +1847,28 @@ class HathorWallet extends EventEmitter {
    *
    * @param tokenId Token uid to get the token details
    *
-   * @memberof HathorWalletServiceWallet
-   * @inner
+   * @return {Promise} token details
    */
   async getTokenDetails(tokenId) {
-    return new Promise((resolve, reject) => {
+    const result = await new Promise((resolve) => {
       return walletApi.getGeneralTokenInfo(tokenId, resolve);
     });
+
+    const { name, symbol, mint, melt, total, transactions_count } = result;
+
+    // Transform to the same format the wallet service facade responds
+    return {
+      totalSupply: total,
+      totalTransactions: transactions_count,
+      tokenInfo: {
+        name,
+        symbol,
+      },
+      authorities: {
+        mint: mint.length > 0,
+        melt: melt.length > 0,
+      },
+    };
   }
 
   isReady() {
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1842,6 +1842,20 @@ class HathorWallet extends EventEmitter {
     }
   }
 
+  /**
+   * Call get token details API
+   *
+   * @param tokenId Token uid to get the token details
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  async getTokenDetails(tokenId) {
+    return new Promise((resolve, reject) => {
+      return walletApi.getGeneralTokenInfo(tokenId, resolve);
+    });
+  }
+
   isReady() {
     return this.state === HathorWallet.READY;
   }
--- a/src/wallet/api/walletApi.ts
+++ b/src/wallet/api/walletApi.ts
@@ -18,7 +18,8 @@ import {
   TxProposalCreateResponseData,
   TxProposalUpdateResponseData,
   UtxoResponseData,
-  AuthTokenResponseData
+  AuthTokenResponseData,
+  TokenDetailsResponseData,
 } from '../types';
 import HathorWalletServiceWallet from '../wallet';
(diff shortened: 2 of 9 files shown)


src/new/wallet.js
```diff
@@ -1842,6 +1842,20 @@ class HathorWallet extends EventEmitter {
     }
   }
 
+  /**
+   * Call get token details API
+   *
+   * @param tokenId Token uid to get the token details
+   *
+   * @memberof HathorWalletServiceWallet
```
pedroferreira1, line 1850: Wrong docstring.

src/wallet/wallet.ts
```diff
@@ -1225,6 +1226,53 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
 
+  /**
+   * Call get token details API
+   *
+   * @param tokenId Token uid to get the token details
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  async getTokenDetails(tokenId: string) {
```
pedroferreira1, line 1237: Missing return type

src/wallet/wallet.ts
```diff
@@ -1225,6 +1226,53 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    const response = await walletApi.getTokenDetails(this, tokenId);
+    const details: TokenDetailsObject = response.details;
+
+    // We should format the response to the same format the thin wallet currently responds
```
pedroferreira1, line 1241: Or the opposite. We designed the wallet service to have the best responses format to serve our wallets, then we've been formatting the old facade returns.

src/wallet/wallet.ts
```diff
@@ -1225,6 +1226,53 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+          [authority, ...meltList],
+        ];
+      } else if (authority.authorities === TOKEN_MINT_MASK) {
+        return [
+          [authority, ...mintList],
+          meltList,
+        ];
+      }
+
+      return acc;
```
pedroferreira1, line 1263: We should never reach here, right?

andreabadesso: We should never reach, but if we do for some weird reason, we will ignore this utxo

I'll add a comment

pedroferreira1: Maybe we should throw an error so we can get possible problems in the wallet service response

src/wallet/api/walletApi.ts
```diff
@@ -93,6 +94,17 @@ const walletApi = {
...
   },
 
+  async getTokenDetails(wallet: HathorWalletServiceWallet, tokenId: string): Promise<TokenDetailsResponseData> {
+    const axios = await axiosInstance(wallet, true);
+    const response = await axios.get(`wallet/tokens/${tokenId}/details`);
+
+    if (response.status === 200 && response.data.success === true) {
+      return response.data;
+    } else {
+      throw new WalletRequestError('Error getting token details.');
```
r4mmer, line 104: On the wallet-service PR we may change to a 400/404 response if the token is not found.
Do you think we should identify here and return "empty" (zero values) if we don't have the token, instead of raising an error?

src/wallet/wallet.ts
andreabadesso: Fixed on aaa03b768fd1c2dfa544fe73f32e58e2512bc3fe

src/new/wallet.js
andreabadesso: Fixed on aaa03b768fd1c2dfa544fe73f32e58e2512bc3fe

src/wallet/wallet.ts
andreabadesso: Refactored on aaa03b768fd1c2dfa544fe73f32e58e2512bc3fe

src/wallet/wallet.ts
andreabadesso: We don't need this anymore as the API has been refactored to return mint and melt as booleans

Refactored in aaa03b768fd1c2dfa544fe73f32e58e2512bc3fe

src/wallet/api/walletApi.ts
andreabadesso: That's a valid point, but we are throwing on every other request when the request is not successful, I think we should keep it this way

r4mmer (approved):

PR #354, opened 2022-04-28
feat: adds address precalculation parameter

tuliomir:
This PR adds a parameter on the wallet initialization `constructor` and `loadAddressHistory` methods, to allow the caller to pre-calculate a wallet's addresses, saving on processing time for the lib.

This is especially useful for testing environments, and is a required feature of HathorNetwork/hathor-wallet-headless#189

### Acceptance Criteria
- The wallet init process must be unchanged when the `preCalculatedAddresses` parameter is not informed
- The wallet init must accept the pre-calculated addresses on the new parameter when informed

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -59,13 +59,19 @@ const ConnectionState = {
  *                          one for each request sent to the server.
  **/
 class HathorWallet extends EventEmitter {
-  /*
-   * connection {ConnectionState} A connection to the server
-   * seed {String} 24 words separated by space
-   * passphrase {String} Wallet passphrase
-   * tokenUid {String} UID of the token to handle on this wallet
-   * password {String} Password to encrypt the seed
-   * pin {String} PIN to execute wallet actions
+  /**
+   * @param param
+   * @param {ConnectionState} param.connection A connection to the server
+   * @param {string} param.seed 24 words separated by space
+   * @param {string} [param.passphrase=''] Wallet passphrase
+   * @param {string} [param.xpriv]
+   * @param {string} [param.xpub]
+   * @param {string} [param.tokenUid] UID of the token to handle on this wallet
+   * @param {string} [param.password] Password to encrypt the seed
+   * @param {string} [param.pinCode] PIN to execute wallet actions
+   * @param {boolean} [param.debug] Activates debug mode
+   * @param {{pubkeys:string[],minSignatures:number}} [param.multisig]
+   * @param {string[]} [param.preCalculatedAddresses] An array of pre-calculated addresses
    */
   constructor({
     connection,
@@ -139,9 +145,7 @@ class HathorWallet extends EventEmitter {
     this.pinCode = pinCode;
     this.password = password;
 
-    if (preCalculatedAddresses) {
-      this.preCalculatedAddresses = preCalculatedAddresses;
-    }
+    this.preCalculatedAddresses = preCalculatedAddresses;
 
     this.store = null;
     if (store) {
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -89,6 +89,7 @@ class HathorWallet extends EventEmitter {
     // Callback to be executed before reload data
     beforeReloadCallback = null,
     multisig = null,
+    preCalculatedAddresses = null,
   } = {}) {
     super();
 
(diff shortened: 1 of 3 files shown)


src/new/wallet.js
```diff
@@ -89,6 +89,7 @@ class HathorWallet extends EventEmitter {
     // Callback to be executed before reload data
     beforeReloadCallback = null,
     multisig = null,
+    preCalculatedAddresses = null,
```
r4mmer, line 92: We need to add this to the docstring

src/new/wallet.js
```diff
@@ -138,6 +139,10 @@ class HathorWallet extends EventEmitter {
     this.pinCode = pinCode;
     this.password = password;
 
+    if (preCalculatedAddresses) {
```
r4mmer, line 142: I don't think we need this `if` since the default value is `null`

src/wallet.js
```diff
@@ -508,7 +508,7 @@ const wallet = {
    * @memberof Wallet
    * @inner
    */
-  loadAddressHistory(startIndex, count, connection = null, store = null) {
+  loadAddressHistory(startIndex, count, connection = null, store = null, preCalculatedAddresses = null) {
```
r4mmer, line 511: We need to add this to the docstring

src/wallet.js
```diff
@@ -523,8 +523,10 @@ const wallet = {
 
       const stopIndex = startIndex + count;
       for (var i=startIndex; i<stopIndex; i++) {
-        // Generate each key from index, encrypt and save
-        const address = this.generateAddress(i);
+        // Generate each key from index, encrypt and save ( if it is not provided pre-calculated )
+        const address = preCalculatedAddresses && preCalculatedAddresses[i]
```
r4mmer, line 527: I think we should use `lodash.get` here, what do you think?

r4mmer (changes requested):

src/wallet.js
tuliomir: I think this is more a matter of code style: `lodash.get` is way less performant, but this is irrelevant in this case. I'm open to the idea.

What do you think of:
```suggestion
        const address = _.get(preCalculatedAddresses[i], this.generateAddress(i).toString());
```

src/wallet.js
tuliomir: Fixed on [5c74a59](https://github.com/HathorNetwork/hathor-wallet-lib/pull/354/commits/5c74a595c99e414d85ee32b924b2159c0e12f5b7)

src/new/wallet.js
tuliomir: Fixed on [5c74a59](https://github.com/HathorNetwork/hathor-wallet-lib/pull/354/commits/5c74a595c99e414d85ee32b924b2159c0e12f5b7).

Also, refactored the whole constructor docstring to the _jsdocs_ format. It was a simple comment before.

src/new/wallet.js
tuliomir: The idea was to avoid clutter on the instance object, but I agree that's a good idea. Done on [5c74a59](https://github.com/HathorNetwork/hathor-wallet-lib/pull/354/commits/5c74a595c99e414d85ee32b924b2159c0e12f5b7).

src/wallet.js
```diff
@@ -502,13 +502,14 @@ const wallet = {
...
    * @param {Store} store Store object to save the data
+   * @param {string[]} preCalculatedAddresses  An array of pre-calculated addresses for this wallet
    *
    * @return {Promise} Promise that resolves when addresses history is finished loading from server
    *
    * @memberof Wallet
    * @inner
    */
-  loadAddressHistory(startIndex, count, connection = null, store = null) {
+  loadAddressHistory(startIndex, count, connection = null, store = null, preCalculatedAddresses = null) {
```
pedroferreira1, line 512: We have tests for the method `executeGenerateWallet`, which calls `loadAddressHistory`. However, there is no way to execute `loadAddressHistory` with `preCalculatedAddresses !== null` with `loadAddressHistory`.

That's why I think you could create a test for this method using `preCalculatedAddresses` parameter.

tuliomir: Good idea!
Created a new test for the `loadAddressHistory` method on [e52e1f0](https://github.com/HathorNetwork/hathor-wallet-lib/pull/354/commits/e52e1f0d06b05981666c2babf9cabc8178141892).

pedroferreira1 (approved):

PR #355, opened 2022-05-02
feat: version data api and changeServer and changeNetwork methods

andreabadesso:
### Acceptance Criteria
- We should have methods on both facades to download version data
- We should be able to change network, base server URL and WebSocket server URL using facade methods
- This PR should fix a bug on the websocket connection that would make the class retry the connection forever even if the instance was destroyed

### Security Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -28,7 +28,7 @@ class Config {
   /**
    * Sets the tx mining service url that will be returned by the config object.
    *
-   * @param {string} url - The url to be set
+   * @param url - The url to be set
    */
   setTxMiningUrl(url: string) {
     this.TX_MINING_URL = url;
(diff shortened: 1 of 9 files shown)


src/new/wallet.js
```diff
@@ -208,6 +208,35 @@ class HathorWallet extends EventEmitter {
...
+      rewardSpendMinBlocks: versionData.reward_spend_min_blocks,
+      maxNumberInputs: versionData.max_number_inputs,
+      maxNumberOutputs: versionData.max_number_outputs,
+    };
+  }
+
+  /**
+   * On this facade, we should call wallet.changeServer and also update the config singleton
+   **/
+  changeServer(newServer) {
```
tuliomir, line 235: Missing parameter type. It's `string`, right?
```suggestion
  changeServer(newServer: string) {
```
Also, I see the functions on `src/wallet/wallet.ts` have the full jsdocs and `@params` info. Maybe it would be good to add them here as well.

src/wallet/wallet.ts
```diff
@@ -163,6 +164,48 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+
+  /**
+   * Gets the stored websocket and base server urls
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  static getServerUrlsFromStorage(): WalletServiceServerUrls {
+    const walletServiceBaseUrl = storage.getItem('wallet:wallet_service:base_server');
+    const walletServiceWsUrl = storage.getItem('wallet:wallet_service:ws_server');
```
tuliomir, line 201: These informations have a similar storage as the 'wallet:network': they are both on the config and on the storage.

Won't we have here also the priority algorithm that we have on `src/config.ts`:`getNetwork()`?

tuliomir (changes requested):

src/config.ts
```diff
@@ -196,6 +198,36 @@ class Config {
...
+
+    /**
+    * Gets the current network
+    *
+    * There is more than one method of setting this.
+    * The priority will be given to the network set using the config object.
+    * If not set, we look next into the storage object keys.
+    * If still not set, the default url is returned
+    */
+    getNetwork() {
```
pedroferreira1, line 216: Missing return type

src/wallet/wallet.ts
```diff
@@ -307,6 +350,13 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     this.clearSensitiveData();
   }
 
+  /**
+   * Returns version data from the connected fullnode
+   **/
+  async getVersionData() {
```
pedroferreira1, line 356: Missing return type

src/new/wallet.js
andreabadesso: I agree, done in 28d1a07c206f6c86b26febbd9d2a4c2cb6c4b42d

src/wallet/api/walletApi.ts
```diff
@@ -42,6 +43,17 @@ const walletApi = {
...
   },
 
+  async getVersionData(wallet: HathorWalletServiceWallet): Promise<FullNodeVersionData> {
+    const axios = await axiosInstance(wallet, false);
+    const response = await axios.get('version');
+    const data = response.data;
+    if (response.status === 200 && data.success) {
+      return data.data;
+    } else {
+      throw new WalletRequestError('Error getting fullnode data.');
```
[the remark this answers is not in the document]
andreabadesso: I don't like the idea of having logs on the lib, but I do agree that we are losing information by just throwing an error message.

I created an issue so we can discuss it there:
https://github.com/HathorNetwork/hathor-wallet-lib/issues/361

src/wallet/wallet.ts
andreabadesso: That's a valid point, but since the method is called getServerUrls**FromStorage**, I think we should always get it from `storage`, even if the config singleton has a different value on it — maybe the user is rolling back to the stored value after a `config.set`

What do you think?

src/websocket/base.ts
```diff
@@ -185,6 +188,24 @@ abstract class BaseWebSocket extends EventEmitter {
     }, this.heartbeatInterval);
   }
 
+  /**
+   * Removes all listeners, ends the connection and removes the setup reconnection timer
+   */
+  close() {
+    this.removeAllListeners();
+    this.endConnection()
```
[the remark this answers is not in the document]
andreabadesso: Thanks, fixed in 463aa23165fd2a513541b31a44baa6a9c829e2ee

src/new/wallet.js
```diff
@@ -208,6 +208,35 @@ class HathorWallet extends EventEmitter {
     return new Network(this.getNetwork());
   }
 
+  async getVersionData() {
```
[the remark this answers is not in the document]
andreabadesso: Thanks! Added on d53965ff68b641c08e6452b2c06adb5f88b6bd0d

src/config.ts
andreabadesso: Fixed in 58e3f03b225b28f2c427fcd5d67b8a4606b4d10b

src/config.ts
```diff
@@ -196,6 +198,36 @@ class Config {
         return DEFAULT_SERVER;
     }
 
+    /**
+     * Sets the current network the wallet is connected to
+     */
+    setNetwork(network: string) {
+      this.NETWORK = network;
```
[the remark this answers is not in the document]
andreabadesso: As we've discussed on slack, I went for the second suggestion. 

Added a `setNetwork` call to the singleton on https://github.com/HathorNetwork/hathor-wallet-lib/pull/355/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R211

Also, extended the Network model on the network singleton to be able to set the config before setting the network on the instance:
https://github.com/HathorNetwork/hathor-wallet-lib/pull/355/files#diff-4f2582c1ea1c15f9c0f0916cefe966d446d7c5154ccdc049b9c0923e0101d3a3R14

What do you think?

src/new/wallet.js
```diff
@@ -208,6 +208,35 @@ class HathorWallet extends EventEmitter {
...
+      minTxWeightK: versionData.min_tx_weight_k,
+      tokenDepositPercentage: versionData.token_deposit_percentage,
+      rewardSpendMinBlocks: versionData.reward_spend_min_blocks,
+      maxNumberInputs: versionData.max_number_inputs,
+      maxNumberOutputs: versionData.max_number_outputs,
+    };
+  }
+
+  /**
+   * On this facade, we should call wallet.changeServer and also update the config singleton
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 58e3f03b225b28f2c427fcd5d67b8a4606b4d10b

src/wallet/wallet.ts
tuliomir: With your explanation it seems ok for me.
Maybe in the future a diagram could explain visually when to use the storage and when to use the singleton for a newcoming developer.

tuliomir (approved):

pedroferreira1 (approved):

src/config.ts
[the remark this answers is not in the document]
pedroferreira1: Great

PR #360, opened 2022-05-06
feat: add support in the sendtransaction method to support data script outputs

pedroferreira1:
### Acceptance Criteria

- We must add support for sending a transaction with one or many data script outputs.
- All other transaction types must continue working as before.
- This feature must be added to both facades.

### Security Checklist

### DIFF
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -476,9 +476,30 @@ const helpers = {
    * @inner
    */
   createNFTOutput(data: string): Output {
+    return this.createDataScriptOutput(data);
+  },
+
+  /**
+   * Create an output with data script
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  createDataScriptOutput(data: string): Output {
     const scriptData = new ScriptData(data);
     // Value 1 and token HTR
     return new Output(1, scriptData.createScript());
+  },
+
+  /**
+   * From the base58 of an address we get the type of it, i.e. 'p2pkh' or 'p2sh'
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  getOutputTypeFromAddress(address: string, network: Network): string {
+    const addressObj = new Address(address, { network });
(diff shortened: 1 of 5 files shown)


andreabadesso (approved):

src/wallet/types.ts
```diff
@@ -173,6 +173,21 @@ export interface OutputRequestObj {
   timelock?: number | null; // output timelock
 }
 
+export interface DataScriptOutputRequestObj {
+  type: string; // output type (in this case will be 'data')
```
luislhl, line 177: Shoud we use a string-literal as type here, since we know this must always `data`?

```ts
export interface DataScriptOutputRequestObj {
  type: 'data';
  data: string;
}
```

src/utils/helpers.ts
```diff
@@ -476,10 +476,31 @@ const helpers = {
...
+   * @inner
+   */
+  createDataScriptOutput(data: string): Output {
     const scriptData = new ScriptData(data);
     // Value 1 and token HTR
     return new Output(1, scriptData.createScript());
   },
+
+  /**
+   * From the base58 of an address we get the type of it, i.e. 'p2pkh' or 'p2sh'
```
luislhl, line 495: Could this method return the type `data` as well, or only `p2pkh` and `p2sh`?

luislhl (commented):
Should we create automated tests for the new behaviors?

src/utils/helpers.ts
pedroferreira1: Actually not. The `data` output does not have an address.

src/wallet/types.ts
pedroferreira1: I like it. Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/360/commits/458117681f174a09d235ec3019f84265332b14e3

luislhl (approved):

PR #365, opened 2022-05-18
feat: create configuration attribute for user agent to set in the requests

pedroferreira1:
### Acceptance Criteria
- We should be able to set a user agent for all requests in the lib

### Security Checklist

### DIFF
--- a/src/api/axiosWrapper.ts
+++ b/src/api/axiosWrapper.ts
@@ -7,6 +7,7 @@
 
 import axios from 'axios';
 import { TIMEOUT } from '../constants';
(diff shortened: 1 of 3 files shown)


andreabadesso (approved):

luislhl (approved):
Should we start document some of these configuration options in the README?

pedroferreira1:
> Should we start document some of these configuration options in the README?

I agree with you on that. Actually the README needs a full change in my opinion. It is one of the tasks of the tech writer to do it but I guess we have many tasks now and it will take some weeks still

PR #368, opened 2022-05-23
fix: return null for no authority utxos

r4mmer:
### Summary

If no authority utxos are found nothing was returned, this meant the caller would get an `undefined`.
The problem is some methods use `!== null` to check if there are any authority utxos and `undefined` makes it true.

### Acceptance Criteria
- Actually return `null` if no authority utxos are found

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1426,7 +1426,7 @@ class HathorWallet extends EventEmitter {
    *    'skipSpent': if should not include spent utxos (default true)
    *  }
    *
-   * @return {Array} Array of objects with {tx_id, index, address, authorities} of the authority output. Returns null in case there are no utxos for this type
+   * @return {Array|null} Array of objects with {tx_id, index, address, authorities} of the authority output. Returns null in case there are no utxos for this type
    **/
   selectAuthorityUtxo(tokenUid, filterUTXOs, options = {}) {
     const newOptions = Object.assign({many: false, skipSpent: true}, options);
@@ -1472,8 +1472,9 @@ class HathorWallet extends EventEmitter {
     }
 
     if (many) {
-      return utxos;
+      return utxos || null;
     }
+    return null;
   }
 
   /**


src/new/wallet.js
```diff
@@ -1426,7 +1426,7 @@ class HathorWallet extends EventEmitter {
    *    'skipSpent': if should not include spent utxos (default true)
    *  }
    *
-   * @return {Array} Array of objects with {tx_id, index, address, authorities} of the authority output. Returns null in case there are no utxos for this type
+   * @return {Array|null} Array of objects with {tx_id, index, address, authorities} of the authority output. Returns null in case there are no utxos for this type
    **/
   selectAuthorityUtxo(tokenUid, filterUTXOs, options = {}) {
```
luislhl, line 1431: There's as place calling this function this way:

```js
    const destroyInputs = this.selectAuthorityUtxo(tokenUid, filterUtxos, { many: true });

    if (destroyInputs.length < count) {
      return Promise.reject({ success: false, message: ErrorMessages.NO_UTXOS_AVAILABLE, errorData: { requestedQuantity: count, availableQuantity: destroyInputs.length } });
    }
```

Shouldn't we treat there the case when it returns `null`?

luislhl (approved):

src/new/wallet.js
luislhl: This is a pre-existing problem, so if this is really a problem, and you think it should be made in another PR, I'm ok.

r4mmer: The `many: true` option will always return an array, an empty array in case none are found.

src/new/wallet.js
```diff
@@ -1474,6 +1474,7 @@ class HathorWallet extends EventEmitter {
     if (many) {
       return utxos;
     }
+    return null;
```
msbrogli, line 1477: Why not an empty array?

msbrogli (changes requested):

src/new/wallet.js
r4mmer: There are some places that expect a `null` response if no authority utxo is found, if they require a list (empty if none are found) they would indicate this with the option `many: true` which makes the response always be an array.

This is an example (and the reason i found this bug) of a place that expects `null`
https://github.com/HathorNetwork/hathor-wallet-lib/blob/3ed53d0dbff27d87da6e267bf16187c5594a32ba/src/new/wallet.js#L609-L610

src/new/wallet.js
```diff
@@ -1426,7 +1426,7 @@ class HathorWallet extends EventEmitter {
    *    'skipSpent': if should not include spent utxos (default true)
    *  }
    *
-   * @return {Array} Array of objects with {tx_id, index, address, authorities} of the authority output. Returns null in case there are no utxos for this type
+   * @return {Array|null} Array of objects with {tx_id, index, address, authorities} of the authority output. Returns null in case there are no utxos for this type
```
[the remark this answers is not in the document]
r4mmer: There are no tests for this method, I could add them if necessary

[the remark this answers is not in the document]
msbrogli: We have another project that will add tests all over the place. I feel we can move forward without tests for now since it's a bugfix.

PR #369, opened 2022-05-30
feat: sort p2sh signatures for multisig

r4mmer:
### Acceptance Criteria

- The signatures to use when creating the input data for a p2sh utxo should be ordered by the account path pubkey
- When creating a p2sh input data we should ignore any signatures past the minimum since they will make the verification fail

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -392,7 +392,7 @@ class HathorWallet extends EventEmitter {
     const xpub = HDPublicKey(accessData.xpubkey);
 
     // deserialize P2SHSignature for all signatures
-    const p2shSignatures = signatures.map(sig => P2SHSignature.deserialize(sig));
+    const p2shSignatures = signatures.sort().map(sig => P2SHSignature.deserialize(sig));
 
     for (const {index, input} of tx.inputs.map((input, index) => ({index, input}))) {
       if (!(input.hash in historyTransactions)) {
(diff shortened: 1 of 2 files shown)


tuliomir (approved):

tuliomir:
From what I read, the code from this PR is ok.
But I think another PR could be opened (or even this one could be changed) to include tests for this new conditional.

Different signatures could be passed in changing orders to ensure the internal sorting and slicing are working as intended.

src/new/wallet.js
```diff
@@ -392,7 +392,7 @@ class HathorWallet extends EventEmitter {
     const xpub = HDPublicKey(accessData.xpubkey);
 
     // deserialize P2SHSignature for all signatures
-    const p2shSignatures = signatures.map(sig => P2SHSignature.deserialize(sig));
+    const p2shSignatures = signatures.sort().map(sig => P2SHSignature.deserialize(sig));
```
pedroferreira1, line 395: I feel we should have a comment here explaining why this work. The signatures must be sorted in the same order as the xpub (as you said in the PR description). The `signatures` array have a P2SHSignature object, which has the xpubkey, then sorting this array will be the same as sorting the signatures in the xpub order

pedroferreira1 (approved):

src/new/wallet.js
r4mmer: Done in [46a30f1](https://github.com/HathorNetwork/hathor-wallet-lib/pull/369/commits/46a30f1da8c379b94475097b8fb075d250476ea2)

PR #370, opened 2022-05-31
feat: HathorWallet emitting its state

tuliomir:
### Summary
The HathorWallet facade currently does not have an event emitter for its state changes. Only pooling is available through the `isReady()` method. This PR adds an event emitter for these state changes.

Some JSDocs were also improved on this PR.

### Motivation
On testing environments pooling is not desirable, since we have potentially dozens of wallets being instantiated in parallel and each test is waiting for the moment it can start interacting with its instance.

Having an event for the wallet ready state allows each test to listen to it and start its assertions as early as possible.

### Acceptance Criteria
- Instances of the HathorWallet class should emit their states on wallet startup

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -281,6 +281,7 @@ class HathorWallet extends EventEmitter {
     if (newState === ConnectionState.CONNECTED) {
       storage.setStore(this.store);
       this.setState(HathorWallet.SYNCING);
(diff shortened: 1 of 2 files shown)


tuliomir:
This event was already implemented on the [setState method](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/new/wallet.js#L1152). Closing this PR.

PR #390, opened 2022-08-01
fix: typo using utxo type

r4mmer:
### Acceptance Criteria
- token field is called `tokenId` on `Utxo` type
- allow deserializing `PartialTxInputData` with no signatures

### Security Checklist

### DIFF
--- a/src/models/partial_tx.ts
+++ b/src/models/partial_tx.ts
@@ -597,8 +597,9 @@ export class PartialTxInputData {
   addSignatures(serialized: string) {
     const arr = serialized.split('|');
     if (arr.length < 2 || arr[0] != PartialTxInputDataPrefix || arr[1] !== this.hash) {
-      // Only the first 2 parts are required
-      // The third onward are the signatures, which can be an empty array
+      // Only the first 2 parts are required, the third onward are the signatures which can be empty
+      // When collecting the input data from atomic-swap participants a participant may not have inputs to sign
+      // allowing the empty input data array case will make this a noop instead of throwing an error.
       throw new SyntaxError('Invalid PartialTxInputData');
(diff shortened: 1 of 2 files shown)


src/models/partial_tx.ts
```diff
@@ -596,7 +596,7 @@ export class PartialTxInputData {
    */
   addSignatures(serialized: string) {
     const arr = serialized.split('|');
-    if (arr.length < 3 || arr[0] != 'PartialTxInputData' || arr[1] !== this.hash) {
+    if (arr.length < 2 || arr[0] != 'PartialTxInputData' || arr[1] !== this.hash) {
```
pedroferreira1, line 599: Don't we have the `PartialTxInputData` string, the hash and the signatures? So we need the array at least with length 3, right?

tuliomir, line 599: Two comments about this line:

Why did you change the amount of elements in this `serialized` array? Is this related to a specific change?

And second, I think it would be better to use the string constant already there:
```suggestion
    if (arr.length < 2 || arr[0] != PartialTxInputDataPrefix || arr[1] !== this.hash) {
```

tuliomir (changes requested):

src/models/partial_tx.ts
r4mmer: There is the case of a participant with no inputs requesting his signatures.
The input data part would be empty so we can either not return the signatures to the user (stating that the input data is empty) or allow the "no signatures" case and it seemed best to just have a noop case when the user "signature" does not contain any input data.

pedroferreira1 (approved):

tuliomir (approved):

PR #394, opened 2022-08-05
fix: add type to change output from old method

r4mmer:
### Security Checklist


src/utils/address.ts
```diff
@@ -0,0 +1,15 @@
+import Address from '../models/address';
```
pedroferreira1, line 1: Missing file header

pedroferreira1 (approved):

andreabadesso (approved):

Issue #397, opened 2022-08-10
fix: HathorWallet.getBalance() returning incorrect number of transactions


tuliomir:
Steps to reproduce:
- Start a new, empty wallet
- Make a single transaction sending HTR funds to this wallet
- Call the `HathorWallet.getBalance()` method with the HTR token

The expected value for the `transactions` property would be 1, but it returns 2.
This can be verified on the integration tests, on the [`getBalance` > `should get the balance for the HTR token` test](https://github.com/HathorNetwork/hathor-wallet-lib/blob/af9f81c7a00e407396d0f5c4664fb77651c50214/__tests__/integration/hathorwallet_facade.test.js#L290-L298).

tuliomir:
In some rare iterations the test actually returns `transactions: 1`, such as with [this test run](https://github.com/HathorNetwork/hathor-wallet-lib/runs/8197438873?check_suite_focus=true).

PR #409, opened 2022-08-24
test: integration xpub and edge cases

tuliomir:
### Acceptance Criteria
- Should make the tests as close to 100% coverage as possible without getting too complex solutions
- The tests coverage threshold is being set on `jest-integration.config.js`

### Notes
**Test Delays**
The arbitrary delay after waiting for received transactions ( `wallet.helper.js` file, line ~260 ) was increased, since the tests were failing consistently in random assertions. When this happens is usually a sign that the system is being overloaded with the amount of tests.

**Integration test `npm` script**
A change was made to the `package.json` file and the usage of `test_network_partial_down` on the main test script was removed. This was done because the successful completion of the `partial_down` script was returning success even when the test itself had failed, giving us false results on the CI.

With this change, the developer will have a bit more work to manually stop the `cpuminer` when making tests locally, until a new solution is implemented.

### Security Checklist


__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -15,18 +16,114 @@ import {
...
 import helpersUtils from '../../src/utils/helpers';
+import walletUtils from '../../src/utils/wallet';
+import { ConnectionState } from '../../src/wallet/types';
+import MemoryStore from '../../src/memory_store'
 
 const fakeTokenUid = '008a19f84f2ae284f19bf3d03386c878ddd15b8b0b604a3a3539aa9d714686e1';
 const sampleNftData = 'ipfs://bafybeiccfclkdtucu6y4yc5cpr6y3yuinr67svmii46v5cfcrkp47ihehy/albums/QXBvbGxvIDEwIE1hZ2F6aW5lIDI3L04=/21716695748_7390815218_o.jpg';
 
 describe('start', () => {
+  it('should reject with invalid inputs', async () => {
```
pedroferreira1, line 34: with invalid parameters?

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -151,6 +248,172 @@ describe('start', () => {
...
+    await expect(hWallet.consolidateUtxos()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.sendTransaction()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.sendManyOutputsTransaction()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.prepareCreateNewToken()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.prepareMintTokensData()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.prepareMeltTokensData()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.prepareDelegateAuthorityData()).rejects.toThrow(WalletFromXPubGuard);
+    await expect(hWallet.prepareDestroyAuthorityData()).rejects.toThrow(WalletFromXPubGuard);
+
+    // Validating that the address generation works as intended
```
pedroferreira1, line 348: I would validate that you can get the wallet balance and utxos correctly

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -151,6 +248,172 @@ describe('start', () => {
...
+  it('should start a wallet via xpub', async () => {
+    const walletData = precalculationHelpers.test.getPrecalculatedWallet();
+    const xpriv = walletUtils.getXPrivKeyFromSeed(
+      walletData.words,
+      { networkName: NETWORK_NAME })
+    const privkey = walletUtils.deriveXpriv(xpriv, "0'/0");
+    const xpub = privkey.xpubkey;
+
+    // Creating a new wallet with a known set of words just to generate the custom token
+    let hWallet = await generateWalletHelper({ xpub });
```
pedroferreira1, line 335: Seeing the code below

```
// Start the wallet
  const walletConfig = {
    seed: walletData.words,
    connection: generateConnection(),
    password: DEFAULT_PASSWORD,
    pinCode: DEFAULT_PIN_CODE,
    preCalculatedAddresses: walletData.addresses,
  };
  if (param) {
    Object.assign(walletConfig, param);
  }
```

It looks like that this wallet will be started with xpub but also with pin/password, what is not correct in the case of an xpub wallet because we use the pin only for actions to access the private key. We must change this to start the wallet only with the xpub and everything should work fine.

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -1135,6 +1461,13 @@ describe('mintTokens', () => {
       100,
     );
 
+    // Should not mint more tokens than the HTR funds allow
```
pedroferreira1, line 1464: Nice one, we should do the same for melt. Try to melt more tokens than you have.

package.json
```diff
@@ -23,7 +23,7 @@
     "test": "jest --env=node --forceExit",
     "build": "babel --presets @babel/preset-react -d lib/ src && tsc",
     "tsc": "tsc",
-    "test_integration": "npm run test_network_up && npm run test_network_integration && npm run test_network_down || npm run test_network_partial_down",
+    "test_integration": "npm run test_network_up && npm run test_network_integration && npm run test_network_down",
```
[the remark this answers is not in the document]
tuliomir: This was a command that was run when there were errors in the test, to deactivate only the `cpuminer` instead of destroying the whole `privatenet`.

Trouble is: it always returned a success for the `cpuminer` shutdown. This was interpreted by GitHub's CI as if the tests succeeded.
So I had to remove it. More info on the PR's description.

src/new/wallet.js
```diff
@@ -20,7 +20,7 @@ import MemoryStore from '../memory_store';
 import config from '../config';
 import SendTransaction from './sendTransaction';
 import Network from '../models/network';
-import { AddressError, WalletError } from '../errors';
+import { AddressError, WalletError, WalletFromXPubGuard } from '../errors';
```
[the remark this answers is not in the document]
tuliomir: Actually this wasn't working on NodeJS 14 and 16: the application threw an error on XPub wallets.
I decided this was too little a change to become its own PR and added it here.

Do you agree with this scope increase?

__tests__/integration/helpers/wallet.helper.js
```diff
@@ -50,10 +50,32 @@ const startedWallets = [];
...
  * @returns {Promise<HathorWallet>}
  */
-export async function generateWalletHelper() {
-  const walletData = precalculationHelpers.test.getPrecalculatedWallet();
+export async function generateWalletHelper(param) {
+  /** @type PrecalculatedWalletData */
+  let walletData = {};
+
+  // Only fetch a precalculated wallet if the input does not offer a specific one
+  if (!param) {
```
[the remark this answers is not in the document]
tuliomir: That way if I didn't want to pass a `preCalculatedAddresses` parameters I would have to explicitly set it to `[]` on the parameters, otherwise the helper would consume another wallet and yield non-obvious results. A test using a `xpub` and `preCalculatedAddresses` without a seed would not at all be possible.

The rational for the current implementation is that:
- If the helper method is called without parameters, it will generate a valid wallet
- If the helper is called with a parameter, the user will have to define the exact configuration for the wallet
- The rule is as easy as possible for a developer to just skim the code and understand it

After all, initializing a wallet without the helper isn't complex. What do you think?

__tests__/integration/hathorwallet_facade.test.js
tuliomir: Good catch. `inputs` could be confusing in this context.

__tests__/integration/hathorwallet_facade.test.js
tuliomir: Added these tests in [cc7fb5d](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/cc7fb5d017b78f0caad96a470122eb0b531b5458).

__tests__/integration/hathorwallet_facade.test.js
tuliomir: Added this test on [0916488](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/0916488c411ea9ebe775714fd2fd57428a925e4a).

src/new/wallet.js
```diff
@@ -71,6 +71,7 @@ class HathorWallet extends EventEmitter {
    * @param {string} [param.password] Password to encrypt the seed
    * @param {string} [param.pinCode] PIN to execute wallet actions
    * @param {boolean} [param.debug] Activates debug mode
+   * @param {MemoryStore} [param.store] Optional store to initialize the wallet with
```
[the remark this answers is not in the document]
tuliomir: Changed to `Storage` on [b169559](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/b16955950f7d51cad35b65881a9e1329824e66ed).

__tests__/integration/hathorwallet_facade.test.js
tuliomir: You're right, this was being instantiated with pin and password.
Removed both these parameters in [b1b919e](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/b1b919ea17d12af980db07820fb083bcf31118c7).

__tests__/integration/helpers/wallet.helper.js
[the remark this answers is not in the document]
pedroferreira1: Now I understand, I got confused by the method docstring

"Generates a Wallet from an available precalculated seed"

I thought you wanted only pre calculated wallets. I just think you should update the docstring.

pedroferreira1 (approved):

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -238,6 +529,63 @@ describe('addresses methods', () => {
...
+    expect(await getMaximumIndexFromAddressGenerator()).toStrictEqual(4)
+
+    // Restore the original gap limit
+    wallet.setGapLimit(originalGapLimit);
+
+    /**
+     * Iterates the getAllAddresses generator and finds the highest address index for the wallet
+     * @return {Promise<number>}
+     */
+    async function getMaximumIndexFromAddressGenerator() {
```
r4mmer, line 567: All addresses are saved on `storage.getItem('wallet:data').keys`, we could check there more easily for the max index.

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -238,6 +529,63 @@ describe('addresses methods', () => {
...
+        const genResults = await addressGenerator.next();
+
+        // If there is an address,
+        if (!genResults.done) {
+          maximumIndex = genResults.value.index;
+        } else {
+          generatorIsActive = false;
+        }
+      }
+      return maximumIndex;
```
r4mmer, line 585: An [example](https://javascript.info/async-iterators-generators) on iterating async generators.

```suggestion
      const addressGenerator = hWallet.getAllAddresses();

      // Results variable
      let maximumIndex = 0;
      for await (const address of addressGenerator) {
        if (address.index > maximumIndex) {
          maximumIndex = address.index;
        }
      }
      return maximumIndex;
```

__tests__/integration/helpers/wallet.helper.js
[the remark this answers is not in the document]
tuliomir: Updated on [2305663](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/23056630d7b67c9881c1194ce588d6c21cf3d9dc).

r4mmer (approved):

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -151,6 +248,200 @@ describe('start', () => {
...
+    // Generating a transaction and validating it shows correctly
+    await GenesisWalletHelper.injectFunds(hWallet.getAddressAtIndex(1), 1);
+
+
+    await expect(hWallet.getBalance(HATHOR_TOKEN_CONFIG.uid)).resolves.toMatchObject([
+      expect.objectContaining({
+        token: expect.objectContaining({ id: HATHOR_TOKEN_CONFIG.uid }),
+        balance: { unlocked: 1, locked: 0 },
+        transactions: 2,
+        // transactions: 1, // TODO: The amount of transactions is 2 but should be 1. Fix this.
```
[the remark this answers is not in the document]
tuliomir: I don't know what's the cause of this behavior. In some rare cases on the CI it even returns the correct value of `1` and makes the test fail.
We have the issue 397 open for this:
- #397

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -151,6 +248,200 @@ describe('start', () => {
...
+    // Starting a new wallet re-using the same words, this time with a specific wallet token
+    hWallet = await generateWalletHelper({
+      seed: walletData.words,
+      preCalculatedAddresses: walletData.addresses,
+      tokenUid,
+    });
+    expect(hWallet.isReady()).toStrictEqual(true); // This operation should work
+
+    // Now testing the methods that use this set tokenUid information
+    // FIXME: No need to explicitly pass the non-boolean `false` as a tokenUid to get this result.
```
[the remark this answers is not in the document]
tuliomir: I think the issue #397 on the `getBalance` method will require enough change on the method's code that this could also be done there.
Do you agree with just increasing its scope instead of creating a new issue?

__tests__/integration/hathorwallet_facade.test.js
tuliomir: Good improvement! Done on [a0e8fdc](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/a0e8fdcf4878d0722f6c2c86d4efe593b8801d1a)

__tests__/integration/hathorwallet_facade.test.js
tuliomir: Added this double-check on [a0e8fdc](https://github.com/HathorNetwork/hathor-wallet-lib/pull/409/commits/a0e8fdcf4878d0722f6c2c86d4efe593b8801d1a)

PR #411, opened 2022-09-12
feat: add sign method to transaction model

r4mmer:
### Acceptance Criteria
- Create a sign method on the transaction model

### Security Checklist

### DIFF
--- a/src/models/transaction.ts
+++ b/src/models/transaction.ts
@@ -16,7 +16,7 @@ import {
   TX_HASH_SIZE_BYTES,
   TX_WEIGHT_CONSTANTS
 } from '../constants'
-import {crypto as cryptoBL, encoding, util} from 'bitcore-lib'
+import {crypto as cryptoBL, encoding, util, PrivateKey} from 'bitcore-lib'
 import {bufferToHex, hexToBuffer, unpackToFloat, unpackToHex, unpackToInt} from '../utils/buffer'
 import helpers from '../utils/helpers'
 import Input from './input'
@@ -90,6 +90,35 @@ class Transaction {
 
     // All inputs sign the same data, so we cache it in the first getDataToSign method call
     this._dataToSignCache = null;
+  }
+
+  /**
+   * Get the signature from the dataToSignHash for a private key
+   *
+   * @return {Buffer}
+   * @memberof Transaction
+   * @static
+   *
+   */
+  static sign(dataToSignHash: Buffer, privateKey: PrivateKey): Buffer {
+    const signature = cryptoBL.ECDSA.sign(dataToSignHash, privateKey, 'little').set({
+      nhashtype: cryptoBL.Signature.SIGHASH_ALL,
+    });
+    return signature.toDER();
+  }
+
+  /**
+   * Get the signature of this transaction for the given private key
+   *
+   * @param {PrivateKey}
+   *
+   * @return {Buffer}
+   * @memberof Transaction
+   * @inner
+   *
+   */
+  sign(privateKey: PrivateKey): Buffer {
+    return Transaction.sign(this.getDataToSignHash(), privateKey);
   }
 
   /**
--- a/src/models/transaction.ts
+++ b/src/models/transaction.ts
@@ -19,7 +19,7 @@ import {
 import {crypto as cryptoBL, encoding, util, PrivateKey} from 'bitcore-lib'
 import {bufferToHex, hexToBuffer, unpackToFloat, unpackToHex, unpackToInt} from '../utils/buffer'
 import helpers from '../utils/helpers'
-import transactionUtils from 'src/utils/transaction'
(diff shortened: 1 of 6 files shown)


tuliomir (commented):
Since you're refactoring the signature flow, maybe it would be a good idea to centralize the signing code too.

Currently we use the `crypto.ECDSA.sign` in four places:
- models/transaction.ts - `sign`
- new/wallet.js - `getAllSignatures` _( uses `.toString()` instead of `.toDER()` )_
- transaction.js - `getSignature`
- wallet/wallet.ts - `getInputData`

It seems to me that we can be confused later with all these different places to sign transaction parts. In this case, the static `Transaction.sign` method could be moved to a _util_ file and then consumed from all these four places.

Do you think this would be a good idea, and a relevant addition to the scope of this PR?

tuliomir (approved):

src/utils/transaction.ts
```diff
@@ -5,11 +5,29 @@
...
-
 import { Utxo } from '../wallet/types';
 import { UtxoError } from '../errors';
+import {crypto as cryptoBL, PrivateKey} from 'bitcore-lib'
 
 const transaction = {
+  /**
+   * Get the signature from the dataToSignHash for a private key
+   *
+   * @param {Buffer} dataToSignHash sha256 of a transaction's dataToSign.
```
pedroferreira1, line 16: This is not only the sha256 of data to sign, we need to do a little bit more. You can say that it's the hash of `dataToSign`

src/models/transaction.ts
```diff
@@ -92,6 +93,32 @@ class Transaction {
...
 
+  /**
+   * Get the signature from the dataToSignHash for a private key
+   *
+   * @return {Buffer}
+   * @memberof Transaction
+   * @static
+   *
+   */
+  static sign(dataToSignHash: Buffer, privateKey: PrivateKey): Buffer {
```
pedroferreira1, line 104: Why do you have this method as well? You could use only the other one, the static method is the same as the utils method, right?

src/models/transaction.ts
```diff
@@ -92,6 +93,32 @@ class Transaction {
...
+   * Get the signature of this transaction for the given private key
+   *
+   * @param {PrivateKey}
+   *
+   * @return {Buffer}
+   * @memberof Transaction
+   * @inner
+   *
+   */
+  sign(privateKey: PrivateKey): Buffer {
```
pedroferreira1, line 118: I would change the method name, this is `getSignature`, you are not signing the transaction.

src/new/wallet.js
```diff
@@ -359,15 +358,11 @@ class HathorWallet extends EventEmitter {
...
       const derivedKey = key.deriveNonCompliantChild(addressIndex);
       const privateKey = derivedKey.privateKey;
 
-      // derive key to address index
-      const sig = crypto.ECDSA.sign(hash, privateKey, 'little').set({
-        nhashtype: crypto.Signature.SIGHASH_ALL
-      });
-
-      signatures[index] = sig.toString();
+      signatures[index] = tx.sign(privateKey).toString('hex');
```
pedroferreira1, line 365: The old code here was doing `sig.toString()` before doing `sig.toDER()`. The new code is doing `sig.toDER().toString('hex')`, is it the same?

src/utils/transaction.ts
```diff
@@ -5,11 +5,29 @@
...
+   * @param {PrivateKey} privateKey Signing key.
+   *
+   * @returns {Buffer}
+   *
+   * @memberof transaction
+   * @inner
+   */
+  sign(dataToSignHash: Buffer, privateKey: PrivateKey): Buffer {
```
pedroferreira1, line 24: I think the same `getSignature` is better for this method, you are not signing anything. What do you think?

src/new/wallet.js
r4mmer: bitcore-lib's [`Signature.toString()`](https://github.com/bitpay/bitcore/blob/master/packages/bitcore-lib/lib/crypto/signature.js#L189-L192) uses `toDER()` then `toString('hex')` internally.
This means we can just use `toString('hex')` here.

src/models/transaction.ts
r4mmer: Yes, it´s exactly the same method. Just kept it here for convenience, do you think we should remove to keep this cleaner?

src/models/transaction.ts
r4mmer: Done

src/utils/transaction.ts
r4mmer: Done.

src/models/transaction.ts
pedroferreira1: I feel we could remove it and keep only the other one.

pedroferreira1 (approved):

PR #416, opened 2022-09-15
chore: improve websocket logs

luislhl:
### Acceptance Criteria
- Log when the websocket connection is established

### Security Checklist

### DIFF
--- a/src/websocket/base.ts
+++ b/src/websocket/base.ts
@@ -186,6 +186,8 @@ abstract class BaseWebSocket extends EventEmitter {
     this.heartbeat = setInterval(() => {
       this.sendPing();
     }, this.heartbeatInterval);
+
+    console.info('Websocket connection opened.');
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/websocket/base.ts
```diff
@@ -186,6 +186,8 @@ abstract class BaseWebSocket extends EventEmitter {
     this.heartbeat = setInterval(() => {
       this.sendPing();
     }, this.heartbeatInterval);
+
+    console.info('Websocket connection opened.');
```
andreabadesso, line 190: I know we are already logging the websocket ping timeout, but I don't like the idea of having logs on the lib, it pollutes the sdout of the clients using it

Maybe we should just emit an event here and log on the client

andreabadesso (changes requested):

src/websocket/base.ts
tuliomir: If we decide to emit an event to log something, I would suggest to create an _util_ `logger` that centralizes all the logging on the lib and implements this event emit with the common interface of `log`, `warn`, `error`, etc.

That way we can have a familiar logging structure within the lib, just as if it was a common application, and have the centralized logger deal with the events. This would also make it easier for future logging configurations.

PR #419, opened 2022-09-23
chore: protect old transaction method from xpub wallet

r4mmer:
### Acceptance Criteria
- Throw `WalletFromXPubGuard` when trying to access the private key of a wallet started with an xpub.

### Security Checklist

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -203,7 +203,7 @@ const wallet = {
 
     if (multisig) {
       const xpub = new HDPublicKey(xpubkey);
-      access['multisig'] = Object.assign({pubkey: xpub.publicKey.toString('hex')}, multisig);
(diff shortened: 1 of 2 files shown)


src/wallet.js
```diff
@@ -1578,6 +1578,7 @@ const wallet = {
     storage.removeItem('wallet:lastGeneratedIndex');
     storage.removeItem('wallet:lastUsedIndex');
     storage.removeItem('wallet:lastUsedAddress');
+    storage.removeItem('wallet:multisig');
```
pedroferreira1, line 1581: Maybe this should be above inside the `if (cleanAccessData) {`, what do you think?

r4mmer: Definetely, I will change it

r4mmer: Done

pedroferreira1 (approved):

andreabadesso (approved):

PR #424, opened 2022-10-18
feat: getSignatures method for HathorWallet

r4mmer:
### Acceptance Criteria
- Create the `getSignatures` method to sign all inputs from the loaded wallet
- Create more tests for the associated signing utils.

### Security Checklist

### DIFF
--- a/__tests__/hathorwallet.test.js
+++ b/__tests__/hathorwallet.test.js
@@ -45,9 +45,10 @@ test('Protected xpub wallet methods', async () => {
   await expect(hWallet.prepareDestroyAuthorityData()).rejects.toThrow(WalletFromXPubGuard);
   expect(hWallet.getAllSignatures).toThrow(WalletFromXPubGuard);
   expect(hWallet.getSignatures).toThrow(WalletFromXPubGuard);
+  expect(hWallet.signTx).toThrow(WalletFromXPubGuard);
 });
 
-test('getSignatures', () => {
+test('signTx', () => {
   // Spy on lib methods used
   const mockStorageGet = jest.spyOn(Storage.prototype, 'getItem').mockReturnValue({
     mainKey: 'mocked-encrypted-privkey',
@@ -69,6 +70,10 @@ test('getSignatures', () => {
     new Output(1, script0.createScript(), { tokenData: 129 }), // Mint authority
     new Output(20, script1.createScript()), // HTR
   ];
+
+  // 1. Common case, 2 inputs from the wallet, check that only these will be signed
+  //    and also confirm the the signature is valid and being set on the transaction.
+
   const tx0 = new Transaction(
     [new Input(inputTx, 0), new Input(inputTx, 1), new Input(inputTx, 2)],
     outputs,
@@ -78,19 +83,18 @@ test('getSignatures', () => {
   // Mock HathorWallet methods used
   const hWallet = new FakeHathorWallet();
   hWallet.isFromXPub.mockReturnValue(false);
-  // hWallet.isFromXPub = () => false;
   hWallet.pinCode = '123';
-  hWallet.getTx = () => ({
+  hWallet.getTx.mockImplementation(() => ({
     outputs: [
       { decoded: { address: mockAddresses[1] } },
       { decoded: { address: mockAddresses[0] } }, // Not from the wallet, will be ignored
       { decoded: { address: mockAddresses[1] } },
     ],
-  });
-  hWallet.isAddressMine = addr => (addr === mockAddresses[1]);
-  hWallet.getAddressIndex = () => 1;
+  }));
+  hWallet.isAddressMine.mockImplementation(addr => (addr === mockAddresses[1]));
+  hWallet.getAddressIndex.mockReturnValue(1);
 
-  let returnedTx = hWallet.getSignatures(tx0);
+  let returnedTx = hWallet.signTx(tx0);
   expect(returnedTx).toBe(tx0);
   // The transaction is filled with the input data
   expect(tx0.inputs[0].data).not.toBe(null);
@@ -125,13 +129,17 @@ test('getSignatures', () => {
   mockInputData.mockClear();
   mockSetData.mockClear();
 
+
+  // 2. Same transaction as 1. but we should use the argument pinCode if given
+  //    Will still check everything to make sure this did not change any behavior.
+
   const tx1 = new Transaction(
     [new Input(inputTx, 0), new Input(inputTx, 1), new Input(inputTx, 2)],
     outputs,
     { version: DEFAULT_TX_VERSION, tokens: [tokenUid] },
   );
   // Calling with a pin code
-  returnedTx = hWallet.getSignatures(tx1, { pinCode: 'another-PIN' });
+  returnedTx = hWallet.signTx(tx1, { pinCode: 'another-PIN' });
   expect(returnedTx).toBe(tx1);
   // The transaction is filled with the input data like before
   expect(tx1.inputs[0].data).not.toBe(null);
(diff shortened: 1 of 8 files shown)


src/new/wallet.js
```diff
@@ -2322,6 +2323,53 @@ class HathorWallet extends EventEmitter {
...
+    const key = HDPrivateKey(privateKeyStr);
+
+    for (const input of tx.inputs) {
+      const inputTx = this.getTx(input.hash);
+      if (!inputTx) {
+        // Input is not from our wallet
+        continue;
+      }
+      const utxo = inputTx.outputs[input.index];
+      if(utxo && this.isAddressMine(utxo.decoded.address)) {
```
pedroferreira1, line 2359: We should have a linter for this later, it's just for consistency

```suggestion
      if (utxo && this.isAddressMine(utxo.decoded.address)) {
```

src/new/wallet.js
```diff
@@ -2322,6 +2323,53 @@ class HathorWallet extends EventEmitter {
...
+   * Sign all inputs
+   *
+   * @param {Transaction} tx The transaction to be signed
+   * @param [options]
+   * @param {string} [options.pincode] PIN to decrypt the private key.
+   *                                   Optional but required if not set in this
+   *
+   * @returns {Transaction} The signed transaction
+   */
+  getSignatures(tx, { pinCode = null } = {}) {
```
pedroferreira1, line 2337: Maybe we should call this method `signTx` because we are not really returning the signatures, we are actually signing the transaction and returning it already signed.

What do you think?

__tests__/hathorwallet.test.js
```diff
@@ -32,5 +43,133 @@ test('Protected xpub wallet methods', async () => {
...
+  const mockAddr0 = new Address(mockAddresses[0]);
+  const script0 = new P2PKH(mockAddr0);
+  const mockAddr1 = new Address(mockAddresses[1]);
+  const script1 = new P2PKH(mockAddr1);
+  const outputs = [
+    new Output(10, script0.createScript(), { tokenData: 1 }), // token funds
+    new Output(1, script0.createScript(), { tokenData: 129 }), // Mint authority
+    new Output(20, script1.createScript()), // HTR
+  ];
+  const tx0 = new Transaction(
```
tuliomir, line 72: I think it would be helpful to have a comment explaining what are the objectives that the test on this tx will have.
In this case, would it be to identify that only the correct inputs are signed?

__tests__/hathorwallet.test.js
```diff
@@ -32,5 +43,133 @@ test('Protected xpub wallet methods', async () => {
...
+  expect(
+    mockSetData.mock.calls[1][0].toString('hex')
+  ).toEqual(mockInputData.mock.results[1].value.toString('hex'));
+
+  // Clear mocks
+  mockPrivkey.mockClear();
+  mockInputData.mockClear();
+  mockSetData.mockClear();
+
+  const tx1 = new Transaction(
```
tuliomir, line 128: This `tx`'s objective is just to ensure that the PIN is accepted on the method call, right?

__tests__/hathorwallet.test.js
```diff
@@ -32,5 +43,133 @@ test('Protected xpub wallet methods', async () => {
...
+  expect(() => {
+   return hWallet.getSignatures(tx0);
+  }).toThrow('Pin is required.');
+
+  // Cleanup
+  mockPrivkey.mockRestore();
+  mockInputData.mockRestore();
+  mockSetData.mockRestore();
+  mockStorageGet.mockRestore();
 });
```
tuliomir, line 175: Github is informing that this file does not end with a newline.
Is our linter configured to also get that? Maybe it would be a good idea to add this rule too.

tuliomir (commented):
I agree with Pedro that the method name could be changed to something closer to `signTx` to better convey its behavior.

Other than that, only comments and a linter suggestion.

src/new/wallet.js
r4mmer: Done

src/new/wallet.js
r4mmer: Changed to `signTx` but created another `getSignatures` that actually return the signatures.

__tests__/hathorwallet.test.js
r4mmer: Added the explanations

__tests__/hathorwallet.test.js
r4mmer: Yes

src/new/wallet.js
```diff
@@ -2322,6 +2304,111 @@ class HathorWallet extends EventEmitter {
...
+    }
+
+    // get private key
+    const accessData = storage.getItem('wallet:accessData');
+    const privateKeyStr = wallet.decryptData(accessData.mainKey, pin);
+    const key = HDPrivateKey(privateKeyStr);
+
+    const signatures = [];
+
+    for (const indices of this.getWalletInputInfo(tx)) {
```
tuliomir, line 2375: typo
```suggestion
    for (const indexes of this.getWalletInputInfo(tx)) {
```

tuliomir (approved):

tuliomir (approved):

src/new/wallet.js
```diff
@@ -2322,6 +2304,111 @@ class HathorWallet extends EventEmitter {
...
+    return signatures;
+  }
+
+  /**
+   * Sign all inputs of the given transaction.
+   *   OBS: only for P2PKH wallets.
+   *
+   * @param {Transaction} tx The transaction to be signed
+   * @param [options]
+   * @param {string} [options.pincode] PIN to decrypt the private key.
```
pedroferreira1, line 2398: `pinCode`*

src/new/wallet.js
```diff
@@ -2322,6 +2304,111 @@ class HathorWallet extends EventEmitter {
...
+    for (const indices of this.getWalletInputInfo(tx)) {
+      const { addressIndex } = indices;
+      // Derive key to addressIndex
+      const derivedKey = key.deriveNonCompliantChild(addressIndex);
+      const privateKey = derivedKey.privateKey;
+      // Get tx signature and populate transaction
+      const sigDER = tx.getSignature(privateKey);
+      signatures.push({
+        signature: sigDER.toString('hex'),
+        pubkey: privateKey.publicKey.toString(),
```
pedroferreira1, line 2384: Is the default of this `toString` hex?

src/new/wallet.js
```diff
@@ -2322,6 +2304,111 @@ class HathorWallet extends EventEmitter {
...
+        const addressPath = this.getAddressPathForIndex(addressIndex);
+        walletInputs.push({ inputIndex, addressIndex, addressPath });
+      }
+    }
+
+    return walletInputs;
+  }
+
+  /**
+   * Get signatures for all inputs of the loaded wallet.
```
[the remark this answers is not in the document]
r4mmer: No, since it returns the ECDSA signature with some metadata (e.g. address index, pubkey) it can be used by both P2PKH and P2SH.
I've also changed the `getAllSignatures` which is P2SH only to use this method.

src/new/wallet.js
r4mmer: Yes, it equivalent to `publicKey.toBuffer().toString('hex')`.

r4mmer: Yes, it equivalent to `publicKey.toBuffer().toString('hex')`.

src/new/wallet.js
r4mmer: It's not a typo since they're both accepted plurals forms of index, but i'll change for consistency with other instances on the lib.

src/new/wallet.js
r4mmer: Fixed

pedroferreira1 (approved):

PR #437, opened 2022-12-08
feat: add pushRegister, pushUpdate, pushUnregister and getTxById methods

alexruzenhack:
### Acceptance Criteria
- implement pushRegister, pushUpdate, and pushUnregister to chat with push notification API
- implement getTxById method to get transaction details

### Security Checklist


__tests__/wallet/wallet.test.ts
```diff
@@ -325,3 +324,111 @@ test('generateCreateWalletAuthData should return correct auth data', async () =>
   expect(xpubMessage.verify(xpubAddress, authData.xpubkeySignature)).toBe(true);
   expect(authXpubMessage.verify(authXpubAddress, authData.authXpubkeySignature)).toBe(true);
 });
+
+test('registerDeviceToPushNotification', async () => {
+  const requestPassword = jest.fn();
+  const network = new Network('testnet');
+  const seed = 'purse orchard camera cloud piece joke hospital mechanic timber horror shoulder rebuild you decrease garlic derive rebuild random naive elbow depart okay parrot cliff';
```
tuliomir, line 331: Since the seed used is the same, how about creating a constant for it and reusing on every test?

tuliomir (approved):

src/wallet/types.ts
```diff
@@ -373,6 +373,56 @@ export interface FullNodeVersionData {
   maxNumberOutputs: number;
 };
 
+export interface PushNotificationResponseData {
+  success: boolean,
+  error?: string,
+  /** This property shows up in case of validation error. */
```
pedroferreira1, line 379: This comment references which code line?

src/wallet/types.ts
```diff
@@ -373,6 +373,56 @@ export interface FullNodeVersionData {
...
+  details?: {message: string, path: string}[]
+};
+
+export interface PushNotificationResult {
+  success: boolean,
+}
+
+export interface PushRegisterRequestData {
+  pushProvider: 'ios' | 'android',
```
pedroferreira1, line 388: Should we create an ENUM for that?

src/wallet/types.ts
```diff
@@ -373,6 +373,56 @@ export interface FullNodeVersionData {
...
+  deviceId: string,
+  enablePush?: boolean,
+  enableShowAmounts?: boolean,
+};
+
+export interface PushRegisterResponseData extends PushNotificationResponseData {};
+
+export interface PushUpdateRequestData {
+  deviceId: string,
+  enablePush: boolean,
```
pedroferreira1, line 398: In the Update it's required, and in the Register it's not?

src/wallet/types.ts
```diff
@@ -373,6 +373,56 @@ export interface FullNodeVersionData {
...
+  enableShowAmounts: boolean,
```
pedroferreira1, line 399: In the Update it's required, and in the Register it's not?

src/wallet/api/walletApi.ts
```diff
@@ -219,6 +227,47 @@ const walletApi = {
...
+    if (response.status === 200 && response.data.success) {
+      return response.data;
+    } else {
+      throw new WalletRequestError('Error updating push notification settings for device.', { cause: response.data });
+    }
+  },
+
+  async pushUnregister(wallet: HathorWalletServiceWallet, payload: PushUnregisterRequestData): Promise<PushUnregisterResponseData> {
+    const axios = await axiosInstance(wallet, true);
+    const response = await axios.post('push/unregister', payload);
```
pedroferreira1, line 254: Should we have here `<PushUnregisterResponseData>`?

src/wallet/api/walletApi.ts
```diff
@@ -22,9 +22,17 @@ import {
...
+  PushRegisterResponseData,
+  PushUpdateRequestData,
+  PushUpdateResponseData,
+  PushUnregisterResponseData,
+  PushUnregisterRequestData,
+  TxByIdTokensResponseData,
 } from '../types';
 import HathorWalletServiceWallet from '../wallet';
 import { WalletRequestError } from '../../errors';
+import Transaction from 'src/models/transaction';
```
pedroferreira1, line 35: Why do you need this?

src/wallet/types.ts
alexruzenhack: Yes. Because in the registration there are default values. In the update, the payload must be complete we should not assume anything. The operation in which the payload can be optional is the `PATCH`.

src/wallet/types.ts
alexruzenhack: Yes. Because in the registration there are default values. In the update, the payload must be complete we should not assume anything. The operation in which the payload can be optional is the PATCH.

src/wallet/api/walletApi.ts
alexruzenhack: No. I've refactor it to send a `DELETE` and the `deviceId` through path param.

src/wallet/api/walletApi.ts
alexruzenhack: We don't.

src/wallet/types.ts
alexruzenhack: The `details?` bellow.

src/wallet/api/walletApi.ts
```diff
@@ -219,6 +223,47 @@ const walletApi = {
       throw new WalletRequestError('Error requesting auth token.');
     }
   },
+
+  async pushRegister(wallet: HathorWalletServiceWallet, payload: PushRegisterRequestData): Promise<PushRegisterResponseData> {
```
tuliomir, line 227: From what we discussed, I understand that these `pushRegister`, `pushUpdate` and `pushUnregister` methods should be moved to a separate utils file - possibly on `src/utils/push.ts`, right?

That way the main `src/wallet` module would not be changed.
What do you think?

src/wallet/wallet.ts
```diff
@@ -59,6 +59,11 @@ import {
   WalletServiceServerUrls,
   FullNodeVersionData,
   WalletAddressMap,
+  PushRegisterRequestData,
+  PushNotificationResult,
+  PushUpdateRequestData,
+  PushUnregisterRequestData,
```
tuliomir, line 65: In this line of thought, these would also not be necessary, right?

src/wallet/api/walletApi.ts
alexruzenhack: Yes, the `wallet` module is not being affected anymore. This is the module `walletApi`. I see the `walletApi` as a client to talk with the `wallet-service`. What do you think?

src/pushNotification.ts
```diff
@@ -0,0 +1,39 @@
+import walletApi from './wallet/api/walletApi';
+import { PushNotificationResult, PushRegisterRequestData, PushUnregisterRequestData, PushUpdateRequestData } from './wallet/types';
+import HathorWalletServiceWallet from './wallet/wallet';
+
+export class PushNotification {
+
+  /**
+   * Register the device to receive push notifications
+   *
+   * @memberof HathorWalletServiceWallet
```
pedroferreira1, line 10:
```suggestion
   * @memberof PushNotification
```

We must update this in all the methods

src/pushNotification.ts
```diff
@@ -0,0 +1,39 @@
...
+   * @inner
+   */
+  static async registerDevice(wallet: HathorWalletServiceWallet, payload: PushRegisterRequestData): Promise<PushNotificationResult> {
```
pedroferreira1, line 13: Why don't we add here `wallet.failIfWalletNotReady()`, like we had before?

src/wallet/types.ts
```diff
@@ -373,6 +373,61 @@ export interface FullNodeVersionData {
...
+};
+
+export interface PushUpdateResponseData extends PushNotificationResponseData {};
+
+export interface PushUnregisterRequestData {
```
pedroferreira1, line 409: We don't have this anymore, right?

src/pushNotification.ts
alexruzenhack: Because this is a private method. Should we turn it public? I don't see any problem about it. However the axiosInstance also checks if the wallet is ready, so we have a redundance here.

pedroferreira1 (approved):

tuliomir (approved):

src/pushNotification.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (response.status === 200 && response.data.success) {
+      return response.data;
+    } else {
+      throw new WalletRequestError('Error unregistering wallet from push notifications.', { cause: response.data });
+    }
+  },
+
+}
+
+export interface PushNotificationResult {
```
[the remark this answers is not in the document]
alexruzenhack: Nice!

src/pushNotification.ts
```diff
@@ -0,0 +1,111 @@
...
+   * @inner
+   */
+  static async unregisterDevice(wallet: HathorWalletServiceWallet, deviceId: string): Promise<PushNotificationResult> {
+    wallet.failIfWalletNotReady();
+    const data = await walletServiceClient.pushUnregister(wallet, deviceId);
+    return data;
+  }
+}
+
+const walletServiceClient = {
```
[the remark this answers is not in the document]
alexruzenhack: I see 2 clients there: the public one `PushNotification` and the private one `walletServiceClient`. Despite they are part of the same process, they can vary in their behavior. This segregation let us evolve them at different paces because they have different needs. In the end, it also facilitates tests, and turn maintenance easier (imo).

PR #439, opened 2022-12-11
feat: instantiate wallet with ws disabled

alexruzenhack:
### Acceptance Criteria
- add possibility to disable websocket on wallet instantiation
- update pushRegister call endpoint to `wallet/push/register`
- add getTxById

### Security Checklist


src/wallet/wallet.ts
```diff
@@ -1709,6 +1722,33 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+   * @returns {boolean} If wallet has websocket connection enabled
+   */
+  isWsEnabled(): boolean {
+    return this._isWsEnabled;
+  }
+
+  /**
+   * This is a test only method, it should not be used in production
+   */
+  async callOnWalletReadyForTest() {
```
andreabadesso, line 1741: Can't we use jest to call `onWalletReady` instead of adding a test method in the facade?

andreabadesso (changes requested):

src/wallet/wallet.ts
alexruzenhack: It is not trivial, but done.

andreabadesso (approved):

src/wallet/connection.ts
```diff
@@ -88,4 +88,32 @@ class WalletServiceConnection extends BaseConnection {
   }
 }
 
-export default WalletServiceConnection;
+/**
+ * This connection is used when the wallet has websocket disabled.
+ */
+export class DummyWalletServiceConnection extends BaseConnection {
```
pedroferreira1, line 94: I'm thinking about this dummy class. I like it in general, to be honest, but it's missing many method. For instance, the wallet class calls `this.conn.stop()` and this dummy class is not overwriting the `BaseConnection` `stop` method, so we would probably get an error.

We also have many public methods in the `BaseConnection`, e.g., `onConnectionChange`, `setState`, `endConnection`, `setup`, ...

I understand that those method shouldn't be used but they are public and we might be using them directly in the wallets.

Not sure what's the best approach here, to be honest, you could overwrite all those method but, if we add a new one to the `BaseConnection` in the future, we would need to remember to also overwrite here, that's why I'm not sure if this is the best thing to do.

What do you think about it?

alexruzenhack: Humm... we can extract an interface for the BaseConnection, this way we do not forget, and for the methods not being used, we can throw errors or do nothing.

This is a class for a very specific case, we can stress it in the class documentation and make it clear it should not be used for other purposes.

The other way to solve it is by doing what is specified in the design, but it interferes more with the wallet code and I avoided it on purpose.

src/wallet/connection.ts
```diff
@@ -88,4 +91,95 @@ class WalletServiceConnection extends BaseConnection {
...
+
+  getCurrentNetwork(): string {
+    return this.currentNetwork;
+  }
+
+  setWalletId(walletId: string) {
+    this.walletId = walletId;
+  }
+
+  isStateClosed(): boolean {
```
pedroferreira1, line 174: You don't have this method in the other connections, why do you need it?

src/wallet/connection.ts
```diff
@@ -88,4 +91,95 @@ class WalletServiceConnection extends BaseConnection {
...
+  }
+
+  start(): void {
+    if (!this.walletId) {
+      throw new Error('Wallet id should be set before connection start.');
+    }
+  }
+
+  stop(): void {
+    this.removeAllListeners();
```
pedroferreira1, line 134: Why do you have to implement all the methods? I thought it would be a dummy connection but here you have change of state.

alexruzenhack: It mimics a real implementation but with no logic underneath. Here I'm removing the listeners because the class extends the EventEmmiter. Eventually, if the wallet set any listener, it will be removed when `stop` is called. But right now the `start` is not even been called. Therefore this implementation is a guard *just in case*.

src/wallet/connection.ts
alexruzenhack: Yes, this method is not part of the interface. It is here just to make the test cover the` close` method, without exposing the internal state in the class, which would be a violation of the *Open Close Principle*.

src/wallet/connection.ts
```diff
@@ -88,4 +91,95 @@ class WalletServiceConnection extends BaseConnection {
...
+    // There is nothing to handle
+  }
+
+  onConnectionChange(value: boolean): void {
+    if (value) {
+      this.setState(ConnectionState.CONNECTED);
+    } else {
+      this.setState(ConnectionState.CONNECTING);
+    }
+  }
```
andreabadesso, line 156: Do you really need to re-implement this method? Where is this used?

src/wallet/connection.ts
```diff
@@ -88,4 +91,95 @@ class WalletServiceConnection extends BaseConnection {
...
+    return this.currentServer;
+  }
+
+  getCurrentNetwork(): string {
+    return this.currentNetwork;
+  }
+
+  setWalletId(walletId: string) {
+    this.walletId = walletId;
+  }
```
andreabadesso, line 172: Same here

src/wallet/wallet.ts
```diff
@@ -581,7 +592,9 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     // We should wait for new addresses before setting wallet to ready
     await this.getNewAddresses(true);
 
-    this.setupConnection();
+    if (this.isWsEnabled()) {
```
andreabadesso, line 595: Do we really need `this.isWsEnabled`?

I mean, the idea of using the `DummyConnection` is to be able to call the methods from it with no effect, like this `this.setupConnection`

andreabadesso (changes requested):

src/wallet/connection.ts
alexruzenhack: As this is a simple state change, it is cheap to implement, and make the dummy usage more consistent overall.

src/wallet/connection.ts
alexruzenhack: Just to make the dummy more consistent.

src/wallet/connection.ts
```diff
@@ -88,4 +91,85 @@ class WalletServiceConnection extends BaseConnection {
...
```
pedroferreira1, line 174: I still don't understand why we need to implement those methods here, they are duplicated code from other classes.

src/wallet/connection.ts
```diff
@@ -88,4 +91,85 @@ class WalletServiceConnection extends BaseConnection {
...
+    this.currentServer = servers[0] || config.getServerUrl();
+  }
+
+  /**
+   * Dummy method to simulate the EventEmitter.on method to avoid errors
+   * @param _event dummy event
+   * @param _listener dummy listener
+   * @returns dummy emitter
+   */
+  on(_event: string, _listener: (...args: any[]) => void): EventEmitter {
```
pedroferreira1, line 131: If you need to implement this method, maybe we should continue extending `EventEmitter`, right? And then we can just remove this method from here. 

Is that correct?

alexruzenhack: Correct. But with this dummy EventEmitter we can avoid side effects by guaranteeing it will do nothing, even if the wallet installs any listener on the connection.

src/wallet/connection.ts
```diff
@@ -10,13 +10,16 @@ import config from '../config';
...
+  IConnection,
 } from '../connection';
 import {
   WsTransaction,
   ConnectionState,
 } from './types';
+import { EventEmitter } from 'events';
 
 export interface WalletServiceConnectionParams extends ConnectionParams {
+  network?: string;
```
pedroferreira1, line 22: Why did you add this here? `ConnectionParams` already has this parameter

src/wallet/connection.ts
```diff
@@ -10,13 +10,16 @@ import config from '../config';
 import BaseConnection, {
   DEFAULT_PARAMS,
   ConnectionParams,
+  IConnection,
 } from '../connection';
 import {
   WsTransaction,
   ConnectionState,
 } from './types';
+import { EventEmitter } from 'events';
```
pedroferreira1, line 19: Are you using this?

src/wallet/wallet.ts
```diff
@@ -109,9 +109,12 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
   // Index of the address to be used by the wallet
   private indexToUse: number;
   // WalletService-ready connection class
+  // A wallet initialized with enableWs=false will have a DummyWalletServiceConnection
```
pedroferreira1, line 112: Old comment

src/wallet/wallet.ts
```diff
@@ -142,6 +146,12 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
 
     // Setup the connection so clients can listen to its events before it is started
     this.conn = new WalletServiceConnection();
+    if (enableWs) {
+      this._isWsEnabled = true;
+    } else {
+      // We use a dummy connection to avoid having to change `conn` declaration
```
pedroferreira1, line 152: Wrong comment

src/pushNotification.ts
```diff
@@ -45,7 +45,7 @@ const walletServiceClient = {
   async pushRegister(wallet: HathorWalletServiceWallet, payload: PushRegisterRequestData): Promise<PushRegisterResponseData> {
     const axios = await axiosInstance(wallet, true);
 
-    const response = await axios.post<PushRegisterResponseData>('push/register', payload);
+    const response = await axios.post<PushRegisterResponseData>('wallet/push/register', payload);
```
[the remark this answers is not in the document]
alexruzenhack: for the WS no, but meanwhile this PR I changed the lambda path and didn't update the lib. Can we use this PR to convey this change?

src/wallet/connection.ts
alexruzenhack: In the dummy, this was an optional param. That's why I added `?`, but sure, I will remove it.

src/wallet/connection.ts
alexruzenhack: No.

src/pushNotification.ts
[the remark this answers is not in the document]
pedroferreira1: It's ok because it's a small change but you should update the acceptance criteria to reflect this, otherwise we may think this is wrong or a leftover from a previous merge

src/wallet/connection.ts
```diff
@@ -17,6 +17,7 @@ import {
 } from './types';
 
 export interface WalletServiceConnectionParams extends ConnectionParams {
+  network: string;
```
pedroferreira1, line 20: I still don't understand what changed in your PR that you need to add this here, as far as I understood, `ConnectionParams` already has this parameter.

setupTests.js
```diff
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+process.env.NODE_ENV = 'test';
```
andreabadesso, line 8: Wouldn't it be better if we set this `ENV` in the package.json command instead of mutating the env on code?

andreabadesso (approved):

src/wallet/connection.ts
alexruzenhack: Removed.

andreabadesso (approved):

pedroferreira1 (approved):

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -30,6 +30,86 @@ import MemoryStore from '../../src/memory_store'
 const fakeTokenUid = '008a19f84f2ae284f19bf3d03386c878ddd15b8b0b604a3a3539aa9d714686e1';
 const sampleNftData = 'ipfs://bafybeiccfclkdtucu6y4yc5cpr6y3yuinr67svmii46v5cfcrkp47ihehy/albums/QXBvbGxvIDEwIE1hZ2F6aW5lIDI3L04=/21716695748_7390815218_o.jpg';
 
+describe('getTxById', () => {
+  afterEach(async () => {
+    await stopAllWallets();
+    await GenesisWalletHelper.clearListeners();
+  });
+
+  it('should return tx token balance', async () => {
```
pedroferreira1, line 39: Should we add more tokens to the test?

pedroferreira1 (approved):

__tests__/hathorwallet.test.js
```diff
@@ -734,3 +735,102 @@ test('getTokenHistoryObject', () => {
...
+          uid: 'A',
+          name: 'Token A',
+          symbol: 'A',
+        },
+      ]
+    },
+    meta: {
+      voided_by: [],
+    },
+  });
```
andreabadesso, line 792:
```suggestion
      outputs: [{
        token: '00',
        value: 942,
        decoded: {
          address: mineAddress,
        },
      }, {
        token: 'A',
        value: 329,
        decoded: {
          address: mineAddress,
        },
      }, {
        token: '00',
        value: 731,
        decoded: {
          address: notMineAddress,
        },
      }],
      tokens: [{
        uid: '00',
        name: 'Hathor',
        symbol: 'HTR',
      }, {
        uid: 'A',
        name: 'Token A',
        symbol: 'A',
      }]
    },
    meta: {
      voided_by: [],
    },
  });
```

__tests__/hathorwallet.test.js
```diff
@@ -734,3 +735,102 @@ test('getTokenHistoryObject', () => {
...
+        timestamp: 1,
+        version: 1,
+        voided: false,
+        weight: 1,
+        tokenId: 'A',
+        tokenName: 'Token A',
+        tokenSymbol: 'A',
+        balance: 329
+      },
+    ],
```
andreabadesso, line 823:
```suggestion
    txTokens: [{
      txId: txId,
      timestamp: 1,
      version: 1,
      voided: false,
      weight: 1,
      tokenId: '00',
      tokenName: 'Hathor',
      tokenSymbol: 'HTR',
      balance: 1673
    }, {
      txId: txId,
      timestamp: 1,
      version: 1,
      voided: false,
      weight: 1,
      tokenId: 'A',
      tokenName: 'Token A',
      tokenSymbol: 'A',
      balance: 329
    }],
```

src/wallet/wallet.ts
```diff
@@ -142,6 +145,11 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
 
     // Setup the connection so clients can listen to its events before it is started
     this.conn = new WalletServiceConnection();
+    if (enableWs) {
+      this._isWsEnabled = true;
+    } else {
+      this._isWsEnabled = false;
+    }
```
andreabadesso, line 152: Can't we do something like:

```suggestion
    this._isWsEnabled = enableWs;
```

Since `enableWs` has a default value on the constructor?

andreabadesso (changes requested):

andreabadesso (approved):

andreabadesso (approved):

__tests__/integration/hathorwallet_facade.test.js
pedroferreira1: This is an example of test that has custom tokens:

```
  it('should get the balance for a custom token', async () => {
    const hWallet = await generateWalletHelper();

    // Validating results for a nonexistant token
    const emptyBalance = await hWallet.getBalance(fakeTokenUid);
    expect(emptyBalance).toHaveLength(1);
    expect(emptyBalance[0]).toMatchObject({
      token: { id: fakeTokenUid },
      balance: { unlocked: 0, locked: 0 },
      transactions: 0,
    });

    // Creating a new custom token
    await GenesisWalletHelper.injectFunds(hWallet.getAddressAtIndex(0), 10);
    const newTokenAmount = getRandomInt(1000, 10);
    const { hash: tokenUid } = await createTokenHelper(
      hWallet,
      'BalanceToken',
      'BAT',
      newTokenAmount,
    );

    const tknBalance = await hWallet.getBalance(tokenUid);
    expect(tknBalance[0]).toMatchObject({
      balance: { unlocked: newTokenAmount, locked: 0 },
      transactions: expect.any(Number),
      // transactions: 1, // TODO: The amount of transactions is often 8 but should be 1. Ref #397
    });

    // Validating that a different wallet (genesis) has no access to this token
    const { hWallet: gWallet } = await GenesisWalletHelper.getSingleton();
    const genesisTknBalance = await gWallet.getBalance(tokenUid);
    expect(genesisTknBalance).toHaveLength(1);
    expect(genesisTknBalance[0]).toMatchObject({
      token: { id: tokenUid },
      balance: { unlocked: 0, locked: 0 },
      transactions: 0,
    });
  });
```

alexruzenhack: [540af9](https://github.com/HathorNetwork/hathor-wallet-lib/pull/439/commits/540af96d83a29d201710ba996f6a465a3654f3f3)

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -30,6 +30,158 @@ import MemoryStore from '../../src/memory_store'
...
+      // transactions: 1, 
+    });
+    // Get balance for the token creation transaction
+    const result = await hWallet.getTxById(tokenUid);
+    expect(result.success).toStrictEqual(true);
+    expect(result.txTokens).toHaveLength(2);
+    expect(result.txTokens).toEqual(expect.arrayContaining([
+      expect.objectContaining({
+        tokenId: HATHOR_TOKEN_CONFIG.uid,
+        balance: expect.any(Number),
```
alexruzenhack, line 165: 💬 It would be nice to have the jest-extended lib here and use the asymmetric matcher [`toBeNegative`](https://jest-extended.jestcommunity.dev/docs/matchers/number/#tobenegative).

pedroferreira1 (approved):

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -112,6 +112,74 @@ describe('getTxById', () => {
...
+    const tknBalance = await hWallet.getBalance(tokenUid);
+    // Assert that only one balance is returned
+    expect(tknBalance).toHaveLength(1);
+    // Assert the balance is equal to the amount generated
+    expect(tknBalance[0]).toMatchObject({
+      token: { id: tokenUid },
+      balance: { unlocked: newTokenAmount, locked: 0 },
+      transactions: expect.any(Number),
+      /**
+       * TODO: The amount of transactions is often 8 but should be 1. Ref #397
```
andreabadesso, line 153: wtf

andreabadesso (approved):

PR #442, opened 2022-12-14
refactor: load history of addresses

r4mmer:
### Acceptance Criteria

- Reduce bottlenecks from the process of load history
- Make the process simpler and less error prone

### Security Checklist

### DIFF
--- a/jest-integration.config.js
+++ b/jest-integration.config.js
@@ -23,10 +23,10 @@ module.exports = {
     },
     // We need a high coverage for the HathorWallet class
     './src/new/wallet.js': {
-      statements: 93,
-      branches: 89,
+      statements: 91,
+      branches: 87,
       functions: 94,
-      lines: 93
+      lines: 91
     }
   },
 };
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -500,6 +500,9 @@ class HathorWallet extends EventEmitter {
    * @inner
    */
   getAddressAtIndex(index) {
+    if (this.preCalculatedAddresses && this.preCalculatedAddresses[index]) {
+      return this.preCalculatedAddresses[index];
+    }
     storage.setStore(this.store);
     return wallet.getAddressAtIndex(index);
   }
@@ -2477,7 +2480,6 @@ class HathorWallet extends EventEmitter {
     const stopIndex = startIndex + count;
     for (let i = startIndex; i < stopIndex; i++) {
       // Generate each key from index (if not provided pre-calculated)
-      // XXX: maybe getAddressAtIndex should get from preCalculated as well
       const address = this.preCalculatedAddresses && this.preCalculatedAddresses[i]
         ? this.preCalculatedAddresses[i]
         : wallet.generateAddress(i);
@@ -2716,7 +2718,7 @@ class HathorWallet extends EventEmitter {
       // Setting last shared address, if necessary
       const candidateIndex = maxIndex + 1;
       if (candidateIndex > lastSharedIndex) {
-        const address = wallet.getAddressAtIndex(candidateIndex);
+        const address = this.getAddressAtIndex(candidateIndex);
         wallet.updateAddress(address, candidateIndex);
       }
     }
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2509,7 +2509,7 @@ class HathorWallet extends EventEmitter {
 
       // Subscribe in websocket to the addresses
       for (let address of addressesToSearch) {
-        wallet.subscribeAddress(address, connection);
+        wallet.subscribeAddress(address, this.conn);
       }
 
       while (hasMore === true) {
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2497,6 +2497,14 @@ class HathorWallet extends EventEmitter {
     return this.fetchTxHistory(addresses);
   }
 
+  /**
+   * Fetch history of addresses from server
+   * then load more addresses if needed to fill the gap limit.
+   *
+   * @param {string[]} addresses Array of addresses to fetch history
+   *
+   * @return {Promise<void>} Promise that resolves when addresses history is finished loading from server
+   */
   async fetchTxHistory(addresses) {
     storage.setStore(this.store);
     // Split addresses array into chunks of at most MAX_ADDRESSES_GET size
@@ -2581,8 +2589,8 @@ class HathorWallet extends EventEmitter {
     }
     // After the addresses have been loaded and properly saved we need to check if we
     // need to load more addresses to keep the gapLimit.
-    // checkGapLimit will resolve when the address history of the gapLimit is loaded
-    await this.checkGapLimit();
+    // fillUntilGapLimit will resolve when the address history of the gapLimit is loaded
+    await this.fillUntilGapLimit();
   }
 
   /**
@@ -2591,7 +2599,7 @@ class HathorWallet extends EventEmitter {
    *
    * @returns {Promise<void>} A promise that will resolve when the addresses are loaded
    */
-  async checkGapLimit() {
+  async fillUntilGapLimit() {
     storage.setStore(this.store);
     const lastGeneratedIndex = wallet.getLastGeneratedIndex();
     const lastUsedIndex = wallet.getLastUsedIndex();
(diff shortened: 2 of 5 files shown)


src/new/wallet.js
```diff
@@ -2464,6 +2459,277 @@ class HathorWallet extends EventEmitter {
...
+    this.preProcessedData = {};
+
+    // Load history from server
+    return this.loadAddresses(0, wallet.getGapLimit());
+  }
+
+    /**
+   * Update the historyTransactions and allTokens from a new array of history that arrived
+   *
+   * @param {Array} newHistory Array of new data that arrived from the server to be added to local data
```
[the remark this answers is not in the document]
r4mmer: It is not a `Transaction` but we have an interface for this as `HistoryTransaction` on `models/type.ts`

src/new/wallet.js
```diff
@@ -505,6 +500,9 @@ class HathorWallet extends EventEmitter {
    * @inner
    */
   getAddressAtIndex(index) {
+    if (this.preCalculatedAddresses && this.preCalculatedAddresses[index]) {
```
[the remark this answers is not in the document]
r4mmer: I think we should think this more thoroughly, there are possible issues, e.g. we have an array with 20 addresses and decide to load the index 100k, the "cache" array will need to be recreated to hold the first 20 addresses and a lot of `undefined`s up to the 100k index.

Although I think we should minimize the derivation operations since they are cpu intensive there is also a tradeoff of increasing the wallet's memory requirements.

src/new/wallet.js
```diff
@@ -2464,6 +2459,277 @@ class HathorWallet extends EventEmitter {
...
+    await this.checkGapLimit();
+  }
+
+  /**
+   * Check if we need to load more addresses to fill the gap limit.
+   * If we need more addresses generate them then load the history.
+   *
+   * @returns {Promise<void>} A promise that will resolve when the addresses are loaded
+   */
+  async checkGapLimit() {
```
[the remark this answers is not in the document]
r4mmer: changed to `fillUntilGapLimit`

src/new/wallet.js
```diff
@@ -2464,6 +2459,277 @@ class HathorWallet extends EventEmitter {
...
+    if (storage.getItem('wallet:address') === null) {
+      wallet.updateAddress(addresses[0], startIndex);
+    }
+
+    wallet.updateLastGeneratedIndex(stopIndex - 1);
+    wallet.setWalletData(walletData);
+    return this.fetchTxHistory(addresses);
+  }
+
+  async fetchTxHistory(addresses) {
```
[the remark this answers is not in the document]
r4mmer: Added docstrings

tuliomir (approved):

jest-integration.config.js
```diff
@@ -23,10 +23,10 @@ module.exports = {
     },
     // We need a high coverage for the HathorWallet class
     './src/new/wallet.js': {
-      statements: 93,
```
pedroferreira1, line 26: Why did you reduce the coverage instead of adding tests?

I know that you've added some but can't we continue with the previous coverage and just add more tests for more cases?

src/wallet.js
```diff
@@ -385,6 +385,22 @@ const wallet = {
...
   },
 
+  /**
+   * Update the last generated address index with a new candidate index if it is
+   * greater than the current last generated index.
+   *
+   * @param {number} newIndex The candidate index for the last generated index
+   * @returns {boolean} Whether the last generated index was updated
+   */
+  updateLastGeneratedIndex(newIndex) {
```
pedroferreira1, line 395: Not sure I like the method name because it may not update the last generated index.

I couldn't think of a good name, to be honest, maybe `conditionallyUpdateLastGeneratedIndex`, or `updateLastGeneratedIndexIfGreater`. Not sure the best approach but the current name is a bit confusing in my opinion.

src/new/wallet.js
```diff
@@ -296,19 +297,13 @@ class HathorWallet extends EventEmitter {
...
       if (this.firstConnection) {
         this.firstConnection = false;
-        promise = wallet.loadAddressHistory(
-          0,
-          wallet.getGapLimit(),
-          this.conn,
-          this.store,
-          this.preCalculatedAddresses
-        );
+        promise = this.loadAddresses(0, wallet.getGapLimit());
```
pedroferreira1, line 300: Does it make sense (in the future) for the gap limit to be configured for each wallet? In that case, we would have a new parameter to set the gap limit for this wallet.

src/new/wallet.js
pedroferreira1, line 503: Why don't we change a bit this and throw an error if you have `this.preCalculatedAddresses` but not with this index?

```
if (this.preCalculatedAddresses) {
  if (!this.preCalculatedAddresses[index]) throw Error();
  return this.preCalculatedAddresses[index];
}
```

Otherwise we would be calling a method to generate a new address but we are using pre calculated addresses

src/new/wallet.js
```diff
@@ -2469,6 +2464,285 @@ class HathorWallet extends EventEmitter {
...
+    storage.setStore(this.store);
+
+    const addresses = [];
+    const walletData = wallet.getWalletData();
+    const stopIndex = startIndex + count;
+    for (let i = startIndex; i < stopIndex; i++) {
+      // Generate each key from index (if not provided pre-calculated)
+      const address = this.preCalculatedAddresses && this.preCalculatedAddresses[i]
+        ? this.preCalculatedAddresses[i]
+        : wallet.generateAddress(i);
```
pedroferreira1, line 2490: You can use `this.getAddressAtIndex`, right? Then you already receive the string format of the address. We don't use the full object in this method anyway. Is that correct?

src/new/wallet.js
```diff
@@ -2469,6 +2464,285 @@ class HathorWallet extends EventEmitter {
...
+      walletData.keys[address.toString()] = { index: i };
+      addresses.push(address.toString());
+      // Address derivation can be cpu intensive so we release the event loop for other possible operations
+      await new Promise(resolve => { setTimeout(resolve, 0) });
+    }
+    if (storage.getItem('wallet:address') === null) {
```
pedroferreira1, line 2496: You can use here `wallet.getCurrentAddress()`, so we prevent this direct storage access.

src/new/wallet.js
```diff
@@ -2469,6 +2464,285 @@ class HathorWallet extends EventEmitter {
...
+            // Prepare parameters for next request
+            firstHash = result.first_hash;
+            const addrIndex = addressesToSearch.indexOf(result.first_address);
+            if (addrIndex === -1) {
+              throw new Error("Invalid address returned from the server.");
+            }
+            addressesToSearch = addressesToSearch.slice(addrIndex);
+          } else {
+            // save the history on storage since the pagination of this chunk is over
+            const historyLoadPartialUpdate = this.saveNewHistory([].concat(...historyBuffer));
```
pedroferreira1, line 2586: Not sure how optimized `[].concat(...historyBuffer)` is, to be honest, because this array might be huge.

Maybe we could try to use some `lodash` methods because they are direct and care for performance (and we already have this lib in the project). 

https://lodash.com/docs/4.17.15#flatten this should do the job, or `flattenDeep`. But as I said, I don't know how to compare the performance of both.

What do you think?

src/new/wallet.js
```diff
@@ -2469,6 +2464,285 @@ class HathorWallet extends EventEmitter {
...
+   * @return {{addressesFound: number, maxIndex: number, historyLength: number}}
+   * @memberof Wallet
+   * @inner
+   */
+  saveNewHistory(newHistory) {
+    storage.setStore(this.store);
+
+    const walletData = wallet.getWalletData();
+    const oldHistoryTransactions = 'historyTransactions' in walletData ? walletData['historyTransactions'] : {};
+    // XXX: Why copy the history?
```
pedroferreira1, line 2663: Not sure the reason, maybe we shouldn't copy. What do you think?

src/new/wallet.js
```diff
@@ -2469,6 +2464,285 @@ class HathorWallet extends EventEmitter {
...
+        wallet.updateAddress(address, candidateIndex);
+      }
+    }
+
+    // Saving to storage before resolving the promise
+    wallet.saveAddressHistory(historyTransactions, allTokens);
+
+    const lastGeneratedIndex = wallet.getLastGeneratedIndex();
+
+    return { addressesFound: lastGeneratedIndex + 1, maxIndex, historyLength: Object.keys(historyTransactions).length };
```
pedroferreira1, line 2744: `newSharedAddress` and `newSharedIndex` that were also returned in the old utils are not used?

src/new/wallet.js
r4mmer: I think we may want to start with the first 50 addresses preCalculated, then load more addresses as needed.

PR #443, opened 2022-12-15
feat: ws processing queue

r4mmer:
### Acceptance Criteria
- Create a queue to process txs that arrived via websocket during initial load
- Process queue after `preProcessWalletData`
- When sync finishes change state to `PROCESSING` instead of `READY`
- `preProcessWalletData` will change state to `READY` when it finishes processing (this includes the ws txs on queue)

### Security Checklist

### DIFF
--- a/__tests__/hathorwallet.test.js
+++ b/__tests__/hathorwallet.test.js
@@ -455,4 +455,7 @@ test('preProcessWalletData', async () => {
       voided: false,
     }],
   });
+  expect(hWallet.getPreProcessedData('balanceByToken')).toEqual({
+    A: { unlocked: 5, locked: 5, transactions: 1 }
+  });
 });
--- a/__tests__/hathorwallet.test.js
+++ b/__tests__/hathorwallet.test.js
@@ -358,17 +358,19 @@ test('getTxBalance', async () => {
   mockAddrCheck.mockRestore();
 });
 
-test('getPreProcessedData', async () => {
+test('getPreProcessedData', () => {
   const hWallet = new FakeHathorWallet();
   hWallet.preProcessedData = {};
 
-  await expect(hWallet.getPreProcessedData('foo')).rejects.toThrow();
+  expect(() => {
+    hWallet.getPreProcessedData('foo')
+  }).toThrow();
 
   hWallet.preProcessedData = {
     foo: 123,
   };
 
-  expect(await hWallet.getPreProcessedData('foo')).toEqual(123);
+  expect(hWallet.getPreProcessedData('foo')).toEqual(123);
 });
 
 test('setState', async () => {
@@ -380,7 +382,6 @@ test('setState', async () => {
   hWallet.setState(HathorWallet.SYNCING);
   expect(hWallet.preProcessWalletData).not.toHaveBeenCalled();
   expect(hWallet.state).toEqual(HathorWallet.SYNCING);
-
 
   hWallet.setState(HathorWallet.PROCESSING);
   expect(hWallet.preProcessWalletData).toHaveBeenCalled();
@@ -394,4 +395,64 @@ test('setState', async () => {
   hWallet.setState(HathorWallet.READY);
   expect(hWallet.preProcessWalletData).not.toHaveBeenCalled();
   expect(hWallet.state).toEqual(HathorWallet.READY);
+});
+
+test('preProcessWalletData', async () => {
+  const hWallet = new FakeHathorWallet();
+  hWallet.preProcessedData = {};
+
+  hWallet.getTxBalance.mockReturnValue(Promise.resolve({
+    'A': 10,
+  }));
+  hWallet.isAddressMine.mockReturnValue(true);
+
+  hWallet._getBalanceRaw.mockReturnValue({ unlocked: 5, locked: 5 });
+
+  hWallet.getFullHistory.mockReturnValue({
+    'txid1': {
+      tx_id: 'txId1',
+      timestamp: 123,
+      is_voided: false,
+      outputs: [
+        {
+          token: 'A',
+          value: 5,
+          decoded: { address: 'addr1' },
+        },
+        {
+          token: 'A',
+          value: 5,
+          decoded: { address: 'addr1', timelock: 127 },
+        }
+      ],
+      inputs: [],
+    },
+  });
+
+  hWallet.processTxQueue.mockReturnValue(Promise.resolve());
+
+  expect(() => {
+    hWallet.getPreProcessedData('tokens');
+  }).toThrow();
+  expect(() => {
+    hWallet.getPreProcessedData('historyByToken');
+  }).toThrow();
+  expect(() => {
+    hWallet.getPreProcessedData('balanceByToken');
+  }).toThrow();
+
+
+  await hWallet.preProcessWalletData();
+  expect(hWallet.processTxQueue).toHaveBeenCalled();
+
+  expect(hWallet.getPreProcessedData('tokens')).toEqual(['A']);
+  expect(hWallet.getPreProcessedData('historyByToken')).toEqual({
+    'A': [{
+      txId: 'txId1',
+      timestamp: 123,
+      tokenUid: 'A',
+      balance: 10,
+      voided: false,
+    }],
+  });
 });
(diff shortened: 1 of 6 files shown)


src/new/wallet.js
```diff
@@ -1122,6 +1131,21 @@ class HathorWallet extends EventEmitter {
...
 
+  /**
+   * Process the transactions on the websocket transaction queue as if they just arrived.
+   *
+   * @memberof HathorWallet
+   * @inner
+   */
+  async processTxQueue() {
+    let wsData = this.wsTxQueue.dequeue();
+    while(wsData !== undefined) {
```
pedroferreira1, line 1142:
```suggestion
    while (wsData !== undefined) {
```

__tests__/hathorwallet.test.js
```diff
@@ -244,3 +245,214 @@ test('getWalletInputInfo', () => {
...
+  expect(() => {
+    hWallet.getPreProcessedData('balanceByToken');
+  }).toThrow();
+
+
+  await hWallet.preProcessWalletData();
+  expect(hWallet.processTxQueue).toHaveBeenCalled();
+
+  expect(hWallet.getPreProcessedData('tokens')).toEqual(['A']);
+  expect(hWallet.getPreProcessedData('historyByToken')).toEqual({
```
pedroferreira1, line 449: Maybe add another check with the balance by token?

pedroferreira1 (approved):

__tests__/hathorwallet.test.js
msbrogli: +1

src/models/queue.ts
```diff
@@ -0,0 +1,75 @@
...
+  value: T,
+  next?: QueueNode<T>;
+};
+
+/**
+ * This is a simple queue using an underlying linked list for O(1) enqueue and dequeue operations.
+ * 
+ * @template [T=Object]
+ */
+export default class Queue<T=Object> {
```
msbrogli, line 18: The implementation seems good to me. But why do we need this? Isn't an array enough?

__tests__/models/queue.test.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(q.size()).toEqual(0);
+    const elements = [1, 2, 3];
+    // Enqueue elements
+    elements.forEach(el => q.enqueue(el));
+    expect(q.size()).toEqual(3);
+    // dequeued elements should in the same order as they were enqueued
+    elements.forEach(expected => {
+        // Peek should return the next element
+        const peekedEl = q.peek();
+        expect(peekedEl).toEqual(expected);
```
msbrogli, line 21: Check the size after peeking and after dequeuing.

src/new/wallet.js
```diff
@@ -566,7 +570,12 @@ class HathorWallet extends EventEmitter {
    **/
   handleWebsocketMsg(wsData) {
     if (wsData.type === 'wallet:address_history') {
-      this.onNewTx(wsData);
+      if ([HathorWallet.PROCESSING, HathorWallet.SYNCING].includes(this.state)) {
```
msbrogli, line 573: I feel it's better to check whether state is not ready. So we can safely add other states in the future.

src/new/wallet.js
```diff
@@ -566,7 +570,12 @@ class HathorWallet extends EventEmitter {
...
+        // Cannot receive a ws message, so we enqueue for later
```
msbrogli, line 574: I guess this comment is imprecise. Maybe "cannot process new transactions from ws when the wallet is not ready. Queuing it to be processed later."?

src/new/wallet.js
```diff
@@ -1279,25 +1305,19 @@ class HathorWallet extends EventEmitter {
...
-      // to support using this facade interchangable with wallet service
-      // facade in both wallets
-
-      // Notice: preProcessWalletData is currently async because `getTxBalance`
-      // is async but it resolves instantly -- we only changed `getTxBalance` to async
-      // to match signatures with the new facade. If we ever need to do something really
-      // async on preProcessWalletData and wait for it before setting state to READY, we
-      // probably should refactor setState to be async
-      this.preProcessWalletData();
+    if (state === HathorWallet.PROCESSING && state !== this.state) {
```
msbrogli, line 1316: I understand the previous code had state transition logic here. I'm not sure this is the best place to add have it, though. For now, maybe we should move this code to a "onEnterStateProcessing()" method?

msbrogli (changes requested):

src/new/wallet.js
```diff
@@ -1122,6 +1131,21 @@ class HathorWallet extends EventEmitter {
...
+   *
+   * @memberof HathorWallet
+   * @inner
+   */
+  async processTxQueue() {
+    let wsData = this.wsTxQueue.dequeue();
+    while(wsData !== undefined) {
+      // process wsData like it just arrived
+      this.onNewTx(wsData);
+      wsData = this.wsTxQueue.dequeue();
```
msbrogli, line 1145: Should we process one transaction per call and just schedule this method to run again in zero seconds? This would unblock the main event loop to handle other stuff if needed.

src/models/queue.ts
r4mmer: The most simple queue implementation with array would use `array.shift` which is O(n) (depending on the engine, firefox managed to make it O(1)).
This also would not be a problem on the current implementation but the original intent was to use it as a queue to process all transactions (the ones comming from the sync and websocket) and in this case it would be problematic since the array would contain all transactions in the history during it's first processing
Shifting all transactions one by one would be O(n^2) this is why it was discarded.

The load refactor may include this feature of unifying the transaction processing, this is why I chose to maintain this queue approach.

src/new/wallet.js
r4mmer: Done

src/new/wallet.js
r4mmer: Done

src/models/queue.ts
```diff
@@ -0,0 +1,75 @@
...
+   * @param {T} value An item to enqueue
+   */
+  enqueue(value: T) {
+    const node: QueueNode<T> = { value, next: undefined };
+    if (this.head) {
+      if (this.last === undefined) {
+        // This shouldn't happen
+        throw new Error('Queue: queue has a start but no end');
+      }
+      this.last.next = node;
```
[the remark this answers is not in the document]
r4mmer: Done

src/new/wallet.js
r4mmer: Done

__tests__/hathorwallet.test.js
r4mmer: Done

__tests__/models/queue.test.js
r4mmer: Done

src/new/wallet.js
r4mmer: Done

src/new/wallet.js
r4mmer: Done.

pedroferreira1 (approved):

src/models/queue.ts
msbrogli: Got it. Thanks! We might consider to use the [collections.js](https://www.collectionsjs.com/) library in the future. The downside is to add a dependency. The upside is to have access to efficient and well-tested data structures.

__tests__/models/queue.test.js
msbrogli: Thanks!

src/new/wallet.js
msbrogli: Great solution! Thanks!

src/new/wallet.js
msbrogli: Great! Thanks!

msbrogli (approved):

src/new/wallet.js
msbrogli: Thanks!

src/new/wallet.js
msbrogli: Thanks!

src/models/queue.ts
r4mmer: I considered using this library but did not want to add a new dependency just for a queue.
But it has some other very interesting data structures like sorted maps which could be very useful

msbrogli (approved):

PR #444, opened 2022-12-15
feat: calculate balance for all tokens

r4mmer:
### Acceptance Criteria
- Create a method to calculate balance for all tokens in the history
- Use this method instead of calculating for individual tokens on `preProcessingWalletData`

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1075,17 +1075,20 @@ class HathorWallet extends EventEmitter {
   /**
    * Get balance for a token (same as old method to be used for compatibility)
    *
-   * @params {string} tokenUid Token uid
+   * @params {string|null} tokenUid Token uid, default will calculate for all tokens
    *
    * @return {Object} Object with balance { available, locked }
    *
    * @memberof HathorWallet
    * @inner
    **/
-  _getBalanceRaw(tokenUid) {
+  _getBalanceRaw(tokenUid = null) {
     storage.setStore(this.store);
     const uid = tokenUid || this.token.uid;
     const historyTransactions = this.getFullHistory();
+    if (tokenUid === null) {
+      return wallet.calculateTokenBalances(Object.values(historyTransactions));
+    }
     return wallet.calculateBalance(Object.values(historyTransactions), uid);
   }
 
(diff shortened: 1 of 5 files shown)


src/new/wallet.js
```diff
@@ -1169,12 +1172,9 @@ class HathorWallet extends EventEmitter {
...
     }
 
-    const tokensBalance = {};
+    const tokensBalance = this._getBalanceRaw();
     for (const tokenUid of Object.keys(tokensHistory)) {
-      const totalBalance = this._getBalanceRaw(tokenUid);
-      totalBalance['transactions'] = transactionCountByToken[tokenUid];
-      // update token total balance
-      tokensBalance[tokenUid] = totalBalance;
+      tokensBalance[tokenUid].transactions = transactionCountByToken[tokenUid];
```
pedroferreira1, line 1177: I think this might lead to a bug. If we have a token with zero balance, it won't return in the `calculateTokenBalances` method and this will try to access the object.

Actually we must return the token in this balance object because we must show that this token exists, has 0 balance and has X transactions.

We could only check this inside this `for` and create an empty balance for the tokens that are not in the `tokensBalance` object yet.

src/new/wallet.js
```diff
@@ -1240,14 +1242,13 @@ class HathorWallet extends EventEmitter {
       }
 
       // Update token balance
-      const totalBalance = this._getBalanceRaw(tokenUid);
```
pedroferreira1, line 1243: Here we might also have the same bug as in the previous method.

pedroferreira1 (changes requested):

src/new/wallet.js
r4mmer: Done

src/new/wallet.js
```diff
@@ -1201,6 +1204,8 @@ class HathorWallet extends EventEmitter {
   async onTxArrived(tx, isNew) {
     const tokensHistory = await this.getPreProcessedData('historyByToken');
     const tokensBalance = await this.getPreProcessedData('balanceByToken');
+    const updatedTokensBalance = this._getBalanceRaw();
```
msbrogli, line 1207: Why do we have to recalculate all balances every new transaction? Can't we just update the values?

msbrogli (changes requested):

msbrogli (approved):

pedroferreira1 (approved):

PR #447, opened 2022-12-19
fix: avoiding a race condition on onOpen and better cleanup of event …

andreabadesso:
…listeners

### Acceptance Criteria
- We should only set `this.connected` after onOpen event listener is called and should check it on the `sendMessage` method, to prevent the error described in https://github.com/HathorNetwork/internal-issues/issues/144

### Security Checklist

### DIFF
--- a/src/websocket/base.ts
+++ b/src/websocket/base.ts
@@ -134,7 +134,8 @@ abstract class BaseWebSocket extends EventEmitter {
           return;
         }
       }
-      this.ws.onclose = () => {};
+
+      this.clearWsListeners();
       this.ws.close();
       this.ws = null;
     }
@@ -146,6 +147,22 @@ abstract class BaseWebSocket extends EventEmitter {
     this.ws.onmessage = (evt) => this.onMessage(evt);
     this.ws.onerror = (evt) => this.onError(evt);
     this.ws.onclose = () => this.onClose();
+
+    this.started = true;
+  }
+
+  /**
+   * Remove all event listeners from the WebSocket instance 
+   **/
+  clearWsListeners() {
+    if (!this.ws) {
+      return;
+    }
+
+    this.ws.onopen = null;
+    this.ws.onclose = null;
+    this.ws.onerror = null;
+    this.ws.onmessage = null;
   }
 
   /**
@@ -180,9 +197,12 @@ abstract class BaseWebSocket extends EventEmitter {
    * Method called when websocket connection is opened
    */
   onOpen() {
+    if (!this.started) {
+      return;
+    }
+
     this.connected = true;
     this.connectedDate = new Date();
-    this.started = true;
     this.heartbeat = setInterval(() => {
       this.sendPing();
     }, this.heartbeatInterval);
@@ -216,7 +236,7 @@ abstract class BaseWebSocket extends EventEmitter {
     this.connectedDate = null;
     this.setIsOnline(false);
     if (this.ws) {
-      this.ws.onclose = () => {};
+      this.clearWsListeners();
       this.ws.close();
       this.ws = null;
     }
@@ -239,7 +259,7 @@ abstract class BaseWebSocket extends EventEmitter {
    * Method called to send a message to the server
    */
   sendMessage(msg: string) {
-    if (!this.started) {
+    if (!this.started || !this.connected) {
       this.setIsOnline(false);
       return;
     }
@@ -283,7 +303,7 @@ abstract class BaseWebSocket extends EventEmitter {
     });
 
     this.onClose();
-  };
+  }
 
   /**
    * Method called to end a websocket connection
@@ -293,7 +313,7 @@ abstract class BaseWebSocket extends EventEmitter {
     this.started = false;
     this.connected = null;
     if (this.ws) {
-      this.ws.onclose = () => {};
+      this.clearWsListeners();
       this.ws.close();
       this.ws = null;
     }
--- a/src/websocket/base.ts
+++ b/src/websocket/base.ts
@@ -8,7 +8,6 @@
 import { EventEmitter } from 'events';
 import _WebSocket from 'isomorphic-ws';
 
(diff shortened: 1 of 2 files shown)


src/websocket/base.ts
```diff
@@ -146,6 +145,29 @@ abstract class BaseWebSocket extends EventEmitter {
...
+    this.ws.onopen = () => {};
+    this.ws.onclose = () => {};
+    this.ws.onerror = () => {};
+    this.ws.onmessage = () => {};
+
+    if (this.ws.readyState === _WebSocket.OPEN) {
+      this.ws.close();
+    }
+
+    this.ws = null;
```
msbrogli, line 170: Should we set `this.started = false;`?

src/websocket/base.ts
```diff
@@ -239,7 +260,7 @@ abstract class BaseWebSocket extends EventEmitter {
    * Method called to send a message to the server
    */
   sendMessage(msg: string) {
-    if (!this.started) {
+    if (!this.started || !this.connected) {
```
msbrogli, line 263: I feel we should add some comments at the beginning of this class explaining the difference between `this.started` and `this.connected`.

In this case, why should we set the wallet as offline if `this.connected === false`? Isn't it already marked as offline?

msbrogli (approved):

src/websocket/base.ts
```diff
@@ -146,6 +145,29 @@ abstract class BaseWebSocket extends EventEmitter {
...
     this.ws.onclose = () => this.onClose();
+
+    this.started = true;
+  }
+
+  /**
+   * Sets all event listeners to noops on the WebSocket instance 
+   * and close it.
+   **/
+  closeWs() {
```
msbrogli, line 156: Should we add a test that calls the methods just like the race condition and checks that it behaved correctly?

src/websocket/base.ts
andreabadesso: Yes we should, done in 37b7bd4f55083042ec5e4dcbf54b75ccdbd126a9

src/websocket/base.ts
andreabadesso: Added the comments in 98ef90b9f68f919b3fe13307b3c738d12f5b4975

I'm not sure why we are setting the wallet as offline, we already do it on the `onClose` event listener, my best guess is that it was done to fix some race condition in the past, since there are no comments explaining this, I'm afraid of removing it. Do you remember anything related to this?

src/websocket/base.ts
andreabadesso: Yes we should, added in a80eb84098e7c4aea5117aa6770cdbc0415c9d89

src/websocket/base.ts
```diff
@@ -146,6 +148,30 @@ abstract class BaseWebSocket extends EventEmitter {
...
+    if (!this.ws) {
+      return;
+    }
+
+    this.ws.onopen = () => {};
+    this.ws.onclose = () => {};
+    this.ws.onerror = () => {};
+    this.ws.onmessage = () => {};
+
+    if (this.ws.readyState === _WebSocket.OPEN) {
```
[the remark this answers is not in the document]
andreabadesso: I agree, done in 3bd925f5a4021e282ecaf752447e899551c21e10

src/websocket/base.ts
```diff
@@ -146,6 +148,30 @@ abstract class BaseWebSocket extends EventEmitter {
...
+      this.ws.close();
+    }
+
+    this.ws = null;
+    this.started = false;
```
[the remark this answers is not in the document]
andreabadesso: Yes we should, done in e1deaa7bdb35eeb90970337f8c63415677e3627c

src/websocket/base.ts
msbrogli: I was reading the code again, and I feel that we call this method from multiple sources. If I got it right, the `this.started` flag indicates whether it should retry to connect if the connection is closed for any reason. I'm reconsidering whether we should really set `this.start = false` here or not. What do you think?

src/websocket/base.ts
msbrogli: Thanks for adding the comments. I feel the `this.started` has a slight different meaning. I mean, even though the websocket object might change, this flag indicates whether we should try to reconnect when the connection is gone. If this is correct, we should just fail to send the message and let the websocket reconnect.

src/websocket/base.ts
msbrogli: Great! Thanks! Does this test fail without the fixes in this PR?

pedroferreira1 (approved):

src/websocket/base.ts
andreabadesso: Actually, what the reconnection mechanism is doing is calling `setup` again on the `retryConnectionInterval`, setInterval is called [here](https://github.com/HathorNetwork/hathor-wallet-lib/blob/851eec942c1406e9a81c992f5b84f7403fb91dc6/src/websocket/base.ts#L225)

The first thing the `setup` method does is checking that it is not already started, so I think it makes sense to set `this.started` to false

src/websocket/base.ts
andreabadesso: I have a different understanding of the `this.started` flag. Apparently, it is currently being used to prevent `this.setup` to be called when a connection is waiting to be established (until `onOpen`) is called by the event listener

src/websocket/base.ts
andreabadesso: Yes it does

msbrogli: Thanks!

src/websocket/base.ts
msbrogli: You're right! Thanks for the explanation.

src/websocket/base.ts
msbrogli: You're right! Thanks for clarifying.

msbrogli (approved):

PR #456, opened 2023-01-16
feat: throwing TxNotFound error when a tx is not found

andreabadesso:
### Acceptance Criteria
- When a tx is not found on the fullnode, we should throw a specific Error so clients can handle it properly

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2472,6 +2472,12 @@ class HathorWallet extends EventEmitter {
     return tx;
   }
 
+  static _txNotFoundGuard(data) {
+    if (get(data, 'message', '') === 'Transaction not found') {
+      throw new TxNotFoundError();
+    }
+  }
+
   /**
    * Queries the fullnode for a transaction
    *
@@ -2489,9 +2495,7 @@ class HathorWallet extends EventEmitter {
     });
 
     if (!tx.success) {
-      if (get(tx, 'message', '') === 'Transaction not found') {
-        throw new TxNotFoundError();
-      }
+      HathorWallet._txNotFoundGuard(tx);
 
       throw new Error(`Invalid transaction ${txId}`);
     }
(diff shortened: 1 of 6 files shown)


src/wallet/api/walletApi.ts
```diff
@@ -247,6 +249,11 @@ const walletApi = {
       return response.data;
     }
 
+    const message = get(response, 'data.message', '');
+    if (message === 'Transaction not found') {
```
pedroferreira1, line 253: Maybe create a method to do this validation, or at least save the string in a variable? In case we change the error message in the future we would need to change in 3 places now.

src/new/wallet.js
```diff
@@ -2488,6 +2489,10 @@ class HathorWallet extends EventEmitter {
     });
 
     if (!tx.success) {
+      if (get(tx, 'message', '') === 'Transaction not found') {
```
pedroferreira1, line 2492: The same as in the other comment, we should have a method for that. In case we want to improve/change this validation somehow in the future.

src/wallet/api/walletApi.ts
andreabadesso: Created a guard method in the last commits, can you check?

src/new/wallet.js
andreabadesso: Created a guard method in the last commits, can you check?

src/wallet/api/walletApi.ts
```diff
@@ -237,6 +239,14 @@ const walletApi = {
     });
   },
 
+  _txNotFoundGuard(data: unknown) {
```
r4mmer, line 242: Maybe this should also be a static like in `HathorWallet` or move this to the error file like `WalletFromXPubGuard`

andreabadesso: It can't be static because `walletApi` is an object and `WalletFromXPubGuard` is an error type, right?

pedroferreira1 (approved):

PR #461, opened 2023-01-18
feat: implement getTxById new wallet facade

alexruzenhack:
### Acceptance Criteria
- Implement getTxById in the "new" wallet facade to match the wallet facade for wallet-service

### Security Checklist

### DIFF
--- a/__tests__/hathorwallet.test.js
+++ b/__tests__/hathorwallet.test.js
@@ -804,7 +804,7 @@ test('getTxById', async () => {
   // happy path
   hWallet.getFullHistory.mockReturnValue(buildValidFullHistory());
   hWallet.getTokenDetails.mockImplementation(localTokenDetails);
-  hWallet.isAddressMine.mockImplementation(localIsAddressMine);
+  hWallet.getTxBalance.mockReturnValue({ '00': 1673, 'A': 329 });
   // act
   await expect(hWallet.getTxById(txId)).resolves.toStrictEqual([
     {
@@ -816,7 +816,7 @@ test('getTxById', async () => {
       tokenId: '00',
       tokenName: 'Hathor',
       tokenSymbol: 'HTR',
-      balance: 942
+      balance: 1673
     },
     {
       txId: txId,
@@ -845,22 +845,13 @@ test('getTxById', async () => {
 
   // none of the address is mine
   hWallet.getFullHistory.mockReturnValue(buildValidFullHistory());
-  hWallet.isAddressMine.mockImplementation((_addr) => false)
+  hWallet.getTxBalance.mockReturnValue({});
   // act
   await expect(hWallet.getTxById(txId)).rejects.toThrow(`Transaction ${txId} does not have any balance for this wallet`);
 
   // error while getting token details
   hWallet.getFullHistory.mockReturnValue(buildValidFullHistory());
-  hWallet.isAddressMine.mockImplementation(localIsAddressMine);
-  hWallet.getTokenDetails.mockImplementation((_token) => {
-    throw new Error('Error while getting token info');
-  });
-  // act
-  await expect(hWallet.getTxById(txId)).rejects.toThrow('Error getting token details');
-
-  // error while getting token details
-  hWallet.getFullHistory.mockReturnValue(buildValidFullHistory());
-  hWallet.isAddressMine.mockImplementation(localIsAddressMine);
+  hWallet.getTxBalance.mockReturnValue({ '00': 1673, 'A': 329 });
   hWallet.getTokenDetails.mockImplementation((_token) => {
     throw new Error('Error while getting token info');
   });
--- a/__tests__/hathorwallet.test.js
+++ b/__tests__/hathorwallet.test.js
@@ -19,6 +19,7 @@ import { HDPrivateKey, crypto, PublicKey } from 'bitcore-lib';
 import transaction from '../src/transaction';
 import { Storage } from '../src/storage';
 import Queue from '../src/models/queue';
+import { FullNodeTxResponse } from '../src/wallet/types';
 
 class FakeHathorWallet {
   constructor() {
@@ -742,13 +743,12 @@ test('getTxById', async () => {
   const mineAddress = 'addr1';
   const notMineAddress = 'addr2';
 
-  const buildValidFullHistory = () => ({
-    [txId]: {
-      tx_id: txId,
+  const buildValidFullHistory = (_txId) => ({
+    tx: {
+      timestamp: 1,
       version: 1,
       weight: 1,
-      timestamp: 1,
-      is_voided: false,
+      parents: [],
       inputs: [],
       outputs: [
         {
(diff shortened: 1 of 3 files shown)


src/new/wallet.js
```diff
@@ -2565,6 +2565,101 @@ class HathorWallet extends EventEmitter {
...
+   */
+  async getTxById(txId) {
+    /**
+     * This function is responsible for calculate the balance of each token in the transaction.
+     * that belongs to the wallet.
+     * @param {{ txId: string, timestamp: number, voided: boolean }} tx
+     * @returns {{ [tokenUid]: number }} Object with the token uid as key and the balance as value
+     */
+    const getTokensBalance = (tx) => {
+      const mineOutputs = tx.outputs.filter((output) => this.isAddressMine(output.decoded.address));
```
andreabadesso, line 2597:
```
const myOutputs = tx.outputs.filter((output) => this.isAddressMine(output.decoded.address));
```

src/new/wallet.js
```diff
@@ -2565,6 +2565,101 @@ class HathorWallet extends EventEmitter {
...
+      const tokensBalance = mineOutputs.reduce((acc, output) => {
```
andreabadesso, line 2598:
```suggestion
      const myOutputs = tx.outputs.filter((output) => this.isAddressMine(output.decoded.address));
      const tokensBalance = myOutputs.reduce((acc, output) => {
```

src/new/wallet.js
```diff
@@ -2565,6 +2565,109 @@ class HathorWallet extends EventEmitter {
...
+   *   tokenId: '00',
+   *   tokenName: 'Hathor',
+   *   tokenSymbol: 'HTR',
+   *   balance: 100,
+   *  },
+   * ]
+   */
+  async getTxById(txId) {
+    /**
+     * This function is responsible for calculate the balance of each token in the transaction.
```
andreabadesso, line 2591:
```suggestion
     * This function is responsible for calculating the balance of each token in the transaction.
```

src/new/wallet.js
```diff
@@ -2565,6 +2565,109 @@ class HathorWallet extends EventEmitter {
...
+     * that belongs to the wallet.
+     * @param {{ txId: string, timestamp: number, voided: boolean }} tx
+     * @returns {{ [tokenUid]: number }} Object with the token uid as key and the balance as value
+     */
+    const getTokensBalance = (tx) => {
+      const mineOutputs = tx.outputs.filter((output) => this.isAddressMine(output.decoded.address));
+      const tokensBalance = mineOutputs.reduce((acc, output) => {
+        const { token, value } = output;
+        if (acc[token]) {
```
andreabadesso, line 2600: Maybe do something like

```typescript
if (token in acc) {
```

So this does not gets evaluated as `false` if `acc[token]` is 0

andreabadesso (changes requested):

src/new/wallet.js
alexruzenhack: I removed this function.

src/new/wallet.js
alexruzenhack: I removed this function.

src/new/wallet.js
alexruzenhack: I removed this function.

src/new/wallet.js
```diff
@@ -2565,6 +2565,91 @@ class HathorWallet extends EventEmitter {
...
+     *   tokenSymbol: 'HTR',
+     *   balance: 100,
+     *  },
+     * ]
+     */
+    const getTokenDetails = async (wallet, tx, tokenBalances) => {
+      const tokenUids = Object.keys(tokenBalances);
+      try {
+        const tokenDetails = await Promise.all(
+          tokenUids.map(async (token) => {
```
andreabadesso, line 2614: This will attempt to download all `tokenDetails` in parallel, right? What happens if there are 255 different tokens? Should we limit the number of parallel downloads to something like, 5?

Not sure if this is a problem though

src/new/wallet.js
```diff
@@ -2565,6 +2565,91 @@ class HathorWallet extends EventEmitter {
...
+     *   voided: false;
+     *   weight: 18.5;
+     *   tokenId: '00',
+     *   tokenName: 'Hathor',
+     *   tokenSymbol: 'HTR',
+     *   balance: 100,
+     *  },
+     * ]
+     */
+    const getTokenDetails = async (wallet, tx, tokenBalances) => {
```
andreabadesso, line 2610: Maybe rename it to `getTxTokensDetails`

andreabadesso (changes requested):

src/new/wallet.js
pedroferreira1: The transactions have at most 16 tokens but this is still a problem in my opinion.

I have a suggestion to improve this method but I feel we could talk first to understand one situation

src/new/wallet.js
```diff
@@ -2565,6 +2565,91 @@ class HathorWallet extends EventEmitter {
 
     return graphvizData;
   }
+
+  /**
+   * This function is responsible for getting the details of each token in the transaction.
+   * @param {Object} wallet the current wallet
+   * @param {{ [tokenUid]: number }} tokenBalances Object with the token uid as key and the balance as value
```
pedroferreira1, line 2572: These two parameters are wrong.

__tests__/hathorwallet.test.js
```diff
@@ -734,3 +734,127 @@ test('getTokenHistoryObject', () => {
   expect(historyObj.balance['00']).toStrictEqual(0);
   expect(historyObj.balance['01']).toStrictEqual(5);
 });
+
```
pedroferreira1, line 737: It would be good to also have an integration test with this method.

src/new/wallet.js
alexruzenhack: Removed.

src/new/wallet.js
alexruzenhack: Removed.

src/new/wallet.js
```diff
@@ -2565,6 +2565,72 @@ class HathorWallet extends EventEmitter {
...
+    const { length: hasBalance } = Object.keys(tokenBalances);
+    if (!hasBalance) {
+      throw new Error(`Transaction ${txId} does not have any balance for this wallet`);
+    }
+
+    const listTokenUid = Object.keys(tokenBalances);
+    const result = listTokenUid.map((uid) => {
+      const getToken = (tokenUid) => {
+        const token = fullTx.tx.tokens.find((token) => token.uid === tokenUid)
+        return token || { uid: '00', name: 'Hathor', symbol: 'HTR' };
```
pedroferreira1, line 2611: I think we could add a protection here in case we have any problems with the API response, maybe something like this, what do you think?

```suggestion
        if (tokenUid === HATHOR_TOKEN_CONFIG.uid) {
          return HATHOR_TOKEN_CONFIG;
        }
        
        const token = fullTx.tx.tokens.find((token) => token.uid === tokenUid)
        if (!token) {
          throw Error('Token not found in tx');
        }
        
        return token;
```

pedroferreira1 (approved):

andreabadesso (approved):

PR #462, opened 2023-01-24
feat: persistent storage

r4mmer:
### Acceptance Criteria
- Implement a store of data using LevelDB
- Implement reverse search indexes to comply with the `IStore` interface
- Deprecate node v8 and v10

### Security Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -4,7 +4,6 @@ on:
     branches:
       - dev
       - master
-      - feat/new-storage-scheme
 
 env:
   TEST_WALLET_START_TIMEOUT: '180000' # 3 minutes
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - dev
       - master
+      - feat/new-storage-scheme
 
 env:
   TEST_WALLET_START_TIMEOUT: '180000' # 3 minutes
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,6 @@ on:
     branches:
     - dev
     - master
-    - feat/new-storage-scheme
 jobs:
   test:
     runs-on: 'ubuntu-latest'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,7 @@ on:
     branches:
     - dev
     - master
+    - feat/new-storage-scheme
 jobs:
   test:
     runs-on: 'ubuntu-latest'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     timeout-minutes: 40  # default is 360
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     timeout-minutes: 40  # default is 360
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [12.x, 14.x]
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
--- a/__tests__/integration/storage/leveldb.test.js
+++ b/__tests__/integration/storage/leveldb.test.js
@@ -35,7 +35,7 @@ async function stopWallets() {
   }
 }
 
-describe("LevelDB persistent store", () => {
+describe('LevelDB persistent store', () => {
   afterEach(async () => {
     await stopWallets();
     await stopAllWallets();
@@ -46,7 +46,6 @@ describe("LevelDB persistent store", () => {
     const DATA_DIR = './testdata.leveldb';
     const walletData = precalculationHelpers.test.getPrecalculatedWallet();
     const xpubkey = walletUtils.getXPubKeyFromSeed(walletData.words, { accountDerivationIndex: '0\'/0' });
-
     const store = new LevelDBStore(DATA_DIR, xpubkey);
     const storage = new Storage(store);
 
--- a/__tests__/integration/storage/leveldb.test.js
+++ b/__tests__/integration/storage/leveldb.test.js
@@ -1,0 +1,83 @@
+import { precalculationHelpers } from '../helpers/wallet-precalculation.helper';
+import { GenesisWalletHelper } from '../helpers/genesis-wallet.helper';
+import {
+  createTokenHelper,
+  DEFAULT_PASSWORD,
+  DEFAULT_PIN_CODE,
+  generateConnection,
+  generateMultisigWalletHelper,
+  generateWalletHelper,
+  stopAllWallets,
+  waitForTxReceived,
+  waitForWalletReady,
+  waitUntilNextTimestamp
+} from '../helpers/wallet.helper';
+import { delay } from '../utils/core.util';
+
+import { LevelDBStore, Storage } from '../../../src/storage';
+import walletUtils from '../../../src/utils/wallet';
+import HathorWallet from '../../../src/new/wallet';
+import { loggers } from '../utils/logger.util';
+
+const startedWallets = [];
+
+/**
+ * Helper to stop wallets started manually in this test file.
+ */
+async function stopWallets() {
+  let hWallet;
+  while (hWallet = startedWallets.pop()) {
+    try {
+      await hWallet.stop({ cleanStorage: true, cleanAddresses: true });
+    } catch (e) {
+      loggers.test.error(e.stack);
+    }
+  }
+}
+
+describe("LevelDB persistent store", () => {
+  afterEach(async () => {
+    await stopWallets();
+    await stopAllWallets();
+    await GenesisWalletHelper.clearListeners();
+  });
+
+  it('should receive and send tokens', async () => {
+    const DATA_DIR = './testdata.leveldb';
+    const walletData = precalculationHelpers.test.getPrecalculatedWallet();
+    const xpubkey = walletUtils.getXPubKeyFromSeed(walletData.words, { accountDerivationIndex: '0\'/0' });
+
+    const store = new LevelDBStore(DATA_DIR, xpubkey);
+    const storage = new Storage(store);
+
+    // Start the wallet
+    const walletConfig = {
+      seed: walletData.words,
+      connection: generateConnection(),
+      password: DEFAULT_PASSWORD,
+      pinCode: DEFAULT_PIN_CODE,
+      preCalculatedAddresses: walletData.addresses,
+      storage,
+    };
+    const hWallet = new HathorWallet(walletConfig);
+    await hWallet.start();
+    startedWallets.push(hWallet);
+    await waitForWalletReady(hWallet);
+
+    // Expect to have an empty list for the full history
+    expect(Object.keys(hWallet.getFullHistory())).toHaveLength(0);
+
+    // Injecting some funds on this wallet
+    await GenesisWalletHelper.injectFunds(await hWallet.getAddressAtIndex(1), 10);
+
+    await delay(100);
+    // Validating the full history increased in one
+    expect(Object.keys(await hWallet.getFullHistory())).toHaveLength(1);
+
+    const tx1 = await hWallet.sendTransaction(await hWallet.getAddressAtIndex(3), 5);
+    await waitForTxReceived(hWallet, tx1.hash);
+    expect(Object.keys(await hWallet.getFullHistory())).toHaveLength(2);
+
+    await expect(store.getTx(tx1.hash)).resolves.not.toBeNull();
+  });
+});
--- a/__tests__/storage/leveldb/leveldb_store.test.js
+++ b/__tests__/storage/leveldb/leveldb_store.test.js
@@ -46,7 +46,6 @@ test('addresses methods', async () => {
   // The order should follow the index ordering
   expect(values).toStrictEqual(['b', 'a', 'c']);
 
-
   await expect(store.getAddress('b')).resolves.toStrictEqual({ base58: 'b', bip32AddressIndex: 1 });
   await expect(store.getAddressMeta('b')).resolves.toMatchObject({ numTransactions: 5 });
   await expect(store.addressCount()).resolves.toEqual(3);
--- a/__tests__/storage/leveldb/leveldb_store.test.js
+++ b/__tests__/storage/leveldb/leveldb_store.test.js
@@ -12,6 +12,7 @@ import walletApi from "../../../src/api/wallet";
 import { HDPrivateKey } from "bitcore-lib";
 import { encryptData } from "../../../src/utils/crypto";
 import { WalletType } from "../../../src/types";
+import { processHistory } from "../../../src/utils/storage";
 
 function _addr_index_key(index) {
   const buf = Buffer.alloc(4);
@@ -117,7 +118,7 @@ test('history methods', async () => {
       symbol: 'CTK',
     });
   });
-  await store.processHistory({ rewardLock: 1 });
+  await processHistory(store, { rewardLock: 1 });
   expect(getTokenApi).not.toHaveBeenCalledWith('00', expect.anything());
   expect(getTokenApi).toHaveBeenCalledWith('01', expect.anything());
   expect(getTokenApi).toHaveBeenCalledWith('02', expect.anything());
@@ -223,17 +224,12 @@ test('token methods', async () => {
   }
   expect(registered).toHaveLength(1);
 
-  await store.deleteTokens(['01', '02']);
-  count = 0;
-  for await (let _ of store.tokenIndex.tokenDB.iterator()) {
-    count += 1;
-  }
-  expect(count).toEqual(1);
-
-  await store.editToken('00', { numTransactions: 10 });
+  await store.editToken('00', { numTransactions: 10, balance: { tokens: { locked: 1, unlocked: 2 } } });
   await expect(store.tokenIndex.getTokenMetadata('00')).resolves.toMatchObject({
     numTransactions: 10,
-    balance: expect.anything(),
+    balance: {
+      tokens: { locked: 1, unlocked: 2 },
+    },
   });
 
   await store.destroy();
--- a/__tests__/storage/leveldb/leveldb_store.test.js
+++ b/__tests__/storage/leveldb/leveldb_store.test.js
@@ -5,62 +5,68 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { GAP_LIMIT } from '../../src/constants';
-import { MemoryStore } from '../../src/storage';
-import tx_history from '../__fixtures__/tx_history';
-import walletApi from '../../src/api/wallet';
-import { HDPrivateKey } from 'bitcore-lib';
-import { encryptData } from '../../src/utils/crypto';
-import { WalletType } from '../../src/types';
-import { processHistory } from '../../src/utils/storage';
+import { GAP_LIMIT, HATHOR_TOKEN_CONFIG } from "../../../src/constants";
+import { LevelDBStore } from "../../../src/storage";
+import tx_history from "../../__fixtures__/tx_history";
+import walletApi from "../../../src/api/wallet";
+import { HDPrivateKey } from "bitcore-lib";
+import { encryptData } from "../../../src/utils/crypto";
+import { WalletType } from "../../../src/types";
 
+function _addr_index_key(index) {
+  const buf = Buffer.alloc(4);
+  buf.writeUint32BE(index);
+  return buf.toString('hex');
+}
 
-test('default values', async () => {
-  const store = new MemoryStore();
-  expect(store.walletData).toMatchObject({
-    lastLoadedAddressIndex: 0,
-    lastUsedAddressIndex: -1,
-    currentAddressIndex: -1,
-    bestBlockHeight: 0,
-    gapLimit: GAP_LIMIT,
-  });
-});
+const DATA_DIR = './data.leveldb';
 
 test('addresses methods', async () => {
-  const store = new MemoryStore();
-  store.addresses.set('a', 2);
-  store.addressIndexes.set(2, 'a');
-  store.addresses.set('b', 1);
-  store.addressIndexes.set(1, 'b');
-  store.addresses.set('c', 3);
-  store.addressIndexes.set(3, 'c');
-  store.addressesMetadata.set('a', 6);
-  store.addressesMetadata.set('b', 5);
+  const xpriv = HDPrivateKey();
+  const store = new LevelDBStore(DATA_DIR, xpriv.xpubkey);
+  const addrBatch = store.addressIndex.addressesDB.batch();
+  addrBatch.put('a', { base58: 'a', bip32AddressIndex: 2 });
+  addrBatch.put('b', { base58: 'b', bip32AddressIndex: 1 });
+  addrBatch.put('c', { base58: 'c', bip32AddressIndex: 3 });
+  await addrBatch.write();
+  const indexBatch = store.addressIndex.addressesIndexDB.batch();
+  indexBatch.put(_addr_index_key(2), 'a');
+  indexBatch.put(_addr_index_key(1), 'b');
+  indexBatch.put(_addr_index_key(3), 'c');
+  await indexBatch.write();
+  const metaBatch = store.addressIndex.addressesMetaDB.batch();
+  metaBatch.put('a', {numTransactions: 6, balance: {}});
+  metaBatch.put('b', {numTransactions: 5, balance: {}});
+  await metaBatch.write();
 
   const values = [];
   for await (const info of store.addressIter()) {
-    values.push(info);
+    values.push(info.base58);
   }
-  expect(values).toStrictEqual([2, 1, 3]);
-  await expect(store.getAddress('b')).resolves.toEqual(1);
-  await expect(store.getAddressMeta('b')).resolves.toEqual(5);
+  // The order should follow the index ordering
+  expect(values).toStrictEqual(['b', 'a', 'c']);
+
+
+  await expect(store.getAddress('b')).resolves.toStrictEqual({ base58: 'b', bip32AddressIndex: 1 });
+  await expect(store.getAddressMeta('b')).resolves.toMatchObject({ numTransactions: 5 });
   await expect(store.addressCount()).resolves.toEqual(3);
+  await expect(store.getAddressAtIndex(3)).resolves.toStrictEqual({ base58: 'c', bip32AddressIndex: 3 });
 
   await expect(store.getCurrentAddress()).rejects.toThrow('Current address is not loaded');
-  expect(store.walletData.currentAddressIndex).toEqual(-1);
-  expect(store.walletData.lastLoadedAddressIndex).toEqual(0);
+  await expect(store.walletIndex.getCurrentAddressIndex()).resolves.toEqual(-1);
+  await expect(store.walletIndex.getLastLoadedAddressIndex()).resolves.toEqual(0);
   await store.saveAddress({
     base58: 'd',
     bip32AddressIndex: 10,
   });
-  expect(store.walletData.currentAddressIndex).toEqual(10);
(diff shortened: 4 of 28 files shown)


.github/workflows/integration-test.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - dev
       - master
+      - feat/new-storage-scheme
```
r4mmer, line 7: This is to make the CI run on the PR, it will be removed before being merged.

.github/workflows/main.yml
```diff
@@ -10,13 +10,14 @@ on:
     branches:
     - dev
     - master
+    - feat/new-storage-scheme
```
r4mmer, line 13: This is to make the CI run on the PR, it will be removed before being merged.

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+    for (const [uid, balance] of meta.balance.entries()) {
+      dbMeta.balance[uid] = balance;
+    }
+    await this.addressesMetaDB.put(address, dbMeta);
+  }
+
+  async getAddressMeta(base58: string): Promise<IAddressMetadata | null> {
+    try {
+      const dbmeta = await this.addressesMetaDB.get(base58);
+      const meta: IAddressMetadata = {numTransactions: dbmeta.numTransactions, balance: new Map<string, IBalance>()};
```
andreabadesso, line 136:
```suggestion
      const meta: IAddressMetadata = {
        numTransactions: dbmeta.numTransactions,
        balance: new Map<string, IBalance>(),
      };
```

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+      for (const [uid, balance] of Object.entries(dbmeta.balance)) {
+        meta.balance.set(uid, balance);
+      }
+      return meta;
+    } catch(err: unknown) {
```
andreabadesso, line 141:
```suggestion
    } catch (err: unknown) {
```

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  async getAddressAtIndex(index: number): Promise<string|null> {
+    if (index < 0) {
+      return null;
+    }
+    try {
+      return await this.addressesIndexDB.get(_index_key(index));
+    } catch(err: unknown) {
```
andreabadesso, line 155:
```suggestion
    } catch (err: unknown) {
```

src/storage/leveldb/errors.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export const KEY_NOT_FOUND_MESSAGE = 'NotFound';
+export const KEY_NOT_FOUND_CODE = 'LEVEL_NOT_FOUND';
+
+export function errorCodeOrNull(err: unknown): string|null {
+  if(typeof err === 'object' && err !== null && 'code' in err && err.code !== undefined) {
```
andreabadesso, line 12:
```suggestion
  if (typeof err === 'object'
      && err !== null
      && 'code' in err
      && err.code !== undefined) {
```

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,149 @@
...
+      if (key !== value.tx_id) {
+        throw new Error('Inconsistent database');
+      }
+
+      // for (const io of [...value.inputs, ...value.outputs]) {
+      //   if (io.decoded.address) {
+      //     addresses.add(io.decoded.address);
+      //   }
+      //   tokens.add(io.token);
+      // }
```
andreabadesso, line 82: Remove?

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      try {
+        await this.tsHistoryDB.get(`${value.timestamp}:${value.tx_id}`);
+      } catch(err: unknown) {
```
andreabadesso, line 86:
```suggestion
      } catch (err: unknown) {
```

src/storage/leveldb/store.ts
```diff
@@ -0,0 +1,501 @@
...
+
+  async destroy(): Promise<void> {
+    await this.close();
+    if (process.version < 'v14.14.0') {
+      // rmSync was introduced in v14.14.0, earlier versions should use rmdirSync
+      // https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fs_rmdirsync_path_options
+      fs.rmdirSync(this.dbpath, { recursive: true });
+    } else {
+      fs.rmSync(this.dbpath, { recursive: true, force: true });
+    }
```
andreabadesso, line 60: This method already returns a promise, why do we need to use the sync methods?

This will only cause clients using our lib and calling destroy() to be slower as it will block the event loop

src/storage/leveldb/store.ts
```diff
@@ -0,0 +1,501 @@
...
+    const address = await this.addressIndex.getAddressAtIndex(index);
+    if (address === null) {
+      return null;
+    }
+    return this.addressIndex.getAddressInfo(address);
+  }
+
+  async saveAddress(info: IAddressInfo): Promise<void> {
+    if (!info.base58) {
+      throw new Error('Invalid address');
```
andreabadesso, line 99: Should we have standard errors like the ones in errors.ts so they can be better handled?

e.g.

`class InvalidAddressError extends Error`

src/storage/leveldb/store.ts
```diff
@@ -0,0 +1,501 @@
...
+    for await (const tx of this.historyIndex.historyIter(tokenUid)) {
+      yield tx;
+    }
+  }
+
+  async historyCount(): Promise<number> {
+    return this.historyIndex.historyCount();
+  }
+
+  async processHistory({ rewardLock }: {rewardLock?: number} = {}): Promise<void> {
```
andreabadesso, line 146: This is a huge method, do you think we can (and should) split it into smaller methods?

src/storage/leveldb/utxo_index.ts
```diff
@@ -0,0 +1,223 @@
...
+    await this.tokenAddressUtxoDB.put(_token_address_utxo_key(utxo), utxo);
+    await this.tokenUtxoDB.put(_token_utxo_key(utxo), utxo);
+  }
+
+  async clear(): Promise<void> {
+    // This should clear all utxos subdbs
+    await this.utxoDB.db.clear();
+    // await this.utxoDB.clear();
+    // await this.tokenUtxoDB.clear();
+    // await this.tokenAddressUtxoDB.clear();
```
andreabadesso, line 221: Don't we need this anymore? Why is it commented out? Remove?

andreabadesso (changes requested):

src/storage/leveldb/store.ts
r4mmer: This is a huge method indeed, but it should not be split into smaller methods since it essentially is a single loop on the transaction history making aggregations and calculating metadata.

src/storage/leveldb/utxo_index.ts
r4mmer: Removed

src/storage/leveldb/store.ts
r4mmer: Started using async version or `rm`

src/storage/leveldb/history_index.ts
r4mmer: Done

src/storage/leveldb/history_index.ts
r4mmer: Removed

src/storage/leveldb/errors.ts
r4mmer: Done

src/storage/leveldb/address_index.ts
r4mmer: Done

src/storage/leveldb/address_index.ts
r4mmer: Done

src/storage/leveldb/address_index.ts
r4mmer: Done

src/api/metadataApi.ts
```diff
@@ -23,9 +24,9 @@ const metadataApi = {
 
     let { retries, retryInterval } = newOptions;
     while (retries >= 0) {
-      const axios = await explorerServiceAxios(network);
+      const client = await explorerServiceAxios(network);
       try {
-        const response = await axios.get(`metadata/dag`, { params: { id }});
+        const response = await client.get(`metadata/dag`, { params: { id }});
```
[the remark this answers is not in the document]
r4mmer: Done

index.js
```diff
@@ -9,7 +9,9 @@ const txMiningApi = require('./lib/api/txMining');
 const versionApi = require('./lib/api/version');
 const axios = require('./lib/api/axiosInstance');
 const metadataApi = require('./lib/api/metadataApi');
-const storage = require('./lib/storage');
+const storage = require('./lib/storage/storage');
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/storage/leveldb/leveldb_store.test.js
```diff
@@ -0,0 +1,362 @@
...
+  metaBatch.put('b', {numTransactions: 5, balance: {}});
+  await metaBatch.write();
+
+  const values = [];
+  for await (const info of store.addressIter()) {
+    values.push(info.base58);
+  }
+  // The order should follow the index ordering
+  expect(values).toStrictEqual(['b', 'a', 'c']);
+
```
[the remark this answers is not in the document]
r4mmer: Removed

__tests__/integration/storage/leveldb.test.js
```diff
@@ -0,0 +1,83 @@
...
+  while (hWallet = startedWallets.pop()) {
+    try {
+      await hWallet.stop({ cleanStorage: true, cleanAddresses: true });
+    } catch (e) {
+      loggers.test.error(e.stack);
+    }
+  }
+}
+
+describe("LevelDB persistent store", () => {
```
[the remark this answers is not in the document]
r4mmer: Done

src/storage/leveldb/utxo_index.ts
```diff
@@ -0,0 +1,220 @@
...
+    const token = options.token || HATHOR_TOKEN_CONFIG.uid;
+    const authorities = options.authorities || 0;
+    const maxUtxos = options.max_utxos || MAX_INPUTS;
+
+    let db: typeof this.utxoDB;
+    const itOptions: ValueIteratorOptions<string, IUtxo> = {};
+    if (options.filter_address !== undefined) {
+      // Use tokenAddressUtxoDB
+      db = this.tokenAddressUtxoDB;
+      let minkey = `${authorities}:${token}:${options.filter_address}:`;
```
pedroferreira1, line 143: Having the it key ending in `:` is ok?

src/storage/leveldb/store.ts
r4mmer: I also believe we should create a standard error, but since it would affect both the persistent storage files and the base storage scheme implementation I think we should make this on its own PR, what do you think?

src/storage/leveldb/wallet_index.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.accessDB = db.sublevel<string, IWalletAccessData>(ACCESS_PREFIX, { valueEncoding: 'json' });
+    this.walletDB = db.sublevel(WALLET_PREFIX);
+    this.genericDB = db.sublevel<string, any>(GENERIC_PREFIX, { valueEncoding: 'json' });
+  }
+
+  async close(): Promise<void> {
+    await this.accessDB.db.close();
+  }
+
+  async _setNumber(dest: 'access'|'wallet'|'generic', key: string, value: number) {
```
pedroferreira1, line 38: Would be great to have docstring to better understand this method with some examples. Actually all the methods in all the leveldb indexes should have docstrings

src/storage/leveldb/wallet_index.ts
```diff
@@ -0,0 +1,210 @@
...
+    buf.writeUInt32BE(value);
+    switch(dest) {
+      case 'access':
+        await this.accessDB.put<string, Buffer>(key, buf, { valueEncoding: 'buffer'});
+        break;
+      case 'wallet':
+        await this.walletDB.put<string, Buffer>(key, buf, { valueEncoding: 'buffer'});
+        break;
+      case 'generic':
+        await this.walletDB.put<string, Buffer>(key, buf, { valueEncoding: 'buffer'});
```
pedroferreira1, line 49:
```suggestion
        await this.genericDB.put<string, Buffer>(key, buf, { valueEncoding: 'buffer'});
```

Is this a bug? We should add a test for this.

src/storage/leveldb/wallet_index.ts
```diff
@@ -0,0 +1,210 @@
...
+      let buf: Buffer;
+      switch(dest) {
+        case 'access':
+          buf = await this.accessDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
+          break;
+        case 'wallet':
+          buf = await this.walletDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
+          break;
+        case 'generic':
+          buf = await this.walletDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
```
pedroferreira1, line 65:
```suggestion
          buf = await this.genericDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
```

Is this a bug? We should add a test for this.

setupTests.js
```diff
@@ -72,4 +72,9 @@ expect.extend({
 });
 
 
-global.window = {};
+// global.window = {};
```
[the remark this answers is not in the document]
r4mmer: It is already removed.

src/storage/leveldb/store.ts
```diff
@@ -0,0 +1,501 @@
...
+import LevelUtxoIndex from './utxo_index';
+import LevelWalletIndex from './wallet_index';
+import LevelTokenIndex from './token_index';
+import walletApi from '../../api/wallet';
+import transaction from '../../utils/transaction';
+import fs from 'fs';
+import { HATHOR_TOKEN_CONFIG } from '../../constants';
+
+export default class LevelDBStore implements IStore {
+  addressIndex: LevelAddressIndex;
```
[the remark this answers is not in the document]
r4mmer: Maybe, having access to the index or even the lower level db helps with testing.

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+      throw err;
+    }
+  }
+
+  async addressExists(base58: string): Promise<boolean> {
+    return (await this.getAddressInfo(base58)) !== null;
+  }
+
+  async * addressIter(): AsyncGenerator<IAddressInfo> {
+    for await (const address of this.addressesIndexDB.values()) {
```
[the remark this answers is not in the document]
r4mmer: Because of how leveldb handles iteration, all entries are sorted by the key value.
Which means that when I iterate on the `addressIndexDB` the order of the returned addresses will be the same as the order of the keys of `addressIndexDB` and since we use the uint32 (in hex format) the addresses will returned in order of the bip32 address index

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+      }
+      yield info;
+    }
+  }
+
+  async setAddressMeta(address: string, meta: IAddressMetadata): Promise<void> {
+    const dbMeta: IAddressMetadataAsRecord = {numTransactions: meta.numTransactions, balance: {}};
+    for (const [uid, balance] of meta.balance.entries()) {
+      dbMeta.balance[uid] = balance;
+    }
```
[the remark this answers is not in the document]
r4mmer: `IAddressMetadata` is defined as:

```ts
export interface IAddressMetadata {
  numTransactions: number;
  balance: Map<string, IBalance>;
}
```
Unfortunately, `Map` is not supported by the json encoding lib of leveldb, so I was required to create a  `IAddressMetadataAsRecord` which uses `Record<string, IBalance>` instead of `Map<string, IBalance>` and I have to change between these types manually when required.

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,142 @@
...
+  historyDB: AbstractSublevel<Level, string | Buffer | Uint8Array, string, IHistoryTx>;
+  tsHistoryDB: AbstractSublevel<Level, string | Buffer | Uint8Array, string, IHistoryTx>;
+  size: number;
+  indexVersion: string = '0.0.1';
+
+  constructor(dbpath: string) {
+    this.dbpath = path.join(dbpath, 'history');
+    const db = new Level(this.dbpath);
+    this.historyDB = db.sublevel<string, IHistoryTx>(HISTORY_PREFIX, { valueEncoding: 'json' });
+    this.tsHistoryDB = db.sublevel<string, IHistoryTx>(TS_HISTORY_PREFIX, { valueEncoding: 'json' });
```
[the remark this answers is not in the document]
r4mmer: It is not needed, this index is used as an iteration index.
With it we can iterate on all transactions ordered by the timestamp.

src/storage/leveldb/utxo_index.ts
r4mmer: Yes, since iteration is sorted on keys.
You can see below that if we don't add a value filter for the amount (`options.amount_smaller_than`) we will calculate the `maxaddr` which is the `options.filter_address` but with the last char 1 higher than the address we are filtering, this actually gives us all keys with the address as a prefix.

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,142 @@
...
+        if (errorCodeOrNull(err) === KEY_NOT_FOUND_CODE) {
+          // Create if index is missing
+          await this.tsHistoryDB.put(_ts_key(value), value);
+          continue;
+        }
+        throw err;
+      }
+    }
+    ret.addresses = Array.from(addresses);
+    ret.tokens = Array.from(tokens);
```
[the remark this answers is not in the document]
r4mmer: The original idea was to validate the addresses and tokens on storage, but this was not necessary so it was removed, I forgot to clean the type.
In the future we may enforce a more strict validation but it is not necessary now.

src/storage/leveldb/wallet_index.ts
r4mmer: Added docstrings

src/storage/leveldb/wallet_index.ts
r4mmer: It is not a bug because this is not used.
The api defined to access the generic data only uses `getItem` and `setItem` with json encoding.

Removed `generic` from possibilities.

src/storage/leveldb/wallet_index.ts
r4mmer: Same as the one above.

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,142 @@
...
+    await this.checkVersion();
+
+    let ret: HistoryIndexValidateResponse = {
+      count: 0,
+      tokens: [],
+      addresses: [],
+    };
+    const addresses = new Set<string>();
+    const tokens = new Set<string>();
+    // Iterate on all addresses and check that we have a corresponding index entry
```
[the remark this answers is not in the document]
r4mmer: Fixed

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+import { Level } from 'level';
+import { AbstractSublevel } from 'abstract-level';
+import { IAddressInfo, IAddressMetadata, IKVAddressIndex, AddressIndexValidateResponse, IAddressMetadataAsRecord, IBalance } from '../../types';
+import { errorCodeOrNull, KEY_NOT_FOUND_CODE } from './errors';
+
+export const ADDRESS_PREFIX = 'address';
+export const INDEX_PREFIX = 'index';
+export const ADDRESS_META_PREFIX = 'meta';
+
+function _index_key(index: number): string {
```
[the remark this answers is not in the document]
r4mmer: Added docstring

src/storage/memory_store.ts
```diff
@@ -399,6 +403,9 @@ export class MemoryStore implements IStore {
     }
 
     for (const uid of allTokens) {
+      if (uid === HATHOR_TOKEN_CONFIG.uid) {
```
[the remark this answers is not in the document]
r4mmer: Moved to add HTR on the constructor, so it should always be on the list.

src/storage/leveldb/store.ts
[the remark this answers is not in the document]
r4mmer: Although they are similar I think we should keep them separate for now, by keeping them separate we can work closer to the store of data (leveldb) and possibly making it faster or use less memory to process the database.

This can be seen on how we use the database directly to store token metadata, address metadata and utxos.
This was done to avoid a burst in memory usage when we are processing the database, specially on wallets with many addresses and utxos.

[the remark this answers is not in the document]
pedroferreira1: I'm not 100% sure I understand. If we split this method into smaller methods, we could reuse them in both methods and have some specific code for specific store implementantion, or not?

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,142 @@
...
+  tsHistoryDB: AbstractSublevel<Level, string | Buffer | Uint8Array, string, IHistoryTx>;
+  size: number;
+  indexVersion: string = '0.0.1';
+
+  constructor(dbpath: string) {
+    this.dbpath = path.join(dbpath, 'history');
+    const db = new Level(this.dbpath);
+    this.historyDB = db.sublevel<string, IHistoryTx>(HISTORY_PREFIX, { valueEncoding: 'json' });
+    this.tsHistoryDB = db.sublevel<string, IHistoryTx>(TS_HISTORY_PREFIX, { valueEncoding: 'json' });
+    this.size = 0;
```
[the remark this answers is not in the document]
pedroferreira1: Why didn't you continue with this external count?

src/storage/leveldb/history_index.ts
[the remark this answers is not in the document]
pedroferreira1: How do you iterate by timestamp in the `MemoryStore`?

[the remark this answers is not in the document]
r4mmer: Since we use `Map`s in `MemoryStore` to store the history instead of `Record`s the iterator guarantees iteration on order of inserts.
To guarantee timestamp ordering on memory store i was thinking of using an auxiliar [SortedMap](https://www.collectionsjs.com/sorted-map) but this would require a new dependency so I still need to check the implementation of this `SortedMap`.

src/storage/leveldb/history_index.ts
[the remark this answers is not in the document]
r4mmer: There were a couple issues during the initial implementation so i went with the simpler approach but we can still change this since its internal to the index.

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  async saveAddress(info: IAddressInfo): Promise<void> {
+    await this.addressesDB.put(info.base58, info);
+    await this.addressesIndexDB.put(_index_key(info.bip32AddressIndex), info.base58);
+  }
+
+  async addressCount(): Promise<number> {
+    // Level is bad at counting addresses addresses
```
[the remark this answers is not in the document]
r4mmer: | no. of addresses | time in ms to count |
| --- | --- |
| 100 | 1 |
| 1000 | 9 |
| 100k | 434 |
| 1M | 5135 |

src/storage/leveldb/token_index.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  async unregisterToken(tokenUid: string): Promise<void> {
+    await this.registeredDB.del(tokenUid);
+  }
+
+  async deleteTokens(tokens: string[]): Promise<void> {
+    for (const uid of tokens) {
+      await this.tokenDB.del(uid);
+      await this.metadataDB.del(uid);
```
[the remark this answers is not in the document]
r4mmer: Method has been removed from `IStore`

src/storage/leveldb/history_index.ts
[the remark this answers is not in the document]
r4mmer: Added a counter on the database with validation on both address and history indexes

src/storage/leveldb/address_index.ts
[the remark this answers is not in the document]
r4mmer: This has been changed to an external count what will only be used when the index is validated.
Before validation we will use the normal count, after we will trust the counter on storage.

| no. of addresses | count time w/o validation | count time after validation |
| --- | --- | --- |
| 100 | 2 | 0 |
| 1000 | 8  | 0 |
| 10k |  52 | 0 |
| 100k | 370 | 0 |
| 1M | 2669 | 0 |

All times are in ms, I used a simple `Date` call before and after the `addressCount` so the count after validation took less then a ms on all cases.

src/storage/leveldb/store.ts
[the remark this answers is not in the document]
r4mmer: A `processHistory` util was created and it is now used.
Both implementations use the same `processHistory` which only use the methods from `IStore` so it can work on any future `IStore` implementation while also allowing for custom implementation if needed.

src/storage/leveldb/address_index.ts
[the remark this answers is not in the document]
pedroferreira1: The external count is great but I feel some of this discussion here and that we had in a conversation should be in the code as comments.

src/storage/leveldb/history_index.ts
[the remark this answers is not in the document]
pedroferreira1: Have you created this issue so we can handle it as soon as it's merged?

src/storage/leveldb/address_index.ts
```diff
@@ -0,0 +1,341 @@
...
+          continue;
+        }
+        throw err;
+      }
+    }
+
+    // We just did a full address count, we can save the size and trust it
+    const buf = Buffer.alloc(4);
+    buf.writeUInt32BE(size);
+    await this.addressesDB.db.put('size', buf.toString('hex'));
```
pedroferreira1, line 146: This code was added in 3 places and you have a method (`_index_key`) that would help a lot with it. I would change the name of the method to be more generic and use it.

src/storage/leveldb/history_index.ts
```diff
@@ -0,0 +1,239 @@
...
+          continue;
+        }
+        throw err;
+      }
+    }
+
+    // We have validated the index, we can now trust the tx count
+    const buf = Buffer.alloc(4);
+    buf.writeUInt32BE(ret.count);
+    await this.historyDB.db.put('size', buf.toString('hex'));
```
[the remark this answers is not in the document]
r4mmer: This has changed to a counter on memory

src/storage/leveldb/address_index.ts
r4mmer: This has changed to a counter on memory

src/storage/leveldb/history_index.ts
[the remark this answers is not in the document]
r4mmer: The issue #480 was created to track this discussion.

src/storage/leveldb/address_index.ts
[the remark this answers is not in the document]
r4mmer: After changing to saving this counter on memory I added a summary of this discussion on the docstring of `addressCount` and `historyCount` methods

pedroferreira1 (approved):

src/storage/leveldb/history_index.ts
```diff
@@ -132,49 +141,19 @@ export default class LevelHistoryIndex implements IKVHistoryIndex {
...
-    const size = await this.historyCount();
-    const buf = Buffer.alloc(4);
-    buf.writeUInt32BE(size + 1);
-    await db.put('size', buf.toString('hex'));
+    return this.size;
   }
 
   /**
    * Run a full count of the txs in the database.
-   * Since we have made a full count we can update the tx count in the database.
```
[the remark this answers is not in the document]
r4mmer: Removed from the other file

src/storage/leveldb/store.ts
andreabadesso: I agree, can you create an issue so we don't forget to refactor this?

andreabadesso (approved):

PR #464, opened 2023-02-22
feat: create endpoint to the atomic swap service

tuliomir:
Implements the initial code for communication with the Atomic Swap Service.

### Acceptance Criteria
- Hashing and encrypting all data before sending it over to the Atomic Swap Service
- Only the `create` endpoint will be interacted with

### Note
No integration tests will be implemented at this stage, as there is no readily available image of the Atomic Swap Service to be added to the `docker-compose.yml` structure necessary for them.

### Security Checklist

### DIFF
--- a/__tests__/wallet/api/swapService.test.ts
+++ b/__tests__/wallet/api/swapService.test.ts
@@ -1,0 +1,55 @@
+
+import { decryptString, encryptString, hashPassword, create } from '../../../src/wallet/api/swapService'
+import config from '../../../src/config';
+import MockAdapter from 'axios-mock-adapter';
+import axios from 'axios';
+
+const mockAxiosAdapter = new MockAdapter(axios);
+
+describe('hashing and encrypting', () => {
+  it('should correctly hash a password', () => {
+    const password = '123';
+    const hashedPassword = hashPassword(password);
+
+    expect(hashedPassword).toHaveLength(64);
+  })
+
+  it('should correctly encrypt and decrypt a string', () => {
+    const originalString = 'PartialTx|123123||';
+    const password = 'strongPassword';
+
+    const encryptedString = encryptString(originalString, password);
+    expect(encryptedString.length !== originalString.length).toBe(true);
+
+    const decryptedString = decryptString(encryptedString, password);
+    expect(decryptedString).toStrictEqual(originalString);
+  })
+})
+
+describe('create api', () => {
+
+  it('should throw an error if the backend base URL was not configured', async () => {
+    await expect(create('PartialTx||', 'abc'))
+      .rejects.toThrowError('Swap service base URL not set.');
+  })
+
+  it('should handle backend errors', async () => {
+    config.setSwapServiceBaseUrl('http://mock-swap-url/')
+
+    mockAxiosAdapter.onPost('/').reply(503)
+    await expect(create('PartialTx||', 'abc'))
+      .rejects.toThrowError('Request failed with status code 503');
+  })
+
+  it('should return the backend results on a successful post', async () => {
+    config.setSwapServiceBaseUrl('http://mock-swap-url/')
+
+    const responseData = {
+      success: true,
+      id: 'proposal-id-123'
+    };
+    mockAxiosAdapter.onPost('/').reply(200, responseData)
+    await expect(create('PartialTx||', 'abc'))
+      .resolves.toStrictEqual(responseData)
+  })
+})
--- a/__tests__/wallet/api/swapService.test.ts
+++ b/__tests__/wallet/api/swapService.test.ts
@@ -26,12 +26,36 @@ describe('hashing and encrypting', () => {
   })
 })
 
-describe('create api', () => {
-
-  it('should throw an error if the backend base URL was not configured', async () => {
-    await expect(create('PartialTx||', 'abc'))
-      .rejects.toThrowError('Swap service base URL not set.');
+describe('base url configuration', () => {
+  it('should throw when no url parameter was offered', () => {
+      expect(() => config.getSwapServiceBaseUrl())
+        .toThrowError('You should either provide a network or call setSwapServiceBaseUrl before calling this.');
   })
+
+  it('should return mainnet address when requested', () => {
+      expect(config.getSwapServiceBaseUrl('mainnet'))
+        .toStrictEqual('https://atomic-swap-service.hathor.network/')
+  })
+
+  it('should return testnet address when requested', () => {
+      expect(config.getSwapServiceBaseUrl('testnet'))
+        .toContain('testnet')
+  })
+
+  it('should throw when an invalid network is requested', () => {
+    // @ts-ignore
+    expect(() => config.getSwapServiceBaseUrl('invalid'))
+      .toThrowError('You should set it explicitly.');
+  })
+
+  it('should return the specified baseURL when it was set', () => {
+    config.setSwapServiceBaseUrl('http://swap-base-url')
+      expect(config.getSwapServiceBaseUrl())
+        .toStrictEqual('http://swap-base-url')
+  })
+})
+
+describe('create api', () => {
 
   it('should handle backend errors', async () => {
     config.setSwapServiceBaseUrl('http://mock-swap-url/')
--- a/__tests__/wallet/api/swapService.test.ts
+++ b/__tests__/wallet/api/swapService.test.ts
@@ -39,13 +39,13 @@ describe('base url configuration', () => {
 
   it('should return testnet address when requested', () => {
       expect(config.getSwapServiceBaseUrl('testnet'))
-        .toContain('testnet')
+        .toStrictEqual('https://atomic-swap-service.testnet.hathor.network/')
   })
 
   it('should throw when an invalid network is requested', () => {
     // @ts-ignore
     expect(() => config.getSwapServiceBaseUrl('invalid'))
-      .toThrowError('You should set it explicitly.');
+      .toThrowError(`Network invalid doesn't have a correspondent Atomic Swap Service url. You should set it explicitly by calling setSwapServiceBaseUrl.`);
   })
 
   it('should return the specified baseURL when it was set', () => {
@@ -57,11 +57,18 @@ describe('base url configuration', () => {
 
 describe('create api', () => {
 
+  it('should throw missing parameter errors', async () => {
+    // @ts-ignore
+    await expect(create()).rejects.toThrowError('Missing serializedPartialTx');
+    // @ts-ignore
+    await expect(create('PartialTx|0001000000000000000000000063f78c0e0000000000||')).rejects.toThrowError('Missing password');
+  })
+
   it('should handle backend errors', async () => {
     config.setSwapServiceBaseUrl('http://mock-swap-url/')
 
     mockAxiosAdapter.onPost('/').reply(503)
-    await expect(create('PartialTx||', 'abc'))
+    await expect(create('PartialTx|0001000000000000000000000063f78c0e0000000000||', 'abc'))
       .rejects.toThrowError('Request failed with status code 503');
   })
 
@@ -73,7 +80,7 @@ describe('create api', () => {
       id: 'proposal-id-123'
     };
     mockAxiosAdapter.onPost('/').reply(200, responseData)
-    await expect(create('PartialTx||', 'abc'))
+    await expect(create('PartialTx|0001000000000000000000000063f78c0e0000000000||', 'abc'))
       .resolves.toStrictEqual(responseData)
   })
 })
--- a/src/config.ts
+++ b/src/config.ts
@@ -16,7 +16,7 @@ const TX_MINING_TESTNET_URL = 'https://txmining.testnet.hathor.network/';
 const EXPLORER_SERVICE_MAINNET_BASE_URL  = 'https://explorer-service.hathor.network/';
 const EXPLORER_SERVICE_TESTNET_BASE_URL  = 'https://explorer-service.testnet.hathor.network/';
 
-// Explorer service URL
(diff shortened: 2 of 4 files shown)


src/config.ts
```diff
@@ -68,7 +73,7 @@ class Config {
 
   /**
    * Gets the configured api key for tx-mining-service
-   *
```
tuliomir, line 71: The linter wouldn't allow me to ignore this line's trailing spaces, so I let it fix it.

src/config.ts
```diff
@@ -99,6 +104,29 @@ class Config {
...
 
+  /**
+   * Returns the base url for swap service
+   * Throws an error if it is not yet set.
+   *
+   * @return The swap service url
+   */
+  getSwapServiceBaseUrl(): string {
+    if (!this.SWAP_SERVICE_BASE_URL) {
+      throw new GetWalletServiceUrlError('Swap service base URL not set.');
```
r4mmer, line 115: Should we check the network to choose an URL before failing?

src/wallet/api/swapService.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Decrypts a string ( PartialTx or Signatures ) from the backend
+ * @param serialized
+ * @param password
+ */
+export function decryptString(serialized: string, password: string): string {
+  const aesDecryptedObject = AES.decrypt(serialized, password);
+  return aesDecryptedObject.toString(CryptoJS.enc.Utf8);
```
r4mmer, line 35: Why not use base64 on both encrypt/decrypt?

src/wallet/api/swapService.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Hashes the password to use it as an authentication on the backend
+ * @param {string} password
+ * @returns {string} hashed password
+ */
+export function hashPassword(password): string {
+  return sha256(password).toString();
```
r4mmer, line 44: Does the `toString` returns the hex value?

r4mmer (changes requested):

src/wallet/api/swapService.ts
tuliomir: We only need `Base64` encoding while the contents are encrypted. As soon as we decrypt the contents, we need it back in `Utf8` so that we can work with it in javascript code. So the final decryption on line 35 should always be to `Utf8`.

If we didn't do this string encoding here, all places that need to use the decrypted object would have to implement the Base64 -> Utf8 conversion themselves.

tuliomir: If your question is related to the line 34, it automatically detects the serialized contents' encoding.

src/config.ts
tuliomir: Good point. I adapted the solution already existant for `getExplorerServiceBaseUrl` and implemented it on [6a5385f](https://github.com/HathorNetwork/hathor-wallet-lib/pull/464/commits/6a5385f38ebae8d5e24022c954dc364763cded64)

src/wallet/api/swapService.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Calls the Atomic Swap Service requesting the creation of a new proposal identifier for the informed partialTx.
+ * @param serializedPartialTx
+ * @param password
+ */
+export const create = async (serializedPartialTx: string, password: string) => {
+  const swapAxios = await axiosInstance();
+  const payload = {
+    partialTx: encryptString(serializedPartialTx, password),
+    authPassword: hashPassword(password),
+  };
```
alexruzenhack, line 75: :speech_balloon: What are the possibilities in this request to go wrong? Should we handle a failure here or let it bubble up a possible error?

src/config.ts
```diff
@@ -99,6 +104,44 @@ class Config {
...
+   * @param network - The name of the network to be used to select the url.
+   * @return The Atomic Swap Service url
+   */
+  getSwapServiceBaseUrl(network?: 'mainnet'|'testnet'): string {
+    if (this.SWAP_SERVICE_BASE_URL) {
+      return this.SWAP_SERVICE_BASE_URL;
+    }
+
+    if (!network) {
+      throw new Error('You should either provide a network or call setSwapServiceBaseUrl before calling this.');
```
alexruzenhack, line 123: :speech_balloon: Why not mension to environment variable alternative as well by configuring the `SWAP_SERVICE_BASE_URL` in the project?

src/config.ts
```diff
@@ -99,6 +104,44 @@ class Config {
...
+    }
+
+    // Keeps the old behavior for cases that don't explicitly set a SWAP_SERVICE_BASE_URL
+    if (network == 'mainnet') {
+      return SWAP_SERVICE_MAINNET_BASE_URL;
+    } else if (network == 'testnet'){
+      return SWAP_SERVICE_TESTNET_BASE_URL;
+    } else {
+      throw new Error(`Network ${network} doesn't have a correspondent Atomic Swap Service url. You should set it explicitly.`);
```
alexruzenhack, line 132:
```suggestion
      throw new Error(`Network ${network} doesn't have a correspondent Atomic Swap Service url. You should set it explicitly by calling setSwapServiceBaseUrl or use a supported network.`);
```

src/wallet/api/swapService.ts
```diff
@@ -0,0 +1,79 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  };
+
+  return axios.create(defaultOptions);
+};
+
+/**
+ * Calls the Atomic Swap Service requesting the creation of a new proposal identifier for the informed partialTx.
```
alexruzenhack, line 66: :speech_balloon: How a `serializedPartialTx` looks like? It would be nice to have an example:

```suggestion
 * Calls the Atomic Swap Service requesting the creation of a new proposal identifier for the informed partialTx.
 * @param serializedPartialTx
 * @param password
 * @return {unknown} data
 * 
 * @example
 * await create('string representing a serialized PartialTx', 'password');
 * // => ...data
```
See https://jsdoc.app/tags-example.html

It would be nice to have the `@return` token describing the data type as well.

src/wallet/api/swapService.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param serializedPartialTx
+ * @param password
+ */
```
alexruzenhack, line 69: :speech_balloon: the `axiosInstance` function can throw errors from `config.getSwapServiceBaseUrl`. It would be nice to have a warning message in the docstring of this function telling the user to make sure the SWAP_SERVICE_BASE_URL is proper configured.

src/config.ts
```diff
@@ -16,9 +16,14 @@ const TX_MINING_TESTNET_URL = 'https://txmining.testnet.hathor.network/';
 const EXPLORER_SERVICE_MAINNET_BASE_URL  = 'https://explorer-service.hathor.network/';
 const EXPLORER_SERVICE_TESTNET_BASE_URL  = 'https://explorer-service.testnet.hathor.network/';
 
+// Explorer service URL
```
alexruzenhack, line 19:
```suggestion
// Swap service URL
```

src/config.ts
```diff
@@ -16,9 +16,14 @@ const TX_MINING_TESTNET_URL = 'https://txmining.testnet.hathor.network/';
...
+const SWAP_SERVICE_MAINNET_BASE_URL  = 'https://atomic-swap-service.hathor.network/';
+const SWAP_SERVICE_TESTNET_BASE_URL  = 'https://atomic-swap-service.testnet.hathor.network/';
+
 class Config {
   TX_MINING_URL?: string;
   TX_MINING_API_KEY?: string;
+  SWAP_SERVICE_BASE_URL?: string;
```
alexruzenhack, line 26: :speech_balloon: It would be nice if this config property have more information about its purpose and even better if have example of utilization.
```suggestion
  /**
   * This property is used to reach out the Swap service,
   * it can incorporate the values of `SWAP_SERVICE_MAINNET_BASE_URL`,
   * `SWAP_SERVICE_TESTNET_BASE_URL`, or one defined by the user through
   * the @see setSwapServiceBaseUrl function.
   */
  SWAP_SERVICE_BASE_URL?: string;
```

__tests__/wallet/api/swapService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  it('should return mainnet address when requested', () => {
+      expect(config.getSwapServiceBaseUrl('mainnet'))
+        .toStrictEqual('https://atomic-swap-service.hathor.network/')
+  })
+
+  it('should return testnet address when requested', () => {
+      expect(config.getSwapServiceBaseUrl('testnet'))
+        .toContain('testnet')
```
alexruzenhack, line 42: :speech_balloon: Why not assert the whole URL for testnet?
```suggestion
        .toContain('https://explorer-service.testnet.hathor.network/')
```

__tests__/wallet/api/swapService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  it('should throw when an invalid network is requested', () => {
+    // @ts-ignore
+    expect(() => config.getSwapServiceBaseUrl('invalid'))
+      .toThrowError('You should set it explicitly.');
```
alexruzenhack, line 48: :speech_balloon: I believe the correct message here is the following:
```suggestion
      .toThrowError("Network invalid doesn't have a correspondent Atomic Swap Service url. You should set it explicitly.");
```

__tests__/wallet/api/swapService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  })
+})
+
+describe('create api', () => {
+
+  it('should handle backend errors', async () => {
+    config.setSwapServiceBaseUrl('http://mock-swap-url/')
+
+    mockAxiosAdapter.onPost('/').reply(503)
+    await expect(create('PartialTx||', 'abc'))
```
alexruzenhack, line 64: :speech_balloon: It would be nice to use a near real PartialTx here, to reinforce the test as a documentation.

alexruzenhack (commented):
:speech_balloon: There is the possibility to add an integration test?

src/config.ts
```diff
@@ -99,6 +104,44 @@ class Config {
...
 
+  /**
+   * Returns the correct base url constant for the Atomic Swap Service.
+   * If the url was explicitly set using the config object, it is always returned.
+   * Otherwise, we return it based on the provided network parameter.
+   *
+   * If the url was not set in the config, and no network is provided, we throw an Error.
+   *
+   * @param network - The name of the network to be used to select the url.
+   * @return The Atomic Swap Service url
```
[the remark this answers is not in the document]
tuliomir: Good point! Added on [e2c58fb](https://github.com/HathorNetwork/hathor-wallet-lib/pull/464/commits/e2c58fb0825269d3ec8d743382145344947b1495)

src/wallet/api/swapService.ts
tuliomir: I skipped that part because typescript already gives this information in a more contextualized way. When I inform the axios post type it automatically adds this type to the `create` function response type:
```ts
await swapAxios.post<{ success: boolean, id: string }>('/', payload); // line 87
```

But since this PR's main objective is to introduce the Atomic Swap Service to the lib, I feel a redundant documentation here is very welcome! Added :+1:

src/wallet/api/swapService.ts
tuliomir: I think it's not a problem to let the eventual failures bubble up here, since they're usually related to `crypto-js` and could have valuable information on where/why the problem happened.

However, it has happened to me once to not receive the `serializedPartialTx` due to a bug and it got me some time to figure it out. Added a parameter validation to the function just in case.

src/config.ts
tuliomir: The headless and desktop wallets will certainly implement this as a environment variable, but I see that this should not be a `wallet-lib` recommendation for all clients.

src/config.ts
tuliomir: Adjusted, but removed the last part to keep the error message not too long.

src/config.ts
tuliomir: Oops... :sweat_smile:

src/config.ts
tuliomir: When I added this I noted that it would be important to also add the same kind of documentation to all the other `Config` class variables, and that would make the code polluted with docstrings.

So I believe it's better to keep just the variable declaration here and the full description of its usage on the `getSwapServiceBaseUrl` and `setSwapServiceBaseUrl` methods, as it is now.

Do you agree?

__tests__/wallet/api/swapService.test.ts
tuliomir: It's valid. Changed it to the full url.

__tests__/wallet/api/swapService.test.ts
tuliomir: Added a real example :+1:

src/wallet/api/swapService.ts
alexruzenhack: Nice addition guarding the `create` function.

src/config.ts
alexruzenhack: Mmm... I'm of the option to keep the documentation, even when it messes with code symmetry. However, it is not a big deal. It is ok to not to have it here.

src/config.ts
alexruzenhack: Mmmm... Well, I believe there is no problem to inform this alternative, after all this is an option the implementer has. If the client for some reason were not able to use an environment variable, this is a client's constraint, not the lib's constraint.

r4mmer (approved):

PR #468, opened 2023-03-01
feat: Swap Service `get`

tuliomir:
Continuation of the Atomic Swap Service integration started on #464 

### Acceptance Criteria
- The lib should interact with the `get` endpoint on the Atomic Swap Service
- All original data should be correctly decrypted on returning the data to the client

### Note
No integration tests will be implemented at this stage, as there is no readily available image of the Atomic Swap Service to be added to the `docker-compose.yml` structure necessary for them.

### Security Checklist
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #473, opened 2023-03-07
feat: swap service `update`

tuliomir:
Continuation of the Atomic Swap Service integration started on #464 and #468 

### Acceptance Criteria
- The lib should interact with the `update` endpoint on the Atomic Swap Service
- All original data should be correctly encrypted on sending the data to the backend

### Note
No integration tests will be added at this stage, as there is no readily available image of the Atomic Swap Service to be added to the `docker-compose.yml` structure necessary for them.
A partial implementation of these tests, however, is available on branch [`feat/swap-service-integration`](https://github.com/HathorNetwork/hathor-wallet-lib/tree/feat/swap-service-integration) ( [diff with `dev`](https://github.com/HathorNetwork/hathor-wallet-lib/compare/dev...feat/swap-service-integration?expand=1) ) and can be executed on a local environment by running the Atomic Swap Service directly from the source code.

### Security Checklist

### DIFF
--- a/src/wallet/api/swapService.ts
+++ b/src/wallet/api/swapService.ts
@@ -185,9 +185,10 @@ interface SwapUpdateParams {
 /**
  * Updates the proposal on the Atomic Swap Service with the parameters informed
  */
-export const update = async ({proposalId, password, partialTx, version, signatures}: SwapUpdateParams):
+export const update = async (params: SwapUpdateParams):
   Promise<{ success: boolean }> => {
   validateParameters();
+  const { proposalId, password, partialTx, version, signatures } = params;
 
   const swapAxios = await axiosInstance();
   const options = {
@@ -209,6 +210,11 @@ export const update = async ({proposalId, password, partialTx, version, signatur
    * @throws {Error} if any version parameter is invalid
    */
   function validateParameters() {
+    if (!params) {
+      throw new Error(`Missing mandatory parameters.`);
+    }
+
+    const { proposalId, password, partialTx, version } = params;
     // Checking for missing parameters
     const missingParameters: String[] = [];
     if (!proposalId) {
@@ -224,7 +230,7 @@ export const update = async ({proposalId, password, partialTx, version, signatur
       missingParameters.push('version');
     }
     if (missingParameters.length > 0) {
-      throw new Error(`Missing mandatory parameters: ${missingParameters.join(', ')}`)
+      throw new Error(`Missing mandatory parameters: ${missingParameters.join(', ')}`);
     }
 
     // Checking for invalid parameters
--- a/src/wallet/api/swapService.ts
+++ b/src/wallet/api/swapService.ts
@@ -187,7 +187,9 @@ interface SwapUpdateParams {
  */
 export const update = async (params: SwapUpdateParams):
   Promise<{ success: boolean }> => {
+  // Validates the input parameters and throws in case of errors
   validateParameters();
+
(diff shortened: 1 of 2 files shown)


src/wallet/api/swapService.ts
```diff
@@ -172,3 +173,70 @@ export const get = async (proposalId: string, password: string): Promise<AtomicS
...
+
+  const { data } = await swapAxios.put<{ success: boolean }>(`/${proposalId}`, payload, options);
+  return { success: data?.success };
+
+  /**
+   * Validates the many mandatory parameters for the `update` method
+   * @throws {Error} if any mandatory parameter is missing
+   * @throws {Error} if any version parameter is invalid
+   */
+  function validateParameters() {
```
alexruzenhack, line 212: 💬 Interesting!

I'm concerned about readability. This structure caused me a little confusion while I was reading. Because of this, I believe it would be better to extract this validation function and do some refactoring to return an array of validation messages, like:

```javascript
function validateUpdateParameters(params) {
  const missingParameters = [];
  if (!params) {
    return { success: false, validationMsg: 'The positional function argument "params" is required' };
  }
  if (!proposalId) {
    missingParameters.push('proposalId');
  }
  // hidden for brevity...
  return { success: true, validationMsg: 'Missing mandatory parameters: ' + missingParameters.join(', ') + '.' };
}
```

This way you avoid throwing exceptions directly from the validation and pass this responsibility to the `update` function, turning the expectations explicit.

The `update` function turns out to be more easy to read and comprehend.

src/wallet/api/swapService.ts
```diff
@@ -172,3 +173,70 @@ export const get = async (proposalId: string, password: string): Promise<AtomicS
...
+
+  const swapAxios = await axiosInstance();
+  const options = {
+    headers: { 'X-Auth-Password': hashPassword(password) }
+  } as AxiosRequestConfig;
+
+  const payload = {
+    partialTx: encryptString(partialTx, password),
+    version,
+    signatures: signatures ? encryptString(signatures, password) : null,
```
alexruzenhack, line 201: 💬 The signature will be validated in the server, right? By validating the signature in the lib we can avoid a round trip to the server. Does it worth the effort to validate it here?

alexruzenhack (commented):
I have some comments about:
- signature validation in the client
- readability refactoring

src/wallet/api/swapService.ts
tuliomir: The server does not validate the signatures: it only accepts whatever value is passed and stores it on the database ( [code reference](https://github.com/HathorNetwork/hathor-atomic-swap-service/pull/4/files#diff-a7489a11dd17a7d3af1075dbca6440d8c18a882f6ec1f077991e2b2f1a8e1478) ).

If the value is `null`, it considers the signatures should be voided and also updates the database's column with `null`.

alexruzenhack: Got it!

src/wallet/api/swapService.ts
tuliomir: It's a good pattern, but would diverge from the other `swapService` methods. All of them _throw_ on invalid parameters and _resolve_ with `{ success: boolean }` when receiving data from the server.

It could be a possibility for a future refactor, but I think it would not be a good idea on this PR.
What do you think?

alexruzenhack (approved):

r4mmer (approved):

PR #475, opened 2023-03-08
feat: `PartialTxProposal` set signatures

tuliomir:
### Acceptance Criteria
- The `PartialTxProposal` should have a new `setSignatures` method
- This method should overwrite the proposal's signatures with the parameters informed

### Motivation
Currently the only way for a `PartialTxProposal` instance to have signatures is having one of the participating wallets sign it.

The `setSignatures` method allows a third party to apply existing signatures to a proposal, or even one of the participating wallets to instantiate the full proposal data without having to re-sign it. This is especially important in the context of the Atomic Swap Service, that allows a fully completed proposal to stay available to any clients before being turned into a transaction.

### Security Checklist


__tests__/wallet/partialTxProposal.test.ts
```diff
@@ -354,6 +355,26 @@ test('resetSignatures', async () => {
   expect(proposal.transaction).toEqual(null);
 });
 
+test('setSignatures', async () => {
```
alexruzenhack, line 358: 💬 In this test it would be nice to have an assert for the `signatures` value, specially after a merge with a different `serializedSignatures`.

src/wallet/partialTxProposal.ts
```diff
@@ -371,6 +371,27 @@ class PartialTxProposal {
...
+      throw new InvalidPartialTxError('Cannot sign incomplete data');
+    }
+    const tx: Transaction = this.partialTx.getTx();
+
+    this.signatures = new PartialTxInputData(
+      tx.getDataToSign().toString('hex'),
+      tx.inputs.length
+    );
+
+    this.signatures.addSignatures(serializedSignatures)
```
alexruzenhack, line 392: 💬 I believe a comment here like `merges with local data` could add more context for the reader.

alexruzenhack (commented):
💬 I have comments about:
- extend the test to cover the merge of serialized signature with the local signature
- increase context by letting a comment in the call `this.signatures.addSignatures(...)`

__tests__/wallet/partialTxProposal.test.ts
tuliomir: The signature adding process is quite complex and would require access to the storage, initializing wallets, etc.

The scope of this unit test is just to make sure the correct `addSignatures` method ( mocked here ) is called with the correct parameters, and trust that this method's results are being well tested elsewhere. Also, there is no merge on this method, only overriding of the local signatures.

The [integration tests](https://github.com/HathorNetwork/hathor-wallet-lib/compare/dev...feat/swap-service-integration) that will be added later will make sure all pieces are working together correctly.

src/wallet/partialTxProposal.ts
alexruzenhack: Nice!

__tests__/wallet/partialTxProposal.test.ts
alexruzenhack: Nice!

alexruzenhack (approved):

src/wallet/partialTxProposal.ts
```diff
@@ -371,6 +371,29 @@ class PartialTxProposal {
...
+    const tx: Transaction = this.partialTx.getTx();
+
+    // Creating an empty signatures object
+    this.signatures = new PartialTxInputData(
+      tx.getDataToSign().toString('hex'),
+      tx.inputs.length
+    );
+
+    // Setting the signatures data from the parameters
+    this.signatures.addSignatures(serializedSignatures)
```
r4mmer, line 394: This method will only work if the signatures are for the same tx (indicated by the `tx.getDataToSign().toString('hex')`).
While this is the expecetd behavior I think we should add a test checking that it fails if we try to add a signature from another tx.

r4mmer (changes requested):

src/wallet/partialTxProposal.ts
tuliomir: Always good to have more validations and improve the error message. It's terrible to have a vague error on these situations.
Added this on [190caaa](https://github.com/HathorNetwork/hathor-wallet-lib/pull/475/commits/190caaa485b7f29dbba783d060570524832eab58)

r4mmer (approved):

alexruzenhack (approved):

Issue #478, opened 2023-03-15
[new-storage] Process locked utxos when the height is updated


r4mmer:
## Motivation

Locked block rewards may become unlocked when the height is updated, this requires the token and address metadata to be updated but we do not have this process yet.

In a similar fashion timelocked utxos become available when the timestamp passes the value in the timelock but we do not have a way to update the balances on token and address metadata. If we process the timelocked utxos when the height is updated we should be checking every 30s (expected time between blocks).

PR #481, opened 2023-03-17
feat: reward lock processing

r4mmer:
### Acceptance Criteria
- We should have a separate list of locked utxos to avoid iterating on all utxos.
- When the height is updated we should process the locked utxos
- Create a method to process locked utxos that became unlocked

### Security Checklist


r4mmer:
closes #478

src/utils/connection.ts
```diff
@@ -10,7 +10,9 @@ import { IStorage } from "../types";
 export function handleWsDashboard(storage: IStorage) {
   return (data: any) => {
     // update network height
-    storage.setCurrentHeight(data.best_block_height as number);
+    const height = data.best_block_height as number;
+    storage.setCurrentHeight(height);
+    storage.unlockUtxos(height);
```
pedroferreira1, line 15: This ws message is sent every second. I think we should check if the height has changed before doing anything here, what do you think?

src/storage/storage.ts
```diff
@@ -46,12 +46,22 @@ export class Storage implements IStorage {
...
   config: Config;
   version: ApiVersion|null;
+  /**
+   * This promise is used to chain the calls to process unlocked utxos.
+   * This way we can avoid concurrent calls.
+   * The best way to do this would be an async queue or a mutex, but to avoid adding
+   * more dependencies we are using this simpler method.
+   *
+   * We can change this implementation to use a mutex or async queue in the future.
+   */
```
pedroferreira1, line 56: Nice comment 👍

src/storage/storage.ts
```diff
@@ -249,6 +259,21 @@ export class Storage implements IStorage {
     await this.store.unregisterToken(tokenUid);
   }
 
+  /**
+   * Process the locked utxos to unlock them if the lock has expired.
+   * Will process both timelocked and heightlocked utxos.
+   *
+   * We will wait for the any previous execution to finish before starting the next one.
```
pedroferreira1, line 266:
```suggestion
   * We will wait for any previous execution to finish before starting the next one.
```

__tests__/storage/storage.test.js
```diff
@@ -143,3 +143,230 @@ test('utxos selected as inputs', async () => {
...
+      address,
+      value: output.value,
+      authorities: 0,
+      timelock,
+      type: tx.version,
+      height: tx.height,
+    };
+  }
+
+  const store = new MemoryStore();
```
pedroferreira1, line 198: Can we do the same test for leveldb store?

__tests__/storage/memory_store.test.js
```diff
@@ -262,6 +262,125 @@ test('utxo methods', async () => {
...
+      type: tx.version,
+      height: tx.height,
+    };
+  }
+
+  jest
+   .useFakeTimers()
+   .setSystemTime(tsCurrent);
+
+  const store = new MemoryStore();
```
pedroferreira1, line 320: Can you do the same tests for the leveldb store?

src/utils/connection.ts
r4mmer: Started to check if the current height is different than the new height before calling `unlockUtxos`

src/storage/storage.ts

__tests__/storage/storage.test.js
r4mmer: Done

__tests__/storage/memory_store.test.js
r4mmer: Done.

__tests__/storage/common_store.test.js
```diff
@@ -0,0 +1,168 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { LevelDBStore, MemoryStore } from "../../src/storage";
+import { HDPrivateKey } from "bitcore-lib";
+import { TOKEN_AUTHORITY_MASK, TOKEN_MINT_MASK } from "../../src/constants";
+
+const DATA_DIR = './testdata.leveldb';
+
+describe("locked utxo methods", () => {
```
r4mmer, line 14: I will start to port the other store tests to this file with a similar approach in another PR.

Basically we should have a common tests that can run with any store implementation to make sure all `IStore` implementations behave the same way.
Ideally we should do the same with `IStorage` implementations and check that any combination of `IStorage` and `IStore` will work as intended.

tuliomir (approved):

__tests__/storage/common_store.test.js
tuliomir: I agree. The test codes got too similar as well, it will be good to have them unified.

r4mmer: https://github.com/HathorNetwork/hathor-wallet-lib/issues/488

src/utils/connection.ts
```diff
@@ -11,8 +11,12 @@ export function handleWsDashboard(storage: IStorage) {
   return (data: any) => {
     // update network height
     const height = data.best_block_height as number;
-    storage.setCurrentHeight(height);
-    storage.unlockUtxos(height);
+    storage.getCurrentHeight().then(async (currentHeight) => {
+      await storage.setCurrentHeight(height);
```
pedroferreira1, line 15: You could move this inside the `if` below as well but it's fine anyway.

src/utils/connection.ts
```diff
@@ -11,8 +11,12 @@ export function handleWsDashboard(storage: IStorage) {
   return (data: any) => {
     // update network height
     const height = data.best_block_height as number;
-    storage.setCurrentHeight(height);
-    storage.unlockUtxos(height);
+    storage.getCurrentHeight().then(async (currentHeight) => {
```
pedroferreira1, line 14: Is there a reason to use promise here?

src/utils/connection.ts
```diff
@@ -11,8 +11,12 @@ export function handleWsDashboard(storage: IStorage) {
...
+      await storage.setCurrentHeight(height);
+      if (height !== currentHeight) {
+        await storage.unlockUtxos(height);
```
pedroferreira1, line 17: Do we need to await?

pedroferreira1 (approved):

src/utils/connection.ts
r4mmer: Its only needed to await the promises below, but they are not actually necessary.

src/utils/connection.ts
r4mmer: Not really, the initial idea was to run `unlockUtxos` before changing the height, but passing the height as a parameter is better, but it made the `await`s unnecessary, so I'm removing them

pedroferreira1 (approved):

PR #483, opened 2023-03-17
feat: make storage optional with SendTransaction facade

r4mmer:
### Description

Among the changes for the new storage scheme we changed the constructor arguments for the SendTransaction facade, due to the abstraction of the storage.
We need to make a small adjustment so that we include the case of sending a pre-prepared transaction that does not require any storage calls since its used on the wallets.

### Acceptance Criteria

- Make storage optional on the SendTransaction
- Fail any calls where storage is required.
- Change tests and internal calls to SendTransaction

### Security Checklist


__tests__/wallet/wallet.test.ts
```diff
@@ -322,7 +322,7 @@ test('generateCreateWalletAuthData should return correct auth data', async () =>
 
 test('getTxById', async () => {
   const wallet = buildWalletToAuthenticateApiCall();
-  spyOn(wallet, 'isReady').and.returnValue(true);
+  jest.spyOn(wallet, 'isReady').mockReturnValue(true);
```
r4mmer, line 325: Not sure how this was passing

package.json
```diff
@@ -51,7 +51,7 @@
     "eslint": "^8.12.0",
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-plugin-import": "^2.26.0",
-    "jest": "^24.8.0",
+    "jest": "^27.0.3",
```
r4mmer, line 54: Need to update jest to start using fake timers to test timelocks.

__tests__/utils/transaction.test.js
```diff
@@ -269,4 +269,110 @@ test('authorities from output', () => {
...
+  const store = new MemoryStore();
+  const storage = new Storage(store);
+
+  jest.spyOn(storage, 'getCurrentHeight').mockReturnValue(Promise.resolve(10));
+  const txSpy = jest.spyOn(storage, 'getTx');
+  const isSelSpy = jest.spyOn(storage, 'isUtxoSelectedAsInput').mockReturnValue(Promise.resolve(false));
+  storage.version = { reward_spend_min_blocks: 5 };
+
+  jest.useFakeTimers()
+  jest.setSystemTime(t3);
```
tuliomir, line 290: It's great we're able to use this feature now! :tada:

src/storage/storage.ts
```diff
@@ -496,6 +496,29 @@ export class Storage implements IStorage {
     }
   }
 
+  /**
+   * Check if an utxo is selected as input.
+   * 
+   * @param {IUtxoId} utxo The utxo we want to check if it is selected as input
+   * @returns {Promise<boolean>}
+   */
```
tuliomir, line 504: I suggest adding an example, so it's easier to remember what is an `IUtxoId` in this context
```suggestion
  /**
   * Check if an utxo is selected as input.
   * 
   * @param {IUtxoId} utxo The utxo we want to check if it is selected as input
   * @returns {Promise<boolean>}
   * @example
   * const isSelected = await isUtxoSelectedAsInput({ txId: 'tx1', index: 0 });
   */
```

src/storage/storage.ts
```diff
@@ -496,6 +496,29 @@ export class Storage implements IStorage {
...
+  async isUtxoSelectedAsInput(utxo: IUtxoId): Promise<boolean> {
+    const utxoId = `${utxo.txId}:${utxo.index}`;
+    return this.utxosSelectedAsInput.has(utxoId);
+  }
+
+  async *utxoSelectedAsInputIter(): AsyncGenerator<IUtxoId, any, unknown> {
```
tuliomir, line 510: Didn't the linter complain about the use of `any`?

tuliomir (approved):

src/storage/storage.ts
r4mmer: I will remove this, but it doesn't complain for `AsyncGenerator`s since the 3 types of the template are the yield type, the return type and the type received by the `next` method and the default values are `any` and `unknown` for the last 2.

Since its the expected (default) behavior the linter will not complain.

src/new/wallet.js
```diff
@@ -800,9 +800,6 @@ class HathorWallet extends EventEmitter {
 
     for await (const utxo of this.storage.selectUtxos(newOptions)) {
       const isLocked = isTimeLocked(utxo.timelock) || isHeightLocked(utxo.height);
-      if (options.only_available_utxos && isLocked) {
-        continue;
-      }
```
pedroferreira1, line 805: Why did you remove this?

src/new/wallet.js
```diff
@@ -850,16 +847,10 @@ class HathorWallet extends EventEmitter {
    * @yields {Utxo} all available utxos
    */
   async * getAllUtxos(options = {}) {
```
pedroferreira1, line 849: Where do we use this method? Would be much better to call it `getAvailableUtxos`

src/storage/storage.ts
```diff
@@ -496,6 +507,31 @@ export class Storage implements IStorage {
...
+   * @returns {Promise<boolean>}
+   * @example
+   * const isSelected = await isUtxoSelectedAsInput({ txId: 'tx1', index: 0 });
+   */
+  async isUtxoSelectedAsInput(utxo: IUtxoId): Promise<boolean> {
+    const utxoId = `${utxo.txId}:${utxo.index}`;
+    return this.utxosSelectedAsInput.has(utxoId);
+  }
+
+  async *utxoSelectedAsInputIter(): AsyncGenerator<IUtxoId> {
```
pedroferreira1, line 523: Missing docstring

src/storage/storage.ts
```diff
@@ -266,9 +266,20 @@ export class Storage implements IStorage {
...
       const utxoId = `${utxo.txId}:${utxo.index}`;
-      return (!this.utxosSelectedAsInput.has(utxoId)) && (options.filter_method ? options.filter_method(utxo) : true);
+      return !this.utxosSelectedAsInput.has(utxoId);
+    }
+    const newFilter = (utxo: IUtxo): boolean => {
+      const optionsFilter = (options.filter_method ? options.filter_method(utxo) : true);
+      const selectedFilter = filterSelected(utxo);
+      if (options.only_available_utxos) {
+        // We need to check if the utxo is selected as an input since we only want available utxos.
+        return selectedFilter && optionsFilter;
```
pedroferreira1, line 278: Don't we need to remove heightlocked and timelocked utxos as well?

src/storage/storage.ts
r4mmer: Docstring added

src/new/wallet.js
r4mmer: Renamed to `getAvailableUtxos`

src/utils/transaction.ts
```diff
@@ -493,6 +493,33 @@ const transaction = {
...
+  async canUseUtxo(utxo: IUtxoId, storage: IStorage): Promise<boolean> {
+    const currentHeight = await storage.getCurrentHeight();
+    const rewardLock = storage.version?.reward_spend_min_blocks || 0;
+    const nowTs = Math.floor(Date.now() / 1000);
+    const tx = await storage.getTx(utxo.txId);
+    if (tx === null) {
+      // This is not our utxo, so we cannot spend it.
+      return false;
+    }
+    const output = tx.outputs[utxo.index];
```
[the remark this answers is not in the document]
r4mmer: Changed to return false if `utxo.index` is greater than outputs length

pedroferreira1 (approved):

Issue #480, opened 2023-03-16
[new storage] Use a SortedMap to ensure timestamp ordering on memory store


r4mmer:
## Motivation

Memory store does not ensure the txs are ordered by timestamps, to ensure this we would need either an auxiliar sorted collection or use a sorted collection as the main data structure.

## Implementation

To avoid implementing this sorting we can use an aditional dependency called [collectionjs](https://github.com/montagejs/collections) which have some interesting data structures that use [splay trees](https://en.wikipedia.org/wiki/Splay_tree) underneath.

The best way to do this would be with a `SortedArray` and pass custom `equal` and `compare` methods since we require a decreasing order on timestamps but an element is equal to the other if the `tx_id` is the same.

If we do not want to include another dependency for this we could make a very simple implementation using the methods available for nodejs.
We could have an array of strings made as `<timestamp>:<tx_id>` where timestamp is the uint32 in hex and we can use the usual `.sort()` to guarantee timestamp ordering.
Since we require decreasing ordering, this can be done by iterating from `array.length-1` to 0, which will be easier than having the array sorted in decreasing order.
Since all transactions enter the storage through `store.saveTx` we can push the new tx to this array and sort it there.

r4mmer:
Done in #485

PR #485, opened 2023-03-21
feat: ensure memory store history order

r4mmer:
### Acceptance Criteria
- solve the issue described in #480 
- force iterations on history to follow the timestamp order

### Security Checklist

### DIFF
--- a/src/storage/memory_store.ts
+++ b/src/storage/memory_store.ts
@@ -364,8 +364,12 @@ export class MemoryStore implements IStore {
     // Protect ordering list from updates on the same transaction
     // We can check the historyTs but it's O(n) and this check is O(1).
     if (!this.history.has(tx.tx_id)) {
-      // Add transaction to the ordering list
-      // and sort it so we ensure the order
+      // Add transaction to the ordering list and sort it
+      // Wallets expect to show users the transactions in order of descending timestamp
+      // This is so wallets can show the most recent transactions to users
+      // XXX: If this becomes a performance bottleneck we can either allow wallets to skip ordering
+      // or have the ordering be done during history processing so that
+      // we don't have to sort the list every time we add a new transaction
       this.historyTs.push(getOrderingKey(tx));
       this.historyTs.sort();
     }
--- a/src/storage/memory_store.ts
+++ b/src/storage/memory_store.ts
@@ -31,6 +31,36 @@ const DEFAULT_WALLET_DATA = {
   gapLimit: GAP_LIMIT,
 };
 
+/**
+ * Get the ordering key for a transaction.
+ * This will be used to sort the transactions by timestamp.
+ *
+ * @param {IHistoryTx} tx The transaction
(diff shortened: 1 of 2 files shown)


src/storage/memory_store.ts
```diff
@@ -31,6 +31,36 @@ const DEFAULT_WALLET_DATA = {
   gapLimit: GAP_LIMIT,
 };
 
+/**
+ * Get the ordering key for a transaction.
+ * This will be used to sort the transactions by timestamp.
+ *
+ * @param {IHistoryTx} tx The transaction
+ * @returns {string} The ordering key for the transaction
```
tuliomir, line 39: Since it's not intuitive what a unsigned it to hexa would look like, an `@example` tag would be welcome here, with a sample of the return value.

src/storage/memory_store.ts
```diff
@@ -302,6 +351,11 @@ export class MemoryStore implements IStore {
   async saveTx(tx: IHistoryTx): Promise<void> {
     this.history.set(tx.tx_id, tx);
 
+    // Add transaction to the ordering list
+    // and sort it so we ensure the order
+    this.historyTs.push(getOrderingKey(tx));
+    this.historyTs.sort();
```
tuliomir, line 357: Should we expect heavy CPU loads from this sorting?

Considering that the new storage scheme is aimed at heavy users, I can imagine this could have an impact on a heavily used exchange wallet, for example. I'm imagining receiving many new transactions per second on a wallet that has thousands of transactions/addresses.

r4mmer: For new transactions, pushing to the end should already be in order since we are ordering by timestamp and the most recent is the higher timestamp.
This should add some overhead when loading the history but it is necessary to ensure this feature.

If this raises an issue later we could add an option to the config to disable this.

src/storage/memory_store.ts
r4mmer: Added examples

src/storage/memory_store.ts
```diff
@@ -300,6 +361,15 @@ export class MemoryStore implements IStore {
    * @returns {Promise<void>}
    */
   async saveTx(tx: IHistoryTx): Promise<void> {
+    // Protect ordering list from updates on the same transaction
+    // We can check the historyTs but it's O(n) and this check is O(1).
+    if (!this.history.has(tx.tx_id)) {
+      // Add transaction to the ordering list
+      // and sort it so we ensure the order
+      this.historyTs.push(getOrderingKey(tx));
+      this.historyTs.sort();
```
pedroferreira1, line 370: I would only add a comment here explaining why we need this list sorted by timestamp and that, if this sort becomes a performance problem in the future, that we can create a pre processed data for each token and remove this sort.

pedroferreira1 (approved):

tuliomir (approved):

PR #486, opened 2023-03-21
feat: leveldb filter locked utxos

r4mmer:
### Acceptance Criteria
- filter for locked utxos only if `only_available_utxos` is true
- filter for timelocks on leveldb

### Security Checklist

### DIFF
--- a/src/storage/leveldb/utxo_index.ts
+++ b/src/storage/leveldb/utxo_index.ts
@@ -186,6 +186,15 @@ export default class LevelUtxoIndex implements IKVUtxoIndex {
       // Heighlocked when network height is lower than block height + reward_spend_min_blocks
       return ((utxo.height || 0) + options.reward_lock) > networkHeight;
     };
+    const nowTs = Math.floor(Date.now() / 1000);
+    const isTimelocked = (utxo: IUtxo) => {
+      if (utxo.timelock === null) {
+        // Not timelocked
+        return false;
+      }
+      // Timelocked when now is lower than timelock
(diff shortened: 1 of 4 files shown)


tuliomir (approved):

src/storage/leveldb/utxo_index.ts
```diff
@@ -186,6 +186,15 @@ export default class LevelUtxoIndex implements IKVUtxoIndex {
       // Heighlocked when network height is lower than block height + reward_spend_min_blocks
       return ((utxo.height || 0) + options.reward_lock) > networkHeight;
```
pedroferreira1, line 187: I think there's a bug here. You are never passing `networkHeight` when you use this method. https://github.com/HathorNetwork/hathor-wallet-lib/blob/dev/src/storage/storage.ts#L292

The method `selectUtxos` from memory store does not have `networkHeight` as parameter, it gets using `this.getCurrentHeight`. Both methods should have the same signature. The interface follows what's in the memory store: https://github.com/HathorNetwork/hathor-wallet-lib/blob/dev/src/types.ts#L288

r4mmer: But this is the utxo index, which is internal to the leveldb store.
We get the network height and pass to this method in the [store](https://github.com/HathorNetwork/hathor-wallet-lib/blob/feat/leveldb-filter-locked-utxos/src/storage/leveldb/store.ts#L227).

pedroferreira1 (approved):

PR #490, opened 2023-03-27
fix: `Transaction` circular dependency

tuliomir:
### Acceptance Criteria
- Should alleviate the circular dependency between `utils/helpers.ts > models/create_token_transaction.ts > models/transaction.ts`
- Extract `intToBytes`, `signedIntToBytes` and `floatToBytes` to the `buffer.ts` helper, instead of the `helpers.ts` that has many other dependencies
- Extract `prettyValue` and `prettyIntegerValue` to a new `numbers.ts` helper without dependencies other than the constants

### Breaking Changes
`utils/helpers.ts` no longer has the following methods:
- `intToBytes`
- `signedIntToBytes`
- `floatToBytes`
- `prettyValue`
- `prettyIntegerValue`

### Security Checklist

### DIFF
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -19,8 +19,7 @@ import Input from '../models/input';
 import Output from '../models/output';
 import Network from '../models/network';
 import Address from '../models/address';
-import { hexToBuffer, unpackToInt, intToBytes, signedIntToBytes, floatToBytes } from './buffer';
-import { prettyValue, prettyIntegerValue } from './numbers';
+import { hexToBuffer, unpackToInt, intToBytes } from './buffer';
 import { crypto, encoding, Address as bitcoreAddress } from 'bitcore-lib';
 import { clone } from 'lodash';
 import { AddressError, OutputValueError, ConstantNotSet, CreateTokenTxInvalid, MaximumNumberInputsError, MaximumNumberOutputsError, ParseError } from '../errors';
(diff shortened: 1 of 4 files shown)


src/utils/helpers.ts
```diff
@@ -56,32 +56,25 @@ const helpers = {
...
    * @memberof Helpers
    * @inner
    */
-  prettyValue(value: number): string {
-    const fixedPlaces = (value/10**DECIMAL_PLACES).toFixed(DECIMAL_PLACES);
-    const integerPart = fixedPlaces.split('.')[0];
-    const decimalPart = fixedPlaces.split('.')[1];
-    return `${this.prettyIntegerValue(parseInt(integerPart))}.${decimalPart}`;
-  },
+  prettyValue: prettyValue,
```
pedroferreira1, line 64: Why do you need to keep this?

tuliomir: For backward compatibility. Lib clients may be using it and now will have a warning because this method now have the `@deprecated` annotation on its docstrings.

After some time we can completely remove this property.

pedroferreira1: We are about to release v1.0.0 that breaks compatibility with almost everything. I think this is the perfect momento to remove this extra code and just document in the release notes that methods A, B, and C were moved to file X, and methods D and E were moved to file Y. What do you think?

tuliomir: Good point! I used to opportunity to also refactor the tests into their own files.

pedroferreira1 (approved):

andreabadesso (approved):

PR #492, opened 2023-03-27
Release v1.0.0-rc1

r4mmer:
### Acceptance Criteria
- Create release candidate 1 for v1.0.0

### Security Checklist


.github/workflows/integration-test.yml
```diff
@@ -33,14 +33,14 @@ jobs:
...
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: test-transacion-logs
           path: tmp
 
       - name: Collect docker logs on failure
         if: failure()
-        uses: jwalton/gh-docker-logs@v1
+        uses: jwalton/gh-docker-logs@v2
```
luislhl, line 43: I would recommend pinning actions to a commit sha instead of tags, as per the recommendations in https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

luislhl (approved):

pedroferreira1 (approved):

.github/workflows/integration-test.yml
r4mmer: Tracking this on https://github.com/HathorNetwork/hathor-wallet-lib/issues/493

PR #499, opened 2023-04-03
feat: split filltx per token

r4mmer:
### Motivation

As observed by @tuliomir on this [comment](https://github.com/HathorNetwork/hathor-wallet-headless/pull/278#discussion_r1153944885) the send transaction facade should be able to configure whether to choose inputs on a per-token basis and this is not possible in the current `fillTx` implementation.

### Acceptance Criteria
- Change `fillTx` to work on a single token.
 - Change `SendTransaction` to configure `chooseInputs` per-token.

### Security Checklist


__tests__/new/sendTransaction.test.js
```diff
@@ -91,6 +91,7 @@ test('prepareTxData', async () => {
         authorities: 0,
       },
       'an-input-from-storage',
+      'an-input-from-storage',
```
r4mmer, line 94: This extra input and output are because we are calling `fillTx` twice, one for HTR and another for the `01` custom token.
Before we would get only 1 since we made only 1 call to `fillTx`.

src/new/sendTransaction.ts
```diff
@@ -132,11 +131,12 @@ class SendTransaction extends EventEmitter {
...
     };
     const chooseInputs = this.inputs.length === 0;
-    const tokenSet = new Set<string>();
+    // Map of token uid to the chooseInputs value of this token
+    const tokenMap = new Map<string, boolean>();
 
     for (const output of this.outputs) {
       if (isDataOutput(output)) {
-        tokenSet.add(HTR_UID);
+        tokenMap.set(HTR_UID, false);
```
pedroferreira1, line 139: Why don't you set this as `true` just like the other?

r4mmer: Done

pedroferreira1 (approved):

src/utils/transaction.ts
```diff
@@ -293,57 +293,52 @@ const transaction = {
    * Or conversely add the melted tokens on the outputs, we will ignore minted/melted funds.
    *
    * @param {IDataTx} tx The transaction we want to calculate the balance.
-   * @returns {Promise<Map<string, Record<'funds'|'mint'|'melt', number>>>} The balance of all tokens on the transaction.
+   * @returns {Promise<Record<'funds'|'mint'|'melt', number>>} The balance of all tokens on the transaction.
```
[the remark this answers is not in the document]
r4mmer: Done

tuliomir (approved):

src/utils/transaction.ts
```diff
@@ -292,58 +292,54 @@ const transaction = {
...
-   * @returns {Promise<Map<string, Record<'funds'|'mint'|'melt', number>>>} The balance of all tokens on the transaction.
+   * @returns {Promise<Record<'funds'|'mint'|'melt', number>>} The balance of the given token on the transaction.
    */
-  async calculateTxBalanceToFillTx(tx: IDataTx): Promise<Map<string, Record<'funds'|'mint'|'melt', number>>> {
-    function getEmptyBalance(): Record<'funds'|'mint'|'melt', number> {
-      return {'funds': 0, 'mint': 0, 'melt': 0};
-    }
-    const balance = new Map<string, Record<'funds'|'mint'|'melt', number>>();
-
+  async calculateTxBalanceToFillTx(token: string, tx: IDataTx): Promise<Record<'funds'|'mint'|'melt', number>> {
```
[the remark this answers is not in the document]
r4mmer: Not really, but it came from a breakdown of the `fillTx` method which is async so by default i made it async.
After thinking about this method, maybe we should include it back into `fillTx` since it can only be used there.
But I think this should be made on its own PR.

PR #501, opened 2023-04-05
feat: utxo selection algo

r4mmer:
### Acceptance Criteria
- Implement the utxo selection algorithm
- Allow utxo selection algorithm to be configurable

### Security Checklist

### DIFF
--- a/__tests__/utils/utxo.test.js
+++ b/__tests__/utils/utxo.test.js
@@ -1,0 +1,178 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { HDPrivateKey } from "bitcore-lib";
+import { LevelDBStore, MemoryStore, Storage } from "../../src/storage";
+import { bestUtxoSelection, fastUtxoSelection } from "../../src/utils/utxo";
+
+const DATA_DIR = './testdata.leveldb';
+
+describe('bestUtxoSelection', () => {
+  const utxos = [
+    {
+      txId: 'tx1',
+      index: 0,
+      value: 100,
+      token: '00',
+      address: 'addr1',
+      authorities: 0,
+    },
+    {
+      txId: 'tx2',
+      index: 0,
+      value: 200,
+      token: '00',
+      address: 'addr2',
+      authorities: 0,
+    },
+    {
+      txId: 'tx3',
+      index: 0,
+      value: 300,
+      token: '00',
+      address: 'addr3',
+      authorities: 0,
+    },
+    {
+      txId: 'tx4',
+      index: 0,
+      value: 400,
+      token: '01',
+      address: 'addr4',
+      authorities: 0,
+    },
+    {
+      txId: 'tx5',
+      index: 0,
+      value: 500,
+      token: '01',
+      address: 'addr5',
+      authorities: 0,
+    },
+    {
+      txId: 'tx6',
+      index: 0,
+      value: 600,
+      token: '01',
+      address: 'addr6',
+      authorities: 0,
+    },
+  ];
+
+  async function addUtxosToStore(store) {
+    for (const utxo of utxos) {
+      await store.saveUtxo(utxo);
+    }
+  }
+
+  /**
+   * Should select the best utxos to use in a transaction
+   * @param {IStore} store 
+   */
+  async function testBestUtxoSelection(store) {
+    await addUtxosToStore(store);
+    await store.setCurrentHeight(1);
+    const storage = new Storage(store);
+
+    // State of the utxos:
+    // 00: 100 + 200 + 300 = 600
+    // 01: 400 + 500 + 600 = 1500
+
+    // Trying to select more than available will return no utxos
+    await expect(bestUtxoSelection(storage, '00', 601)).resolves.toMatchObject({ utxos: [], amount: 0 });
+    await expect(bestUtxoSelection(storage, '01', 1501)).resolves.toMatchObject({ utxos: [], amount: 0 });
+
+    // Trying to select the available amount will return all utxos of that token
+    await expect(bestUtxoSelection(storage, '00', 600)).resolves.toMatchObject({ utxos: [utxos[2], utxos[1], utxos[0]], amount: 600 });
+    await expect(bestUtxoSelection(storage, '01', 1500)).resolves.toMatchObject({ utxos: [utxos[5], utxos[4], utxos[3]], amount: 1500 });
+
+    // Trying to select an amount and an utxo with that amount exists will return only that utxo
+    await expect(bestUtxoSelection(storage, '00', 100)).resolves.toMatchObject({ utxos: [utxos[0]], amount: 100 });
+    await expect(bestUtxoSelection(storage, '00', 200)).resolves.toMatchObject({ utxos: [utxos[1]], amount: 200 });
+    await expect(bestUtxoSelection(storage, '00', 300)).resolves.toMatchObject({ utxos: [utxos[2]], amount: 300 });
+    await expect(bestUtxoSelection(storage, '01', 400)).resolves.toMatchObject({ utxos: [utxos[3]], amount: 400 });
+    await expect(bestUtxoSelection(storage, '01', 500)).resolves.toMatchObject({ utxos: [utxos[4]], amount: 500 });
+    await expect(bestUtxoSelection(storage, '01', 600)).resolves.toMatchObject({ utxos: [utxos[5]], amount: 600 });
+
+    // Trying to select an amount that can be fulfilled with 1 utxo will select the smallest utxo that can fulfill it
+    await expect(bestUtxoSelection(storage, '00', 101)).resolves.toMatchObject({ utxos: [utxos[1]], amount: 200 });
+    await expect(bestUtxoSelection(storage, '00', 201)).resolves.toMatchObject({ utxos: [utxos[2]], amount: 300 });
+    await expect(bestUtxoSelection(storage, '01', 300)).resolves.toMatchObject({ utxos: [utxos[3]], amount: 400 });
+    await expect(bestUtxoSelection(storage, '01', 401)).resolves.toMatchObject({ utxos: [utxos[4]], amount: 500 });
+
+    // Trying to select an amount that can be fulfilled with 2 utxos will select the smallest number of utxos that can fulfill it
+    await expect(bestUtxoSelection(storage, '00', 301)).resolves.toMatchObject({ utxos: [utxos[2], utxos[1]], amount: 500 });
+    await expect(bestUtxoSelection(storage, '00', 501)).resolves.toMatchObject({ utxos: [utxos[2], utxos[1], utxos[0]], amount: 600 });
+    await expect(bestUtxoSelection(storage, '01', 601)).resolves.toMatchObject({ utxos: [utxos[5], utxos[4]], amount: 1100 });
+    await expect(bestUtxoSelection(storage, '01', 1101)).resolves.toMatchObject({ utxos: [utxos[5], utxos[4], utxos[3]], amount: 1500 });
+  }
+
+  test('bestUtxoSelection with memory store', async () => {
+    const store = new MemoryStore();
+    await testBestUtxoSelection(store);
+  });
+
+  test('bestUtxoSelection with indexeddb store', async () => {
+    const xpriv = new HDPrivateKey();
+    const store = new LevelDBStore(DATA_DIR, xpriv.xpubkey);
+    await testBestUtxoSelection(store);
+  });
+
+  /**
+   * Should select the highest utxos until the amount is fulfilled
+   * @param {IStore} store 
+   */
+  async function testFastUtxoSelection(store) {
+    await addUtxosToStore(store);
+    await store.setCurrentHeight(1);
+    const storage = new Storage(store);
+
+    // State of the utxos:
+    // 00: 100 + 200 + 300 = 600
+    // 01: 400 + 500 + 600 = 1500
+
+    // Trying to select more than available will return no utxos
+    await expect(fastUtxoSelection(storage, '00', 601)).resolves.toMatchObject({ utxos: [], amount: 0 });
+    await expect(fastUtxoSelection(storage, '01', 1501)).resolves.toMatchObject({ utxos: [], amount: 0 });
+
+    // Trying to select the available amount will return all utxos of that token, sorted by value descending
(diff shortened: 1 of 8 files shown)


src/new/sendTransaction.ts
```diff
@@ -494,4 +509,162 @@ class SendTransaction extends EventEmitter {
...
+  selectedToken: string,
+): Promise<{success: boolean, message: string}> {
+  const tx = await storage.getTx(input.txId);
+  if (tx === null) {
+    return { success: false, message: `Transaction [${input.txId}] does not exist in the wallet` };
+  }
+  if (tx.is_voided) {
+    return { success: false, message: `Transaction [${input.txId}] is voided` };
+  }
+  if (tx.outputs.length - 1 > input.index) {
```
pedroferreira1, line 637: Shouldn't it be the opposite? 

```suggestion
  if (tx.outputs.length - 1 < input.index) {
```

src/new/sendTransaction.ts
```diff
@@ -494,4 +509,162 @@ class SendTransaction extends EventEmitter {
...
+     */
+    return { success: false, message: `Output [${input.index}] of transaction [${input.txId}] is an authority output` };
+  }
+
+  if (txout.decoded.address) {
+    if (txout.decoded.address !== input.address) {
+      return { success: false, message: `Output [${input.index}] of transaction [${input.txId}] does not have the same address as the provided input` };
+    }
+    if (!await storage.isAddressMine(txout.decoded.address)) {
+      return { success: false, message: `Output [${input.index}] of transaction [${input.txId}] is not yours` };
```
pedroferreira1, line 654:
```suggestion
      return { success: false, message: `Output [${input.index}] of transaction [${input.txId}] is not from the wallet` };
```

src/new/sendTransaction.ts
```diff
@@ -494,4 +509,162 @@ class SendTransaction extends EventEmitter {
...
+  options: IUtxoSelectionOptions = {},
+): Promise<Pick<IDataTx, 'inputs'|'outputs'>> {
+  const token = options.token || HATHOR_TOKEN_CONFIG.uid;
+  const utxoSelection = options.utxoSelectionMethod || bestUtxoSelection;
+  const newtxData: Pick<IDataTx, 'inputs'|'outputs'> = { inputs: [], outputs: [] };
+  let outputAmount = 0;
+
+  // Calculate balance for the token on the transaction
+  for (const output of dataTx.outputs) {
+    if (isDataOutputCreateToken(output)) {
```
pedroferreira1, line 532: With this line are you ignoring melt and mint authorities? Or should you check `isAuthorityOutput(output)`?

src/new/sendTransaction.ts
```diff
@@ -494,4 +509,162 @@ class SendTransaction extends EventEmitter {
...
+        token: utxo.token,
+        address: utxo.address,
+      } as IDataInput;
+    });
+
+    if (newUtxos.amount > outputAmount) {
+      // We need to create a change output
+      const changeAddress = options.changeAddress || await storage.getCurrentAddress();
+      const changeOutput: IDataOutput = {
+        type: 'p2pkh', //get from wallet
```
pedroferreira1, line 568: This should be fixed, right? Otherwise will be wrong for multisig. (the same for the other change)

pedroferreira1 (commented):
I think all new methods from utxo utils and sendTransaction should have unit tests for all possible situations

src/types.ts
```diff
@@ -230,7 +230,7 @@ export interface IMultisigData {
 export interface IUtxoFilterOptions {
   token?: string; // default to HTR
   authorities?: number; // default to 0 (funds)
-  max_utxos?: number; // default to 255 (MAX_INPUTS)
+  max_utxos?: number; // default to unlimited
```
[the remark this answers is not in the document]
r4mmer: By default the selectUtxos was stopping at 255 utxos even if we wanted to iterate on all utxos.
I prefer using the `selectUtxos` method but if we want we can revert this and use `getAllUtxos` instead.

src/types.ts
```diff
@@ -241,6 +241,15 @@ export interface IUtxoFilterOptions {
   reward_lock?: number,
 }
 
+export type UtxoSelectionAlgorithm = (storage: IStorage, token: string, amount: number) => Promise<{ utxos: IUtxo[], amount: number }>;
+
+export interface IUtxoSelectionOptions {
+  token?: string,
+  changeAddress?: string,
+  chooseInputs?: boolean,
+  utxoSelectionMethod?: UtxoSelectionAlgorithm,
```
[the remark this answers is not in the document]
r4mmer: I will create one because it will help with the organization later on, but we still have to decide how will this be configured by the wallet, or if the `SendTransaction` facade will receive the algorithm as an optional argument.

src/utils/utxo.ts
```diff
@@ -0,0 +1,129 @@
...
+        amount: amount,
+      };
+    }
+
+    utxos.push(utxo);
+    utxosAmount += utxo.value;
+
+    if (utxo.value > amount) {
+      // We want to select the smallest utxo that is bigger than the amount
+      if (selectedUtxo === null || utxo.value <= selectedUtxo.value) {
```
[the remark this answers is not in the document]
r4mmer: There is no special case, specially now that we are ordering the utxos.

src/new/sendTransaction.ts
r4mmer: The outputs provided by the user should be in the form of an `ISendOutput` which does not allow authorities.
The only other way to provide outputs would be to provide the `Transaction` instance which does not use this method.

Here we are ignoring "minted custom tokens" and "HTR melted withdrawals" in the outputs provided by the user.
This is to ensure the typescript compiler of the type of the output since `IDataOutputCreateToken` does not have a token property.

src/utils/utxo.ts
```diff
@@ -0,0 +1,129 @@
...
+  let utxosAmount = 0;
+
+  const options = {
+    token,
+    authorities: 0,
+    target_amount: amount,
+    only_available_utxos: true,
+  };
+
+  for await (const utxo of storage.selectUtxos(options)) {
```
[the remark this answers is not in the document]
r4mmer: It will now since we created the `order_by_value` option.

__tests__/utils/utxo.test.js
```diff
@@ -0,0 +1,178 @@
...
+
+    // Selection will always return the highest utxos first until the amount is fulfilled
+    await expect(fastUtxoSelection(storage, '00', 100)).resolves.toMatchObject({ utxos: [utxos[2]], amount: 300 });
+    await expect(fastUtxoSelection(storage, '00', 200)).resolves.toMatchObject({ utxos: [utxos[2]], amount: 300 });
+    await expect(fastUtxoSelection(storage, '00', 300)).resolves.toMatchObject({ utxos: [utxos[2]], amount: 300 });
+    await expect(fastUtxoSelection(storage, '01', 400)).resolves.toMatchObject({ utxos: [utxos[5]], amount: 600 });
+    await expect(fastUtxoSelection(storage, '01', 500)).resolves.toMatchObject({ utxos: [utxos[5]], amount: 600 });
+    await expect(fastUtxoSelection(storage, '01', 600)).resolves.toMatchObject({ utxos: [utxos[5]], amount: 600 });
+
+    // Trying to select an amount that can be fulfilled with 1 utxo will select the smallest utxo that can fulfill it
```
tuliomir, line 154:
```suggestion
    // Trying to select an amount that can be fulfilled with 1 utxo will select the biggest utxo that can fulfill it
```

src/utils/utxo.ts
```diff
@@ -0,0 +1,159 @@
...
+    };
+  } else if (utxosAmount < amount) {
+    // We don't have enough funds
+    return {
+      utxos: [],
+      amount: 0,
+    };
+  } else {
+    // We need to ensure we use the smallest number of utxos and avoid hitting the maximum number of inputs
+    // This can be done by ordering the utxos by value and selecting the highest values first untill the amount is fulfilled
```
tuliomir, line 142:
```suggestion
    // This can be done by ordering the utxos by value and selecting the highest values first until the amount is fulfilled
```

src/utils/utxo.ts
```diff
@@ -0,0 +1,159 @@
...
+    // But since the store ensures the utxos are ordered by value descending, we can just iterate on them to achieve the same result
+    // This is emsured due to options.order_by_value = 'desc' on the selectUtxos method.
```
tuliomir, line 144:
```suggestion
    // This is ensured due to options.order_by_value = 'desc' on the selectUtxos method.
```

src/utils/utxo.ts
```diff
@@ -0,0 +1,159 @@
...
+      if (selectedUtxo === null || utxo.value < selectedUtxo.value) {
+        selectedUtxo = utxo;
+      }
+    }
+
+    if ((utxo.value < amount) && (selectedUtxo !== null)) {
+      // We already have an utxo that is bigger than the amount required
+      // with the lowest possible value.
+      // We don't need to iterate more
+      break;
```
pedroferreira1, line 125: Why don't you return here directly?

```
  return {
    utxos: [selectedUtxo],
    amount: selectedUtxo.value,
  };
```

src/utils/utxo.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
```
pedroferreira1, line 126: If you add here:

```
if (utxo.value < amount) && utxosAmount >= amount) {
  break;
}
```

You guarantee that the `utxos` array will have the ones you want and prevent you from iterating again, doesn't it?

src/new/sendTransaction.ts
```diff
@@ -494,4 +509,182 @@ class SendTransaction extends EventEmitter {
...
+  }
+
+  if (options.chooseInputs) {
+    if (outputAmount === 0) {
+      // We cannot process a target amount of 0 tokens.
+      throw new Error('Invalid amount of tokens to send.');
+    }
+
+    // We will choose the inputs to fill outputAmount.funds
+    const newUtxos = await utxoSelection(storage, token, outputAmount);
```
pedroferreira1, line 552: If `newUtxos.inputs === []`, it means we can't fulfill this request for utxos, so should we throw an error?

src/new/sendTransaction.ts
```diff
@@ -494,4 +509,162 @@ class SendTransaction extends EventEmitter {
...
+): Promise<Pick<IDataTx, 'inputs'|'outputs'>> {
+  const token = options.token || HATHOR_TOKEN_CONFIG.uid;
+  const utxoSelection = options.utxoSelectionMethod || bestUtxoSelection;
+  const newtxData: Pick<IDataTx, 'inputs'|'outputs'> = { inputs: [], outputs: [] };
+  let outputAmount = 0;
+
+  // Calculate balance for the token on the transaction
+  for (const output of dataTx.outputs) {
+    if (isDataOutputCreateToken(output)) {
+      // This is a mint output
```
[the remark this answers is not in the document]
r4mmer: This is only because the `output.token` property does not exist on `type: 'mint'` but it should also be ignored if the output was a `type: 'melt'` which would make this an HTR withdrawal, so no input would be needed for the withdrawal.

This check is mostly for the type since the facade does not allow for token operations or authority inputs/outputs of any kind when using the inputs and outputs format, so we will not have a mint or melt operation here in any way.

src/utils/utxo.ts
r4mmer: It will be returned just after the loop, so I thought we could just break instead of returning the selectedUtxo in 2 places

src/utils/utxo.ts
r4mmer: Done

src/utils/utxo.ts
r4mmer: Done

__tests__/utils/utxo.test.js
r4mmer: Done

src/utils/utxo.ts
r4mmer: Done

src/new/sendTransaction.ts
r4mmer: Did the same check but with `newUtxos.amount < outputAmount`

src/utils/utxo.ts
r4mmer: Done

src/storage/leveldb/utxo_index.ts
```diff
@@ -278,7 +293,7 @@ export default class LevelUtxoIndex implements IKVUtxoIndex {
       utxoNum += 1;
       sumAmount += utxo.value;
 
-      if ((options.target_amount && sumAmount >= options.target_amount) || (utxoNum >= maxUtxos)) {
+      if ((options.target_amount && sumAmount >= options.target_amount) || (options.max_utxos && utxoNum >= options.max_utxos)) {
```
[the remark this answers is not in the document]
r4mmer: Fixed by breaking the line

src/new/sendTransaction.ts
```diff
@@ -188,22 +190,33 @@ class SendTransaction extends EventEmitter {
...
+      partialTxData.inputs.push(...proposedData.inputs);
+      partialTxData.outputs.push(...proposedData.outputs);
+    }
+
+    let outputs: IDataOutput[];
+    if (partialTxData.outputs.length === 0) {
+      outputs = txData.outputs;
+    } else {
+      // Shuffle outputs, so we don't have change output always in the same index
+      outputs = shuffle([...txData.outputs, ...partialTxData.outputs])
```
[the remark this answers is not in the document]
r4mmer: I'm not sure, the idea of shuffling is so that we can prevent someone from knowing 2 addresses are from the same wallet by identifying an output as change.

So not having the change makes the shuffle not really necessary

src/new/sendTransaction.ts
r4mmer: Done

src/new/sendTransaction.ts
```diff
@@ -494,4 +509,182 @@ class SendTransaction extends EventEmitter {
...
+        token: utxo.token,
+        address: utxo.address,
+      } as IDataInput;
+    });
+
+    if (newUtxos.amount > outputAmount) {
+      // We need to create a change output
+      const changeAddress = options.changeAddress || await storage.getCurrentAddress();
+      const walletType = await storage.getWalletType();
+      let outputType: 'p2pkh'|'p2sh';
```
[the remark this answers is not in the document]
r4mmer: Done

src/utils/utxo.ts
pedroferreira1: I saw you did this break but with it can't you remove the loop below?

pedroferreira1 (approved):

PR #505, opened 2023-04-12
fix: check input token in sendTransaction

r4mmer:
### Acceptance Criteria
- Check provided input token before checking validity

### Security Checklist


src/new/sendTransaction.ts
```diff
@@ -599,6 +599,11 @@ export async function prepareSendTokensData(
   } else {
     let inputAmount = 0;
     for (const input of dataTx.inputs) {
+      if (input.token != token) {
```
tuliomir, line 602: Could we use `!==` instead? It's more restrictive and safe for validations

tuliomir (approved):

PR #509, opened 2023-04-18
fix: account path not checking for null before decrypting

andreabadesso:
### Acceptance Criteria
- `getAcctPathXprivKey` should return null if it doesn't exist in storage

### Security Checklist

### DIFF
--- a/src/wallet.js
+++ b/src/wallet.js
@@ -34,7 +34,7 @@ import helpers from './helpers';
 import helperUtils from './utils/helpers';
 import walletUtils from './utils/wallet';
 import { getAddressType } from './utils/address';
-import { AddressError, ConstantNotSet, OutputValueError, WalletTypeError } from './errors';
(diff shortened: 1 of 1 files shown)


src/wallet.js
```diff
@@ -2894,6 +2894,11 @@ const wallet = {
 
     const accessData = this.getWalletAccessData();
     const encryptedXPriv = accessData.acctPathMainKey;
+
```
pedroferreira1, line 2897: I think we should just update the docstring to make it clear that the return can be `null`

pedroferreira1 (approved):

src/wallet.js
andreabadesso: Done in 2717da25822ec6b1fb20eeaaf99381fcc0680638

msbrogli (approved):

PR #510, opened 2023-04-19
feat: validate user provided change address

r4mmer:
### Acceptance Criteria
- Create a method to validate user provided change address and throw appropriate error
  - If no address is provided by the user, get one from the wallet.
- Change utils and functions that may use user provided change or choose one to use this new method.
- Expose number utils since it should be used on the wallets

### Security Checklist

### DIFF
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -142,8 +142,8 @@ export class Storage implements IStorage {
   }
 
   /**
-   * Get a change address to use, if one is provided we need to check we own its
-   * If not provided the current address will be used instead.
+   * Get a change address to use, if one is provided we need to check if we own it
+   * If not provided, the current address will be used instead.
    *
    * @param {Object} [options={}]
    * @param {string|null|undefined} [options.changeAddress=undefined] User provided change address to use
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -142,6 +142,25 @@ export class Storage implements IStorage {
   }
 
   /**
+   * Get a change address to use, if one is provided we need to check we own its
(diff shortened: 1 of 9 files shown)


src/storage/storage.ts
```diff
@@ -141,6 +141,25 @@ export class Storage implements IStorage {
     return this.store.getCurrentAddress(markAsUsed);
   }
 
+  /**
+   * Get a change address to use, if one is provided we need to check we own its
+   * If not provided the current address will be used instead.
```
tuliomir, line 146:
```suggestion
   * Get a change address to use, if one is provided we need to check if we own it
   * If not provided, the current address will be used instead.
```

src/storage/storage.ts
```diff
@@ -141,6 +141,25 @@ export class Storage implements IStorage {
...
+   *
+   * @param {Object} [options={}]
+   * @param {string|null|undefined} [options.changeAddress=undefined] User provided change address to use
+   * @returns {Promise<string>} The change address to use
+   */
+  async getChangeAddress({ changeAddress }: { changeAddress?: string | null | undefined; } = {}): Promise<string> {
+    if (changeAddress) {
```
tuliomir, line 153: As a suggestion for [flattening the code](https://refactoring.guru/replace-nested-conditional-with-guard-clauses) with guard clauses:
```ts
if (!changeAddress) {
  return getCurrentAddress();
}

if (!isAddressMine) {
  throw "Not from wallet"
}

return changeAddress;
```

tuliomir (approved):

pedroferreira1 (approved):

PR #512, opened 2023-04-19
feat: generate access data with account path key

r4mmer:
### Acceptance Criteria
- Generate account path key on the access data where possible.
- storage should have a method to get the account path key.

### Security Checklist

### DIFF
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -726,7 +726,7 @@ export class Storage implements IStorage {
    */
   async getAcctPathXPrivKey(pinCode: string): Promise<string> {
     const accessData = await this._getValidAccessData();
-    if (accessData.acctPathKey === undefined) {
+    if (!accessData.acctPathKey) {
       throw new Error('Private key is not present on this wallet.');
     }
 
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -724,6 +724,27 @@ export class Storage implements IStorage {
   }
 
   /**
+   * Get account path xprivkey if available.
+   *
+   * @param {string} pinCode
+   * @returns {Promise<string>}
+   */
+  async getAcctPathXPrivKey(pinCode: string): Promise<string> {
+    const accessData = await this._getValidAccessData();
+    if (accessData.acctPathKey === undefined) {
+      throw new Error('Private key is not present on this wallet.');
+    }
+
+    try {
+      // decryptData handles pin validation
+      return decryptData(accessData.acctPathKey, pinCode);
+    } catch(err: unknown) {
+      // FIXME: check error type to not hide crypto errors
+      throw new Error('Invalid PIN code.');
+    }
+  }
+
+  /**
    * Decrypt and return the auth private key of the wallet.
    *
    * @param {string} pinCode Pin to unlock the private key
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -714,13 +714,8 @@ export class Storage implements IStorage {
       throw new Error('Private key is not present on this wallet.');
     }
 
-    try {
-      // decryptData handles pin validation
-      return decryptData(accessData.mainKey, pinCode);
-    } catch(err: unknown) {
-      // FIXME: check error type to not hide crypto errors
-      throw new Error('Invalid PIN code.');
-    }
+    // decryptData handles pin validation
+    return decryptData(accessData.mainKey, pinCode);
   }
 
   /**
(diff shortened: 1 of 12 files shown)


src/storage/storage.ts
```diff
@@ -699,6 +699,27 @@ export class Storage implements IStorage {
...
 
+  /**
+   * Get account path xprivkey if available.
+   *
+   * @param {string} pinCode
+   * @returns {Promise<string>}
+   */
+  async getAcctPathXPrivKey(pinCode: string): Promise<string> {
+    const accessData = await this._getValidAccessData();
+    if (accessData.acctPathKey === undefined) {
```
andreabadesso, line 710: Should we also check for `null`?

src/storage/storage.ts
```diff
@@ -699,6 +699,27 @@ export class Storage implements IStorage {
...
+      throw new Error('Private key is not present on this wallet.');
+    }
+
+    try {
+      // decryptData handles pin validation
+      return decryptData(accessData.acctPathKey, pinCode);
+    } catch(err: unknown) {
+      // FIXME: check error type to not hide crypto errors
```
andreabadesso, line 718: Can we do it?

src/storage/storage.ts
```diff
@@ -699,6 +699,27 @@ export class Storage implements IStorage {
...
+  /**
+   * Get account path xprivkey if available.
+   *
+   * @param {string} pinCode
+   * @returns {Promise<string>}
+   */
+  async getAcctPathXPrivKey(pinCode: string): Promise<string> {
+    const accessData = await this._getValidAccessData();
+    if (accessData.acctPathKey === undefined) {
+      throw new Error('Private key is not present on this wallet.');
```
andreabadesso, line 711: The current implementation of the `getAcctPathXPrivKey` returns `null` when it's not able to get it from the `accessData`

I agree with refactoring it to throw an error instead, but we must remember to update the mobile-wallet accordingly [here](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/25519b9d6e068270cb96f9480082e2f595dcd5ee/src/screens/PinScreen.js#L164)

src/utils/wallet.ts
```diff
@@ -448,11 +448,14 @@ const wallet = {
 
     const argXpriv = new HDPrivateKey(xprivkey);
     let xpriv: HDPrivateKey;
+    let acctXpriv: HDPrivateKey|null = null;
```
andreabadesso, line 451:
```suggestion
    let acctXpriv: HDPrivateKey | null = null;
```

src/utils/wallet.ts
```diff
@@ -475,13 +478,21 @@ const wallet = {
...
-    return {
+    const accessData: IWalletAccessData = {
       walletType,
       multisigData,
       mainKey: encryptedMainKey,
       xpubkey: xpriv.xpubkey,
       walletFlags: 0,
     };
+
+    if (acctXpriv !== null) {
```
andreabadesso, line 489: Should we also check for `undefined`?

Maybe `acctXpriv != null`

andreabadesso (dismissed):

src/storage/storage.ts
r4mmer: We can but `IWalletAccessData` does not allow null as a value for this.

src/utils/wallet.ts
r4mmer: The variable is initialized as null and bitcore-libs derivation does not return undefined.

src/storage/storage.ts

src/utils/wallet.ts
r4mmer: Done

src/utils/wallet.ts
```diff
@@ -475,13 +478,21 @@ const wallet = {
...
+      // Account path key will only be available if the provided key is a root key
+      const encryptedAuthPathKey = encryptData(acctXpriv.xprivkey, pin);
```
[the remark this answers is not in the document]
r4mmer: Done

src/utils/wallet.ts
```diff
@@ -475,13 +499,26 @@ const wallet = {
...
+      // Account path key will only be available if the provided key is a root key
+      const encryptedAcctPathKey = encryptData(acctXpriv.xprivkey, pin);
+      accessData.acctPathKey = encryptedAcctPathKey;
+    }
+
+    if (!!(seed && password)) {
+      const encryptedWords = encryptData(seed, password);
+      accessData.words = encryptedWords;
+    }
+
```
pedroferreira1, line 520: In order to use this method instead of `generateAccessDataFromSeed`, we would need to also add `authKey`. Does it make sense to add it?

r4mmer: Makes sense, added authKey when starting with xpriv as well.
But since the path to the authKey is special, we can only do so when starting from the root xprivkey

tuliomir (approved):

PR #516, opened 2023-04-27
feat: new token utils

r4mmer:
### Acceptance Criteria
Add token utilities required by the mobile wallet
- check that a token is registered by the uid
- get deposit percentage configuration for the current network
- validate token configuration string
- util to check for duplicate name or symbol on the registered tokens

### Security Checklist

### DIFF
--- a/__tests__/utils/helpers.test.js
+++ b/__tests__/utils/helpers.test.js
@@ -267,3 +267,9 @@ test('handlePrepareDataError', () => {
 
   expect(() => { helpers.handlePrepareDataError(err) }).toThrow(err);
 });
+
+test('cleanupString', () => {
+  expect(helpers.cleanupString('str 1')).toEqual('str 1');
+  expect(helpers.cleanupString('str  2')).toEqual('str 2');
+  expect(helpers.cleanupString('str\t 3')).toEqual('str 3');
+});
--- a/__tests__/utils/helpers.test.js
+++ b/__tests__/utils/helpers.test.js
@@ -272,6 +272,6 @@ test('cleanupString', () => {
   expect(helpers.cleanupString('str 1')).toEqual('str 1');
   expect(helpers.cleanupString('str  2')).toEqual('str 2');
   expect(helpers.cleanupString('str\t 3')).toEqual('str 3');
-  expect(helpers.cleanupString('  Str\t 4   \t')).toEqual(' str 4 ');
+  expect(helpers.cleanupString('  Str\t 4   \t')).toEqual('str 4');
   expect(helpers.cleanupString('STR 5')).toEqual('str 5');
 });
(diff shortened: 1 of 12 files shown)


src/storage/storage.ts
```diff
@@ -72,6 +72,21 @@ export class Storage implements IStorage {
...
+   */
+  getTokenDepositPercentage(): number {
+    if (this.version && this.version.token_deposit_percentage) {
+      return this.version.token_deposit_percentage;
+    }
+    /**
+     *  When using wallet-service facade we do not update the version constants
+     *  Since this data is important for the wallets UI we will return the default value here.
+     */
+    return TOKEN_DEPOSIT_PERCENTAGE;
```
glevco, line 87: If the `token_deposit_percentage` is `0`, should we return `0` or the default?

Also, could this be simplified?

```suggestion
    return this.version?.token_deposit_percentage ?? TOKEN_DEPOSIT_PERCENTAGE
```

(either `??` or `||` should be used, depending on the first question)

glevco (approved):

src/storage/storage.ts
r4mmer: `token_deposit_percentage` is returned from the fullnode api, it won't be 0 on our networks but I think a private network could have 0 for it.

pedroferreira1, line 87: The wallet service has a table for constants, and it also needs this percentage information. How was it handled before?

__tests__/utils/helpers.test.js
```diff
@@ -267,3 +267,9 @@ test('handlePrepareDataError', () => {
 
   expect(() => { helpers.handlePrepareDataError(err) }).toThrow(err);
 });
+
+test('cleanupString', () => {
+  expect(helpers.cleanupString('str 1')).toEqual('str 1');
+  expect(helpers.cleanupString('str  2')).toEqual('str 2');
+  expect(helpers.cleanupString('str\t 3')).toEqual('str 3');
```
pedroferreira1, line 274: You could add spaces in the begin and in the end, and also uppercase

src/storage/storage.ts
r4mmer: Good observation, after checking I saw that the `getDepositPercentage` from the token utils would retrieve the deposit from storage (it was set by the version api, which the wallet-service does not use, hence the default) and this would be used internally by other token utils, this is why we had the default in the method instead of the wallet-service.

I'm changing so that the value can be received via argument, and use the default if not present, all instances on the wallet should retrieve from storage and pass as an argument and the other token utils also receive the storage as a parameter so they can get it from there as well, sounds good?

__tests__/utils/helpers.test.js
r4mmer: Done

pedroferreira1 (approved):

PR #518, opened 2023-05-04
feat: leveldb dir name

r4mmer:
### Acceptance Criteria
- allow leveldb store to use any directory name

### Security Checklist

### DIFF
--- a/src/storage/leveldb/store.ts
+++ b/src/storage/leveldb/store.ts
@@ -152,10 +152,12 @@ export default class LevelDBStore implements IStore {
         }
       }
     }
-    if ((await this.walletIndex.getCurrentAddressIndex()) < maxIndex) {
-      await this.walletIndex.setCurrentAddressIndex(Math.min(maxIndex + 1, await this.walletIndex.getLastLoadedAddressIndex()));
+    if (maxIndex >= 0) {
+      if ((await this.walletIndex.getCurrentAddressIndex()) < maxIndex) {
+        await this.walletIndex.setCurrentAddressIndex(Math.min(maxIndex + 1, await this.walletIndex.getLastLoadedAddressIndex()));
+      }
+      await this.walletIndex.setLastUsedAddressIndex(maxIndex);
     }
-    await this.walletIndex.setLastUsedAddressIndex(maxIndex);
   }
 
   async getTx(txId: string): Promise<IHistoryTx | null> {
@@ -314,7 +316,7 @@ export default class LevelDBStore implements IStore {
     await this.walletIndex.setItem(key, value);
   }
 
-  async cleanStorage(cleanHistory?: boolean | undefined, cleanAddresses?: boolean | undefined): Promise<void> {
+  async cleanStorage(cleanHistory: boolean = false, cleanAddresses: boolean = false): Promise<void> {
     if (cleanHistory) {
       await this.tokenIndex.clear();
       await this.historyIndex.clear();
--- a/src/storage/leveldb/store.ts
+++ b/src/storage/leveldb/store.ts
@@ -323,8 +323,6 @@ export default class LevelDBStore implements IStore {
   }
 
   async cleanStorage(cleanHistory?: boolean | undefined, cleanAddresses?: boolean | undefined): Promise<void> {
-    // set access data to null
-    await this.walletIndex.cleanAccessData();
     if (cleanHistory) {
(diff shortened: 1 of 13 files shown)


src/storage/leveldb/store.ts
```diff
@@ -21,13 +20,9 @@ export default class LevelDBStore implements IStore {
...
   dbpath: string;
-  xpubkey: string;
 
-  constructor(dbroot: string, xpubkey: string) {
-    // The xpubkey in the account or change path?
-    this.xpubkey = xpubkey;
-    const xpub = HDPublicKey.fromString(xpubkey);
-    const dbpath = path.join(dbroot, xpub.publicKey.toString());
+  constructor(dirpath: string, dbroot: string = 'hathor.data') {
+    const dbpath = path.join(dbroot, dirpath);
```
tuliomir, line 25: If we add characters that are not accepted by the filesystem, will be error thrown here be easy to understand?

If the error message is too vague or misleading, I would suggest adding a validation and/or a `snakeCase(deburr(input))` kind of operation here ( reference [link](https://lodash.com/docs/4.17.15#deburr) )

src/storage/leveldb/store.ts
```diff
@@ -156,10 +151,12 @@ export default class LevelDBStore implements IStore {
         }
       }
     }
-    if ((await this.walletIndex.getCurrentAddressIndex()) < maxIndex) {
-      await this.walletIndex.setCurrentAddressIndex(Math.min(maxIndex + 1, await this.walletIndex.getLastLoadedAddressIndex()));
+    if (maxIndex >= 0) {
```
tuliomir, line 154: I feel it would be nice to have a comment about why this `maxIndex >= 0` validation is necessary

If I understand correctly, we're just trying to avoid calculating addresses when the wallet has never used any address beyond index 0.

src/storage/leveldb/wallet_index.ts
```diff
@@ -53,6 +53,9 @@ export default class LevelWalletIndex implements IKVWalletIndex {
    * @param {number} value The value to set.
    */
   async _setNumber(dest: 'access'|'wallet', key: string, value: number) {
+    if (value < 0) {
+      return;
```
tuliomir, line 57: Shouldn't we throw here, instead of silently ignoring the call?

src/storage/storage.ts
```diff
@@ -734,7 +734,7 @@ export class Storage implements IStorage {
     }
     this.version = null;
     if (cleanStorage || cleanAddresses) {
-      this.cleanStorage(cleanStorage, cleanAddresses);
+      await this.cleanStorage(cleanStorage, cleanAddresses);
```
tuliomir, line 737: The `cleanStorage` method calls the first parameter `cleanHistory`, which I think is more accurate than `cleanStorage`.

What do you think about renaming this here as well? Would it be within the scope of this PR?

src/storage/leveldb/store.ts
r4mmer: Well, we use -1 as a starting point and if no address is used (should not happen, but it did on some tests) we will try to save -1 on the leveldb, but we encode it as unsigned int, which throws an error since -1 is negative.

This check is only to avoid this error, since negative address index should be impossible in bip39 we are not avoiding any actual cases.

src/storage/leveldb/wallet_index.ts
r4mmer: This is to avoid errors on the tests, but maybe we should let this throw and make checks before this happens.

src/storage/storage.ts
r4mmer: We should make it outside this PR, but I agree this is a better name

src/storage/leveldb/wallet_index.ts
alexruzenhack: I agree it would be better to have the assert and exception here. However, it would affect many other functions not related with theme of the PR. I suggest to make this change in another PR.

alexruzenhack (approved):
I only suggest to make the `_setNumber` refactoring in another PR.

src/storage/leveldb/store.ts
r4mmer: This is a good idea, but I think we should have this on a separate PR since it could also be used in other places.

tuliomir (approved):

PR #521, opened 2023-05-13
feat: add new parameter to set authority address when minting/melting tokens

pedroferreira1:
### Acceptance Criteria
- Add parameter to set new created authority address when minting/melting tokens
- This address by default must belong to the same wallet but another parameter was added to allow external addresses.
- I decided not to add it to the wallet service facade because it's a specific request and it will be added to both facades in the release candidate version
- Added the stable master branch to the github workflow files just to trigger the CI, it will be removed before merging

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -1845,7 +1845,7 @@ class HathorWallet extends EventEmitter {
    *                                   (if not sent we choose the next available address to use)
    * @param {boolean} [options.startMiningTx=true] boolean to trigger start mining (default true)
    * @param {boolean} [options.createAnotherMint] boolean to create another mint authority or not
-   *                                              for the wallet
+   * @param {string} [options.mintAuthorityAddress] the address to send the new mint authority created
    * @param {string} [options.pinCode] pin to decrypt xpriv information.
    *                                   Optional but required if not set in this
    *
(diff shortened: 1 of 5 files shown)


r4mmer (approved):

src/new/wallet.js
```diff
@@ -1903,7 +1907,18 @@ class HathorWallet extends EventEmitter {
...
-    const ret = tokens.generateMeltData(meltInput[0], tokenUid, amount, pin, newOptions.createAnotherMelt, { depositAddress: newOptions.address, changeAddress: newOptions.changeAddress });
+    const ret = tokens.generateMeltData(
+      meltInput[0],
+      tokenUid,
+      amount,
+      pin,
+      newOptions.createAnotherMelt,
+      {
+        depositAddress: newOptions.address,
+        changeAddress: newOptions.changeAddress,
```
msbrogli, line 1918: Maybe add an option `allowExternalAddress: true/false`? I guess it would prevent operational mistakes.

msbrogli (changes requested):

src/new/wallet.js
pedroferreira1: Added the parameter with a different name because there are 3 addresses as parameters.

msbrogli: I just checked the changes. I understand that there might be more than one address in several cases. In this case, I'd rather use an `allowExternalAddresses: string[]` (or a set) that contains a list of authorized external addresses to be used anywhere in the method.

The current code base is good and we can also move forward with it and change it later.

What do you think? Change now or later?

msbrogli (approved):

r4mmer (approved):

src/new/wallet.js
msbrogli: We just had a conversation about it and I agree that the current implementation is better than my suggestion.

PR #522, opened 2023-05-15
fix created leveldb dir name

r4mmer:
### Acceptance Criteria

Some tests merged after we changed how we name the leveldb store used the old naming system.
This only affects some tests but when running they will leave a directory with the xpub name instead of creating inside the `testdata.leveldb` like other tests.

- Use the correct parameters when instantiating the leveldb store on tests

### Security Checklist

### DIFF
--- a/__tests__/storage/common_store.test.js
+++ b/__tests__/storage/common_store.test.js
@@ -8,7 +8,7 @@
 import { LevelDBStore, MemoryStore, Storage } from "../../src/storage";
 import { HDPrivateKey } from "bitcore-lib";
 import { TOKEN_AUTHORITY_MASK, TOKEN_MINT_MASK } from "../../src/constants";
-import { crypto } from "bitcore-lib";
+import walletUtils from "../../src/utils/wallet";
 
 const DATA_DIR = './testdata.leveldb';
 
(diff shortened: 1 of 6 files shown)


__tests__/storage/common_store.test.js
```diff
@@ -177,7 +177,8 @@ describe('registered tokens', () => {
 
   it('should work with leveldb store', async () => {
     const xpriv = new HDPrivateKey();
-    const store = new LevelDBStore(DATA_DIR, xpriv.xpubkey);
+    const walletId = crypto.Hash.sha256(Buffer.from(xpriv.xpubkey)).toString('hex');
```
pedroferreira1, line 180: Can we wait the other PR and use the util created here?

tuliomir: If we do, I think it's important to also refactor the other instances of similar code in the tests ( about 14, on my latest search )

tuliomir (approved):

__tests__/utils/utxo.test.js
```diff
@@ -1,13 +1,14 @@
 /**
  * Copyright (c) Hathor Labs and its affiliates.
- *
+ *wallet
```
pedroferreira1, line 3: Remove?

pedroferreira1 (approved):

__tests__/utils/utxo.test.js
r4mmer: Done

PR #524, opened 2023-05-16
feat: wallet service storage argument

r4mmer:
### Acceptance Criteria
- Allow starting a wallet-service facade passing the storage as argument
- When starting the wallet-service facade, use the accessData present on the storage if it exists.

### Security Checklist

### DIFF
--- a/__tests__/wallet/wallet.test.ts
+++ b/__tests__/wallet/wallet.test.ts
@@ -1348,4 +1348,16 @@ test('start', async () => {
     authKey: expect.anything(),
   });
   await wallet.stop();
+
+  // Check we throw an error when giving an invalid authxpriv
+  expect(() => {
+    return new HathorWalletServiceWallet({
+      requestPassword,
+      xpriv: acctKey,
+      authxpriv: 'invalid-xprivkey',
+      network,
+      passphrase: '',
+      storage,
+    });
+  }).toThrow('authxpriv parameter is an invalid hd privatekey');
 });
--- a/__tests__/wallet/wallet.test.ts
+++ b/__tests__/wallet/wallet.test.ts
@@ -21,6 +21,11 @@ import { TxNotFoundError, SendTxError } from '../../src/errors';
 import SendTransactionWalletService from '../../src/wallet/sendTransactionWalletService';
 import transaction from '../../src/utils/transaction';
 import { TOKEN_MELT_MASK, TOKEN_MINT_MASK } from '../../src/constants';
+import { MemoryStore, Storage } from '../../src/storage';
+import walletApi from '../../src/wallet/api/walletApi';
+import { WalletStatus } from '../../src/wallet/types';
+import wallet from '../../src/utils/wallet';
+import { resolve } from 'path';
 
 // Mock SendTransactionWalletService class so we don't try to send actual transactions
 // TODO: We should refactor the way we use classes from inside other classes. Using dependency injection would facilitate unit tests a lot and avoid mocks like this.
@@ -798,7 +803,7 @@ test('prepareDestroyAuthority', async () => {
   // Clear mocks
   spy1.mockRestore();
   spy2.mockRestore();
-  spy3.mockRestore();  
+  spy3.mockRestore();
 });
 
 test('destroyAuthority should throw if wallet is not ready', async () => {
@@ -1260,4 +1265,50 @@ test('createNFTs', async () => {
   spy2.mockRestore();
   spy3.mockRestore();
   spy4.mockRestore();
+});
+
+test('start', async () => {
+  const requestPassword = jest.fn();
+  const network = new Network('testnet');
+  const seed = defaultWalletSeed;
+  const store = new MemoryStore();
+  const storage = new Storage(store);
+
+  jest.spyOn(HathorWalletServiceWallet.prototype, 'pollForWalletStatus').mockImplementation(() => Promise.resolve());
+  jest.spyOn(HathorWalletServiceWallet.prototype, 'setupConnection').mockImplementation(jest.fn());
+  jest.spyOn(walletApi, 'getNewAddresses')
+    .mockImplementation(() => Promise.resolve({ success: true, addresses: [] }));
+  jest.spyOn(walletApi, 'createWallet')
+    .mockImplementation(() => Promise.resolve({
+      success: true,
+      status: {
+        walletId: 'id',
+        xpubkey: 'xpub',
+        status: 'creating',
+        maxGap: 20,
+        createdAt: 0,
+        readyAt: 0,
+      },
+    }));
+
+  const wallet1 = new HathorWalletServiceWallet({
+    requestPassword,
+    seed,
+    network,
+    passphrase: '',
+    storage,
+  });
+  await wallet1.start({ pinCode: '1234', password: '1234' });
+  await wallet1.stop();
+
+  // The storage should have the accessData
+  const wallet2 = new HathorWalletServiceWallet({
+    requestPassword,
+    seed,
+    network,
+    passphrase: '',
+    storage,
+  });
+  await wallet2.start({ pinCode: '1234', password: '1234' });
+  await wallet2.stop();
 });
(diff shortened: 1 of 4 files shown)


__tests__/wallet/wallet.test.ts
```diff
@@ -746,4 +751,51 @@ test('sendTransaction', async () => {
...
+
+  // The storage should have the accessData
+  const wallet2 = new HathorWalletServiceWallet({
+    requestPassword,
+    seed,
+    network,
+    passphrase: '',
+    storage,
+  });
+  await wallet2.start({ pinCode: '1234', password: '1234' });
```
tuliomir, line 798: I feel we should have some assertions here that confirm the storage's access data is being actually read from, and not re-created from scratch.

src/wallet/wallet.ts
```diff
@@ -126,14 +126,16 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    storage = null,
   }: {
     requestPassword: Function,
     seed?: string | null,
     xpriv?: string | null,
     xpub?: string | null,
     network: Network,
     passphrase?: string,
-    enableWs: boolean,
+    enableWs?: boolean,
```
tuliomir, line 137: If this is an optional boolean, could we consider the default value to be `false`?

__tests__/wallet/wallet.test.ts
r4mmer: Done

src/wallet/wallet.ts
```diff
@@ -94,6 +95,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
   private xpub: string | null;
   // Xpriv of the wallet on the account derivation path
   private xpriv: string | null;
+  private authxpriv: string | null;
```
pedroferreira1, line 98: Missing comment, which path is this xpriv from?

pedroferreira1, line 98: Is this the same as the attribute below? `authPrivKey`

r4mmer: Removed from the properties, but it is still on the constructor arguments.

The reson for this is that we used to expect the auth xpriv to be persisted through the access data on storage, but if the access data is persisted, we do not need to use the xpriv to derive the access data since the persisted storage will have the entire access data.

Since we want to preserve this feature we need to have the auth xpriv either derived or passed as an argument.
And this facade expects the xpriv to be the account path xpriv which cannot be used to derive the auth path xpriv.
So the only option is to require the auth key xpriv in the constructor.

I removed the `authxpriv` property since as you pointed out, we already have a property for this.

r4mmer: This property has been removed and we will use the `authPrivKey` instead

src/wallet/wallet.ts
```diff
@@ -302,13 +315,59 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+        const changeLevelPrivKey = accountLevelPrivKey.deriveNonCompliantChild(0);
+
+        accessData = walletUtils.generateAccessDataFromXpriv(
+          changeLevelPrivKey.xprivkey,
+          {
+            pin: pinCode,
+            // multisig: not implemented on wallet service yet
+          },
+        );
+        accessData.authKey = encryptData(this.authPrivKey.xprivkey!, pinCode);
```
pedroferreira1, line 364: Why don't we move this line to the `generateAccessDataFromXpriv` method? We already handle this for other accessData keys there, right?

src/wallet/wallet.ts
```diff
@@ -165,6 +178,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     this.xpriv = xpriv;
     this.seed = seed;
     this.xpub = xpub;
+    this.authPrivKey = authxpriv ? bitcore.HDPrivateKey(authxpriv) : null;
```
andreabadesso, line 181: Should we validate that `authxpriv` is a valid xprivkey and throw a proper error?

We can use [`bitcore.HDPrivateKey.isValidSerialized`](https://github.com/bitpay/bitcore/blob/88318365e65509a386376f39cd6b4579063cf654/packages/bitcore-lib/lib/hdprivatekey.js#L300)

r4mmer: Done

src/wallet/wallet.ts
```diff
@@ -302,13 +315,59 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      const accessData = await this.storage.getAccessData();
+      hasAccessData = !!accessData;
+    } catch (err) {
+      if (err instanceof UninitializedWalletError) {
+        hasAccessData = false;
+      } else {
+        throw err;
+      }
+    }
+    if (!hasAccessData) {
```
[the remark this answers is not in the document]
r4mmer: Done

src/wallet/wallet.ts
```diff
@@ -302,13 +315,59 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+        accessData = walletUtils.generateAccessDataFromXpriv(
+          changeLevelPrivKey.xprivkey,
+          {
+            pin: pinCode,
+            authXpriv: this.authPrivKey.xprivkey!,
+            // multisig: not implemented on wallet service yet
+          },
+        );
+      } else {
+        throw new Error('This should never happen.');
```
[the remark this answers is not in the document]
r4mmer: Done

src/wallet/wallet.ts
```diff
@@ -302,13 +315,59 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      }
+      await this.storage.saveAccessData(accessData);
```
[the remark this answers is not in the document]
r4mmer: Done

src/wallet/wallet.ts
r4mmer: Moved to the wallet utils as requested

src/utils/wallet.ts
```diff
@@ -471,7 +471,14 @@ const wallet = {
...
       seed,
       password,
-    }: {multisig?: IMultisigData, pin: string, seed?: string, password?: string},
+      authXpriv,
+    }: {
+      multisig?: IMultisigData,
+      pin: string,
+      seed?: string,
+      password?: string,
+      authXpriv?: string,
```
pedroferreira1, line 480: Missing in the docstring

src/utils/wallet.ts
```diff
@@ -528,10 +535,14 @@ const wallet = {
       accessData.acctPathKey = encryptedAcctPathKey;
     }
 
-    if (authXpriv) {
-      // Auth path key will only be available if the provided key is a root key.
-      const encryptedAuthPathKey = encryptData(authXpriv.xprivkey, pin);
-      accessData.authKey = encryptedAuthPathKey;
+    if (authXpriv || derivedAuthKey) {
```
pedroferreira1, line 538: Is it ok if we have both? Or should we protect in case the user passes the `authXpriv` as parameter with a root key?

r4mmer: I don't think it should be an issue to have the auth key as a parameter and derived from the root key.
We should choose which one has priority, and I think having the argument as priority is the best choice.
What do you think?

src/wallet/wallet.ts
```diff
@@ -165,6 +177,10 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     this.xpriv = xpriv;
     this.seed = seed;
     this.xpub = xpub;
+    if (authxpriv && !bitcore.HDPrivateKey.isValidSerialized(authxpriv)) {
+      throw new Error('authxpriv parameter is an invalid hd privatekey');
```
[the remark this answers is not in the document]
r4mmer: Added a test for an invalid authxpriv parameter

src/utils/wallet.ts
r4mmer: Done

andreabadesso (approved):

pedroferreira1 (approved):

PR #525, opened 2023-05-17
feat: hardware wallet support

r4mmer:
### Acceptance Criteria
- Allow starting hardware wallets with the new storage scheme.

### Security Checklist

### DIFF
--- a/__tests__/integration/hathorwallet_facade.test.js
+++ b/__tests__/integration/hathorwallet_facade.test.js
@@ -8,6 +8,7 @@ import {
   generateConnection,
   generateMultisigWalletHelper,
   generateWalletHelper,
+  generateWalletHelperRO,
   stopAllWallets,
   waitForTxReceived,
   waitForWalletReady,
@@ -2933,12 +2934,9 @@ describe('storage methods', () => {
 
   it('should return if the wallet is a hardware wallet', async () => {
     const hWallet = await generateWalletHelper();
-    const accessData = await hWallet.storage.getAccessData();
-
     await expect(hWallet.isHardwareWallet()).resolves.toBe(false);
 
-    accessData.walletFlags |= WALLET_FLAGS.HARDWARE;
-    await hWallet.storage.saveAccessData(accessData);
-    await expect(hWallet.isHardwareWallet()).resolves.toBe(true);
+    const hWalletRO = await generateWalletHelperRO({ hardware: true });
+    await expect(hWalletRO.isHardwareWallet()).resolves.toBe(true);
   });
 });
--- a/__tests__/integration/hathorwallet_facade.test.js
+++ b/__tests__/integration/hathorwallet_facade.test.js
@@ -26,7 +26,7 @@ import transaction from '../../src/utils/transaction';
 import Mnemonic from 'bitcore-mnemonic/lib/mnemonic';
 import { P2PKH_ACCT_PATH } from '../../src/constants';
 import Network from '../../src/models/network';
-import { WalletType } from '../../src/types';
+import { WalletType, WALLET_FLAGS } from '../../src/types';
 
 const fakeTokenUid = '008a19f84f2ae284f19bf3d03386c878ddd15b8b0b604a3a3539aa9d714686e1';
 const sampleNftData = 'ipfs://bafybeiccfclkdtucu6y4yc5cpr6y3yuinr67svmii46v5cfcrkp47ihehy/albums/QXBvbGxvIDEwIE1hZ2F6aW5lIDI3L04=/21716695748_7390815218_o.jpg';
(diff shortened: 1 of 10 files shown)


src/storage/storage.ts
```diff
@@ -861,4 +861,9 @@ export class Storage implements IStorage {
   async getGapLimit(): Promise<number> {
     return this.store.getGapLimit();
   }
+
+  async isHardwareWallet(): Promise<boolean> {
```
andreabadesso, line 865: Missing docstring

andreabadesso (changes requested):

src/storage/storage.ts
r4mmer: Added docstring

andreabadesso (approved):

src/storage/storage.ts
```diff
@@ -862,4 +862,13 @@ export class Storage implements IStorage {
   async getGapLimit(): Promise<number> {
     return this.store.getGapLimit();
   }
+
+  /**
+   * Return if the loaded wallet was started from a hardware wallet.
+   * @returns {Promise<boolean>}
+   */
+  async isHardwareWallet(): Promise<boolean> {
```
pedroferreira1, line 870: Does it make sense to create a method in both facades like this?

```
async isHardwareWallet() {
  return this.storage.isHardwareWallet();
}
```

If we don't do that, the wallets would need to access a storage method, which I prefer to avoid. What do you think?

src/utils/wallet.ts
```diff
@@ -395,7 +395,13 @@ const wallet = {
    * @param {{ multisig: IMultisigData }} [options={}] Options to generate the access data.
    * @returns {IWalletAccessData}
    */
-  generateAccessDataFromXpub(xpubkey: string, {multisig}: {multisig?: IMultisigData} = {}): IWalletAccessData {
+  generateAccessDataFromXpub(
+    xpubkey: string,
+    { multisig, hardware = false }: { multisig?: IMultisigData, hardware?: boolean} = {}): IWalletAccessData {
```
pedroferreira1, line 400: Missing new parameter in the docstring

__tests__/storage/storage.test.js
```diff
@@ -708,3 +709,19 @@ describe('checkPin and checkPassword', () => {
...
+  const storage = new Storage(store);
+  const accessDataSpy = jest.spyOn(storage, '_getValidAccessData');
+
+  accessDataSpy.mockReturnValue(Promise.resolve({
+    walletFlags: WALLET_FLAGS.HARDWARE,
+  }));
+  await expect(storage.isHardwareWallet()).resolves.toBe(true);
+
+  accessDataSpy.mockReturnValue(Promise.resolve({
+    walletFlags: 0,
```
pedroferreira1, line 724: Maybe add a test with READONLY flag to validate it will return false as well.

src/storage/storage.ts
r4mmer: Done

src/utils/wallet.ts
r4mmer: Added

__tests__/storage/storage.test.js
r4mmer: Done

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -2930,4 +2930,15 @@ describe('storage methods', () => {
     await expect(mshWallet.getAccessData()).resolves.toEqual(mshAccessData);
     await expect(mshWallet.getMultisigData()).resolves.toEqual(mshAccessData.multisigData);
   });
+
+  it('should return if the wallet is a hardware wallet', async () => {
```
pedroferreira1, line 2934: Do we have a method in the integration tests to create a readonly wallet (from an xpub)? I prefer creating the wallet with the normal process and then check if it saved correctly the hardware flag, instead of manually change the access data

src/new/wallet.js
```diff
@@ -2499,6 +2499,14 @@ class HathorWallet extends EventEmitter {
   async checkPinAndPassword(pin, password) {
     return await this.checkPin(pin) && await this.checkPassword(password);
   }
+
+  /**
+   * Check if the wallet is a hardware wallet.
+   * @returns {Promise<boolean>}
+   */
+  async isHardwareWallet() {
```
pedroferreira1, line 2507: We need the same method in the wallet service facade and also to update the interface.

__tests__/integration/hathorwallet_facade.test.js
r4mmer: Created a method to generate a readonly wallet from the pre-calculated wallets.
This way the generated wallet will come with the readonly flag and optionally we can enable the hardware flag to use in this test.

src/new/wallet.js
r4mmer: Done.
But the method in the wallet-service facade will always return `false` since we currently do not support hardware wallets with the wallet-service.

PR #533, opened 2023-05-19
feat: Websocket connection to the Atomic Swap Service

tuliomir:
### Acceptance Criteria
- Establishes the connection with the Atomic Swap Service via a websocket protocol

_Note_: The subscribing and unsubscribing of proposals will be made at a future PR. This one interacts with the server foundation provided by the backend on HathorNetwork/hathor-atomic-swap-service#5

### Security Checklist


src/new/swapConnection.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * It has the following states:
+ * - CLOSED: When it is disconnected from the server.
+ * - CONNECTING: When it is connecting to the server.
+ * - CONNECTED: When it is connected.
+ *
+ * You can subscribe for the following events:
+ * - update-atomic-swap-proposal: Fired when the state of a listened proposal changes
+ **/
+export class AtomicSwapServiceConnection extends BaseConnection {
```
r4mmer, line 28: Why are we extending `BaseConnection` when `BaseConnection` was meant to be a ws connection with the fullnode?

src/websocket/atomic-swap.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+   * Handle message receiving from websocket
+   *
+   * @param {Object} evt Event that has data (evt.data) sent in the websocket
+   */
+  onMessage(evt) {
+    const message = JSON.parse(evt.data)
+    const _type = message.type.split(':')[0]
```
r4mmer, line 30: Why are we splitting the type with `:`?

src/new/swapConnection.ts
tuliomir: Good point! Opened #535 to discuss the refactoring possibilities for this, including the Wallet Service websocket that is already implemented using it.

src/websocket/atomic-swap.ts
tuliomir: It wasn't necessary. Removed on [d1d252f](https://github.com/HathorNetwork/hathor-wallet-lib/pull/533/commits/d1d252f142c486672c57747daa09e63a54d144b2)

r4mmer (approved):

src/websocket/atomic-swap.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.timeoutTimer = setTimeout(() => this.onConnectionDown(), this.connectionTimeout);
+      }
+    }
+    this.emit(_type, message)
+  }
+
+  /**
+   * Returns a JSON stringified ping message
+   */
+  getPingMessage() {
```
pedroferreira1, line 48: Do you have a pong in the swap service ws?

src/websocket/atomic-swap.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Handles websocket connections and message transmission
+ *
+ * This class extends the base websocket class.
+ *
+ * @class
+ * @name AtomicSwapWebSocket
+ */
+class AtomicSwapWebSocket extends BaseWebSocket {
```
pedroferreira1, line 18: Is there any difference between this class and `WalletWebSocket`? The only one I've seen is the type split, which could be a parameter.

src/new/swapConnection.ts
pedroferreira1: I agree with @r4mmer that it's a different use. The `BaseConnection` is not meant to be used with full node only but with wallet context.

I think it's bad to use this as base class, you should refactor it before merging, it's weird that you have your `ConnectionParams` options with `network` and `servers`, which makes no sense for the swap service.

src/new/swapConnection.ts
```diff
@@ -0,0 +1,95 @@
+/**
```
pedroferreira1, line 1: I don't think the folder `new` is the right folder for it. I would create a new folder for `swapService`, or something like that, to keep the swap service files.

src/websocket/atomic-swap.ts
tuliomir: Yes, [we do](https://github.com/HathorNetwork/hathor-atomic-swap-service/blob/a0245dbf90612f4c01d96570f8c4912082070ac7/src/functions/websocket/handler.ts#LL50C1-L55C3).

src/websocket/atomic-swap.ts
tuliomir: Good point! I refactored all my code to use the `WalletWebSocket` instead, even the tests.
And since this class no longer serves only the wallet, I renamed it to `GenericWebSocket` on [bb5f334](https://github.com/HathorNetwork/hathor-wallet-lib/pull/533/commits/bb5f334dca068f64c250e8f41943d23ae2eb0795).

Do you agree with these changes?

__tests__/new/swapConnection.test.ts
```diff
@@ -0,0 +1,108 @@
+import { AtomicSwapServiceConnection } from '../../src/swapService/swapConnection';
```
pedroferreira1, line 1: The folder of the test should change as well?

src/swapService/swapConnection.ts
```diff
@@ -0,0 +1,118 @@
...
+      for (const proposalId of proposalsIds) {
+        const msg = JSON.stringify({ type: 'subscribe_proposal', proposalId });
+        this.websocket.sendMessage(msg);
+      }
+    }
+  }
+
+  unsubscribeProposal(proposalId: string) {
+    if (this.websocket) {
+      const msg = JSON.stringify({type: 'unsubscribe_proposal', proposalId});
```
pedroferreira1, line 114:
```suggestion
      const msg = JSON.stringify({ type: 'unsubscribe_proposal', proposalId });
```

pedroferreira1 (approved):

__tests__/new/swapConnection.test.ts
tuliomir: Fixed on [b6207ef](https://github.com/HathorNetwork/hathor-wallet-lib/pull/533/commits/b6207efa1e23251ee6b77e9fc0039459d6347594)

src/swapService/swapConnection.ts
tuliomir: Fixed on [f205976](https://github.com/HathorNetwork/hathor-wallet-lib/pull/533/commits/f2059766809fa1fb89d1f649058d2d346f49de5b)

PR #536, opened 2023-05-29
feat: babel async generator preset

r4mmer:
### Acceptance Criteria

- Include the babel plugin `@babel/plugin-transform-async-generator-functions`
- include number balance on the token history reponse of `getTxHistory`

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "test": "jest --env=node --forceExit",
     "test:watch": "jest --watch --env=node",
-    "build": "babel --presets @babel/preset-react -d lib/ src && tsc",
+    "build": "babel -d lib/ src && tsc",
     "tsc": "tsc",
     "test_integration": "npm run test_network_up && npm run test_network_integration && npm run test_network_down",
     "test_network_up": "docker-compose -f ./__tests__/integration/configuration/docker-compose.yml up -d && mkdir -p tmp && cp ./__tests__/integration/configuration/precalculated-wallets.json ./tmp/wallets.json",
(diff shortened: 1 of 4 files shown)


andreabadesso (approved):

package.json
```diff
@@ -39,6 +39,7 @@
     "@babel/eslint-parser": "^7.18.2",
     "@babel/node": "^7.13.0",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
+    "@babel/plugin-transform-async-generator-functions": "^7.22.3",
```
pedroferreira1, line 42: Missing package lock for this dependency

src/new/wallet.js
```diff
@@ -673,11 +673,13 @@ class HathorWallet extends EventEmitter {
       if (count <= 0) {
         break;
       }
+      const txbalance = await this.getTxBalance(tx);
       txs.push({
         txId: tx.tx_id,
         timestamp: tx.timestamp,
         tokenUid: uid,
         voided: tx.is_voided,
+        balance: txbalance[uid] || 0,
```
pedroferreira1, line 682: Not sure if we need in one of the wallets but the wallet service facada also has the tx version in this return object. And it doesn't have the `tokenUid`. If we are using it, the wallet service won't have this information.

pedroferreira1, line 682: Missing this information in the docstring.

package.json
r4mmer: Added

pedroferreira1 (approved):

PR #546, opened 2023-06-05
feat: sign message

andreabadesso:
### Acceptance Criteria
- We should have a `signMessageWIthAddress` method on both facades

### Security Checklist

### DIFF
--- a/__tests__/wallet/wallet.test.ts
+++ b/__tests__/wallet/wallet.test.ts
@@ -28,7 +28,7 @@ import { TOKEN_MELT_MASK, TOKEN_MINT_MASK } from '../../src/constants';
 import { MemoryStore, Storage } from '../../src/storage';
 import walletApi from '../../src/wallet/api/walletApi';
 import walletUtils from '../../src/utils/wallet';
-import { decryptData } from '../../src/utils/crypto';
+import { decryptData, verifyMessage } from '../../src/utils/crypto';
 import { WALLET_SERVICE_AUTH_DERIVATION_PATH } from '../../src/constants';
 
 // Mock SendTransactionWalletService class so we don't try to send actual transactions
@@ -1457,4 +1457,55 @@ test('getAddressPrivKey', async () => {
     'WUgjC47cFz5z9Uag92MKdnL8XgHdCfscNx',
     'WPRp9yj7Tjj9praWMy1whXuUc5zi1TdQtm'
   ]);
+});
+
+test('signMessageWithAddress', async () => {
+  const requestPassword = jest.fn();
+  const network = new Network('testnet');
+  const seed = defaultWalletSeed;
+  let store = new MemoryStore();
+  let storage = new Storage(store);
+  const accessData = walletUtils.generateAccessDataFromSeed(seed, {
+    networkName: 'testnet',
+    password: '1234',
+    pin: '1234',
+  });
+
+  jest.spyOn(HathorWalletServiceWallet.prototype, 'pollForWalletStatus').mockImplementation(() => Promise.resolve());
+  jest.spyOn(HathorWalletServiceWallet.prototype, 'setupConnection').mockImplementation(jest.fn());
+  jest.spyOn(walletApi, 'getNewAddresses')
+    .mockImplementation(() => Promise.resolve({ success: true, addresses: [] }));
+  jest.spyOn(walletApi, 'createWallet')
+    .mockImplementation(() => Promise.resolve({
+      success: true,
+      status: {
+        walletId: 'id',
+        xpubkey: 'xpub',
+        status: 'creating',
+        maxGap: 20,
+        createdAt: 0,
+        readyAt: 0,
+      },
+    }));
+
+  let wallet = new HathorWalletServiceWallet({
+    requestPassword,
+    seed,
+    network,
+    storage,
+  });
+
+  await wallet.start({ pinCode: '1234', password: '1234' });
+
+  const message = 'sign-me-please';
+  const addressIndex = 2;
+  const address = 'WPfG7P4YQDJ4MpwTS6qrfGW4fvYvAhPpV7';
+  const signedMessage = await wallet.signMessageWithAddress(
+    message,
+    addressIndex,
+    '1234',
+  );
+
+  expect(verifyMessage(message, signedMessage, address))
+    .toBeTruthy();
 });
--- a/__tests__/wallet/wallet.test.ts
+++ b/__tests__/wallet/wallet.test.ts
@@ -15,7 +15,11 @@ import {
   CreateWalletAuthData,
 } from '../../src/wallet/types';
 import config from '../../src/config';
-import { buildSuccessTxByIdTokenDataResponse, buildWalletToAuthenticateApiCall, defaultWalletSeed } from '../__mock_helpers/wallet-service.fixtures';
+import {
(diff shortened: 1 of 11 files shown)


src/new/wallet.js
```diff
@@ -1339,6 +1341,46 @@ class HathorWallet extends EventEmitter {
...
+   * @param {number} addressIndex - The address' index to fetch
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  async getAddressPrivKey(pinCode, addressIndex) {
+    const mainXPrivKey = await this.storage.getMainXPrivKey(pinCode);
+    const addressXPriv = new bitcore.HDPrivateKey(mainXPrivKey).derive(addressIndex);
+
+    return addressXPriv.privateKey;
```
r4mmer, line 1357: Why not return the `HDPrivateKey`? this way we will have access to both private and public keys if needed

src/utils/crypto.ts
```diff
@@ -6,10 +6,14 @@
...
 import CryptoJS from 'crypto-js';
+import bitcore from 'bitcore-lib';
 import { DecryptionError, InvalidPasswdError, UnsupportedHasherError } from '../errors';
-import { HASH_ITERATIONS, HASH_KEY_SIZE } from '../constants';
+import { HATHOR_MAGIC_BYTES, HASH_ITERATIONS, HASH_KEY_SIZE } from '../constants';
 import { IEncryptedData } from '../types';
 
+// Monkey-patch MAGIC_BYTES to use Hathor's
+bitcore.Message.MAGIC_BYTES = Buffer.from(HATHOR_MAGIC_BYTES);
+
```
r4mmer, line 16: Do we know why bitcore uses these magic bytes?
And does bitcore do anything else to the data other than prefix with the magic bytes?
I say this because we may have to implement this signature and verification on other places that cannot use bitcore's library (e.g. Ledger app).

So we may have to understand the whole signature process and be able to duplicate it on other places using other languages.

__tests__/wallet/wallet.test.ts
```diff
@@ -1402,3 +1406,106 @@ test('start', async () => {
...
+  });
+  await wallet.start({ pinCode: '1234', password: '1234' });
+
+  expect([
+    await wallet.getAddressPrivKey('1234', 1),
+    await wallet.getAddressPrivKey('1234', 2),
+    await wallet.getAddressPrivKey('1234', 3),
+    await wallet.getAddressPrivKey('1234', 4),
+    // We need to pass the network because bitcore-lib forces bitcoin network
+  ].map((privKey) => privKey.toAddress(network.getNetwork()).toString()))
```
r4mmer, line 1453:
```suggestion
  ].map((privKey) => privKey.toAddress(network).toString()))
```

r4mmer (changes requested):

src/utils/crypto.ts
andreabadesso: From my investigation, it is a convention that was [first proposed by bitcoin-core](https://github.com/bitcoin/bitcoin/pull/524) and that bitcore-lib also uses

The idea of having the message prefixed with an known string is to prevent someone from tricking a user into signing a transaction instead of a simple text message

Looking at the [implementation](https://github.com/bitpay/bitcore-message/blob/master/lib/message.js), I don't see anything preventing us from implementing it on Ledger

__tests__/wallet/wallet.test.ts
andreabadesso: That won't work as `toAddress` expects a `bitcoreNetwork`

r4mmer (approved):

src/new/wallet.js
```diff
@@ -1339,6 +1341,48 @@ class HathorWallet extends EventEmitter {
...
+
+    return addressHDPrivKey;
+  }
+
+  /**
+   * Returns a base64 encoded signed message with an address' private key given an
+   * andress index
+   *
+   * @param {string} message - The message to sign
+   * @param {number} pinCode - The address index to sign with
```
pedroferreira1, line 1367:
```suggestion
   * @param {number} index - The address index to sign with
```

src/new/wallet.js
```diff
@@ -1339,6 +1341,48 @@ class HathorWallet extends EventEmitter {
...
 
+  /**
+   * Returns an address' HDPrivateKey given an index and the encryption password
+   *
+   * @param {string} pinCode - The PIN used to encrypt data in accessData
+   * @param {number} addressIndex - The address' index to fetch
+   *
+   * @returns {Promise<HDPrivateKey>} Promise that resolves with the HDPrivateKey
+   *
+   * @memberof HathorWalletServiceWallet
```
pedroferreira1, line 1352:
```suggestion
   * @memberof HathorWallet
```

src/new/wallet.js
```diff
@@ -1339,6 +1341,48 @@ class HathorWallet extends EventEmitter {
...
+   * Returns a base64 encoded signed message with an address' private key given an
+   * andress index
+   *
+   * @param {string} message - The message to sign
+   * @param {number} pinCode - The address index to sign with
+   * @param {string} pinCode - The PIN used to encrypt data in accessData
+   *
+   * @return {Promise} Promise that resolves with the signed message
+   *
+   * @memberof HathorWalletServiceWallet
```
pedroferreira1, line 1372:
```suggestion
   * @memberof HathorWallet
```

pedroferreira1 (approved):

src/new/wallet.js
andreabadesso: Fixed

src/new/wallet.js
andreabadesso: Fixed, thanks

src/utils/crypto.ts
```diff
@@ -6,10 +6,14 @@
...
 
 import CryptoJS from 'crypto-js';
+import bitcore from 'bitcore-lib';
 import { DecryptionError, InvalidPasswdError, UnsupportedHasherError } from '../errors';
-import { HASH_ITERATIONS, HASH_KEY_SIZE } from '../constants';
+import { HATHOR_MAGIC_BYTES, HASH_ITERATIONS, HASH_KEY_SIZE } from '../constants';
 import { IEncryptedData } from '../types';
 
+// Monkey-patch MAGIC_BYTES to use Hathor's
+bitcore.Message.MAGIC_BYTES = Buffer.from(HATHOR_MAGIC_BYTES);
```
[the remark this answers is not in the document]
pedroferreira1: > We can and probably should use the new utils/crypto.ts method, but I suggest we don't refactor it in this PR as we also need to change wallet-service to expect Hathor Signed Message\n instead of Bitcoin Signed Message\n it currently expect.

That's my main concern. If we changed the magic bytes for the lib, the current wallets we have in the wallet service wouldn't work anymore, right? Because they were generated using the old magic bytes, or not?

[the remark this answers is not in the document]
andreabadesso: I checked and this is indeed a problem, since we are monkey patching the Message class, it will affect the signMessage for the wallet-service.

I will merge this and work on a fix on the wallet-service side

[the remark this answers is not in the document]
andreabadesso: @pedroferreira1 PR opened in https://github.com/HathorNetwork/hathor-wallet-service/pull/426

PR #547, opened 2023-06-06
feat: wallet desktop utils

r4mmer:
### Acceptance Criteria

- Do not delete registered tokens when cleaning the wallet storage
- Add utilities required by the wallet desktop
- emit the correct event for partial ui update

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "axios": "^0.21.4",
     "bitcore-lib": "^8.25.10",
     "bitcore-mnemonic": "^8.25.10",
+    "buffer": "^6.0.3",
     "crypto-js": "^3.1.9-1",
     "isomorphic-ws": "^4.0.1",
     "level": "^8.0.0",
--- a/src/storage/leveldb/token_index.ts
+++ b/src/storage/leveldb/token_index.ts
@@ -254,9 +254,15 @@ export default class LevelTokenIndex implements IKVTokenIndex {
 
   /**
    * Clear all entries from the database.
+   * @param {boolean} skipRegistered skip deleting the registered tokens
    * @returns {Promise<void>}
    */
-  async clear(): Promise<void> {
-    await this.tokenDB.db.clear();
+  async clear(skipRegistered: boolean = false): Promise<void> {
+    if (skipRegistered) {
+      await this.tokenDB.clear();
+      await this.metadataDB.clear();
+    } else {
+      await this.tokenDB.db.clear();
+    }
   }
 }
--- a/src/storage/leveldb/wallet_index.ts
+++ b/src/storage/leveldb/wallet_index.ts
@@ -11,6 +11,7 @@ import { AbstractSublevel } from 'abstract-level';
 import { IKVWalletIndex, IWalletData, IWalletAccessData } from '../../types';
 import { GAP_LIMIT } from '../../constants';
 import { errorCodeOrNull, KEY_NOT_FOUND_CODE } from './errors';
+import { Buffer } from 'buffer';
 
 export const ACCESS_PREFIX = 'access';
 export const WALLET_PREFIX = 'wallet';
@@ -81,19 +82,25 @@ export default class LevelWalletIndex implements IKVWalletIndex {
       // valueEncoding of buffer can be Buffer of Uint8Array depending on the level implementation
       // classic-level will return Buffer
       // browser-level will return Uint8Array
-      let tmp: Buffer|Uint8Array;
+      let buf: Buffer|Uint8Array;
       switch(dest) {
         case 'access':
-          tmp = await this.accessDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
+          buf = await this.accessDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
           break;
         case 'wallet':
-          tmp = await this.walletDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
+          buf = await this.walletDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
           break;
       }
 
-      const buf = Buffer.from(tmp)
+      if (Buffer.isBuffer(buf)) {
+        return buf.readUint32BE(0);
+      } else if (buf instanceof Uint8Array) {
+        const newBuf = new Uint32Array(buf.buffer);
+        return newBuf[0];
+      } else {
+        throw new Error('Cannot read value received from the database');
+      }
 
-      return buf.readUint32BE(0);
     } catch (err: unknown) {
       if (errorCodeOrNull(err) === KEY_NOT_FOUND_CODE) {
         return null;
--- a/src/storage/leveldb/wallet_index.ts
+++ b/src/storage/leveldb/wallet_index.ts
@@ -11,7 +11,6 @@ import { AbstractSublevel } from 'abstract-level';
 import { IKVWalletIndex, IWalletData, IWalletAccessData } from '../../types';
 import { GAP_LIMIT } from '../../constants';
 import { errorCodeOrNull, KEY_NOT_FOUND_CODE } from './errors';
-import { Buffer } from 'buffer';
 
 export const ACCESS_PREFIX = 'access';
 export const WALLET_PREFIX = 'wallet';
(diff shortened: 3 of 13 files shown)


src/storage/leveldb/token_index.ts
```diff
@@ -254,9 +254,15 @@ export default class LevelTokenIndex implements IKVTokenIndex {
...
    */
-  async clear(): Promise<void> {
-    await this.tokenDB.db.clear();
+  async clear(skipRegistered: boolean = false): Promise<void> {
+    if (skipRegistered) {
+      await this.tokenDB.clear();
+      await this.metadataDB.clear();
+    } else {
+      await this.tokenDB.db.clear();
+    }
```
r4mmer, line 266: `this.tokenDB.db.clear()` will delete all token keys, including the registered tokens.
When cleaning the storage we would wish to persist the registered tokens, we would only like to clean all tokens when completely reseting the wallet, which is why we have the boolean parameter here.

src/utils/storage.ts
```diff
@@ -93,9 +93,9 @@ export async function syncHistory(startIndex: number, count: number, storage: IS
         foundAnyTx = true;
       }
       // update UI
-      connection.emit('wallet-partial-load-history', {
-        addressesFound: storage.store.addressCount(),
-        historyLength: storage.store.historyCount(),
+      connection.emit('wallet-load-partial-update', {
+        addressesFound: await storage.store.addressCount(),
+        historyLength: await storage.store.historyCount(),
```
r4mmer, line 98: The event name was incorrect and the values should be awaited

src/storage/leveldb/wallet_index.ts
```diff
@@ -48,47 +48,29 @@ export default class LevelWalletIndex implements IKVWalletIndex {
...
-          buf = await this.accessDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
-          break;
-        case 'wallet':
-          buf = await this.walletDB.get<string, Buffer>(key, { valueEncoding: 'buffer'});
-          break;
-      }
-
-      return buf.readUint32BE(0);
+      const tmp = await this.walletDB.get(key);
+      return Number('0x'+tmp);
```
tuliomir, line 73: Just to be extra sure on this implicit conversion results, I'd suggest using string templates:
```suggestion
      return Number(`$0x${tmp}`);
```

src/utils/helpers.ts
```diff
@@ -528,6 +529,43 @@ const helpers = {
   cleanupString(s: string): string {
     return s.replace(/\s\s+/g, ' ').trim().toLowerCase();
   },
+
+  /**
+   * Axios fails merging this configuration to the default configuration because it has an issue
+   * with circular structures: https://github.com/mzabriskie/axios/issues/370
+   * Got this code from https://github.com/softonic/axios-retry/blob/master/es/index.js#L203
```
tuliomir, line 536: Typo on the filename
```suggestion
   * Got this code from https://github.com/softonic/axios-retry/blob/master/es/index.mjs#L203
```

package.json
```diff
@@ -13,6 +13,7 @@
     "axios": "^0.21.4",
     "bitcore-lib": "^8.25.10",
     "bitcore-mnemonic": "^8.25.10",
+    "buffer": "^6.0.3",
```
tuliomir, line 16: I think it would be good to add just a short explanation to this PR's description on why did we add the `buffer` lib now, instead of relying on the runtime's native one.

tuliomir (approved):

src/utils/transaction.ts
```diff
@@ -529,6 +540,35 @@ const transaction = {
...
+   *
+   * @param {Pick<IHistoryTx, 'version'>} tx Object to get the type
+   *
+   * @return {string} Type of the object
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  getTxType(tx: Pick<IHistoryTx, 'version'>): string {
+    if (this.isBlock(tx)) {
```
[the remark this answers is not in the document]
r4mmer: I agree, but this is the same logic from the old util, the idea was to port the utility so i didn't look into improvements.

src/utils/helpers.ts
r4mmer: This comment was taken from the old lib, maybe the file has been renamed since.

src/storage/leveldb/wallet_index.ts
r4mmer: Why the extra `$` at the start?

package.json
r4mmer: The buffer was already present on our lib but it was a dependency of some other package.
The issue is that since we actually import it on our lib we should have it explicitly on the package.json file.

It may be important later as a polyfill if needed but currently we use it as a dependency.

src/storage/leveldb/wallet_index.ts
```diff
@@ -48,47 +48,29 @@ export default class LevelWalletIndex implements IKVWalletIndex {
...
-  async _setNumber(dest: 'access'|'wallet', key: string, value: number) {
+  async _setNumber(key: string, value: number) {
     if (value < 0) {
       throw new Error(`Invalid unsigned int ${value} being set on ${key}`);
     }
-    const buf = Buffer.alloc(4);
-    buf.writeUInt32BE(value);
-    switch(dest) {
-      case 'access':
-        await this.accessDB.put<string, Buffer>(key, buf, { valueEncoding: 'buffer'});
```
pedroferreira1, line 63: You don't use this method for access db anymore?

src/utils/helpers.ts
```diff
@@ -528,6 +529,43 @@ const helpers = {
...
+   * @param {AxiosRequestConfig} config New axios config
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  fixAxiosConfig(axios: AxiosInstance, config: AxiosRequestConfig) {
+    if (axios.defaults.httpAgent === config.httpAgent) {
+      delete config.httpAgent;
+    }
+    if (axios.defaults.httpsAgent === config.httpsAgent) {
```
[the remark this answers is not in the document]
r4mmer: We are doing it, on line [545](https://github.com/HathorNetwork/hathor-wallet-lib/blob/161873006e6d35502bc4c5ffd07a21ab07f544df/src/utils/helpers.ts#L545)

src/storage/leveldb/wallet_index.ts
r4mmer: There is only 1 information saved as a buffer on leveldb, since we would be required to always treat for both `Buffer` and `Uint8Array` I think the best decision is to not rely on the leveldb buffer encoding and use the string encoding.
The data will be encoded in our format, which I decided on the hex value since it gives a deterministic size without the need of a encoding/decoding lib.

package.json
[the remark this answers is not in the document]
r4mmer: There are no changes on the package-lock.
The buffer was already an indirect dependency, but we were importing it on the source code so I added as a direct dependency.

src/storage/leveldb/token_index.ts
```diff
@@ -254,9 +254,15 @@ export default class LevelTokenIndex implements IKVTokenIndex {
 
   /**
    * Clear all entries from the database.
+   * @param {boolean} skipRegistered skip deleting the registered tokens
    * @returns {Promise<void>}
    */
-  async clear(): Promise<void> {
-    await this.tokenDB.db.clear();
+  async clear(skipRegistered: boolean = false): Promise<void> {
+    if (skipRegistered) {
```
alexruzenhack, line 261: I believe this check should contain a negative, right?
```suggestion
    if (!skipRegistered) {
```

src/storage/leveldb/wallet_index.ts
```diff
@@ -48,47 +48,29 @@ export default class LevelWalletIndex implements IKVWalletIndex {
    * Convert a number to a uint32 buffer before saving on database.
    * Internal helper method, since this logic is used in multiple places.
```
alexruzenhack, line 49: I believe this doc string need an update, right?

alexruzenhack (changes requested):
I have this comment to be confirmed first: https://github.com/HathorNetwork/hathor-wallet-lib/pull/547/files#r1265439448

src/storage/leveldb/token_index.ts
r4mmer: It's correct, `this.tokenDB.db` refers to the database containing all token databases or sublevels (token, metadata and registered tokens).
If we want to skip the registered tokens sublevel we have to manually clear the token and metadata ones.

PR #549, opened 2023-06-13
feat(feature-activation): implement signal bits

glevco:
### Acceptance Criteria

- Implement new `signalBits` field in all transaction models, occupying the first byte of a transaction
- Change the `version` field in all transaction models, so it only occupies the second byte of a transaction

### Security Checklist

### DIFF
--- a/src/models/create_token_transaction.ts
+++ b/src/models/create_token_transaction.ts
@@ -10,6 +10,7 @@ import {
   TOKEN_INFO_VERSION,
   MAX_TOKEN_NAME_SIZE,
   MAX_TOKEN_SYMBOL_SIZE,
+  DEFAULT_SIGNAL_BITS,
 } from '../constants'
 import { unpackToInt, unpackLen, intToBytes } from '../utils/buffer';
 import Input from './input';
@@ -23,6 +24,7 @@ import ScriptData from "./script_data";
 import {OutputType} from "../wallet/types";
 
 type optionsType = {
+  signalBits?: number,
   weight?: number,
   nonce?: number,
   timestamp?: number | null,
@@ -38,6 +40,7 @@ class CreateTokenTransaction extends Transaction {
 
   constructor(name: string, symbol: string, inputs: Input[], outputs: Output[], options: optionsType = {}) {
     const defaultOptions: optionsType = {
+      signalBits: DEFAULT_SIGNAL_BITS,
       weight: 0,
       nonce: 0,
       timestamp: null,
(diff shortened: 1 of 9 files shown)


r4mmer:
We shoudl make this change on the helpers as well.
For instance:

- utils/helpers.ts
  - [createTxFromBytes](https://github.com/HathorNetwork/hathor-wallet-lib/blob/047ed60a6434188bb48c16ffba355b311e138f51/src/utils/helpers.ts#L230)
  - [createTxFromData](https://github.com/HathorNetwork/hathor-wallet-lib/blob/047ed60a6434188bb48c16ffba355b311e138f51/src/utils/helpers.ts#L348)
  - [createTxFromHistoryObject](https://github.com/HathorNetwork/hathor-wallet-lib/blob/047ed60a6434188bb48c16ffba355b311e138f51/src/utils/helpers.ts#L382)
- utils/transaction
  - [createTransactionFromData](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/utils/transaction.ts#L429).

These are some places where we expect the version as a 2 byte number.
If this change affects the transaction being returned from the full-node or if we will have to specify the signal byte when building transactions there are other places that require change, even the tests should change.

glevco:
@r4mmer done for the functions you mentioned, in https://github.com/HathorNetwork/hathor-wallet-lib/pull/549/commits/e9f955ce7a2cc32d7574a71ad5669b5575dea08e

> If this change affects the transaction being returned from the full-node or if we will have to specify the signal byte when building transactions there are other places that require change, even the tests should change.

For now this PR is intended only to keep compatibility. It shouldn't be necessary to actually interact with signal bits. Even if we didn't make any changes, it probably wouldn't break anything, since the current `version` occupies 2 bytes, but actually just uses one, so the first one is ignored.

I guess for now we can just add the existence of signal bits, and change `version` to use only one byte. I'll update the acceptance criteria accordingly.

r4mmer (approved):

src/models/create_token_transaction.ts
```diff
@@ -55,7 +55,7 @@ class CreateTokenTransaction extends Transaction {
 
   /**
    * Serialize funds fields
-   * version, len inputs, len outputs, inputs, outputs and token info
+   * signal bits, version, len inputs, len outputs, inputs, outputs and token info
```
pedroferreira1, line 58: Comment unrelated with the line. I think you should add the `signalBits` field to this class's `optionsType` and `defaultOptions`. The `version` field does not appear there because a transaction that creates tokens has a fixed version value but the signal bits might change.

pedroferreira1 (commented):
We do have two tests that check the serialization/deserialization of the `Transaction` object. I know it's not in use the new field but we could add the line there that validates that `expect(tx.signalBits).toBe(0)`.

The tests are in `__tests__/models/transaction.test.js`

What do you think?

src/models/create_token_transaction.ts
glevco: Done!

glevco:
> The tests are in `__tests__/models/transaction.test.js`
> 
> What do you think?

@pedroferreira1 sure! I've added `expect`s for that

pedroferreira1 (approved):

PR #565, opened 2023-08-22
fix: check for block using utils

r4mmer:
### Acceptance Criteria
- Include merged mining blocks when checking for heightlock

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -784,7 +784,7 @@ class HathorWallet extends EventEmitter {
         // XXX: we currently do not check heightlock on the helper, checking here for compatibility
         const nowHeight = await this.storage.getCurrentHeight();
         const rewardLock = this.storage.version?.reward_spend_min_blocks;
-        const is_height_locked = tx.height && nowHeight && rewardLock && (nowHeight < (tx.height + rewardLock) );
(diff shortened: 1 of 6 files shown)


tuliomir (approved):

msbrogli (changes requested):

src/utils/transaction.ts
```diff
@@ -230,7 +248,7 @@ const transaction = {
     const nowTs = Math.floor(Date.now() / 1000);
     const nowHeight = await storage.getCurrentHeight();
     const rewardLock = storage.version?.reward_spend_min_blocks;
-    const isHeightLocked = (!(rewardLock && tx.height)) ? false : ((tx.height + rewardLock) < nowHeight);
```
pedroferreira1, line 233: This was a bug before, right? It's the opposite comparison

pedroferreira1 (approved):

src/utils/transaction.ts
r4mmer: Yes, but the only place it is used is the wallet facade's `getTxBalance` which returns the sum of locked and unlocked balances, so this error was being overlooked.

src/new/wallet.js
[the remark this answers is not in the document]
r4mmer: Created the tests

msbrogli (approved):

PR #569, opened 2023-08-24
feat: use best utxo selection algorithm by default for tokens methods (create, mint, melt)

pedroferreira1:
### Acceptance Criteria
- Create token, mint tokens, and melt tokens should use the best utxo selection algorithm by default.

### Security Checklist

### DIFF
--- a/src/new/sendTransaction.ts
+++ b/src/new/sendTransaction.ts
@@ -571,7 +571,7 @@ export async function prepareSendTokensData(
     if (newUtxos.amount < outputAmount) {
       throw new Error(`Token: ${token}. Insufficient amount of tokens to fill the amount.`);
     }
-    newTxData.inputs = newUtxos.utxos.map(helpers.getDataInputFromUtxo);
+    newtxData.inputs = newUtxos.utxos.map(helpers.getDataInputFromUtxo);
 
     if (newUtxos.amount > outputAmount) {
       // We need to create a change output
--- a/src/new/sendTransaction.ts
+++ b/src/new/sendTransaction.ts
@@ -571,9 +571,7 @@ export async function prepareSendTokensData(
     if (newUtxos.amount < outputAmount) {
       throw new Error(`Token: ${token}. Insufficient amount of tokens to fill the amount.`);
     }
-    newtxData.inputs = newUtxos.utxos.map((utxo) => {
-      return helpers.getDataInputFromUtxo(utxo);
-    });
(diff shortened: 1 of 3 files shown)


andreabadesso (approved):

src/utils/tokens.ts
```diff
@@ -411,9 +418,9 @@ const tokens = {
     const withdrawAmount = this.getWithdrawAmount(amount, depositPercent);
 
     // get inputs that amount to requested melt amount
-    let foundAmount = 0;
-    for await (const utxo of storage.selectUtxos({token, target_amount: amount})) {
-      foundAmount += utxo.value;
+    const selectedUtxos = await utxoSelection(storage, token, amount);
+    const foundAmount = selectedUtxos.amount;
+    for (const utxo of selectedUtxos.utxos) {
```
r4mmer, line 423: Maybe we should use the authorities from the utxo instead of `authorities: 0`, if there was an error selecting utxos, this would hide if we chose an authority utxo.

In the same vein, this is actually a conversion from `IUtxo` to `IDataInput` which have very just a few properties of difference, maybe we should have it as an util since this conversion should be very common.

pedroferreira1: Done: https://github.com/HathorNetwork/hathor-wallet-lib/pull/569/commits/733cbd75571c0d7c2d2d3c08f529bafda956fee2

src/new/sendTransaction.ts
```diff
@@ -572,14 +572,7 @@ export async function prepareSendTokensData(
...
     newtxData.inputs = newUtxos.utxos.map((utxo) => {
-      return {
-        txId: utxo.txId,
-        index: utxo.index,
-        value: utxo.value,
-        authorities: utxo.authorities,
-        token: utxo.token,
-        address: utxo.address,
-      } as IDataInput;
+      return helpers.getDataInputFromUtxo(utxo);
```
r4mmer, line 575: Here we can just do:
```js
newTxData.inputs = newUtxos.utxos.map(helpers.getDataInputFromUtxo);
```

r4mmer (approved):

src/new/sendTransaction.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/569/commits/527d4366e062f205de5d2f952338ae8302e7d561

PR #570, opened 2023-08-30
fix: change server

r4mmer:
### Acceptance Criteria
- Do not stop wallet when running changeServer

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -256,8 +256,7 @@ class HathorWallet extends EventEmitter {
    * @memberof HathorWallet
    * @inner
    **/
-  async changeServer(newServer) {
-    await this.stop({cleanStorage: true, cleanAddresses: true});
+  changeServer(newServer) {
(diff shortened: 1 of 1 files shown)


alexruzenhack (approved):

src/new/wallet.js
```diff
@@ -256,8 +256,7 @@ class HathorWallet extends EventEmitter {
    * @memberof HathorWallet
    * @inner
    **/
-  async changeServer(newServer) {
-    await this.stop({cleanStorage: true, cleanAddresses: true});
+  changeServer(newServer) {
```
luislhl, line 259: There was no reason at all for stopping it? Won't we experience any drawback?

The code is very simple, but I don't have enough context right now to understand the real consequences of this change, that's why I'm asking. If we have a concrete motivation for doing this change, it could help in the contextualization as well.

luislhl (approved):

luislhl (approved):

PR #572, opened 2023-08-31
feat: retry token api on history sync

r4mmer:
### Acceptance Criteria
- Retry any failures on the token api when syncing history

### Security Checklist

### DIFF
--- a/src/utils/storage.ts
+++ b/src/utils/storage.ts
@@ -251,6 +251,60 @@ export async function processHistory(storage: IStorage, { rewardLock }: { reward
 }
 
 /**
+ * Fetch and save the data of the token set on the storage
+ * @param {IStorage} storage - Storage to save the tokens.
+ * @param {Set<string>} tokens - set of tokens to fetch and save.
+ * @returns {Promise<void>}
+ */
+async function updateTokensData(storage: IStorage, tokens: Set<string>): Promise<void> {
+  async function fetchTokenData(uid: string): Promise<{
+    success: true,
+    name: string,
+    symbol: string,
+  } | { success: false, message: string }> {
+    let retryCount = 0;
+
+    while (retryCount <= 5) {
+      try {
+        // Fetch and return the api response
+        let result: {
+          success: true;
+          name: string;
+          symbol: string;
+        } | { success: false, message: string } = await new Promise((resolve) => {
+          return walletApi.getGeneralTokenInfo(uid, resolve);
+        });
+        return result;
+      } catch (err: unknown) {
+        // Increase the retry counter and try again
+        retryCount += 1;
+        continue;
+      }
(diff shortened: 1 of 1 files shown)


src/utils/storage.ts
```diff
@@ -250,6 +250,60 @@ export async function processHistory(storage: IStorage, { rewardLock }: { reward
...
+          name: string;
+          symbol: string;
+        } | { success: false, message: string } = await new Promise((resolve) => {
+          return walletApi.getGeneralTokenInfo(uid, resolve);
+        });
+        return result;
+      } catch (err: unknown) {
+        // Increase the retry counter and try again
+        retryCount += 1;
+        continue;
```
luislhl, line 281: I think we should have a way to log what error is happening in the fullnode, to help debugging.

But I know that we don't have a good way to do this from the lib currently, and putting a `console.log` in it is not the best option.

So I'm letting this comment here just to increase awareness for this problem of not having a good way to log things from the wallet-lib.

src/utils/storage.ts
```diff
@@ -250,6 +250,60 @@ export async function processHistory(storage: IStorage, { rewardLock }: { reward
...
+ */
+async function updateTokensData(storage: IStorage, tokens: Set<string>): Promise<void> {
+  async function fetchTokenData(uid: string): Promise<{
+    success: true,
+    name: string,
+    symbol: string,
+  } | { success: false, message: string }> {
+    let retryCount = 0;
+
+    while (retryCount <= 5) {
```
luislhl, line 267: Would it make sense to use an exponential backoff for the retries?

luislhl (approved):

src/utils/storage.ts
r4mmer: We can, an idea would be to use:
```js
catch (err) {
  retryCount += 1;
  const delay = 10 * (2 ** retryCount);
  await new Promise(resolve => setTimeout(resolve, delay));
  continue;
}
```

Which would give us 10ms, 20ms, 40ms, 80ms and 160ms of intervals.
We could use another multiplier instead of 10ms to increase the intervals as well.
What do you think?

src/utils/storage.ts
r4mmer: Done

src/utils/storage.ts
luislhl: The idea of the exponential backoff would be to give some time for the fullnode to recover. I don't think it would recover in 160ms.

I was thinking of longer delays, like 0.5s, 1s, 2s, 4s, 8s. Unless waiting so long is bad for the use case that needs this response.

r4mmer: Done

tuliomir (approved):

PR #577, opened 2023-09-25
feat: compile entrypoint file with typescript

r4mmer:
### Acceptance Criteria
- Convert index to typescript so we can have consistent typing of the wallet-lib

### Security Checklist


src/index.ts
```diff
@@ -0,0 +1,103 @@
+import * as constants  from './constants';
+import dateFormatter  from './utils/date';
+import websocket  from './websocket';
+import * as errors  from './errors';
+import * as ErrorMessages  from './errorMessages';
```
r4mmer, line 5: We could just import the `ErrorMessages` enum but the wallets use the `hathorLib.ErrorMessages.ErrorMessages...` so I'll keep this "unnecessary encapsulation" so we don't have to refactor the wallets

src/lib.ts
```diff
@@ -0,0 +1,103 @@
...
+import transactionUtils  from './utils/transaction';
+import * as bufferUtils  from './utils/buffer';
+import HathorWalletServiceWallet  from './wallet/wallet';
+import walletServiceApi  from './wallet/api/walletApi';
+import SendTransactionWalletService  from './wallet/sendTransactionWalletService';
+import config  from './config';
+import * as PushNotification  from './pushNotification';
+import { WalletType }  from './types';
+
+import {PartialTx, PartialTxInputData}  from './models/partial_tx';
```
pedroferreira1, line 47:
```suggestion
import { PartialTx, PartialTxInputData }  from './models/partial_tx';
```

src/index.ts
```diff
@@ -0,0 +1,4 @@
+import * as hathorLib from './lib';
+
+export default hathorLib;
```
pedroferreira1, line 3: Why do you need both exports?

r4mmer: The current exported hathorlib enables importing as both named import and default import and we use both on the wallets all the time.
It's very common to use any of the import styles below:
```js
import hathorLib from '@hathor/wallet-lib';
import { HathorWallet } from '@hathor/wallet-lib';
import hathorLib, { HathorWallet } from '@hathor/wallet-lib';
```

The issue with ES6 exports is that it does allow for default export and named export in the same file, but not for the same object, meaning we would have to duplicate the object with all exports for the named and default export.
But if we add an intermediate file (i.e. `lib.ts`) we can export the lib exactly like it currently is exported.

src/lib.ts
r4mmer: Formatted the whole file.

src/index.ts
r4mmer: Apparently this is a design difference between ES6 and CommonJS, because CommonJS (`module.exports`) uses objects as namespaces for modules but it also has to work with objects as default exports.
This is why when importing from ES6 (`import..from` syntax) the CommonJS export works as both named exports and default export, but when exporting from ES6 we have to specify that we want to use both types of exports.

andreabadesso (approved):

pedroferreira1 (approved):

PR #578, opened 2023-09-27
feat: new address scanning policy

r4mmer:
### Acceptance Criteria

- Create a new address scanning policy called index-limit
  - This should be configured with a start and end index and should keep only these addresses loaded.
- Create methods on storage to configure scanning policies
- Change utils to check the scanning policy when loading addresses or processing history
- Allow the wallet facade to configure the scanning policy
- Create methods on the wallet facade to load more addresses when using the index-limit scanning policy

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -285,7 +285,10 @@ class HathorWallet extends EventEmitter {
       return;
     }
 
-    const limits = this.storage.getIndexLimit();
+    const limits = await this.storage.getIndexLimit();
+    if (!limits) {
+      return;
+    }
     const newEndIndex = limits.endIndex + count;
     await this.indexLimitSetEndIndex(newEndIndex);
   }
@@ -302,7 +305,10 @@ class HathorWallet extends EventEmitter {
       return;
     }
 
-    const limits = this.storage.getIndexLimit();
+    const limits = await this.storage.getIndexLimit();
+    if (!limits) {
+      return;
+    }
 
     if (endIndex <= limits.endIndex) {
       // Cannot unload addresses from storage.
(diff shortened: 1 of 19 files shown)


src/new/wallet.js
```diff
@@ -343,38 +409,40 @@ class HathorWallet extends EventEmitter {
...
-        await syncHistory(0, walletData.gapLimit, this.storage, this.conn);
-      } else {
-        if (this.beforeReloadCallback) {
-          this.beforeReloadCallback();
+        // If it's the first connection we just load the history
+        // otherwise we are reloading data, so we must execute some cleans
+        // before loading the full data again
+        if (this.firstConnection) {
+          this.firstConnection = false;
+          const addressesToLoad = await scanPolicyStartAddresses(this.storage);
```
r4mmer, line 423: This is agnostic to the configured scanning policy, so to add more scanning policies we should configure them on the `scanPolicyStartAddresses`

src/new/wallet.js
```diff
@@ -1144,10 +1212,15 @@ class HathorWallet extends EventEmitter {
...
-      await syncHistory(fillGapLimit.nextIndex, fillGapLimit.count, this.storage, this.conn);
+    // check address scanning policy and load more addresses if needed
+    const loadMoreAddresses = await checkScanningPolicy(this.storage);
+    if (loadMoreAddresses !== null) {
+      await syncHistory(
+        loadMoreAddresses.nextIndex,
+        loadMoreAddresses.count,
+        this.storage,
+        this.conn,
+      );
```
r4mmer, line 1223: Here we check if the new transaction received requires the wallet to load more addresses.
The index-limit will not load more addresses unless configured.

PR #583, opened 2023-10-25
feat: index limit address scanning policy

r4mmer:
### Acceptance Criteria

- Implement the new address scanning policy called index-limit
- This new scanning policy will be manually controlled and no new address will be loaded unless the user wants it.
- Adapt the wallet facade to be able to start wallets with the new scanning policy or the default one (the gap-limit policy)
- implement utils to abstract the scanning policy logic

The wallet service handles address generation so this PR will not handle any changes on the wallet-service facade.

### Security Checklist


src/new/wallet.js
```diff
@@ -1106,12 +1169,28 @@ class HathorWallet extends EventEmitter {
...
+      // but it schedule the next iteration to run after other threads.
+      await new Promise(resolve => { setTimeout(resolve, 0); });
+    }
+
+    await this.storage.processHistory();
+  }
+
+  async checkScanningPolicy() {
+    // check address scanning policy and load more addresses if needed
+    const loadMoreAddresses = await checkScanningPolicy(this.storage);
```
tuliomir, line 1186: question: I didn't understand this recursion. Is there another method with this same name?

src/new/wallet.js
```diff
@@ -1122,15 +1201,21 @@ class HathorWallet extends EventEmitter {
...
+    }
   }
 
   setState(state) {
     if (state === HathorWallet.PROCESSING && state !== this.state) {
       // XXX: will not await this so we can process history on background.
-      this.onEnterStateProcessing();
+      this.onEnterStateProcessing().catch(e => {
+        console.error(e);
+        this.setState(HathorWallet.ERROR);
```
tuliomir, line 1217: The `onEnterStateProcessing()` already handles the wallet state, right?
Should we repeat this `setState()` command here?

src/utils/storage.ts
```diff
@@ -197,15 +200,103 @@ export async function *loadAddressHistory(addresses: string[], storage: IStorage
...
+        throw new Error('Index limit is not configured');
+      }
+      return {
+        nextIndex: limits.startIndex,
+        count: limits.endIndex - limits.startIndex + 1,
+      };
+    default:
+    case 'gap-limit':
+      return {
+        nextIndex: 0,
```
tuliomir, line 224: question: This got me a bit confused: the purpose of this method is to return the _start_ address, right?

So why call this property `nextIndex`? It gives me the impression that this wallet has no transactions on all its mapped addresses ( which would be the interpretation of receiving `0` from `getNextAddress()` )

Should we call this `startIndex` instead?

tuliomir (commented):
issue: CodeCov is indicating a lack of coverage on the current tests that block the CI from passing. I'd suggest increasing the test coverage with its recommendations.

__tests__/storage/common_store.test.js
```diff
@@ -189,3 +189,63 @@ describe('registered tokens', () => {
...
+    await storage.setGapLimit(127);
+    await expect(storage.getGapLimit()).resolves.toEqual(127);
+    await expect(storage.getScanningPolicy()).resolves.toEqual('gap-limit');
+    await expect(storage.getScanningPolicyData()).resolves.toEqual({
+      policy: 'gap-limit',
+      gapLimit: 127,
+    });
+
+    // Setting index-limit
+    await storage.setScanningPolicyData({
```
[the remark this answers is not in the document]
r4mmer: There is no set behavior for `getGapLimit` when configured for index-limit.

src/new/wallet.js
r4mmer: This is just a safeguard agains unhandled errors, since we use unawaited promises I think we should at least enforce this.

src/utils/storage.ts
r4mmer: The name is just to fit with the interface and I also think we should use the same types in similar methods.

src/new/wallet.js
```diff
@@ -269,13 +270,75 @@ class HathorWallet extends EventEmitter {
...
+  async indexLimitLoadMore(count) {
+    const scanPolicy = await this.storage.getScanningPolicy();
+    if (scanPolicy !== 'index-limit') {
+      // This is a no-op
+      return;
+    }
+
+    const limits = await this.storage.getIndexLimit();
+    if (!limits) {
+      return;
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/utils/storage.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('scanning policy methods', () => {
+  it('start addresses', async () => {
+    const store = new MemoryStore();
+    const storage = new Storage(store);
+    const gapLimit = 27;
+    jest.spyOn(storage, 'getGapLimit').mockReturnValue(Promise.resolve(gapLimit));
+    const policyMock = jest.spyOn(storage, 'getScanningPolicy');
+
+    policyMock.mockReturnValue(Promise.resolve('gap-limit'));
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/storage/common_store.test.js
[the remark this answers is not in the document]
tuliomir: nitpick (non-blocking): Precisely! That's why I think this should be tested. 

This scanning policy feature changes some deep behaviors of the wallet. I think that taking care of the responses of those existing methods helps avoid breaking changes to the app.

Even though I think this is important, it's not necessary for this PR.

src/new/wallet.js
tuliomir: nitpick(non-blocking): I still think it's a bit confusing. A way to explicit this is being called on the storage would help here.

Just a comment I believe would be enough to avoid this confusion, but another more complex alternative would be importing `storageUtils` as a dedicated variable and calling `storageUtils.checkScanningPolicy`.

src/utils/storage.ts
tuliomir: I agree it's better to keep the same types.

__tests__/utils/storage.test.js
```diff
@@ -0,0 +1,40 @@
...
+    await expect(checkScanningPolicy(storage)).resolves.toEqual({
+      nextIndex: 1,
+      count: 19,
+    });
```
tuliomir, line 35: question: If this wallet is pristine and the gapLimit is `27`, why is the `count` only `19`?
Isn't it supposed to be `26`? ( [source code](https://github.com/HathorNetwork/hathor-wallet-lib/blob/7e06f89d2a62abbe01656d0dcb1313b28008f54f/src/utils/storage.ts#L304) )
```js
lastUsedAddressIndex + gapLimit - lastLoadedAddressIndex,
// -1 + 27 - 0
```

r4mmer: Nice catch, the test was incorrectly mocking the value.

src/new/wallet.js
r4mmer: Changed the name of the method to `scanAdressesToLoad` it will use the scanning policy util to determine which addresses need to be loaded and sync the history of them.

Also added a docstring to the method.

tuliomir (approved):

src/storage/memory_store.ts
```diff
@@ -750,7 +770,44 @@ export class MemoryStore implements IStore {
...
+      if (!this.walletData.scanPolicyData.gapLimit) {
+        this.walletData.scanPolicyData = {
+          policy: 'gap-limit',
+          gapLimit: GAP_LIMIT,
+        };
+      }
+      return this.walletData.scanPolicyData.gapLimit;
+    }
+    // Return default gap limit
+    return GAP_LIMIT;
```
pedroferreira1, line 783: Should we return `null` if scan policy is `index-limit`?

r4mmer: The `getGapLimit` of the Storage method will throw if the scan policy is index-limit.
This means the store method can safely return a `number` instead of `number|null`.

src/new/wallet.js
```diff
@@ -269,13 +270,65 @@ class HathorWallet extends EventEmitter {
...
   }
 
+  /**
+   * Load more addresses if configured to index-limit scanning policy.
+   * If not, do nothing.
+   * @param {number} count Number of addresses to load
+   */
+  async indexLimitLoadMore(count) {
+    const scanPolicy = await this.storage.getScanningPolicy();
+    if (scanPolicy !== 'index-limit') {
```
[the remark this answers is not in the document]
r4mmer: Same as [here](https://github.com/HathorNetwork/hathor-wallet-lib/pull/583#discussion_r1410989842)

src/new/wallet.js
```diff
@@ -269,13 +270,65 @@ class HathorWallet extends EventEmitter {
     return this.storage.setGapLimit(value);
   }
 
+  /**
+   * Load more addresses if configured to index-limit scanning policy.
+   * If not, do nothing.
+   * @param {number} count Number of addresses to load
+   */
+  async indexLimitLoadMore(count) {
```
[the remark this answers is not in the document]
r4mmer: Done

src/constants.ts
```diff
@@ -256,3 +258,8 @@ export const P2PKH_ACCT_PATH = `m/44'/${HATHOR_BIP44_CODE}'/0'`;
  * String to be prefixed before signed messages using bitcore-message
  */
 export const HATHOR_MAGIC_BYTES = 'Hathor Signed Message:\n';
+
+/**
+ * Default address scanning policy
+ */
+export const DEFAULT_ADDRESS_SCANNING_POLICY: AddressScanPolicy = 'gap-limit';
```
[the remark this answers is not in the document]
pedroferreira1: Can we create an ENUM to use in the types and in the code? It's really bad having the hardcoded strings everywhere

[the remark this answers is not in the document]
r4mmer: Done

src/types.ts
```diff
@@ -215,6 +215,11 @@ export interface IWalletAccessData {
   walletFlags: number;
 }
 
+export enum SCANNING_POLICY {
+  GAP_LIMIT = 'gap-limit',
+  INDEX_LIMIT = 'index-limit',
+};
+
 export interface IGapLimitAddressScanPolicy {
   policy: 'gap-limit';
```
pedroferreira1, line 224: Can we use the ENUM here and in the other places of this file as well?

pedroferreira1 (approved):

PR #589, opened 2023-11-07
tests: wait until the first block is mined before start the integration tests, to release the genesis reward

pedroferreira1:
### Acceptance Criteria
- We must mine at least one block before starting the integration tests, so the genesis reward is unlocked.

### Security Checklist

### DIFF
--- a/__tests__/integration/helpers/wallet.helper.js
+++ b/__tests__/integration/helpers/wallet.helper.js
@@ -396,8 +396,26 @@ export async function waitNextBlock(storage) {
   const currentHeight = await storage.getCurrentHeight();
   let height = currentHeight;
 
+  const timeout = 120000; // 120s of timeout to find the next block
+  let timeoutReached = false;
+  // Timeout handler
+  const timeoutHandler = setTimeout(() => {
+    timeoutReached = true;
+  }, timeout);
+
   while (height === currentHeight) {
+    if (timeoutReached) {
+      break;
+    }
+
     await delay(1000);
     height = await storage.getCurrentHeight();
+  }
+
+  // Clear timeout handler
+  clearTimeout(timeoutHandler);
+
+  if (timeoutReached) {
+    throw new Error('Timeout reached when waiting for the next block.');
   }
 }
--- a/__tests__/integration/helpers/wallet.helper.js
+++ b/__tests__/integration/helpers/wallet.helper.js
@@ -381,3 +381,23 @@ export async function waitUntilNextTimestamp(hWallet, txId) {
   loggers.test.log(`Waiting for ${timeToWait}ms for the next timestamp.`);
   await delay(timeToWait);
 }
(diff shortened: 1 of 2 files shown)


__tests__/integration/helpers/wallet.helper.js
```diff
@@ -381,3 +381,23 @@ export async function waitUntilNextTimestamp(hWallet, txId) {
...
+ * assume that if the height changes, a new block has arrived
+ *
+ * @param {Storage} storage
+ * @returns {Promise<void>}
+ */
+export async function waitNextBlock(storage) {
+  const currentHeight = await storage.getCurrentHeight();
+  let height = currentHeight;
+
+  while (height === currentHeight) {
```
andreabadesso, line 399: Should we implement a timeout here to prevent the CI from running indefinitely? Or does jest timeout by itself?

Maybe we can do something like:

 ```javascript
let elapsedTime = 0;
while (height === currentHeight) {
  if (elapsedTime > TIMEOUT) {
    throw new Error('Timed out waiting for next block.');
  }
  ...
}
```

**Review by tuliomir — APPROVED** [review]

File: __tests__/integration/helpers/wallet.helper.js (hunk 1 above)
Comment (pedroferreira1) [reply __tests__/integration/helpers/wallet.helper.js:406]: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/589/commits/2bab139d0273aa21654d1c5010dc4bb8d1c67e41

File: __tests__/integration/helpers/wallet.helper.js [hunk 2]
```diff
@@ -381,3 +381,41 @@ export async function waitUntilNextTimestamp(hWallet, txId) {
...
+    height = await storage.getCurrentHeight();
+  }
+
+  // Clear timeout handler
+  clearTimeout(timeoutHandler);
+
+  if (timeoutReached) {
+    throw new Error('Timeout reached when waiting for the next block.');
+  }
+}
```
andreabadesso, line 421:
```suggestion
  let timeoutReached = false;
  let elapsedTime = 0;
  
  while (height === currentHeight) {
    if (elapsedTime >= timeout) {
      timeoutReached = true;
      break;
    }

    await delay(1000);
    elapsedTime += 1000;
    height = await storage.getCurrentHeight();
  }

  if (timeoutReached) {
    throw new Error('Timeout reached when waiting for the next block.');
  }
}
```

andreabadesso (approved):

__tests__/integration/helpers/wallet.helper.js
pedroferreira1: As we discussed privately the proposed implementation has one big problem if the asynchronous calls inside the loop take a lot of time. In this case, this solution would wrongly increase only 1s in the timeout.

Because of that we decided to keep the same code.

PR #590, opened 2023-11-07
tests: start marking utxos as spent when tx is received for integration tests

pedroferreira1:
### Acceptance Criteria
- All inject funds calls must wait for the destination wallet to receive the new transaction, besides the genesis.
- Refactor of `waitForTxReceived` integration tests helper method:
  - Depending on the ws event to handle the tx received could lead to bugs, since the event might arrive even before the `waitForTxReceived`. This used to be handled with a `getTx` method but this may also cause bugs, since the tx must be added AND processed before considered to fully arrived, so all the metadata is updated.
  - Create a new attribute for transactions that arrive from the websocket `processingStatus`, so we know when these transactions have finished processing and its metadata is fully updated.
- When expecting a tx to be received in the websocket, we expect its metadata to be updated. However, the tx also affects the metadata of other transactions, e.g., the utxos that become spent, so we must consider the tx as received only after the utxos are marked as spent.
- Removed many random `delay` calls we used to have to mask these async bugs.

### Security Checklist

### DIFF
--- a/__tests__/integration/hathorwallet_facade.test.js
+++ b/__tests__/integration/hathorwallet_facade.test.js
@@ -47,8 +47,7 @@ describe('getTxById', () => {
 
     // Injecting some funds on this wallet
     const fundDestinationAddress = await hWallet.getAddressAtIndex(0);
-    const tx1 = await GenesisWalletHelper.injectFunds(fundDestinationAddress, 10);
-    await delay(1000);
+    const tx1 = await GenesisWalletHelper.injectFunds(hWallet, fundDestinationAddress, 10);
     // Validating the full history increased in one
     expect(Object.keys(await hWallet.getFullHistory())).toHaveLength(1);
 
@@ -133,7 +132,7 @@ describe('getTxById', () => {
     // Test case: custom token with funds
     const address = await hWallet.getAddressAtIndex(0);
     // Inject 10 HTR into the wallet
-    await GenesisWalletHelper.injectFunds(address, 10);
+    await GenesisWalletHelper.injectFunds(hWallet, address, 10);
     // Generate a random amount of new tokens
     const newTokenAmount = getRandomInt(1000, 10);
     // Create a new custom token with the generated amount
@@ -306,9 +305,6 @@ describe('start', () => {
   it('should start a wallet with a transaction history', async () => {
     // Send a transaction to one of the wallet's addresses
     const walletData = precalculationHelpers.test.getPrecalculatedWallet();
-    const injectAddress = walletData.addresses[0];
-    const injectValue = getRandomInt(10, 1);
-    const injectionTx = await GenesisWalletHelper.injectFunds(injectAddress, injectValue);
 
     // Start the wallet
     const walletConfig = {
@@ -321,6 +317,10 @@ describe('start', () => {
     const hWallet = new HathorWallet(walletConfig);
     await hWallet.start();
     await waitForWalletReady(hWallet);
+
+    const injectAddress = walletData.addresses[0];
+    const injectValue = getRandomInt(10, 1);
+    const injectionTx = await GenesisWalletHelper.injectFunds(hWallet, injectAddress, injectValue);
 
     // Validate that it has transactions
     const txHistory = await hWallet.getTxHistory();
@@ -401,7 +401,7 @@ describe('start', () => {
       seed: walletData.words,
       preCalculatedAddresses: walletData.addresses,
     });
-    await GenesisWalletHelper.injectFunds(await hWallet.getAddressAtIndex(0), 2);
+    await GenesisWalletHelper.injectFunds(hWallet, await hWallet.getAddressAtIndex(0), 2);
     const { hash: tokenUid } = await createTokenHelper(
       hWallet,
       'Dedicated Wallet Token',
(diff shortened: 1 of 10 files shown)


andreabadesso (changes requested):

__tests__/integration/helpers/wallet.helper.js
```diff
@@ -274,82 +273,94 @@ export function waitForWalletReady(hWallet) {
...
   const startTime = Date.now().valueOf();
-  let alreadyResponded = false;
-  const existingTx = await hWallet.getTx(txId);
+  let timeoutHandler;
+  let timeoutReached = false;
+  const timeoutPeriod = timeout || TX_TIMEOUT_DEFAULT;
+  // Timeout handler
+  timeoutHandler = setTimeout(() => {
+    timeoutReached = true;
+  }, timeoutPeriod);
```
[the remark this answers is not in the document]
pedroferreira1: As explained in the other thread, we will continue with the `setTimeout`

__tests__/integration/hathorwallet_others.test.js
```diff
@@ -899,10 +905,11 @@ describe('consolidateUtxos', () => {
       { address: addr2, value: 40, token: HATHOR_TOKEN_CONFIG.uid },
     ]);
     await waitForTxReceived(hWallet1, fundTx.hash);
+    await waitForTxReceived(hWallet2, fundTx.hash);
 
     // Sending transaction and validating the method response
     const consolidateTx = await hWallet1.consolidateUtxos(
-      await await hWallet1.getAddressAtIndex(4),
+      await hWallet1.getAddressAtIndex(4),
```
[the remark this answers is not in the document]
pedroferreira1: It seems it was working, I don't know how

__tests__/integration/helpers/wallet.helper.js
```diff
@@ -274,82 +273,94 @@ export function waitForWalletReady(hWallet) {
...
-  return new Promise(async (resolve, reject) => {
-    // Event listener
-    const handleNewTx = newTx => {
-      // Ignore this event if we didn't receive the transaction we expected.
-      if (newTx.tx_id !== txId) {
-        return;
-      }
+  // We consider that the tx was received after it's in the storage
+  // and the history processing is finished
+  while (!storageTx || storageTx.processingStatus !== 'finished') {
```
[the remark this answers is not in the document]
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/590/commits/f3ea02d1b0c363a5b097be8f4a6e9dbcb1c43384

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -306,9 +305,6 @@ describe('start', () => {
   it('should start a wallet with a transaction history', async () => {
     // Send a transaction to one of the wallet's addresses
     const walletData = precalculationHelpers.test.getPrecalculatedWallet();
-    const injectAddress = walletData.addresses[0];
-    const injectValue = getRandomInt(10, 1);
-    const injectionTx = await GenesisWalletHelper.injectFunds(injectAddress, injectValue);
```
tuliomir, line 311: With this change, the concept of this test was modified: it was supposed to validate if a wallet _already containing transactions_ in its history would be loaded correctly.

In the way it's built now, it starts _an empty wallet_ just like the test above.

The way I see it, we could either:
- Reformulate the test to stop the wallet and start it again ( we would need to get its precalculated data and store it )
- Remove this test, as it is a behavior already tested by the genesis wallet and the miner wallets helpers

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -1539,6 +1534,7 @@ describe('sendManyOutputsTransaction', () => {
     // Confirm that now the balance is available
     const sendTx = await hWallet.sendTransaction(await hWallet.getAddressAtIndex(4), 8);
     expect(sendTx).toHaveProperty('hash');
+    await waitForTxReceived(hWallet, sendTx.hash);
```
tuliomir, line 1537: We probably don't need to wait for this transaction, as the test is focused on the response object alone. It is not checking the side effects on wallet history.

__tests__/integration/helpers/wallet.helper.js
```diff
@@ -274,82 +274,94 @@ export function waitForWalletReady(hWallet) {
...
+  }
 
-      // No need to respond if the event listener worked.
-      if (alreadyResponded) {
-        return;
-      }
+  if (timeoutReached) {
+    // Throw error in case of timeout
+    throw new Error(`Timeout of ${timeoutPeriod}ms without receiving the tx with id ${txId}`);
+  } else {
```
tuliomir, line 317: We don't need this `else`. From line 316 on on we're sure `timeoutReached` is false

andreabadesso (approved):

__tests__/integration/hathorwallet_facade.test.js
pedroferreira1: I decided to continue with the test using another method to send the transaction

https://github.com/HathorNetwork/hathor-wallet-lib/pull/590/commits/4ec53859fa1be9798162eb0f735d7a83875e0d7b

What do you think?

__tests__/integration/hathorwallet_facade.test.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/590/commits/6ca745ce49d22bb18ee519203059db62f7700a4a

__tests__/integration/helpers/wallet.helper.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/590/commits/5c1085e54005ee64543b98b8f0a8d69358e6cc4b

tuliomir (approved):

PR #592, opened 2023-11-22
fix: call getHDPrivateKeyFromAddress with options as second param

alexruzenhack:
This function extracts the pin from options or get the default value from the wallet. However, it causes an error in the wallet-mobile because the wallet doesn't have a default pin.

### Acceptance Criteria
- Pass the right argument to the `getHDPrivateKeyFromAddress` function.

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2642,7 +2642,7 @@ class HathorWallet extends EventEmitter {
     }
 
     // Get private key that will sign the nano contract transaction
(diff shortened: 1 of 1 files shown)


r4mmer (approved):
Unrelated to this PR, but I checked the `getHDPrivateKeyFromAddress` method and it could use the storage method (getMainXPrivKey) to get the main private key instead of manually decrypting the key.

pedroferreira1 (approved):

PR #596, opened 2023-12-18
feat: add support for multiple data script outputs when creating a token

pedroferreira1:
### Motivation

A use case needs to create tokens with data script outputs (more than one in the same create token transaction, if possible). This is similar to the Create NFT transaction, however we would be marking this token as an NFT in our metadata bucket, which is bad for the use case.

Because of that, we decided to add support to data script outputs in the lib methods adding them in the last positions of outputs, because the NFT standard expects the data output to be in the first output.

This is a short term solution and a new project will be created to define better how to better identify an NFT token, without fixing where the data output must be.

### Acceptance Criteria
- Add support for one or many data script outputs in a token create transaction.
  - This token can't be identified as an NFT by our wallet service lambda.
- Change `getNFTCreationFee` method name to `getDataScriptOutputFee`.
  - Even though this is a breaking change and not critical for the new feature, this improves the method name.
- The method to create NFT continues accepting only one data script output, so we don't change the method signature.

### Security Checklist

### DIFF
--- a/__tests__/integration/hathorwallet_facade.test.js
+++ b/__tests__/integration/hathorwallet_facade.test.js
@@ -29,6 +29,7 @@ import Mnemonic from 'bitcore-mnemonic/lib/mnemonic';
 import { P2PKH_ACCT_PATH } from '../../src/constants';
 import Network from '../../src/models/network';
 import { WalletType, WALLET_FLAGS } from '../../src/types';
+import { parseScriptData } from '../../src/utils/scripts';
 
 const fakeTokenUid = '008a19f84f2ae284f19bf3d03386c878ddd15b8b0b604a3a3539aa9d714686e1';
 const sampleNftData = 'ipfs://bafybeiccfclkdtucu6y4yc5cpr6y3yuinr67svmii46v5cfcrkp47ihehy/albums/QXBvbGxvIDEwIE1hZ2F6aW5lIDI3L04=/21716695748_7390815218_o.jpg';
@@ -2431,6 +2432,40 @@ describe('destroyAuthority', () => {
 
     // Trying to melt and validating its error object
     await expect(hWallet.meltTokens(tokenUid, 100)).rejects.toThrow('authority output');
+  });
+});
+
+describe('create token with data outputs', () => {
+  afterEach(async () => {
+    await stopAllWallets();
+    await GenesisWalletHelper.clearListeners();
+  });
+
+  it('should create a token with data outputs', async () => {
+    const hWallet = await generateWalletHelper();
+    await GenesisWalletHelper.injectFunds(hWallet, await hWallet.getAddressAtIndex(0), 10);
+    const tx = await createTokenHelper(
+      hWallet,
+      'Token with data outputs',
+      'DOUT',
+      100,
+      {
+        data: ['test1', 'test2']
+      }
+    );
+
+    // Make sure the last 2 outputs are the data outputs
+    const lastOutput = tx.outputs[tx.outputs.length - 1];
+    expect(lastOutput.value).toBe(1);
+    expect(lastOutput.tokenData).toBe(0);
+    const lastOutputScript = parseScriptData(lastOutput.script);
+    expect(lastOutputScript.data).toBe('test2');
+
+    const outputBeforeLast = tx.outputs[tx.outputs.length - 2];
+    expect(outputBeforeLast.value).toBe(1);
+    expect(outputBeforeLast.tokenData).toBe(0);
+    const outputBeforeLastScript = parseScriptData(outputBeforeLast.script);
+    expect(outputBeforeLastScript.data).toBe('test1');
   });
 });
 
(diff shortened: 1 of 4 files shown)


__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -2434,6 +2435,40 @@ describe('destroyAuthority', () => {
...
+    expect(lastOutput.tokenData).toBe(0);
+    const lastOutputScript = parseScriptData(lastOutput.script);
+    expect(lastOutputScript.data).toBe('test2');
+
+    const outputBeforeLast = tx.outputs[tx.outputs.length - 2];
+    expect(outputBeforeLast.value).toBe(1);
+    expect(outputBeforeLast.tokenData).toBe(0);
+    const outputBeforeLastScript = parseScriptData(outputBeforeLast.script);
+    expect(outputBeforeLastScript.data).toBe('test1');
+  });
```
tuliomir, line 2469: I suggest adding another step ( or another test ) using the `CreateTokenTransaction` model `validateNft` helper method on this transaction.

This extra step would ensure that this transaction will not be incorrectly classified as an NFT by our external service.

This assertion would apply for this current solution proposed by this PR and also for the final solution that will be developed, because both would continue relying on the `validateNft` function.

A sample of code that already does this validation can be found on the Wallet Service repository ( [link](https://github.com/HathorNetwork/hathor-wallet-service/blob/ba42e76abbf4064c66feffc8bc97a0deb9d8b9bd/src/utils/nft.utils.ts#L38) )

src/utils/tokens.ts
```diff
@@ -481,7 +483,8 @@ const tokens = {
    * @param {string} [options.mintAuthorityAddress] the address to send the mint authority created
    * @param {boolean} [options.createMelt=true] Whether to create a melt output
    * @param {string} [options.meltAuthorityAddress] the address to send the melt authority created
-   * @param {string|null} [options.nftData=null] NFT data to create an NFT token
+   * @param {string[]|null} [options.data=null] list of data strings to add each as a data script output
```
r4mmer, line 486: question(if-minor): Do you think we should have the string be encoded?

For NFTs we used utf8 encoding since the data would be a string anyway, but now that we are allowing other types of data output should we have a different encoding, for instance base64 since its safe for utf8 strings and binary data.
Even if we keep using utf8 I think we should add the expected encoding on the docstrings, to make sure the caller knows how to pass the data.

src/utils/tokens.ts
```diff
@@ -239,10 +239,10 @@ const tokens = {
   },
 
   /**
-   * Get the HTR value of the fee to create a NFT
-   * @returns {number} The fee to create a NFT
+   * Get the HTR value of the fee to add a data script output
+   * @returns {number} The fee to have a data script output
    */
-  getNFTCreationFee(): number {
+  getDataScriptOutputFee(): number {
```
r4mmer, line 245: issue(if-minor): This may be used by the other wallets.

If it is being used we should remember to bump the lib and make this change on the wallets after this is released

pedroferreira1: I thought we would be using it in the wallets but we actually don't. We have method in the wallet utils but we should change to use the lib.

https://github.com/HathorNetwork/hathor-wallet/blob/master/src/utils/tokens.js#L111

src/utils/tokens.ts
```diff
@@ -527,22 +531,25 @@ const tokens = {
...
+          type: 'data',
+          data: dataString,
+          value: 1,
+          token: HATHOR_TOKEN_CONFIG.uid,
+          authorities: 0,
+        } as IDataOutput;
+
+        // We currently have an external service that identifies NFT tokens with the first output as the data output
+        // that's why we are keeping like this
+        // However, this will change after a new project is completed to better identify an NFT token
```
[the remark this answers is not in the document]
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/596/commits/987a1c35fab6d4a4e2593028e2db5b5ad7fba510

__tests__/integration/hathorwallet_facade.test.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/596/commits/010f3ae5d911a6b6263aef378d038e58e0a44663

src/utils/tokens.ts
pedroferreira1: As discussed privately, we will keep the utf-8 encoding for now (added the info in the docs https://github.com/HathorNetwork/hathor-wallet-lib/pull/596/commits/8ed69441ff119f739a7f0437dd566184b2ec9532) and discuss this support in another issue (https://github.com/HathorNetwork/hathor-wallet-lib/issues/597)

r4mmer (approved):

tuliomir (approved):

PR #598, opened 2024-01-02
fix: address index method should return null in case the address does not belong to the wallet

pedroferreira1:
### Motivation

Calling `getAddressIndex` with an address that does not belong to the wallet is throwing an error but should return `null`, according to the documentation.

```
  /**
   * Get index of address
   * Returns null if address does not belong to the wallet
   *
   * @param {string} address Address to get the index
   *
   * @return {Promise<number | null>}
   **/
```

### Acceptance Criteria
- Return `null` for the `getAddressIndex` method, if the address does not belong to the wallet.

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2215,10 +2215,7 @@ class HathorWallet extends EventEmitter {
    **/
   async getAddressIndex(address) {
     const addressInfo = await this.storage.getAddressInfo(address);
-    if (!addressInfo) {
-      return null;
-    }
-    return addressInfo.bip32AddressIndex;
+    return get(addressInfo, 'bip32AddressIndex', null);
   }
 
   /**
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2211,7 +2211,7 @@ class HathorWallet extends EventEmitter {
    *
    * @param {string} address Address to get the index
    *
-   * @return {Promise<number>}
+   * @return {Promise<number | null>}
    **/
   async getAddressIndex(address) {
     const addressInfo = await this.storage.getAddressInfo(address);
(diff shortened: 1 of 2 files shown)


src/new/wallet.js
```diff
@@ -2211,11 +2211,14 @@ class HathorWallet extends EventEmitter {
...
+   * @return {Promise<number | null>}
    **/
   async getAddressIndex(address) {
-    const addresInfo = await this.storage.getAddressInfo(address);
-    return addresInfo.bip32AddressIndex;
+    const addressInfo = await this.storage.getAddressInfo(address);
+    if (!addressInfo) {
+      return null;
+    }
+    return addressInfo.bip32AddressIndex;
```
andreabadesso, line 2221:
```
import { get } from 'lodash';
```

```suggestion
    return get(addressInfo, 'bip32AddressIndex', null);
```

r4mmer, line 2221: suggestion(if-minor)

```suggestion
    return addressInfo?.bip32AddressIndex ?? null;
```

The [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is just to safely access the index, but it returns `undefined` so the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) can return null (and it is safe for zero value unline `||`).

r4mmer (approved):

src/new/wallet.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/598/commits/210c596022cfe1c02fd375fea05add2c4dd29277

pedroferreira1: I ended up doing @andreabadesso's suggestion (https://github.com/HathorNetwork/hathor-wallet-lib/pull/598/commits/210c596022cfe1c02fd375fea05add2c4dd29277) which is pretty similar. And added a test for this method which was missing

andreabadesso (approved):

PR #618, opened 2024-03-04
chore: storage method to sign tx with custom override

r4mmer:
### Acceptance Criteria

- Allow the user to override the tx signature method.
- Default to `transactionUtils.signTransaction`, our main signature method.

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2365,11 +2365,15 @@ class HathorWallet extends EventEmitter {
    */
   async getSignatures(tx, { pinCode = null } = {}) {
     const pin = pinCode || this.pinCode;
-    const signatures = await this.storage.signTxP2PKH(tx, pin);
-    return signatures.map(sigData => ({
-      ...sigData,
-      addressPath: this.getAddressPathForIndex(sigData.addressIndex),
-    }));
+    const signatures = await this.storage.signTx(tx, pin);
+    const ret = [];
+    for (const sigData of signatures) {
+      ret.push({
+        ...sigData,
+        addressPath: await this.getAddressPathForIndex(sigData.addressIndex),
+      });
+    }
+    return ret;
   }
 
   /**
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -505,7 +505,7 @@ class HathorWallet extends EventEmitter {
     // only works because the serialized signature starts with the account path pubkey.
     const p2shSignatures = signatures.sort().map(sig => P2SHSignature.deserialize(sig));
 
-    for await (const {tx: spentTx, input, index} of this.storage.getSpentTxs(tx.inputs)) {
+    for await (const { tx: spentTx, input, index } of this.storage.getSpentTxs(tx.inputs)) {
       const spentUtxo = spentTx.outputs[input.index];
       const storageAddress = (await this.storage.getAddressInfo(spentUtxo.decoded.address));
       if (storageAddress === null) {
@@ -524,7 +524,7 @@ class HathorWallet extends EventEmitter {
           sigs.push(hexToBuffer(p2shSig.signatures[index]));
         } catch (e) {
           // skip if there is no signature, or if it's not hex
-          continue
+          continue;
         }
       }
       const inputData = walletUtils.getP2SHInputData(sigs, redeemScript);
@@ -2370,6 +2370,7 @@ class HathorWallet extends EventEmitter {
     for (const sigData of signatures) {
       ret.push({
         ...sigData,
+        signature: sigData.signature.toString('hex'),
         addressPath: await this.getAddressPathForIndex(sigData.addressIndex),
       });
     }
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -90,7 +90,6 @@ class HathorWallet extends EventEmitter {
    * @param {{pubkeys:string[],numSignatures:number}} [param.multisig]
    * @param {string[]} [param.preCalculatedAddresses] An array of pre-calculated addresses
    * @param {import('../types').AddressScanPolicyData} [param.scanPolicy] config specific to
-   * @param {boolean} [param.isSignedExternally=false] Whether we should disable the xpub guard
    * the address scan policy.
    */
   constructor({
@@ -116,7 +115,6 @@ class HathorWallet extends EventEmitter {
     multisig = null,
     preCalculatedAddresses = null,
     scanPolicy = null,
-    isSignedExternally = false,
   } = {}) {
     super();
 
@@ -205,7 +203,7 @@ class HathorWallet extends EventEmitter {
     this.newTxPromise = Promise.resolve();
 
     this.scanPolicy = scanPolicy;
-    this.isSignedExternally = isSignedExternally;
+    this.isSignedExternally = this.storage.hasTxSignatureMethod();
   }
 
   /**
@@ -2375,6 +2373,9 @@ class HathorWallet extends EventEmitter {
       throw new WalletFromXPubGuard('getSignatures');
     }
     const pin = pinCode || this.pinCode;
+    if (!pin) {
+      throw new Error(ERROR_MESSAGE_PIN_REQUIRED);
+    }
     const signatures = await this.storage.getTxSignatures(tx, pin);
     const sigInfoArray = [];
     for (const sigData of signatures) {
@@ -2750,6 +2751,15 @@ class HathorWallet extends EventEmitter {
     // Derive key to addressIndex
     const derivedKey = key.deriveNonCompliantChild(addressIndex);
     return derivedKey.privateKey;
+  }
+
+  /**
+   * Set the external tx signing method.
+   * @param {EcdsaTxSign} method
+   */
+  setExternalTxSigningMethod(method) {
+    this.isSignedExternally = true;
+    this.storage.setTxSignatureMethod(method);
   }
 }
 
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2331,7 +2331,7 @@ class HathorWallet extends EventEmitter {
    * addressPath: string,
    * }[]>} List of indexes and their associated address index
    */
-   async getWalletInputInfo(tx) {
+  async getWalletInputInfo(tx) {
     const walletInputs = [];
 
     for await (const {tx: spentTx, input, index} of this.storage.getSpentTxs(tx.inputs)) {
@@ -2364,34 +2364,12 @@ class HathorWallet extends EventEmitter {
    * }>} Input and signature information
    */
   async getSignatures(tx, { pinCode = null } = {}) {
-    if (await this.storage.isReadonly()) {
-      throw new WalletFromXPubGuard('getSignatures');
-    }
     const pin = pinCode || this.pinCode;
-    if (!pin) {
-      throw new Error(ERROR_MESSAGE_PIN_REQUIRED);
-    }
-
-    const xprivkey = await this.storage.getMainXPrivKey(pin);
-    const key = HDPrivateKey(xprivkey);
-
-    const signatures = [];
-    const dataToSignHash = tx.getDataToSignHash();
-    for (const indexes of await this.getWalletInputInfo(tx)) {
-      const { addressIndex } = indexes;
-      // Derive key to addressIndex
-      const derivedKey = key.deriveNonCompliantChild(addressIndex);
-      const privateKey = derivedKey.privateKey;
-      // Get tx signature and populate transaction
-      const sigDER = transactionUtils.getSignature(dataToSignHash, privateKey);
-      signatures.push({
-        signature: sigDER.toString('hex'),
-        pubkey: privateKey.publicKey.toString(),
-        ...indexes,
-      });
-    }
-
-    return signatures;
+    const signatures = await this.storage.signTxP2PKH(tx, pin);
+    return signatures.map(sigData => ({
+      ...sigData,
+      addressPath: this.getAddressPathForIndex(sigData.addressIndex),
+    }));
   }
 
   /**
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2755,10 +2755,10 @@ class HathorWallet extends EventEmitter {
 
   /**
    * Set the external tx signing method.
-   * @param {EcdsaTxSign} method
+   * @param {EcdsaTxSign|null} method
    */
   setExternalTxSigningMethod(method) {
-    this.isSignedExternally = true;
+    this.isSignedExternally = !!method;
     this.storage.setTxSignatureMethod(method);
   }
 }
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2366,16 +2366,16 @@ class HathorWallet extends EventEmitter {
   async getSignatures(tx, { pinCode = null } = {}) {
     const pin = pinCode || this.pinCode;
     const signatures = await this.storage.signTx(tx, pin);
-    const ret = [];
+    const sigInfoArray = [];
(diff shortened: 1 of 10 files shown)


src/new/wallet.js
```diff
@@ -1571,7 +1571,7 @@ class HathorWallet extends EventEmitter {
         isCreateNFT: newOptions.isCreateNFT,
       },
     );
-    return await transactionUtils.prepareTransaction(
+    return transactionUtils.prepareTransaction(
```
r4mmer, line 1574: note: These are just to remove unnecessary awaits.

src/new/wallet.js
```diff
@@ -2355,34 +2355,12 @@ class HathorWallet extends EventEmitter {
...
-        ...indexes,
-      });
-    }
-
-    return signatures;
+    const signatures = await this.storage.signTxP2PKH(tx, pin);
+    return signatures.map(sigData => ({
+      ...sigData,
+      addressPath: this.getAddressPathForIndex(sigData.addressIndex),
+    }));
```
r4mmer, line 2363: note: This will allow us to sign transactions for P2SH wallets using either the bitcore signing from the storage mainKey or the custom signing method.

src/storage/storage.ts
```diff
@@ -78,6 +84,17 @@ export class Storage implements IStorage {
...
+    this.txSignFunc = txSign;
+  }
+
+  async signTx(tx: Transaction, pinCode: string): Promise<IInputSignature[]> {
+    if (this.txSignFunc) {
+      return this.txSignFunc(tx, this, pinCode);
+    }
+    return transactionUtils.getSignatureForTx(tx, this, pinCode);
+  }
+
```
r4mmer, line 97: note: This is the core feature of this review.

Any signing on the wallet should be done from this method, this will allow us to maintain support for external signing methods going forward.

src/storage/storage.ts
```diff
@@ -134,6 +151,20 @@ export class Storage implements IStorage {
...
+      // public key already cached on address info
+      return addressInfo.publicKey;
+    }
+
+    // derive public key from xpub
+    const accessData = await this._getValidAccessData();
+    const hdpubkey = new HDPublicKey(accessData.xpubkey);
+    const key: HDPublicKey = hdpubkey.deriveChild(index);
+    return key.publicKey.toString('hex');
+  }
```
r4mmer, line 166: note: We only got the pubkey from the derived mainKey, this does not work for external signing where we only have the xpubkey.

src/utils/transaction.ts
```diff
@@ -123,12 +123,13 @@ const transaction = {
     return signature.toDER();
   },
 
-  async signTransaction(tx: Transaction, storage: IStorage, pinCode: string): Promise<Transaction> {
+  async getSignatureForTx(tx: Transaction, storage: IStorage, pinCode: string): Promise<IInputSignature[]> {
```
r4mmer, line 126: note: This new util is the default signing method for the wallet-lib for now.

src/storage/storage.ts
tuliomir: question: Why does codecov warns us of lack of coverage here, since this is now the main form of signing transactions on the lib? Shouldn't this be covered by most of all of our transaction tests?

tuliomir (commented):
thought(blocking): It troubles me that we're changing a central protection ( read only guards ) and removing the tests that would assure us it's working as intended later.

I'd suggest splitting this feature differently:
- First with a refactor for the transaction utils with all new signature methods. All tests would have to be green still
- Then adding a flag to allow for a full wallet with external cryptography ( the scope of #619 , as I understand ). The tests could be adapted accordingly, with almost no new tests needed.
- Lastly, implement the feature to actually initialize the wallet without any `xPriv` and having an external signing method. New tests would be made for this new feature.

What do you think of this approach?

src/storage/storage.ts
r4mmer: The integration test does not upload the coverate artifact when it fails.

The tests are actually passing, its just the wallet facade coverate that dropped slightly but it will be back once the xpub guard PR is merged

src/storage/storage.ts
```diff
@@ -134,6 +151,20 @@ export class Storage implements IStorage {
...
+    if (addressInfo?.publicKey) {
+      // public key already cached on address info
+      return addressInfo.publicKey;
+    }
+
+    // derive public key from xpub
+    const accessData = await this._getValidAccessData();
+    const hdpubkey = new HDPublicKey(accessData.xpubkey);
+    const key: HDPublicKey = hdpubkey.deriveChild(index);
+    return key.publicKey.toString('hex');
```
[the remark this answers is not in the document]
r4mmer: This is meant to be used to get the pubkey of the path `m/44'/280'/0'/0/index` where index is the address index.

When we sign the transaction with an external method we still need the actual pubkey to build the input data (on both multisig and p2pkh) but we cannot use the main key to derive it (which is the method we use on the lib).

Also, If the public key is generated during the initial address derivation we will use it (on the `addressInfo.publicKey`).

src/storage/storage.ts
```diff
@@ -134,6 +151,20 @@ export class Storage implements IStorage {
     return this.store.getAddressAtIndex(index);
   }
 
+  async getAddressPubkey(index: number): Promise<string> {
```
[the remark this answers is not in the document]
r4mmer: Done

src/new/wallet.js
```diff
@@ -2355,34 +2355,18 @@ class HathorWallet extends EventEmitter {
...
-      const derivedKey = key.deriveNonCompliantChild(addressIndex);
-      const privateKey = derivedKey.privateKey;
-      // Get tx signature and populate transaction
-      const sigDER = transactionUtils.getSignature(dataToSignHash, privateKey);
-      signatures.push({
-        signature: sigDER.toString('hex'),
-        pubkey: privateKey.publicKey.toString(),
-        ...indexes,
+    const signatures = await this.storage.signTx(tx, pin);
+    const ret = [];
```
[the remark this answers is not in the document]
r4mmer: Renamed to `sigInfoArray`

src/storage/storage.ts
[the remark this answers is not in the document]
tuliomir: Is there any downside to reusing the `deriveAddressFromXPubP2PKH()` function here instead of deriving manually?

https://github.com/HathorNetwork/hathor-wallet-lib/blob/91118d5335c2b63afb559e5deddc7a81f73cbc13/src/utils/address.ts#L27-L31 ?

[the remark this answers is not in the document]
r4mmer: This is meant to be used by both p2pkh and p2sh wallets.
If you really want to minimize repetition the best approach would be to create an util `getAddressPubkeyFromXpub` and use it on both the `deriveAddressFromXPubP2PKH` and this new method.
But I think this is creates a new method for a very simple derive operation.

src/new/wallet.js
```diff
@@ -515,7 +522,7 @@ class HathorWallet extends EventEmitter {
           sigs.push(hexToBuffer(p2shSig.signatures[index]));
         } catch (e) {
           // skip if there is no signature, or if it's not hex
-          continue
+          continue;
```
tuliomir, line 525: note(non-blocking): Codecov is raising one case that was not tested: having invalid p2sh signatures.

Since P2SH transactions involve message exchange and direct user input I think it's something actually _plausible_ to happen. Maybe we could open an issue to remind us of this test case.

src/storage/storage.ts
```diff
@@ -78,6 +84,27 @@ export class Storage implements IStorage {
     this.version = version;
   }
 
+  /**
+   * Set the tx signing function
+   * @param {EcdsaTxSign} txSign The signing function
+   */
+  setTxSign(txSign: EcdsaTxSign): void {
+    this.txSignFunc = txSign;
```
tuliomir, line 92: question: Would it be too much of an effort to create a `txSignFunc` and test it? The best case would be on an integration test, but I believe just a unit test would be enough for now.

That way we would ensure the override process is actually working.

src/storage/storage.ts
```diff
@@ -134,6 +161,26 @@ export class Storage implements IStorage {
...
   }
 
+  /**
+   * Get the address public key, if not available derive from xpub
+   * @param {number} index
+   * @async
+   * @returns {Promise<string>} The public key DER encoded in hex
+   */
+  async getAddressPubkey(index: number): Promise<string> {
+    const addressInfo = await this.store.getAddressAtIndex(index);
```
tuliomir, line 171: chore: We have no coverage on the new `getAddressPubkey` method. I feel this is important to validate the new feature.

src/storage/storage.ts
```diff
@@ -78,6 +84,27 @@ export class Storage implements IStorage {
...
+    this.txSignFunc = txSign;
+  }
+
+  /**
+   * Sign the transaction
+   * @param {Transaction} tx The transaction to sign
+   * @param {string} pinCode The pin code
+   * @returns {Promise<IInputSignature[]>} The signatures
+   */
+  async signTx(tx: Transaction, pinCode: string): Promise<IInputSignature[]> {
```
tuliomir, line 101: suggestion: This method does not sign the transaction, but only returns an array of signatures for another method to inject it into the `Transaction` object.

I'd suggest renaming this method to `getTxSignatures(tx, pinCode)` to better reflect that it is a pure function.

src/utils/transaction.ts
```diff
@@ -140,13 +141,27 @@ const transaction = {
         continue;
       }
       const xpriv = xprivkey.deriveNonCompliantChild(addressInfo.bip32AddressIndex);
+      signatures.push({
+        inputIndex,
+        addressIndex: addressInfo.bip32AddressIndex,
+        signature: this.getSignature(dataToSignHash, xpriv.privateKey),
+        pubkey: xpriv.publicKey.toDER(),
```
pedroferreira1, line 148: We used to have `xpriv.publicKey.toBuffer()` for this. Is there any difference between both codes?

src/storage/storage.ts
```diff
@@ -134,6 +161,26 @@ export class Storage implements IStorage {
     return this.store.getAddressAtIndex(index);
   }
 
+  /**
+   * Get the address public key, if not available derive from xpub
+   * @param {number} index
+   * @async
+   * @returns {Promise<string>} The public key DER encoded in hex
+   */
+  async getAddressPubkey(index: number): Promise<string> {
```
pedroferreira1, line 170: Are you using this method somewhere?

src/storage/storage.ts
r4mmer: `getSignatureForTx` is the perfect example, it fetches/derives the mainKey and uses the storage to iterate on the inputs that are owned by the wallet, signing only those that belong to the wallet.

Any other test method would essentially need to have to do the same.
During tests what I would do is create a function that prints something to stdout and call `getSignatureForTx`, so I would know it called the correct method and that it works as expected.

src/types.ts
```diff
@@ -6,9 +6,24 @@
...
 
+export interface IInputSignature {
+  inputIndex: number,
+  addressIndex: number,
+  signature: Buffer,
+  pubkey: Buffer,
+}
+
+/**
+ * Signature that receives the data to sign.
```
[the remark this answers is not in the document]
r4mmer: Yes, the initial implementation would receive the data to sign and the address index.
During the implementation I changed since this is not ideal, because it would require retrieving the key from storage and decrypting + signing it for each input instead of fetching once and signing multiple times.

This would be bad specially for HSM since we would need to start a session for each address key, even when we have 2 inputs for the same address.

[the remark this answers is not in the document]
r4mmer: I'll fix the docstring

src/storage/storage.ts
tuliomir: I'm not sure if I understand: `getSignatureForTx` is an example of signing function - actually the default one used by the lib, right?

CoveCov is informing that we have not tested this line `92`. As I understand, to increase this coverage in unit tests we would just need to call `setTxSign` with a mocked function and ensure it's actually invoked whenever a signature is required within the scope of this storage/wallet.

Integration tests for an actual valid external signature could be left to another moment.

src/storage/storage.ts
```diff
@@ -78,6 +84,27 @@ export class Storage implements IStorage {
     this.version = version;
   }
 
+  /**
+   * Set the tx signing function
+   * @param {EcdsaTxSign} txSign The signing function
+   */
+  setTxSign(txSign: EcdsaTxSign): void {
```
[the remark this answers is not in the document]
r4mmer: Having the storage hold this method makes more sense since this method will be responsible for fetching the private key, identifying which inputs to sign (and the address indexes to use) then signing.
The storage is expected to hold the private keys and addresses information.
In the future the wallets can just override this method on the storage and all internals of the wallet-lib will work.

It also makes the wallet-lib implementation simpler since some internal utilities use the storage instance and not the wallet facade instance, we would need to change them and take care to always pass the wallet facade anywhere we would need to sign anything.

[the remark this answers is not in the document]
r4mmer: If you think we should, we can also have a facade method to set this on the storage.
What do you think?

src/utils/transaction.ts
r4mmer: No, internally `toBuffer()` is the same method as `toDER()`.
[reference](https://github.com/bitpay/bitcore/blob/v10.0.28/packages/bitcore-lib/lib/publickey.js#L327)

src/storage/storage.ts
r4mmer: Not in the wallet-lib but HSM custom signature method will require this.

src/storage/storage.ts
r4mmer: Done

src/storage/storage.ts
r4mmer: Added tests

src/storage/storage.ts
r4mmer: Created an integration test showcasing this.
It will also increase the coverage since it tests the custom signature method case.

src/new/wallet.js
r4mmer: Can you create an issue for this?
I think its important, but it is beyond the scope of this PR.

src/new/wallet.js
```diff
@@ -2364,34 +2371,21 @@ class HathorWallet extends EventEmitter {
    * }>} Input and signature information
    */
   async getSignatures(tx, { pinCode = null } = {}) {
-    if (await this.storage.isReadonly()) {
+    if (await this.isReadonly()) {
       throw new WalletFromXPubGuard('getSignatures');
     }
     const pin = pinCode || this.pinCode;
-    if (!pin) {
```
pedroferreira1, line 2371: Why did you remove this?

src/storage/storage.ts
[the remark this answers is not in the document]
pedroferreira1: I agree that it's good to have it on storage because of your explanation but I think it's good also to have it on the facade. 

I think that any external client that wants to use this method shouldn't have to call the storage. So I like your suggestion, to create a method in the facade that calls this storage method.

src/new/wallet.js
r4mmer: Added the pin check back, but the rest was moved to the new signing method.

src/storage/storage.ts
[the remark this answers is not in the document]

pedroferreira1 (approved):

src/new/wallet.js
tuliomir: Opened #618 requesting this test

PR #630, opened 2024-03-29
fix: getTxHistory [1]

alexruzenhack:
### Motivation

It causes error when calling getTxHistory in the saga effect to fetch tx history on wallet-mobile.

### Acceptance Criteria
- Fix `getTxHistory` by passing network object to `getAddressFromPubkey`

### Security Checklist

### DIFF
--- a/__tests__/integration/nanocontracts/bet.test.js
+++ b/__tests__/integration/nanocontracts/bet.test.js
@@ -302,9 +302,9 @@ describe('full cycle of bet nano contract', () => {
       expect(txIds).toContain(tx.hash);
     }
 
-    // Get tx history, it must be greater then zero
+    // Get tx history with success
     const txHistory = await wallet.getTxHistory();
-    expect(txHistory).not.toHaveLength(0);
+    expect(txHistory).toHaveLength(4);
   }
 
   it('bet deposit', async () => {
--- a/__tests__/integration/nanocontracts/bet.test.js
+++ b/__tests__/integration/nanocontracts/bet.test.js
@@ -301,6 +301,10 @@ describe('full cycle of bet nano contract', () => {
     for (const tx of ncHistory2.history) {
       expect(txIds).toContain(tx.hash);
     }
+
+    // Get tx history, it must be greater then zero
+    const txHistory = await wallet.getTxHistory();
+    expect(txHistory).not.toHaveLength(0);
   }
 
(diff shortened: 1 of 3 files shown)


src/new/wallet.js
```diff
@@ -768,7 +768,7 @@ class HathorWallet extends EventEmitter {
         version: tx.version,
         ncId: tx.nc_id,
         ncMethod: tx.nc_method,
-        ncCaller: tx.nc_pubkey && getAddressFromPubkey(tx.nc_pubkey),
+        ncCaller: tx.nc_pubkey && getAddressFromPubkey(tx.nc_pubkey, this.getNetworkObject()),
```
pedroferreira1, line 771: We should have an integration test for this, since it was a bug before.

pedroferreira1 (changes requested):

src/new/wallet.js
r4mmer: Do we have a unit test at least?
If so, why did it not warn us of this?

r4mmer (approved):

src/new/wallet.js
pedroferreira1: We can use the integration tests that we already have for nano contracts at https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/__tests__/integration/nanocontracts/bet.test.js to call this method at least once to validate it. 

What do you think?

alexruzenhack: Done.

r4mmer (approved):

__tests__/integration/nanocontracts/bet.test.js
```diff
@@ -301,6 +301,10 @@ describe('full cycle of bet nano contract', () => {
     for (const tx of ncHistory2.history) {
       expect(txIds).toContain(tx.hash);
     }
+
+    // Get tx history, it must be greater then zero
+    const txHistory = await wallet.getTxHistory();
+    expect(txHistory).not.toHaveLength(0);
```
pedroferreira1, line 307: Why don't you expect the correct number of transactions, which is 5?

alexruzenhack: I can update it, but the purpose was only to call `getTxHistory` with success, and this expect is sufficient for this purpose.

PR #641, opened 2024-05-01
fix: Failing tests on CI

tuliomir:
This PR aims to fix minor bugs on the CI tests before upgrading to NodeJS 20 on #635 :

After upgrading to NodeJS 20, the `mineTransaction.test.ts` unit tests started failing. This was due to a stricter policy by Node of crashing whenever there is an unhandled promise in the application, instead of just emitting a warning as it was on NodeJS 14.

### Acceptance Criteria
- Promises on `MineTransaction` tests should always be handled
- The CI must be passing for all PRs

### Security Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -41,8 +41,10 @@ jobs:
           name: test-transacion-logs
           path: tmp
       - name: Upload coverage
-        # https://github.com/codecov/codecov-action/releases/tag/v3.1.1
-        uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
+        # https://github.com/codecov/codecov-action/releases/tag/v4.3.0
+        uses: codecov/codecov-action@84508663e988701840491b86de86b666e8a86bed
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         if: ${{ matrix.node-version }} == 14.x
         with:
           verbose: true
--- a/src/storage/leveldb/store.ts
+++ b/src/storage/leveldb/store.ts
@@ -415,6 +415,6 @@ export default class LevelDBStore implements IStore {
    * @async
    */
   async unregisterNanoContract(ncId: string): Promise<void> {
-    await this.nanoContractIndex.unregisterNanoContract(ncId);
+    return this.nanoContractIndex.unregisterNanoContract(ncId);
   }
 }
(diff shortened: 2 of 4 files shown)


.github/workflows/integration-test.yml
```diff
@@ -41,8 +41,10 @@ jobs:
           name: test-transacion-logs
           path: tmp
       - name: Upload coverage
-        # https://github.com/codecov/codecov-action/releases/tag/v3.1.1
-        uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
+        # https://github.com/codecov/codecov-action/releases/tag/v4.3.0
+        uses: codecov/codecov-action@84508663e988701840491b86de86b666e8a86bed
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
tuliomir, line 47: This is the same solution applied on #638

src/wallet/mineTransaction.ts
```diff
@@ -67,6 +67,9 @@ class MineTransaction extends EventEmitter {
 
     // Promise that resolves when push tx finishes with success
     // or rejects in case of an error
+    // TODO: If this promise is not handled properly we crash the client application with a
```
alexruzenhack, line 70: Do we have an issue to track this?

alexruzenhack (approved):

src/wallet/mineTransaction.ts
tuliomir: That's a good idea. Opened 642 to track it:
- https://github.com/HathorNetwork/hathor-wallet-lib/issues/642

src/storage/leveldb/store.ts
```diff
@@ -415,6 +415,6 @@ export default class LevelDBStore implements IStore {
    * @async
    */
   async unregisterNanoContract(ncId: string): Promise<void> {
-    this.nanoContractIndex.unregisterNanoContract(ncId);
+    await this.nanoContractIndex.unregisterNanoContract(ncId);
```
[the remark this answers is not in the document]
pedroferreira1: We don't need to `await` here, we should await where we are using this `unregisterNanoContract` method. Where are we using it that is breaking?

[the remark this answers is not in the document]
tuliomir: As we discussed, we need to somehow link the `this.nanoContractIndex` timing with the return of `unregisterNanoContract`.

Changed the `await` to a `return` to make the relationship clearer. The end product will be a `Promise<void>`, so there is no change to the caller.

pedroferreira1 (approved):

PR #644, opened 2024-05-03
feat: add first_block to ITxHistory and getTxHistory

alexruzenhack:
### Acceptance Criteria
- It should add first_block property to ITxHistory
- It should return firstBlock property on getTxHistory

### Security Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -93,7 +93,7 @@ export interface IHistoryTx {
   nc_method?: string; // For nano contract
   nc_args?: string; // For nano contract. Args in hex
   nc_pubkey?: string; // For nano contract. Pubkey in hex
-  first_block?: string; // For nano contract. It confirms the execution
+  first_block?: string;
 }
 
 export enum TxHistoryProcessingStatus {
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -93,6 +93,7 @@ export interface IHistoryTx {
   nc_method?: string; // For nano contract
   nc_args?: string; // For nano contract. Args in hex
   nc_pubkey?: string; // For nano contract. Pubkey in hex
+  first_block?: string; // For nano contract. It confirms the execution
```
pedroferreira1, line 96: I would just remove this "For nano contract". I think it's my bad, when I wrote this in the comments above, I wanted to make it clear that the attributes `nc_...` were returned only for nano contract transactions, which is not the case for the `first_block` attribute.

pedroferreira1 (dismissed):

src/types.ts
alexruzenhack: Oh! I though it was to add context. Ok, I will remove.

alexruzenhack: Cleaned: https://github.com/HathorNetwork/hathor-wallet-lib/pull/644/commits/08f33fd6a6919d27236053f6c6a94241431109f0

pedroferreira1 (approved):

tuliomir (approved):

PR #647, opened 2024-05-09
feat: add support for optional parameters in nano contract serializer/deserializer

pedroferreira1:
### Acceptance Criteria
- Add support in the serializer and deserializer class for optional parameter.
- Fix bug in the signed data deserializer, where the bytes value was not being converted to hex.

**Note:** Unfortunately, we currently don't have any blueprints with optional parameters, so we can't test this in the integration tests, thus only unit tests were added.

### Security Checklist

### DIFF
--- a/src/nano_contracts/deserializer.ts
+++ b/src/nano_contracts/deserializer.ts
@@ -22,6 +22,12 @@ class Deserializer {
    * @inner
    */
   deserializeFromType(value: Buffer, type: string): any {
+    if (type.endsWith('?')) {
+      // It's an optional
+      const optionalType = type.slice(0, -1);
+      return this.toOptional(value, optionalType);
+    }
+
     if (type.startsWith('SignedData[')) {
       return this.toSigned(value, type);
     }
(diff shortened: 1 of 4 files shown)


src/nano_contracts/deserializer.ts
```diff
@@ -22,6 +22,12 @@ class Deserializer {
    * @inner
    */
   deserializeFromType(value: Buffer, type: string): any {
+    if (type.endsWith('?')) {
```
r4mmer, line 25: Do we have a list of types or modifiers?
If so I think we should add a link to it with the [see tag](https://jsdoc.app/tags-see).

r4mmer (approved):

src/nano_contracts/deserializer.ts
tuliomir: Another alternative to a `@see` tag, since we're using Typescript on this repository, would be to add a type for the `type` parameter.

It's a developer-friendly way to have this information always at hand, and static code validation, on compatible IDEs.

tuliomir (approved):

src/nano_contracts/deserializer.ts
pedroferreira1: As we discussed in private, we currently don't have this documentation link. Also, the list of possible types is not static, so the documentation link is better.

I created an issue so we can track this when we have the documentation link. https://github.com/HathorNetwork/hathor-wallet-lib/issues/649

PR #657, opened 2024-06-03
feat: add methods to update nano contract address in storage and store

pedroferreira1:
### Acceptance Criteria
- Add methods in storages and default stores to update the address associated with a registered nano contract.
- Add `getRegisteredNanoContracts` to the `IStorage`.

### Security Checklist

### DIFF
--- a/__tests__/storage/storage.test.js
+++ b/__tests__/storage/storage.test.js
@@ -64,9 +64,11 @@ describe('handleStop', () => {
       outputs: [],
     });
     await storage.registerToken(testToken);
+    const address0 = await storage.getAddressAtIndex(0);
+    const address1 = await storage.getAddressAtIndex(1);
     const testNano = {
       ncId: 'abc',
-      address: 'nanoAddress',
+      address: address0.base58,
       blueprintId: 'blueprintId',
       blueprintName: 'blueprintName',
     };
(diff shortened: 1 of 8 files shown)


src/storage/storage.ts
```diff
@@ -1023,4 +1023,13 @@ export class Storage implements IStorage {
...
   }
+
+  /**
+   * Update nano contract registered address
+   * @param ncId Nano Contract ID.
+   * @param address New registered address
+   */
+  async updateNanoContractRegisteredAddress(ncId: string, address: string): Promise<void> {
+    return this.store.updateNanoContractRegisteredAddress(ncId, address);
+  }
```
r4mmer, line 1034: The registered address may not be a wallet address.
Do you think we should use the address index instead? This would ensure the address was ours.

We can also check here with `this.isAddressMine(address)` so we don't add an address from outside the wallet (or an invalid one).

tuliomir: suggestion: Even if the address should not necessarily be from the current wallet, an assertion that the parameter is at least a _valid Base58 address_ is important.

pedroferreira1: This validation is currently done in the wallet using this method, but I guess we can add a protection here as well

pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/657/commits/6064a6882a946cc3aa8cc86d54fb90f3032aa667

__tests__/storage/storage.test.js
```diff
@@ -85,12 +87,12 @@ describe('handleStop', () => {
...
-    await expect(store.getNanoContract('abc')).resolves.toMatchObject({ address: 'nanoAddress' });
-    await storage.updateNanoContractRegisteredAddress('abc', 'nanoAddress2');
-    await expect(store.getNanoContract('abc')).resolves.toMatchObject({ address: 'nanoAddress2' });
+    await expect(store.getNanoContract('abc')).resolves.toMatchObject({ address: address0.base58 });
+    await storage.updateNanoContractRegisteredAddress('abc', address1.base58);
+    await expect(store.getNanoContract('abc')).resolves.toMatchObject({ address: address1.base58 });
 
     // Go back to default address
-    await storage.updateNanoContractRegisteredAddress('abc', 'nanoAddress');
+    await storage.updateNanoContractRegisteredAddress('abc', address0.base58);
```
r4mmer, line 95: Can we add another test that using `nanoAddress` fails with an error?

r4mmer (changes requested):

__tests__/storage/storage.test.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/657/commits/9cfca2b4cf24e858d7e6a0f8b5e19259c23e7dcd

tuliomir (approved):

r4mmer (approved):

PR #670, opened 2024-06-13
Chore: Eslint automatic fixes

tuliomir:
### Acceptance Criteria
- All automatic adjustments by `prettier` should be applied
- All automatic fixes by `eslint --fix` should be applied
- A few manually implemented fixes for Typescript errors revealed by the automatic changes ( https://github.com/HathorNetwork/hathor-wallet-lib/pull/670/commits/5803f3e0721fa7f9d4d7d3d0e17f02f93aea4e80 )
- A necessary adaptation in a function call because its type was revealed to be incorrect by `eslint --fix` ( https://github.com/HathorNetwork/hathor-wallet-lib/pull/670/commits/81c994d405adfc6218843b2ca97a9647d8703b0e )

### Security Checklist


src/wallet/wallet.ts
```diff
@@ -1010,7 +1010,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     this.failIfWalletNotReady();
     const newOptions = {
       token: '00',
-      changeAddress: null,
+      changeAddress: undefined,
```
tuliomir, line 1013: Passing a `changeAddress` as null was not expected by the `sendManyOutputsTransaction` method and had to be changed to `undefined`, otherwise the build phase would break.

The tests had to be changed accordingly. Since this is a static type issue, it's not expected that it will cause any issues with the wallet-lib consumers.

src/wallet/wallet.ts
```diff
@@ -945,12 +1002,17 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    value: number,
+    options: { token?: string; changeAddress?: string; pinCode?: string } = {}
+  ): Promise<Transaction> {
     this.failIfWalletNotReady();
-    const newOptions = Object.assign({
+    const newOptions = {
       token: '00',
-      changeAddress: null
-    }, options);
+      changeAddress: undefined,
```
tuliomir, line 1013: Passing a `changeAddress` as null was not expected by the `sendManyOutputsTransaction` method and had to be changed to `undefined` to comply with the expected type `changeAddress?: string;`. Otherwise the build phase would break.

The tests had to be changed accordingly. Since this is a static type issue, it's not expected that it will cause any issues with the wallet-lib consumers.

### Note
Another alternative solution would be adding `null` to the valid types for this parameter: `changeAddress?: string | null`.

r4mmer (approved):

andreabadesso (approved):

PR #674, opened 2024-06-14
feat: token admin data

r4mmer:
### Acceptance Criteria
- Allow data outputs on mint and melt operations

### Security Checklist

### DIFF
--- a/__tests__/integration/hathorwallet_facade.test.js
+++ b/__tests__/integration/hathorwallet_facade.test.js
@@ -1970,6 +1970,40 @@ describe('mintTokens', () => {
     const tokenBalance4 = await hWallet.getBalance(tokenUid);
     const expectedAmount4 = expectedAmount2 + 100;
     expect(tokenBalance4[0]).toHaveProperty('balance.unlocked', expectedAmount4);
+
+    const mintResponse5 = await hWallet.mintTokens(tokenUid, 100, { data: ['foobar'] });
+    expect(mintResponse5.hash).toBeDefined();
+    await waitForTxReceived(hWallet, mintResponse5.hash);
+
+    // Validating there is a correct reference to the custom token
+    expect(mintResponse5).toHaveProperty('tokens.length', 1);
+    expect(mintResponse5.tokens[0]).toEqual(tokenUid);
+
+    // Validating custom token balance
+    const tokenBalance5 = await hWallet.getBalance(tokenUid);
+    const expectedAmount5 = expectedAmount4 + 100;
+    expect(tokenBalance5[0]).toHaveProperty('balance.unlocked', expectedAmount5);
+
+    const dataOutput5 = mintResponse5.outputs[mintResponse5.outputs.length - 1];
+    expect(dataOutput5).toHaveProperty('value', 1);
+    expect(dataOutput5).toHaveProperty('script', Buffer.from([6, 102, 111, 111, 98, 97, 114, 172]));
+
+    const mintResponse6 = await hWallet.mintTokens(tokenUid, 100, { unshiftData: true, data: ['foobar'] });
+    expect(mintResponse6.hash).toBeDefined();
+    await waitForTxReceived(hWallet, mintResponse6.hash);
+
+    // Validating there is a correct reference to the custom token
+    expect(mintResponse6).toHaveProperty('tokens.length', 1);
+    expect(mintResponse6.tokens[0]).toEqual(tokenUid);
+
+    // Validating custom token balance
+    const tokenBalance6 = await hWallet.getBalance(tokenUid);
+    const expectedAmount6 = expectedAmount5 + 100;
+    expect(tokenBalance6[0]).toHaveProperty('balance.unlocked', expectedAmount6);
+
+    const dataOutput6 = mintResponse6.outputs[0];
+    expect(dataOutput6).toHaveProperty('value', 1);
+    expect(dataOutput6).toHaveProperty('script', Buffer.from([6, 102, 111, 111, 98, 97, 114, 172]));
   });
 
   it('should deposit correct HTR values for minting', async () => {
@@ -2039,14 +2073,14 @@ describe('meltTokens', () => {
 
   it('should melt tokens', async () => {
     const hWallet = await generateWalletHelper();
-    await GenesisWalletHelper.injectFunds(hWallet, await hWallet.getAddressAtIndex(0), 10);
+    await GenesisWalletHelper.injectFunds(hWallet, await hWallet.getAddressAtIndex(0), 15);
 
     // Creating the token
     const { hash: tokenUid } = await createTokenHelper(
       hWallet,
       'Token to Melt',
       'TMELT',
-      300,
+      500,
     );
 
     // Should not melt more than there is available
@@ -2060,7 +2094,7 @@ describe('meltTokens', () => {
 
     // Validating custom token balance
     const tokenBalance = await hWallet.getBalance(tokenUid);
-    const expectedAmount = 300 - meltAmount;
+    const expectedAmount = 500 - meltAmount;
     expect(tokenBalance[0]).toHaveProperty('balance.unlocked', expectedAmount);
 
     // Melt tokens with defined melt authority address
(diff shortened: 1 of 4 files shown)


src/utils/tokens.ts
```diff
@@ -531,28 +631,6 @@ const tokens = {
...
-        // However, this will change after a new project is completed to better identify an NFT token
-        // the method that validates the NFT is in src/models/CreateTokenTransaction.validateNft
-        if (isCreateNFT) {
-          txData.outputs.unshift(outputData);
-        } else {
-          txData.outputs.push(outputData);
-        }
-      }
-    }
-
```
r4mmer, line 555: This was moved to the `prepareMintTxData`

src/utils/tokens.ts
```diff
@@ -361,6 +364,28 @@ const tokens = {
       });
     }
 
+    if (data !== null) {
```
andreabadesso, line 367: suggestion(non-blocking): Can you move this to the if in the line 310 (reordering whatever is needed)?

src/utils/tokens.ts
```diff
@@ -361,6 +364,28 @@ const tokens = {
...
+          data: dataString,
+          value: 1,
+          token: HATHOR_TOKEN_CONFIG.uid,
+          authorities: 0,
+        } as IDataOutput;
+
+        // We currently have an external service that identifies NFT tokens with the first output as the data output
+        // that's why we are keeping like this
+        // However, this will change after a new project is completed to better identify an NFT token
+        // the method that validates the NFT is in src/models/CreateTokenTransaction.validateNft
```
andreabadesso, line 380: I don't get it, we already have the `validateNft` method. What is missing?

If something is really missing, can you improve this comment to be clear on what it is?

src/utils/tokens.ts
```diff
@@ -410,7 +441,25 @@ const tokens = {
     const outputs: IDataOutput[] = [];
     const tokens = [authorityMeltInput.token];
     const depositPercent = storage.getTokenDepositPercentage();
-    const withdrawAmount = this.getWithdrawAmount(amount, depositPercent);
+    let withdrawAmount = this.getWithdrawAmount(amount, depositPercent);
+    // The deposit amount will be the quantity of data strings in the array
+    // multiplied by the fee or 0 if there are no data outputs
+    let depositAmount = data != null ? this.getDataScriptOutputFee() * data.length : 0;
```
andreabadesso, line 447: Why are you using `!=` null here? It's safe to use `!==` as we are defaulting it to `null`, right?

src/utils/tokens.ts
```diff
@@ -410,7 +441,25 @@ const tokens = {
...
+
+    // We only make these calculations if we are creating data outputs
```
andreabadesso, line 449:
```suggestion
    // We only make these calculations if we are creating data outputs because the transaction needs to deposit the fee
```

src/utils/tokens.ts
```diff
@@ -410,7 +441,25 @@ const tokens = {
...
+    if (depositAmount > 0) {
+      // If we are creating data outputs the withdrawal amount may be used to create the data outputs
+      // This may prevent finding HTR inputs to meet the deposit amount if we are creating HTR with the melt.
+      if (withdrawAmount >= depositAmount) {
+        // We can use part of the withdraw tokens as deposit
+        withdrawAmount -= depositAmount;
+        depositAmount = 0;
```
andreabadesso, line 456: Clever

src/utils/tokens.ts
```diff
@@ -438,6 +487,34 @@ const tokens = {
...
+          type: getAddressType(cAddress, storage.config.getNetwork()),
+          address: cAddress,
+          value: foundAmount - depositAmount,
+          timelock: null,
+          token: HATHOR_TOKEN_CONFIG.uid,
+          authorities: 0,
+          isChange: true,
+        });
+      }
+    }
```
andreabadesso, line 516: I'm guessing this part of the code is also present on other methods that involve depositing tokens...

Maybe open a KTLO to create an util method

src/utils/tokens.ts
```diff
@@ -462,6 +539,28 @@ const tokens = {
...
+        // that's why we are keeping like this
+        // However, this will change after a new project is completed to better identify an NFT token
+        // the method that validates the NFT is in src/models/CreateTokenTransaction.validateNft
+        if (unshiftData) {
+          outputs.unshift(outputData);
+        } else {
+          outputs.push(outputData);
+        }
+      }
+    }
```
andreabadesso, line 562: Can this be a helper?

andreabadesso (changes requested):

src/utils/tokens.ts
```diff
@@ -410,7 +441,25 @@ const tokens = {
...
+    // We only make these calculations if we are creating data outputs
+    if (depositAmount > 0) {
+      // If we are creating data outputs the withdrawal amount may be used to create the data outputs
+      // This may prevent finding HTR inputs to meet the deposit amount if we are creating HTR with the melt.
+      if (withdrawAmount >= depositAmount) {
+        // We can use part of the withdraw tokens as deposit
+        withdrawAmount -= depositAmount;
+        depositAmount = 0;
+      } else {
+        // Deposit is greater than withdraw, we will use all withdrawed tokens and still need to find utxos to meet deposit
```
pedroferreira1, line 458:
```suggestion
        // Deposit is greater than withdraw, we will use all withdrawn tokens and still need to find utxos to meet deposit
```

src/utils/tokens.ts
```diff
@@ -462,6 +539,28 @@ const tokens = {
...
+          data: dataString,
+          value: 1,
+          token: HATHOR_TOKEN_CONFIG.uid,
+          authorities: 0,
+        } as IDataOutput;
+
+        // We currently have an external service that identifies NFT tokens with the first output as the data output
+        // that's why we are keeping like this
+        // However, this will change after a new project is completed to better identify an NFT token
+        // the method that validates the NFT is in src/models/CreateTokenTransaction.validateNft
```
pedroferreira1, line 555: We can keep the `unshift` parameter for consistency but this comment is not needed in this case. It's needed in the other because it's used for creating an NFT but here we don't create a token, it's just melt.

pedroferreira1 (commented):
I think we should have integration tests for this new feature

src/utils/tokens.ts
r4mmer: The expected behavior is that data outputs should be at the end (unless it is an NFT).
We have some tests that expect this behavior.

I could move but then any additional output would need to be spliced and this behavior changes with `unshiftData`.
This makes the code simpler in my opinion, what do you think?

src/utils/tokens.ts
r4mmer: This comment was already present on the `prepareCreateTokenData` where we were adding the data outputs.
I just moved the code to the `prepareMintTxData` because we now want this behavior on both methods.

src/utils/tokens.ts
r4mmer: I like the DRY (Don't Repeat Yourself) thinking but since this logic is only used twice in the lib I don't mind leaving it repeated.

src/utils/tokens.ts
r4mmer: Done

src/utils/tokens.ts
r4mmer: Done

src/utils/tokens.ts
r4mmer: Done

src/utils/tokens.ts
r4mmer: Changed

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -1970,6 +1970,46 @@ describe('mintTokens', () => {
     const tokenBalance4 = await hWallet.getBalance(tokenUid);
     const expectedAmount4 = expectedAmount2 + 100;
     expect(tokenBalance4[0]).toHaveProperty('balance.unlocked', expectedAmount4);
+
+    // Delegate mint back to wallet 1
+    const delegateResponse = await hWallet2.delegateAuthority(tokenUid, 'mint', address0);
+    expect(delegateResponse.hash).toBeDefined();
+    await waitForTxReceived(hWallet, delegateResponse.hash);
+    await waitForTxReceived(hWallet2, delegateResponse.hash);
```
r4mmer, line 1978: We need to start with the delegate because the last call sends the authority to `hWallet2`, so we need to send it back and continue the test

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -2039,14 +2079,14 @@ describe('meltTokens', () => {
...
-    await GenesisWalletHelper.injectFunds(hWallet, await hWallet.getAddressAtIndex(0), 10);
+    await GenesisWalletHelper.injectFunds(hWallet, await hWallet.getAddressAtIndex(0), 15);
 
     // Creating the token
     const { hash: tokenUid } = await createTokenHelper(
       hWallet,
       'Token to Melt',
       'TMELT',
-      300,
+      500,
```
r4mmer, line 2089: We have 2 new requests that melt 100 each, so we need to mint more tokens before melting them

__tests__/integration/hathorwallet_facade.test.js
```diff
@@ -1970,6 +1970,46 @@ describe('mintTokens', () => {
...
+    expect(mintResponse5.tokens[0]).toEqual(tokenUid);
+
+    // Validating custom token balance
+    const tokenBalance5 = await hWallet.getBalance(tokenUid);
+    const expectedAmount5 = expectedAmount4 + 100;
+    expect(tokenBalance5[0]).toHaveProperty('balance.unlocked', expectedAmount5);
+
+    const dataOutput5 = mintResponse5.outputs[mintResponse5.outputs.length - 1];
+    expect(dataOutput5).toHaveProperty('value', 1);
+    expect(dataOutput5).toHaveProperty('script', Buffer.from([6, 102, 111, 111, 98, 97, 114, 172]));
```
r4mmer, line 1995: This script is a data output for "foobar", starting with the length (6) and ending with the opcode

__tests__/new/hathorwallet.test.js
```diff
@@ -1003,6 +1063,153 @@ describe('prepare transactions without signature', () => {
...
+        value: 1,
+      }),
+      expect.objectContaining({
+        tokenData: 129,
+        value: 2,
+      }),
+    ]));
+  });
+
+  test('prepareMeltTokensData with data outputs and selecting utxos', async () => {
```
r4mmer, line 1133: This test is very similar to the one above, but we create 2 data outputs, so we need to find 0,01 HTR to add on the inputs, the utxo found has 0.02 HTR so we need to add the change on the outputs.
So we add 0.02 HTR on the inputs and 2 new outputs (data + change)

__tests__/integration/hathorwallet_facade.test.js
pedroferreira1, line 1995: Instead of doing this, what do you think about doing the following to test if the script is correct?

```
const data = 'foobar';
const dataScript = new ScriptData(data);
expect(dataOutput5).toHaveProperty('script', dataScript.createScript());
```

pedroferreira1 (approved):

andreabadesso (approved):

PR #679, opened 2024-06-17
style: Least intrusive lint rules

tuliomir:
### Acceptance Criteria
- Fix linting errors and warnings that cause the least impact possible on the code
- Each commit should handle a single linting rule to facilitate the reviewing process
- Each `eslint-disable-*` command should have a comment explaining why it was used

### Note
To fix the `no-undef` rule it was necessary to upgrade `tsconfig.json` target from `es6` to `es2018`.

### Security Checklist


src/lib.ts
```diff
@@ -91,7 +90,7 @@ export {
   Network,
   addressUtils,
   cryptoUtils,
-  dateUtils,
+  dateFormatter as dateUtils,
```
r4mmer, line 93: We already export dateFormatter, maybe we should remove one of these and change on the projects if they use the removed name

src/models/create_token_transaction.ts
```diff
@@ -127,19 +127,23 @@ class CreateTokenTransaction extends Transaction {
     array.push(symbolBytes);
   }
 
-  getTokenInfoFromBytes(buf: Buffer): Buffer {
+  getTokenInfoFromBytes(_buf: Buffer): Buffer {
```
r4mmer, line 130: Maybe choose another name since underscore indicates that the variable is not used

src/models/create_token_transaction.ts
```diff
@@ -127,19 +127,23 @@ class CreateTokenTransaction extends Transaction {
...
     let tokenInfoVersion;
     let lenName;
     let lenSymbol;
     let bufName;
     let bufSymbol;
+    // Copies buffer locally, not to change the original parameter
+    let buf = Buffer.from(_buf);
 
+    // eslint-disable-next-line prefer-const -- To split this declaration would be confusing
```
r4mmer, line 139: How confusing would it be?
And maybe instead of disable-next-line, we should disable prefer-const on the function or file so we dont have so many disable comments

src/models/p2sh_signature.ts
```diff
@@ -57,6 +57,7 @@ class P2SHSignature {
     const signatures: Record<number, string> = {};
     for (const sig of arr.slice(1)) {
       const parts = sig.split(':');
+      // eslint-disable-next-line prefer-destructuring -- Destructuring would make this harder to read
```
r4mmer, line 60: Maybe we should use destructuring

src/nano_contracts/serializer.ts
```diff
@@ -22,6 +22,7 @@ class Serializer {
    * @memberof Serializer
    * @inner
    */
+  // eslint-disable-next-line class-methods-use-this -- XXX: This method should be made static
```
r4mmer, line 25: Maybe disable this on the whole class

src/utils/tokens.ts
```diff
@@ -415,12 +415,12 @@ const tokens = {
...
     }
 
-    const tokens = token !== null ? [token] : [];
+    const tokensArray = token !== null ? [token] : [];
 
     return {
       inputs,
       outputs,
-      tokens,
+      tokens: tokensArray,
```
r4mmer, line 423: Why the name changes in this file?

src/utils/wallet.ts
```diff
@@ -73,11 +73,11 @@ const wallet = {
       .toLowerCase();
     const wordsArray = newWordsString.split(' ');
 
-    const getInvalidWords = (words: string[]): string[] => {
+    const getInvalidWords = (_words: string[]): string[] => {
       const wordlist = Mnemonic.Words.ENGLISH;
       const errorList: string[] = [];
 
-      for (const word of words) {
+      for (const word of _words) {
```
r4mmer, line 80: why?

r4mmer (changes requested):

.eslintrc.js
```diff
@@ -42,12 +42,20 @@ module.exports = {
...
     'no-use-before-define': 0,
-    'no-unused-vars': [1, { vars: 'all', args: 'none' }],
+    'no-unused-vars': 'off', // This must be disabled for the typescript rule below to work
+    '@typescript-eslint/no-unused-vars': [
+      'warn',
+      {
+        vars: 'all',
+        args: 'none', // We should change to args: after-used
+      },
+    ],
```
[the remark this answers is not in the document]
tuliomir: Yes it will. This is in the rule [documentation instructions](https://typescript-eslint.io/rules/no-unused-vars/#how-to-use).

The only change I left documented for later is the `args: after-used` option, that would require too many refactorings and possibly breaking changes.

__tests__/__mock_helpers/wallet-service.fixtures.ts
```diff
@@ -35,6 +35,7 @@ export const buildSuccessTxByIdTokenDataResponse = (): string => {
     txTokens,
   });
 };
+// eslint-disable-next-line -- 'arrow-parens' rule conflicts with typescript syntax here
```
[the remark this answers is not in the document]
tuliomir: [`arrow-parens`](https://eslint.org/docs/latest/rules/arrow-parens) requires the parenthesis to be removed when there is only one parameter.

But since the single parameter is optional in typescript, the resulting code would be something like `const fun = param? => 'result'`, which is invalid syntax for typescript.

However, this rule is useful - especially because it can run automatically in many cases - and I believe it's still good to add it.

src/api/axiosWrapper.ts
```diff
@@ -18,18 +18,20 @@ import config from '../config';
...
+ * @param resolve (UNUSED) Callback to be stored and used in case of a retry after a fail
+ * @param _timeout Timeout in milliseconds for the request
+ * @param _additionalHeaders Headers to be sent with the request
  */
 export const axiosWrapperCreateRequestInstance = (
   url: string,
-  resolve?: Function | null,
-  timeout?: number | null,
-  additionalHeaders = {}
+  resolve?: undefined | null, // XXX: We should remove or use this parameter
```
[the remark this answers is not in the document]
tuliomir: This would be a breaking change to a method that is used in many places, both in the lib and in its clients: I took the path of just mapping the situation instead of acting now.

Do you agree with this approach?

src/api/metadataApi.ts
```diff
@@ -23,7 +23,8 @@ const metadataApi = {
       ...options,
     };
 
-    let { retries, retryInterval } = newOptions;
+    const { retryInterval } = newOptions;
+    let { retries } = newOptions;
```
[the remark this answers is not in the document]
tuliomir: Because `retryInterval` is never changed, but `retries` is. So one is a `const` to comply with `prefer-const`.

src/lib.ts
tuliomir: I agree. Opened #686 to track this removal.

src/utils/tokens.ts
tuliomir: Because of the [`no-shadow`](https://eslint.org/docs/latest/rules/no-shadow) rule. Those names were already declared in a scope above this one, so they should not be used again.

src/utils/wallet.ts
tuliomir: Because of the [`no-shadow`](https://eslint.org/docs/latest/rules/no-shadow) rule. This name was already declared in a scope above this one, so it should not be used again.

src/models/create_token_transaction.ts
tuliomir: Good point. Renamed in [3eef8f2](https://github.com/HathorNetwork/hathor-wallet-lib/pull/679/commits/3eef8f27f90e242fb2f5de5b75249d1419eb8d85)

src/models/create_token_transaction.ts
tuliomir: We would need to, for each call of `unpackToInt()`, devise a way to store the first element with a `const` and the second element in a `let`.

This could be done by either:
- Calling the method once and storing its full response in a variable and then using two destructuring assignments from it, or;
- Calling the method twice and using a different destructuring assignment for each call

I think this is not a reasonable way to write, so I prefer disabling this rule when calling `unpackToInt` for now.

As for the disabling command, it was refactored to be less repetitive on https://github.com/HathorNetwork/hathor-wallet-lib/pull/679/commits/2c120faba11bf808b123074a66bbf743748d3d75

src/models/p2sh_signature.ts
tuliomir: Done on https://github.com/HathorNetwork/hathor-wallet-lib/pull/679/commits/de3e7750df496b6c9575101ae2049de29f652e55

src/nano_contracts/serializer.ts
tuliomir: Good point. Done on https://github.com/HathorNetwork/hathor-wallet-lib/pull/679/commits/2c120faba11bf808b123074a66bbf743748d3d75

src/utils/wallet.ts
r4mmer: As commented in another file `_words` may be interpreted as not being used, can we rename to `seed` or something else?

r4mmer (dismissed):

andreabadesso (approved):

r4mmer (approved):

andreabadesso (approved):

PR #680, opened 2024-06-19
fix: consistent-return rule

tuliomir:
### Acceptance Criteria
- Fixes lint [consistent-return](https://eslint.org/docs/latest/rules/consistent-return) rule

### Security Checklist

### DIFF
--- a/src/api/metadataApi.ts
+++ b/src/api/metadataApi.ts
@@ -12,6 +12,7 @@ import { METADATA_RETRY_LIMIT, DOWNLOAD_METADATA_RETRY_INTERVAL } from '../const
 import { GetDagMetadataApiError } from '../errors';
 
 const metadataApi = {
+  // eslint-disable-next-line consistent-return -- The return is consistent, but too complex for the rule
   async getDagMetadata(id: string, network: string, options = {}) {
     type optionsType = {
       retries: number;
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -634,6 +634,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    */
   async pollForWalletStatus(): Promise<void> {
     return new Promise((resolve, reject) => {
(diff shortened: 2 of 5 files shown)


src/api/metadataApi.ts
```diff
@@ -12,6 +12,7 @@ import { METADATA_RETRY_LIMIT, DOWNLOAD_METADATA_RETRY_INTERVAL } from '../const
 import { GetDagMetadataApiError } from '../errors';
 
 const metadataApi = {
+  // eslint-disable-next-line consistent-return -- The return is consistent, but too complex for the rule
```
r4mmer, line 15: Instead of ignoring the rule maybe we could have a type for the return of this method and `return response.data as newType`.

This would make callers aware of the type and would solve the issue with the rule

src/wallet/wallet.ts
```diff
@@ -634,6 +634,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    */
   async pollForWalletStatus(): Promise<void> {
     return new Promise((resolve, reject) => {
+      // eslint-disable-next-line consistent-return -- Polling function does not need to return always
```
r4mmer, line 637: Instead of disabling the rule, why not change the return statements?

`return resolve();` to
```ts
resolve();
return;
```

Then we can add a return on the end of the method, creating a consistent return in this method.

r4mmer (changes requested):

src/api/metadataApi.ts
tuliomir: Actually, the problem with this rule wasn't the type, but the fact that some code branches would never explicitly `return`, when others did.

To fix this, a more intrusive refactor was necessary with potential to add bugs that would be hard to fetch. Since this method had no unit tests I decided to solve this in another PR:
- #702 

Would you agree in leaving this `eslint-disable` here and solving it through this other PR?

r4mmer (approved):

andreabadesso (approved):

PR #685, opened 2024-06-19
fix: no-explicit-any rule

tuliomir:
This PR is composed of two separate commits for an easier scoped review: one commit for the nano-contracts code and the other for the remaining lib code.

### Acceptance Criteria
- Fixes [`@typescript-eslint/no-explicit-any`](https://typescript-eslint.io/rules/no-explicit-any/) rule

### Security Checklist

### DIFF
--- a/src/nano_contracts/builder.ts
+++ b/src/nano_contracts/builder.ts
@@ -18,7 +18,12 @@ import {
 import Serializer from './serializer';
 import HathorWallet from '../new/wallet';
 import { NanoContractTransactionError } from '../errors';
-import { NanoContractActionType, NanoContractAction, MethodArgInfo } from './types';
+import {
+  NanoContractActionType,
+  NanoContractAction,
+  MethodArgInfo,
+  NanoContractArgumentType,
+} from './types';
 import ncApi from '../api/nano';
 import { validateBlueprintMethodArgs } from './utils';
 
@@ -34,7 +39,7 @@ class NanoContractTransactionBuilder {
 
   caller: Buffer | null;
 
-  args: any[] | null;
+  args: NanoContractArgumentType[] | null;
 
   transaction: NanoContract | null;
 
@@ -94,7 +99,7 @@ class NanoContractTransactionBuilder {
    * @memberof NanoContractTransactionBuilder
    * @inner
    */
-  setArgs(args: any[]) {
+  setArgs(args: NanoContractArgumentType[]) {
     this.args = args;
     return this;
   }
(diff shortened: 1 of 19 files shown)


r4mmer (approved):

src/nano_contracts/types.ts
```diff
@@ -11,6 +11,8 @@ export enum NanoContractActionType {
   WITHDRAWAL = 'withdrawal',
 }
 
+export type NanoContractArgumentType = string | Buffer | number | boolean;
```
pedroferreira1, line 14: This can be `null` as well.

src/nano_contracts/types.ts
```diff
@@ -73,7 +75,7 @@ export interface NanoContractHistoryAPIResponse {
 
 interface StateValueSuccess {
   // State value return
-  value: any;
+  value: NanoContractArgumentType;
```
pedroferreira1, line 78: This can't be `Buffer` because it comes from the API

src/nano_contracts/builder.ts
```diff
@@ -34,7 +39,7 @@ class NanoContractTransactionBuilder {
 
   caller: Buffer | null;
 
-  args: any[] | null;
+  args: NanoContractArgumentType[] | null;
```
pedroferreira1, line 42: We don't expect `Buffer` here

src/nano_contracts/serializer.ts
```diff
@@ -160,7 +161,7 @@ class Serializer {
    * @memberof Serializer
    * @inner
    */
-  fromOptional(value: any, type: string): Buffer {
+  fromOptional(value: NanoContractArgumentType, type: string): Buffer {
```
pedroferreira1, line 164: This `value` can be `null` as well

src/nano_contracts/types.ts
tuliomir: Added on https://github.com/HathorNetwork/hathor-wallet-lib/pull/685/commits/ae048f821e498c422ba6450dd422a10f0cd7198c

src/nano_contracts/types.ts
tuliomir: On https://github.com/HathorNetwork/hathor-wallet-lib/pull/685/commits/ae048f821e498c422ba6450dd422a10f0cd7198c I added another type named `NanoContractArgumentApiInputType` that has no Buffer. I believe this solves this issue

src/nano_contracts/builder.ts
tuliomir: Changed to the new type mentioned above.

src/nano_contracts/serializer.ts
tuliomir: This was solved by editing the type directly, as said above.

pedroferreira1 (approved):

r4mmer (approved):

PR #688, opened 2024-06-21
feat: use htr uid const where possible

r4mmer:
### Acceptance Criteria

- Change all instances of `HATHOR_TOKEN_CONFIG.uid` to `HATHOR_TOKEN_UID`

### Security Checklist

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -155,7 +155,10 @@ export const AUTHORITY_TOKEN_DATA = TOKEN_AUTHORITY_MASK | 1;
  */
 export const HATHOR_TOKEN_CONFIG = { name: 'Hathor', symbol: 'HTR', uid: '00' };
 
-export const HATHOR_TOKEN_UID = '00';
+/**
+ * Hathor token uid
+ */
+export const HATHOR_TOKEN_UID: string = '00';
 
 /**
  * Hathor token default index
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -158,7 +158,7 @@ export const HATHOR_TOKEN_CONFIG = { name: 'Hathor', symbol: 'HTR', uid: '00' };
 /**
  * Hathor token uid
  */
-export const HATHOR_TOKEN_UID: string = '00';
+export const NATIVE_TOKEN_UID: string = '00';
 
 /**
  * Hathor token default index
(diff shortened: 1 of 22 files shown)


src/constants.ts
```diff
@@ -155,6 +155,11 @@ export const AUTHORITY_TOKEN_DATA = TOKEN_AUTHORITY_MASK | 1;
  */
 export const HATHOR_TOKEN_CONFIG = { name: 'Hathor', symbol: 'HTR', uid: '00' };
 
+/**
+ * Hathor token uid
+ */
+export const HATHOR_TOKEN_UID: string = '00';
```
pedroferreira1, line 161: Why don't we call it `DEFAULT_TOKEN_UID`?

src/constants.ts
```diff
@@ -155,6 +155,11 @@ export const AUTHORITY_TOKEN_DATA = TOKEN_AUTHORITY_MASK | 1;
  */
 export const HATHOR_TOKEN_CONFIG = { name: 'Hathor', symbol: 'HTR', uid: '00' };
```
pedroferreira1, line 156:
```suggestion
export const HATHOR_TOKEN_CONFIG = { name: 'Hathor', symbol: 'HTR', uid: HATHOR_TOKEN_UID };
```

src/constants.ts
tuliomir: Another idea on top of this is to keep the `HATHOR_TOKEN_UID` as `00`, because that's what the Hathor Network actually uses.

Another constant could be `NATIVE_TOKEN_UID`, and this one could be obtained from the `config.ts` instead of `constants.js`. That way one would not need to recompile the application just to refer to a different network.

src/constants.ts
tuliomir, line 156: question: Would it be a good idea to keep the `HATHOR_TOKEN_CONFIG` unchanged and add all information about the network's native token to the `config.ts`?

All references to `HATHOR_TOKEN_CONFIG`, not only its `uid` would need to be changed by that.

src/constants.ts
r4mmer: Renamed to `NATIVE_TOKEN_UID` as discussed.

src/constants.ts
r4mmer: I don't this this is an improvement since this config will not be on the constants for long

r4mmer: As discussed, only the native token name and symbol may change on private networks and side-dags.
Leanving the `NATIVE_TOKEN_UID` a constant

tuliomir (approved):

pedroferreira1 (approved):

PR #689, opened 2024-06-24
feat: native token config

r4mmer:
### Description

This is part of development of this [issue](https://github.com/HathorNetwork/internal-issues/issues/330).

### Acceptance Criteria

- Create a way to configure the native token.
- Remove the constant `HATHOR_TOKEN_CONFIG`.
- Remove the default hathor token config from storage.

### Security Checklist

### DIFF
--- a/__tests__/storage/storage.test.js
+++ b/__tests__/storage/storage.test.js
@@ -19,6 +19,8 @@ import {
   TOKEN_MINT_MASK,
   WALLET_SERVICE_AUTH_DERIVATION_PATH,
   GAP_LIMIT,
+  DEFAULT_NATIVE_TOKEN_CONFIG,
+  NATIVE_TOKEN_UID,
 } from '../../src/constants';
 import * as cryptoUtils from '../../src/utils/crypto';
 import { InvalidPasswdError } from '../../src/errors';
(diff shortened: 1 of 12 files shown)


src/utils/storage.ts
```diff
@@ -403,6 +403,15 @@ async function updateTokensData(storage: IStorage, tokens: Set<string>): Promise
   > {
     let retryCount = 0;
 
+    if (uid === NATIVE_TOKEN_UID) {
```
pedroferreira1, line 406: Why do you need this now? Was it a bug?

src/storage/leveldb/token_index.ts
```diff
@@ -49,9 +48,6 @@ export default class LevelTokenIndex implements IKVTokenIndex {
     this.tokenDB = db.sublevel<string, ITokenData>(TOKEN_PREFIX, { valueEncoding: 'json' });
     this.metadataDB = db.sublevel<string, ITokenMetadata>(META_PREFIX, { valueEncoding: 'json' });
     this.registeredDB = db.sublevel<string, ITokenData>(REGISTER_PREFIX, { valueEncoding: 'json' });
-
-    // Add HTR to the database
-    this.saveToken(HATHOR_TOKEN_CONFIG);
```
pedroferreira1, line 54: How are we handling the default token on storage now? We need to save the native token on it, don't we?

r4mmer: Yes, the storage will start empty and the wallet will need to add the token(s) it wants the wallet to start with.

src/utils/storage.ts
r4mmer: No, its just that before the storage would always start with HTR but this changed and if the wallet does not add the native token it can reach a state where we ask the fullnode for the native token details.
This is just to save the call to the fullnode by using the configured native token.

src/constants.ts
```diff
@@ -153,12 +153,12 @@ export const AUTHORITY_TOKEN_DATA = TOKEN_AUTHORITY_MASK | 1;
...
 
 /**
- * Hathor token config
- * @property {'Hathor'} name
- * @property {'HTR'} symbol
- * @property {'00'} uid
+ * Default HTR token config
  */
-export const HATHOR_TOKEN_CONFIG = { name: 'Hathor', symbol: 'HTR', uid: '00' };
+export const DEFAULT_TOKEN_CONFIG = {
```
pedroferreira1:
```suggestion
export const DEFAULT_NATIVE_TOKEN_CONFIG = {
```

src/constants.ts
```diff
@@ -153,12 +153,12 @@ export const AUTHORITY_TOKEN_DATA = TOKEN_AUTHORITY_MASK | 1;
 export const NATIVE_TOKEN_UID: string = '00';
 
 /**
- * Hathor token config
- * @property {'Hathor'} name
- * @property {'HTR'} symbol
- * @property {'00'} uid
+ * Default HTR token config
```
pedroferreira1, line 156:
```suggestion
 * Default Native token config
```

pedroferreira1 (approved):

src/new/wallet.js
```diff
@@ -2155,7 +2156,7 @@ class HathorWallet extends EventEmitter {
   getTokenData() {
     if (this.tokenUid === NATIVE_TOKEN_UID) {
       // Hathor token we don't get from the full node
-      this.token = HATHOR_TOKEN_CONFIG;
+      this.token = this.storage.getNativeTokenData();
```
tuliomir, line 2159: suggestion(non-blocking): It would be nice to have those two changed lines inside `src/new/wallet` tested as well, even though they're not as critical.

r4mmer: Under codecov the facade lines seems to be tested

src/storage/storage.ts
```diff
@@ -95,6 +99,30 @@ export class Storage implements IStorage {
     this.version = version;
   }
 
+  /**
+   * Set the native token config
+   * @param {Omit<ITokenData, 'uid'>} tokenData The native token config
+   */
+  async setNativeTokenData(tokenData: Omit<ITokenData, 'uid'>|null|undefined): Promise<void> {
+    if (tokenData) {
+      this.nativeTokenData = tokenData;
```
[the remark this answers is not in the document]

pedroferreira1 (dismissed):

__tests__/storage/storage.test.js
```diff
@@ -193,6 +195,35 @@ describe('config version', () => {
...
+    expect(storage.getNativeTokenData()).toEqual({ name: 'Native', symbol: 'N', uid: NATIVE_TOKEN_UID });
+    storage.setApiVersion(null);
+    expect(storage.getNativeTokenData()).toEqual({ ...DEFAULT_NATIVE_TOKEN_CONFIG, uid: NATIVE_TOKEN_UID });
+  });
+
+  it('should save HTR when no custom native token is on version', async () => {
+    const store = new MemoryStore();
+    const storage = new Storage(store);
+    await expect(storage.getToken(NATIVE_TOKEN_UID)).resolves.toEqual(null);
+    await storage.saveNativeToken();
```
tuliomir, line 214: suggestion(non-blocking): The last nice to have test here would be to call `storage.saveNativeToken()` twice, to ensure its conditional is working properly.

tuliomir (dismissed):

pedroferreira1 (approved):

tuliomir (approved):

PR #692, opened 2024-06-28
feat: Workflow to sync release into master

tuliomir:
## Motivation
We sometimes forgot to merge the contents of the `release` branch back into `master` after a successful release. This workflow aims to automate the creation of the PR for this.

### Security considerations
There is no simple command on the GitHub CLI to add a card to a specific column. To do this it's necessary to execute mutation queries directly to GitHub API, with `write` permissions for the entire project, which raise a flag about the width of the necessary permissions.

The available actions that could execute this are still in beta, or were made obsolete by the GitHub Projects v2. So it was decided to implement the queries and mutation directly inside the workflow.

### Future possibilities
To facilitate the use of this functionality in other repositories, we should implement a GitHub Action inside the HathorNetwork organization with this code.

### Additional tasks
To run this workflow it's necessary to create a new access token containing the following permissions:
- Organization: Read and Write to organization projects
- Repository: Read and Write to pull requests for this specific repository ( or a limited set of repositories that can reuse this functionality )

This token should be added to this repository secrets with the name `RELEASE_TOKEN`

## Acceptance Criteria
- When the `publish release` action is taken, a new PR should be created
- This new PR should be assigned to the person who triggered the release
- This new PR should be added to HathorNetwork's engineering kanban project board
- This PR should be moved to the "In Progress (WIP)" column

### Security Checklist

### DIFF
--- a/.github/workflows/release-sync.yml
+++ b/.github/workflows/release-sync.yml
@@ -1,0 +1,162 @@
+name: Sync Release with Master
+
+# Triggered when a new version is released publicly
+on:
+  workflow_dispatch:
+  release:
+    types: [ released ]
+
+jobs:
+  create-sync-pull-request:
+    name: Create release sync PR
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+      ORG_NAME: HathorNetwork
+      PROJECT_NUMBER: 15
+      COLUMN_NAME: "In Progress (WIP)"
+
+    steps:
+      - name: Checkout
+        id: checkout
+        # https://github.com/actions/checkout/releases/tag/v4.1.6
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
+
+      - name: Create Pull Request
+        id: create_pull
+        run: |
+          # The assignee will be the user that manually published the release
+          PR_URL=$(gh pr create \
+            --title "chore: [${{ github.ref_name }}] Merge release into master" \
+            --body "Automated PR to merge `release` branch into `master` based on release event." \
+            --base "master" \
+            --head "release" \
+            --assignee "@me")
+
+          # Store value in step output
+          echo "PR_URL=$PR_URL" >> "$GITHUB_OUTPUT"
+
+      - name: Associate PR with project
+        env:
+          PR_URL: ${{ steps.create_pull.outputs.PR_URL }}
+        run: |
+          gh project item-add $PROJECT_NUMBER \
+            --owner $ORG_NAME \
+            --url "$PR_URL" \
+
+      - name: Fetch project and column data
+        id: proj_columns
+        env:
+          PR_URL: ${{ steps.create_pull.outputs.PR_URL }}
+        # There is no direct command to move a card inside a project. Here we start gathering id data about the project
+        # to do this operation through the API
+        run: |
+          QUERY="
+          query {
+            organization(login: \"$ORG_NAME\") {
+              projectV2(number: $PROJECT_NUMBER) {
+                id
+                field(name: \"Status\") {
+                  __typename
+                  ...on ProjectV2SingleSelectField {
+                    id
+                    options {
+                      id
+                      name
+                    }
+                  }
+                }
+              }
+            }
+          }
+          "
+          PROJ_RESPONSE=$(gh api graphql -f query="$QUERY")
+
+          # Parse json response and fetch necessary data identifiers from it
+          PROJ_DATA=$(echo "$PROJ_RESPONSE" | jq -r --arg column_name "$COLUMN_NAME" '
+            .data.organization.projectV2.id as $projectId |
+            .data.organization.projectV2.field.id as $fieldId |
+            .data.organization.projectV2.field.options[] |
+            select(.name == $column_name) |
+            {
+              projectId: $projectId,
+              fieldId: $fieldId,
+              optionId: .id
+            }')
+          PROJECT_ID=$(echo "$PROJ_DATA" | jq -r '.projectId')
+          FIELD_ID=$(echo "$PROJ_DATA" | jq -r '.fieldId')
+          OPTION_ID=$(echo "$PROJ_DATA" | jq -r '.optionId')
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release-sync.yml
```diff
@@ -0,0 +1,162 @@
...
+      - name: Fetch PR card Id
+        id: proj_cardId
+        env:
+          PR_URL: ${{ steps.create_pull.outputs.PR_URL }}
+        run: |
+          QUERY="
+            query {
+              organization(login: \"$ORG_NAME\") {
+                projectV2(number: $PROJECT_NUMBER) {
+                  items(first: 10) {
```
luislhl, line 104: Will this still work in case there are more than 10 cards in the column, and the one you're looking for is not among the first 10?

.github/workflows/release-sync.yml
```diff
@@ -0,0 +1,162 @@
...
+                  }
+                }
+              }
+            }
+          }
+          "
+          PROJ_RESPONSE=$(gh api graphql -f query="$QUERY")
+
+          # Parse json response and fetch necessary data identifiers from it
+          PROJ_DATA=$(echo "$PROJ_RESPONSE" | jq -r --arg column_name "$COLUMN_NAME" '
```
luislhl, line 76: suggestion (about all the code, not only this line):

Maybe writing this as a Python script, instead of bash, could improve readability and maintainability.

I think it would be easier to parse the result, and easier to use pagination in the queries in case we need it.

It would still be possible to run the command `gh api graphql` using a system shell and get the response from it.

.github/workflows/release-sync.yml
```diff
@@ -0,0 +1,162 @@
...
+                }
+              }
+            ) {
+              projectV2Item {
+                id
+              }
+            }
+          }
+          "
+          PROJ_RESPONSE=$(gh api graphql -f query="$QUERY")
```
luislhl, line 162: Is it possible that this command will fail? Or return a response indicating failure?

Should we treat this and make the workflow fail to warn us?

.github/workflows/release-sync.yml
tuliomir: That's a very important point. Our main project currently has just under 400 items, and that query surely wouldn't work.

I'll look into a way of restricting the query only for the desired repository.

.github/workflows/release-sync.yml
tuliomir: Good point. I'll add an error handler step

.github/workflows/release-sync.yml
tuliomir: I fixed this by using another approach to the query: instead of querying the project for the specific PR, I queried the PR for its related project item. This was done in https://github.com/HathorNetwork/hathor-wallet-lib/pull/692/commits/e7a592034a628f64e7cdc1a127dfff78cf61130a

.github/workflows/release-sync.yml
tuliomir: After some experimentation, I believe it will better to first release this workflow with the sync functionality and then add the error handling to it.

The main reason is because the workflow already has all the error handling and logging features we need, we just need a better channel to be informed about it: that would be adding an integration with Slack.

Since this PR already has a number of repository secrets, access tokens and configurations to be done, I prefer leaving the additional effort to another PR that will complement this one.

What do you think about this?

.github/workflows/release-sync.yml
tuliomir: I agree that Python would be a better approach to this feature.

However this solution evolved from the proof of concept I was building in bash, and converting it now would take a considerable amount of additional time. I'd prefer leaving this to a future moment when we decide to optimize this solution.

.github/workflows/release-sync.yml
luislhl: LGTM

.github/workflows/release-sync.yml
luislhl: LGTM

luislhl (dismissed):

.github/workflows/release-sync.yml
```diff
@@ -0,0 +1,157 @@
...
+
+jobs:
+  create-sync-pull-request:
+    name: Create release sync PR
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+      ORG_NAME: HathorNetwork
+      PROJECT_NUMBER: 15
+      COLUMN_NAME: "In Progress (WIP)"
```
pedroferreira1, line 17: Why not In Progress (Done)?

.github/workflows/release-sync.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Create Pull Request
+        id: create_pull
+        run: |
+          # The assignee will be the user that manually published the release
+          PR_URL=$(gh pr create \
+            --title "chore: [${{ github.ref_name }}] Merge release into master" \
+            --body "Automated PR to merge `release` branch into `master` based on release event." \
+            --base "master" \
+            --head "release" \
+            --assignee "@me")
```
pedroferreira1, line 34: Will this `@me` work? 

And can we set one of the reviewers to be `@me` as well? If the PR is automatically created, we can have this dev as one of the reviewers, I think

.github/workflows/release-sync.yml
tuliomir: Good point! Changed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/692/commits/92cfca0f48f294aadb2c0be7d2cedd78e8550d9c

.github/workflows/release-sync.yml
tuliomir: Yes, `@me` works for this command, but unfortunately can't be set as a reviewer.

This is because the actor who triggered the action will also be the one set as the _creator_ of the PR, and the creator can't be set as a reviewer.

A solution to this would be to build a dedicated bot like @dependabot , that would create the PR and then set the assignee and reviewer to the actor who triggered the release.

pedroferreira1 (approved):

luislhl (approved):

PR #706, opened 2024-07-16
feat: streaming history sync

r4mmer:
### Streamming history sync

The client will send a `request:history:<method>` where method is either `xpub` or `manual` to start syncing the history of a wallet through the websocket connection with the fullnode.
The difference between `xpub` and `manual` modes is that in `xpub` mode we send the xpubkey and the addresses are derived on the fullnode and on `manual` mode we derive the addresses and send the list of addresses the fullnode has to iterate over.

Once the stream starts the fullnode will start streaming the addresses and history.

It will send `stream:history:address` with an address, this marks the end of the history from the last address.
All transactions after this will be on this address history (until the next address event is received).

It will send `stream:history:vertex` with a transaction on the wallet history.

It will send `stream:history:end` when all addresses and transactions were sent.

It will send a `stream:history:error` when an error happens on the fullnode and we should stop the streamming.

### Acceptance Criteria

- Allow configuring a wallet to use another method of history sync.
- Implement a streamming history sync method (client to the API implemented [here](https://github.com/HathorNetwork/hathor-core/pull/1082)).
- Create tests with a mocked websocket server implementation.

### Security Checklist


src/storage/storage.ts
```diff
@@ -344,6 +356,29 @@ export class Storage implements IStorage {
...
+    startIndex: number,
+    count: number,
+    connection: FullNodeConnection,
+    shouldProcessHistory: boolean = false
+  ): Promise<void> {
+    if (
+      [HistorySyncMode.STREAM_XPUB, HistorySyncMode.STREAM_MANUAL].includes(this.historySyncMode)
+    ) {
+      const walletType = await this.getWalletType();
+      if (walletType !== WalletType.P2PKH) {
```
msbrogli, line 369: What would be the other wallet types? I'm wondering why the stream API would not work for them.

src/storage/storage.ts
msbrogli, line 1: Why are placing the selector in the storage? Why not create a separate `HistorySynchronizer` interface that would have two implementations: `PollingHttpAPI` and `StreamingWebSocketAPI`.

src/types.ts
```diff
@@ -22,6 +22,12 @@ export interface IInputSignature {
   pubkey: Buffer;
 }
 
+export enum HistorySyncMode {
+  API = 'api',
```
msbrogli, line 26: I'd use `POLLING_HTTP_API`, `MANUAL_STREAM_WS`, and `XPUB_STREAM_WS`.

src/utils/storage.ts
```diff
@@ -76,6 +96,9 @@ export async function reloadStorage(
   storage: IStorage,
   connection: FullnodeConnection
 ): Promise<void> {
+  // Stop a stream on this connection if one is running.
+  await connection.stopStream();
```
msbrogli, line 100: Maybe create a generic `connection.onReload()`?

src/new/connection.ts
```diff
@@ -31,10 +31,15 @@ class WalletConnection extends BaseConnection {
 
   static CONNECTED: number = 2;
 
+  currentStreamId: string | null = null;
```
msbrogli, line 34: Why does the connection keep this information? I guess it should belong to the streaming manager (and not the connection itself). Same for the `streamAbortController`.

src/utils/stream.ts
msbrogli, line 1: Maybe move from `utils/stream.ts` to `sync/stream.ts`?

src/utils/stream.ts
```diff
@@ -0,0 +1,399 @@
...
+export async function xpubStreamSyncHistory(
+  startIndex: number,
+  _count: number,
+  storage: IStorage,
+  connection: FullnodeConnection,
+  shouldProcessHistory: boolean = false
+) {
+  let firstIndex = startIndex;
+  const scanPolicyData = await storage.getScanningPolicyData();
+  if (isGapLimitScanPolicy(scanPolicyData)) {
```
msbrogli, line 90: What if it is not? What are the other policies? Should we support them too?

src/utils/stream.ts
```diff
@@ -0,0 +1,399 @@
...
+export async function manualStreamSyncHistory(
+  startIndex: number,
+  _count: number,
+  storage: IStorage,
+  connection: FullnodeConnection,
+  shouldProcessHistory: boolean = false
+) {
+  let firstIndex = startIndex;
+  const scanPolicyData = await storage.getScanningPolicyData();
+  if (isGapLimitScanPolicy(scanPolicyData)) {
```
msbrogli, line 114: Same here.

src/new/connection.ts
msbrogli: I guess I got a better understanding of what you're doing. Maybe create just one attribute that stores all information about the current streaming and also knows how to abort?

src/utils/stream.ts
```diff
@@ -0,0 +1,399 @@
...
+          // Generate next batch if needed
+          batchGenerationPromise = batchGenerationPromise.then(async () => {
+            await generateNextBatch();
+          });
+        }
+        if (isStreamSyncHistoryEnd(wsData)) {
+          // cleanup and stop the method.
+          // console.log(`Stream end at ${Date.now()}`);
+          // This will resolve the promise and clean the listener on the connection
+          abortController.abort();
```
msbrogli, line 342: Why would you abort if the streamer signaled that it had been successfully completed?

src/utils/stream.ts
```diff
@@ -0,0 +1,399 @@
...
+ * Start a stream to sync the history of the wallet on `storage`.
+ * Since there is a lot of overlap between xpub and manual modes this method was created to accomodate both.
+ * @param {number} startIndex Index to start loading addresses
+ * @param {IStorage} storage The storage to load the addresses
+ * @param {FullnodeConnection} connection Connection to the full node
+ * @param {boolean} shouldProcessHistory If we should process the history after loading it.
+ * @param {HistorySyncMode} mode The mode of the stream
+ * @returns {Promise<void>}
+ */
+export async function streamSyncHistory(
```
msbrogli, line 139: I think this method is "doing too much" which makes it too complex and error prone.

msbrogli (changes requested):

src/storage/storage.ts
r4mmer: The other wallet type is P2SH for MultiSig wallets.
We could have support for them on manual stream mode but I believe we can add this support later.
What do you think?

src/storage/storage.ts
r4mmer: We can refactor this to be a class based implementation but do you want to move this configuration to the wallet facade instead of the storage?

src/utils/stream.ts
r4mmer: Yes, but I kept the funtion implementation to be consistent with our existing `syncHistory`.
I think once we have a functional implementation and integration testing we could add a refactor task to make these class based implementations like you suggested.

What do you think?

src/types.ts
r4mmer: Done

src/utils/storage.ts
r4mmer: Done

src/new/connection.ts
r4mmer: Done

src/utils/stream.ts
r4mmer: Done

src/utils/stream.ts
r4mmer: Changed to only resolve.

The idea was that the abort controller has some other responsabilities like stopping the client from generating more addresses and removing the event listener to avoid memory leaks.
We already have a call to the abort method on the finally so we can just resolve the promise, and let the cleanup be done when the function ends.

src/new/connection.ts
```diff
@@ -99,6 +113,89 @@ class WalletConnection extends BaseConnection {
...
+      if (this.streamController === null) {
+        // There is no active stream.
+        resolve();
+        return;
+      }
+      // Create a timeout so we do not wait indefinetely
+      // It it reaches here we should reject since something went wrong.
+      const timer = setTimeout(() => {
+        reject();
+      }, 10000);
```
msbrogli, line 180: Hard coded parameter. I guess you should make it at least a constant in this file.

src/new/connection.ts
```diff
@@ -99,6 +113,89 @@ class WalletConnection extends BaseConnection {
...
+
+  lockStream(streamId: string): boolean {
+    if (this.streamController === null) {
+      this.streamController = new StreamController(streamId);
+      return true;
+    }
+    return false;
+  }
+
+  startStreamingHistory(id: string, firstIndex: number, xpubkey: string, gapLimit: number = -1) {
```
msbrogli, line 130: Maybe rename to `sendStartXPubStreamingHistory`?

src/new/connection.ts
```diff
@@ -99,6 +113,89 @@ class WalletConnection extends BaseConnection {
...
+    if (this.streamController?.streamId !== id) {
+      throw new Error('There is an on-going stream, cannot start a second one');
+    }
+    if (this.websocket) {
```
msbrogli, line 134: Maybe raise an exception in the else block?

src/new/connection.ts
```diff
@@ -99,6 +113,89 @@ class WalletConnection extends BaseConnection {
...
+    id: string,
+    firstIndex: number,
+    addresses: [number, string][],
+    first: boolean,
+    gapLimit: number = -1
+  ) {
+    if (this.streamController?.streamId !== id) {
+      throw new Error('There is an on-going stream, cannot start a second one');
+    }
+    if (this.websocket) {
```
msbrogli, line 156: Maybe raise an exception in the else block?

src/new/connection.ts
```diff
@@ -99,6 +113,89 @@ class WalletConnection extends BaseConnection {
...
+      }
+      // Create a timeout so we do not wait indefinetely
+      // It it reaches here we should reject since something went wrong.
+      const timer = setTimeout(() => {
+        reject();
+      }, 10000);
+
+      // We have an active stream.
+      // We will wait for the stream to end then resolve.
+      this.once('stream-end', () => {
```
msbrogli, line 184: An `stream-error` message would also close the stream.

src/storage/storage.ts
msbrogli: Got it. But wouldn't a better verification logic already include it? Something like:

```typescript
const supportedSyncMode = getSupportedSyncMode(this.getWalletType())
if not supportedSyncMode.includes(this.historySyncMode) {
    throw new Error('...')
}
```

Where:

```typescript
getSupportedSyncMode(walletType): set<historySyncMode> {
  if (walletType !== WalletType.P2PKH) {
    return set([HistorySyncMode.XPUB_STREAM_WS, HistorySyncMode.MANUAL_STREAM_WS])
  else if (walletType == WalletType.P2SH) {
    return set([HistorySyncMode.MANUAL_STREAM_WS])
  } else {
    throw new Error('invalid wallet type')
  }
}
```

src/utils/storage.ts
```diff
@@ -19,18 +19,38 @@ import {
...
+ */
+export function getHistorySyncMethod(mode: HistorySyncMode): HistorySyncFunction {
+  switch (mode) {
+    case HistorySyncMode.MANUAL_STREAM_WS:
+      return manualStreamSyncHistory;
+    case HistorySyncMode.XPUB_STREAM_WS:
+      return xpubStreamSyncHistory;
+    default:
+      // case HistorySyncMode.POLLING_HTTP_API
+      return apiSyncHistory;
```
msbrogli, line 50:
```suggestion
    case HistorySyncMode.POLLING_HTTP_API:
    default:
      return apiSyncHistory;
```

msbrogli (changes requested):

src/utils/stream.ts
msbrogli: Got it. Great! I'm wondering whether just StreamController would be a better name since it manages other stuff.

src/utils/stream.ts
msbrogli: I agree. Let's add it as a task for the next phase of this project. Then we can prioritize it later.

src/utils/stream.ts
msbrogli: Thanks!

src/new/connection.ts
msbrogli: Thanks!

src/utils/storage.ts
msbrogli: Thanks!

src/types.ts
msbrogli: Thanks!

src/storage/storage.ts
msbrogli: I'm wondering why the sync mode would matter to the storage. We can refactor it in a future phase of this project if you think it would be a bigger refactor.

src/utils/stream.ts
msbrogli: I'm ok moving forward like this if we need to speed up the release of this feature. Otherwise, I guess we can create a new structure that is transparent to the caller regardless of the sync mode. I don't necessarily think we need a class based implementation. My feeling is that the current implementation is confusing and error prone.

src/new/connection.ts
r4mmer: There is no `stream-error` event.
The method syncHistory fails with an error when the fullnode returns an error, but we emit a `stream-end` event to the connection instance.

src/utils/stream.ts
r4mmer: I can rename the variable it but [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) is from javascript so we can't rename the class.

src/new/connection.ts
r4mmer: Done

src/new/connection.ts
r4mmer: Done

src/new/connection.ts
r4mmer: Done

src/new/connection.ts
r4mmer: Done

src/utils/storage.ts
r4mmer: Done

src/utils/stream.ts
r4mmer: Made some changes to move the responsabilities of the client to a stream manager instance and the `streamSyncHistory` now only makes use of the manager to start, receive and cleanup streamming resources.

src/storage/storage.ts
r4mmer: Makes sense, moved the sync configuration to the wallet facade

src/storage/storage.ts

src/new/connection.ts
```diff
@@ -99,6 +118,98 @@ class WalletConnection extends BaseConnection {
...
+      if (this.streamController === null) {
+        // There is no active stream.
+        resolve();
+        return;
+      }
+      // Create a timeout so we do not wait indefinetely
+      // It it reaches here we should reject since something went wrong.
+      const timer = setTimeout(() => {
+        reject();
+      }, STREAM_ABORT_TIMEOUT);
```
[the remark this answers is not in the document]
r4mmer: `reject` type does not match the `setTimeout` first parameter, so typescript would fail to build.

src/sync/stream.ts
```diff
@@ -0,0 +1,486 @@
...
+  }
+
+  const manager = new StreamManager(
+    firstIndex,
+    storage,
+    connection,
+    HistorySyncMode.MANUAL_STREAM_WS
+  );
+  await streamSyncHistory(manager, shouldProcessHistory);
+}
```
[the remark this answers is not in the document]
r4mmer: That is what `streamSyncHistory` was made for.

I think that the logic to build the manager is simple enough to be duplicated for both modes without issue.
What do you think?

src/new/connection.ts
```diff
@@ -99,6 +118,98 @@ class WalletConnection extends BaseConnection {
...
+
+  async stopStream() {
+    await new Promise<void>((resolve, reject) => {
+      if (this.streamController === null) {
+        // There is no active stream.
+        resolve();
+        return;
+      }
+      // Create a timeout so we do not wait indefinetely
+      // It it reaches here we should reject since something went wrong.
```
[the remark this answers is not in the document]
r4mmer: Fixed

src/new/connection.ts
```diff
@@ -99,6 +118,98 @@ class WalletConnection extends BaseConnection {
...
+    }
+
+    const data = JSON.stringify({
+      id,
+      first,
+      addresses,
+      type: 'request:history:manual',
+      'first-index': firstIndex,
+      'gap-limit': gapLimit,
+    });
```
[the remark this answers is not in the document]
r4mmer: Done

src/new/connection.ts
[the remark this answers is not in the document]
glevco: Thanks!

src/new/connection.ts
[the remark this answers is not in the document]
glevco: Ok!

src/new/connection.ts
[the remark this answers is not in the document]
glevco: Thanks!

src/sync/stream.ts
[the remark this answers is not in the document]
glevco: I think it's always better to prevent duplication, but no problem

glevco (dismissed):

src/new/connection.ts
msbrogli: Got it. Thanks!

src/sync/stream.ts
```diff
@@ -0,0 +1,486 @@
...
+function buildListener(manager: StreamManager, resolve: () => void) {
+  return (wsData: IStreamSyncHistoryData) => {
+    // Early return if the stream is aborted
+    // This will prevent any tx or address from being added when we want to stop the stream.
+    if (manager.signal.aborted) {
+      return;
+    }
+    // Only process the message if it is from our stream, this error should not happen.
+    if (wsData.id !== manager.streamId) {
+      // Check that the stream id is the same we sent
```
msbrogli, line 396: Should we log a warning? It seems this condition should never happen.

src/sync/stream.ts
```diff
@@ -0,0 +1,486 @@
...
+    }
+    if (isStreamSyncHistoryEnd(wsData)) {
+      // cleanup and stop the method.
+      resolve();
+    }
+    // An error happened on the fullnode, we should stop the stream
+    if (isStreamSyncHistoryError(wsData)) {
+      console.error(`Stream error: ${wsData.errmsg}`);
+      manager.abortWithError(wsData.errmsg);
+    }
```
msbrogli, line 419: Isn't it better to use the following structure:

```
if (a) {
} else if (b) {
} else if (c) {
} else {
    throw new "should never happen";
}

src/storage/storage.ts
msbrogli: Thanks!

src/storage/storage.ts
msbrogli: Thanks!

src/utils/stream.ts
msbrogli: You can close this thread after you add a new task to get this done later in the project.

src/utils/stream.ts
msbrogli: Got it. Ok. Thanks for trying.

src/utils/stream.ts
msbrogli: I guess it's way better. Good job! Thanks!

src/new/connection.ts
msbrogli: Thanks!

src/new/connection.ts
msbrogli: Thanks!

src/utils/storage.ts
msbrogli: Thanks!

src/new/connection.ts
msbrogli: Thanks!

src/new/connection.ts
msbrogli: Thanks!

msbrogli (approved):

src/utils/stream.ts
r4mmer: The issue https://github.com/HathorNetwork/hathor-wallet-lib/issues/709 was created to track this.

src/sync/stream.ts
r4mmer: Done

src/sync/stream.ts
r4mmer: Done

src/sync/stream.ts
msbrogli: Thanks!

src/sync/stream.ts
msbrogli: Thanks!

msbrogli (dismissed):

glevco (approved):

msbrogli (approved):

PR #708, opened 2024-07-23
feat: add support for maximum output values

glevco:
Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/799

### Motivation

This is the last PR in the `bigint` refactor, and it changes the `OutputValueType` to `bigint`. It leverages the structures implemented in previous PRs to support this change.

### Acceptance Criteria

- Change `OutputValueType` from `number` to `bigint`.
- Change number literals to bigint literals (`123` -> `123n`). This is the bulk of this PR as tests have a lot of literals.
- Add explicit conversions to/from BigInt/Number where necessary.
- Some other related changes such as using `-` instead of `Math.abs` as the `Math` module does not support `bigint`.

### Security Checklist


.github/workflows/integration-test.yml
```diff
@@ -5,6 +5,7 @@ on:
       - release
       - release-candidate
       - master
+      - refactor/ts-tests # TODO: REMOVE THIS
```
glevco, line 8: TODO: Remove

.github/workflows/main.yml
```diff
@@ -12,6 +12,7 @@ on:
     - release
     - release-candidate
     - master
+    - refactor/ts-tests # TODO: REMOVE THIS
```
glevco, line 15: TODO: Remove

.github/workflows/integration-test.yml
```diff
@@ -5,7 +5,7 @@ on:
       - release
       - release-candidate
       - master
-      - refactor/typing-improvements # TODO: REMOVE THIS
+      - feat/bigint-functions # TODO: REMOVE THIS
```

.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ on:
     - release
     - release-candidate
     - master
-    - refactor/typing-improvements # TODO: REMOVE THIS
+    - feat/bigint-functions # TODO: REMOVE THIS
```

.github/workflows/integration-test.yml
```diff
@@ -12,6 +12,7 @@ on:
       - release
       - release-candidate
       - master
+      - chore/remove-node-v20 # TODO: REMOVE THIS
```
glevco, line 15: Remove this before merging

.github/workflows/lint.yml
```diff
@@ -12,6 +12,7 @@ on:
     - release
     - release-candidate
     - master
+    - chore/remove-node-v20 # TODO: REMOVE THIS
```
glevco, line 15: Remove this before merging

.github/workflows/unit-test.yml
```diff
@@ -12,6 +12,7 @@ on:
     - release
     - release-candidate
     - master
+    - chore/remove-node-v20 # TODO: REMOVE THIS
```
glevco, line 15: Remove this before merging

src/utils/tokens.ts
```diff
@@ -271,15 +271,15 @@ const tokens = {
     mintAmount: OutputValueType,
     depositPercent: number = TOKEN_DEPOSIT_PERCENTAGE
   ): OutputValueType {
-    return Math.ceil(depositPercent * mintAmount);
+    return BigInt(Math.ceil(depositPercent * Number(mintAmount)));
```
pedroferreira1, line 274: Won't this `Number(mintAmount)` affect the deposit amount if the `mintAmount` is too big?

src/constants.ts
```diff
@@ -105,13 +105,12 @@ export const TOKEN_INFO_VERSION = 1;
...
  */
-export const MAX_OUTPUT_VALUE_32: OutputValueType = 2 ** 31 - 1;
+export const MAX_OUTPUT_VALUE_32: OutputValueType = 2n ** 31n - 1n;
 
 /**
  * Max accepted value for an output
- * Because of a precision problem in javascript we don't handle all 8 bytes of value
  */
-export const MAX_OUTPUT_VALUE: OutputValueType = 2 ** 43;
+export const MAX_OUTPUT_VALUE: OutputValueType = 2n ** 63n;
```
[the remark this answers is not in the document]
glevco: We do, because the first bit is used to signal the value size, so when using 8 bytes the maximum value is not `2**64-1`, but rather `2**63`. This is enforced in `hathor-core`.

src/constants.ts
```diff
@@ -133,13 +132,13 @@ export const TOKEN_AUTHORITY_MASK: number = 0b10000000;
  * Mask to check if it's mint UTXO (last bit indicates it)
  * For further information: https://gitlab.com/HathorNetwork/rfcs/blob/master/text/0004-tokens.md
  */
-export const TOKEN_MINT_MASK: OutputValueType = 0b00000001;
+export const TOKEN_MINT_MASK: OutputValueType = 0b00000001n;
```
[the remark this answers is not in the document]
glevco: The mint and melt masks are applied to the output value, which is bigint, but the authority and index masks are applied to the `token_data`, which is still a number. It's convenient to have the masks match the types of what they mask, so we don't have to convert between them.

src/models/output.ts
```diff
@@ -93,12 +92,12 @@ class Output {
       throw new OutputValueError('Output value must be positive');
     }
     if (this.value > MAX_OUTPUT_VALUE) {
-      throw new OutputValueError(`Maximum value is ${prettyValue(MAX_OUTPUT_VALUE)}`);
+      throw new OutputValueError(`Maximum value is ${MAX_OUTPUT_VALUE}`);
     }
```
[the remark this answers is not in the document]
glevco: Please see https://github.com/HathorNetwork/hathor-wallet-lib/pull/708#discussion_r1857453757

src/models/transaction.ts
```diff
@@ -292,7 +292,10 @@ class Transaction {
       txSize += 64;
     }
 
-    let sumOutputs = this.getOutputsSum();
+    // Here we have to explicitly convert a bigint to a number (a double), which loses precision,
+    // but is exactly compatible with the reference Python implementation because of the calculations below.
+    let sumOutputs = Number(this.getOutputsSum());
```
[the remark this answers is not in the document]
glevco: We could keep the `sumOutputs` variable as a `BigInt`, but that wouldn't be useful because its only purpose is to calculate the `amount` below, which is then used to calculate the weight:

https://github.com/HathorNetwork/hathor-wallet-lib/blob/107d93e76c374d570e5e7227d49b8bb3a41f67e0/src/models/transaction.ts#L297-L309

Since the `weight` is a `Number` and the constants used to calculate it are too, `amount` must also be a `Number`, otherwise JS would error (`TypeError: Cannot mix BigInt and other types, use explicit conversions`).

We have the same problem in Python, but Python does the conversion automatically, so we don't have to do it explicitly:

https://github.com/HathorNetwork/hathor-core/blob/4da0fb9186793ee1116eab39915441909fcf1c28/hathor/daa.py#L201-L208

In that case, `tx.sum_outputs` is an `int` but `amount` is a `float`.

This does mean that we lose precision, but as stated in code comment, this is compatible with the weight calculated by the full node.

src/utils/tokens.ts
glevco: Yes it will. This is exactly the same problem as described in https://github.com/HathorNetwork/hathor-wallet-lib/pull/708#discussion_r1857469627, and again it conforms to the protocol implementation in Python, which does this conversion implicitly, losing precision:

https://github.com/HathorNetwork/hathor-core/blob/4da0fb9186793ee1116eab39915441909fcf1c28/hathor/transaction/util.py#L54-L59

In practice, this can never be a problem in mainnet because the maximum number of existing tokens does not reach the boundary where a double loses integer precision. For Ekvilibro, I think it's possible that there exists a number of tokens that would lose precision here, but again it's not too big of a problem because this is compatible with what the network (hathor-core) expects.

The worse that could happen is that they could deposit a bit more tokens than they withdraw after a mint and melt. This is already the expected behavior for low amounts which would require deposits of less than 1 HTR (you deposit 1 and withdraw 0). Even then it's not an issue because we created an amount of native tokens that more than covers the amount of custom tokens they expect to mint. To be safe, we should keep this is mind for other use cases, though.

__tests__/integration/hathorwallet_facade.test.ts
```diff
@@ -2179,29 +2185,27 @@ describe('meltTokens', () => {
...
 
     // Creating the token
-    const { hash: tokenUid } = await createTokenHelper(hWallet, 'Token to Melt', 'TMELT', 500);
+    const { hash: tokenUid } = await createTokenHelper(hWallet, 'Token to Melt', 'TMELT', 500n);
 
     // Should not melt more than there is available
-    await expect(hWallet.meltTokens(tokenUid, 999)).rejects.toThrow(
-      'Not enough tokens to melt: 999 requested, 500 available'
-    );
+    await expect(hWallet.meltTokens(tokenUid, 999n)).rejects.toThrow('Not enough tokens to melt');
```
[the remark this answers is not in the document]
glevco: Not sure if there was any reason to remove it, as this was done a while ago. Readded in https://github.com/HathorNetwork/hathor-wallet-lib/pull/708/commits/31b2c2dd9a43dd1de50e2d4c46b86214800bdabf

src/utils/tokens.ts
pedroferreira1: Makes sense, I ran the test with python and js

```python
>>> math.ceil(abs(2**63 * 0.01))
92233720368547760
```

```js
> Math.ceil(Number(2n**63n) * 0.01)
92233720368547760
```

Can you just add a comment regarding this in the code, please?

pedroferreira1 (approved):

src/utils/tokens.ts
glevco: Sure! Done in https://github.com/HathorNetwork/hathor-wallet-lib/pull/708/commits/3f0df644ff4b75edac3dd0e800d5963cb25f1f5f

r4mmer (approved):

pedroferreira1 (approved):

pedroferreira1 (dismissed):

r4mmer (dismissed):

r4mmer (approved):

PR #720, opened 2024-07-31
refactor: typing improvements

glevco:
Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/719

### Motivation

This PR makes some simple improvements in typing in multiple files that will aid implementing the `OutputValueType` refactor from `number` to `bigint`.

### Acceptance Criteria

- Add some missing parameter/return types in functions.
- Add some missing properties in test fixtures.
- Rename `api/txApi.js` and `api/wallet.js` files to `.ts`, enabling typing in them.
- Fix typing of some `IHistoryTx` properties.
- No behavior should change.

### Security Checklist

### DIFF
--- a/src/api/axiosInstance.ts
+++ b/src/api/axiosInstance.ts
@@ -21,7 +21,7 @@ import axiosWrapperCreateRequestInstance from './axiosWrapper';
  * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
  * @param {number} timeout Timeout in milliseconds for the request
  */
-export const defaultCreateRequestInstance = (resolve, timeout) => {
+export const defaultCreateRequestInstance = (resolve?: null, timeout?: number) => {
   return axiosWrapperCreateRequestInstance(config.getServerUrl(), resolve, timeout);
 };
 
@@ -31,6 +31,6 @@ export const registerNewCreateRequestInstance = fn => {
   _createRequestInstance = fn;
 };
 
-export const createRequestInstance = (resolve, timeout) => {
+export const createRequestInstance = (resolve?: null, timeout?: number) => {
   return _createRequestInstance(resolve, timeout);
 };
--- a/src/api/txApi.ts
+++ b/src/api/txApi.ts
@@ -52,7 +52,8 @@ const txApi = {
    * @inner
    */
   getTransactions(type, count, timestamp, hash, page, resolve) {
-    const data = { type, count };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data: any = { type, count };
     if (hash) {
       data.hash = hash;
       data.timestamp = timestamp;
--- a/src/api/wallet.ts
+++ b/src/api/wallet.ts
@@ -27,7 +27,8 @@ const walletApi = {
    * @inner
    */
   getAddressHistory(addresses, hash, resolve) {
-    const data = { addresses, paginate: true };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data: any = { addresses, paginate: true };
     if (hash) {
       data.hash = hash;
     }
@@ -65,7 +66,8 @@ const walletApi = {
    * @inner
    */
   getAddressHistoryForAwait(addresses, hash) {
-    const data = { addresses, paginate: true };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data: any = { addresses, paginate: true };
     if (hash) {
       data.hash = hash;
     }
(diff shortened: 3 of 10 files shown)


.github/workflows/integration-test.yml
```diff
@@ -5,6 +5,7 @@ on:
       - release
       - release-candidate
       - master
+      - refactor/output-value-type # TODO: REMOVE THIS
```

.github/workflows/main.yml
```diff
@@ -12,6 +12,7 @@ on:
     - release
     - release-candidate
     - master
+    - refactor/output-value-type # TODO: REMOVE THIS
```

.github/workflows/integration-test.yml
glevco: Done

.github/workflows/main.yml
glevco: Done

msbrogli (dismissed):

src/api/axiosInstance.ts
```diff
@@ -21,7 +21,7 @@ import axiosWrapperCreateRequestInstance from './axiosWrapper';
  * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
  * @param {number} timeout Timeout in milliseconds for the request
  */
-export const defaultCreateRequestInstance = (resolve, timeout) => {
+export const defaultCreateRequestInstance = (resolve?: null, timeout?: number) => {
```
r4mmer, line 24: `resolve` is not `null`.

r4mmer (changes requested):

src/types.ts
```diff
@@ -93,18 +93,18 @@ export interface IAuthoritiesBalance {
 
 export interface IHistoryTx {
   tx_id: string;
-  signalBits: number;
+  signalBits?: number;
```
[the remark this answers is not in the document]
glevco: Indeed `signal_bits` should always be present, however the `/thin_wallet/address_history` API does not include it. In the full node, this API [uses the `to_json_extended()`](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/wallet/resources/thin_wallet/address_history.py#L223) method that doesn't send `signal_bits`. You can verify this in this [example API call](https://node1.mainnet.hathor.network/v1a/thin_wallet/address_history?addresses[]=HJB2yxxsHtudGGy3jmVeadwMfRi2zNCKKD).

The full node API should probably be updated to include this, but meanwhile this is the correct type that represents the return value.

src/types.ts
```diff
@@ -93,18 +93,18 @@ export interface IAuthoritiesBalance {
...
   version: number;
   weight: number;
   timestamp: number;
   is_voided: boolean;
-  nonce: number;
+  nonce?: number;
```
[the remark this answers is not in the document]
glevco: Same as commented in https://github.com/HathorNetwork/hathor-wallet-lib/pull/720#discussion_r1707960326, the `/thin_wallet/address_history`  full node API does not currently return those values: `nonce` is never set, and `tokens` is only set for token creation transactions, and therefore optional.

src/api/axiosInstance.ts
glevco: `defaultCreateRequestInstance` only uses it as an argument to `axiosWrapperCreateRequestInstance`, defined [here](https://github.com/HathorNetwork/hathor-wallet-lib/blob/09a75f7f5700ae50a7ed8b0a03162a8c185b1d80/src/api/axiosWrapper.ts#L28) as `null`.

src/api/txApi.ts
```diff
@@ -52,7 +52,8 @@ const txApi = {
    * @inner
    */
   getTransactions(type, count, timestamp, hash, page, resolve) {
-    const data = { type, count };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data: any = { type, count };
```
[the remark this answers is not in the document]
glevco: I agree, however I did not want to do this unrelated change in this PR. Those files where `.js` before, and therefore we didn't have any type checking for them. I changed them to `.ts` to leverage only the return type of those functions (which will be set in the next PR). However, after changing it to TS, the compiler complained about those `const data` and I set them as `any` in this first step.

Later, I think this whole file should be correctly typed, beginning by setting the argument types in functions. Then, correctly typing `data` will be easier.

src/api/wallet.ts
```diff
@@ -27,7 +27,8 @@ const walletApi = {
    * @inner
    */
   getAddressHistory(addresses, hash, resolve) {
-    const data = { addresses, paginate: true };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
glevco: Same as commented in https://github.com/HathorNetwork/hathor-wallet-lib/pull/720#discussion_r1707976388

r4mmer (dismissed):

msbrogli (approved):

r4mmer (approved):

PR #723, opened 2024-07-31
style: `jest/expect-expect` rule

tuliomir:
### Acceptance Criteria
- Fixes the [`expect-expect` rule](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/expect-expect.md)

### Future possibilities
We could refactor all the assertion helper functions to have the same prefix or suffix, so that we could replace the many exact names in `eslintrc.js` with a single entry that would cover all of them.

It would also be more straightforward to add new helper functions using the same pattern instead of adding them to the `eslintrc.js`.

### Security Checklist

### DIFF
--- a/__tests__/integration/nanocontracts/bet.test.ts
+++ b/__tests__/integration/nanocontracts/bet.test.ts
@@ -309,6 +309,7 @@ describe('full cycle of bet nano contract', () => {
     expect(txHistory).toHaveLength(4);
   };
 
+  /* eslint-disable jest/expect-expect -- All assertions are on the helper function */
   it('bet deposit', async () => {
     await executeTests(hWallet);
   });
@@ -318,6 +319,7 @@ describe('full cycle of bet nano contract', () => {
   it.skip('bet deposit with multisig wallet', async () => {
     await executeTests(mhWallet);
   });
+  /* eslint-enable jest/expect-expect */
 
   it('handle errors', async () => {
     const address0 = await hWallet.getAddressAtIndex(0);
--- a/__tests__/integration/nanocontracts/bet.test.ts
+++ b/__tests__/integration/nanocontracts/bet.test.ts
@@ -309,7 +309,6 @@ describe('full cycle of bet nano contract', () => {
     expect(txHistory).toHaveLength(4);
   };
 
-  /* eslint-disable jest/expect-expect -- All assertions are on the helper function */
   it('bet deposit', async () => {
     await executeTests(hWallet);
   });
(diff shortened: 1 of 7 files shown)


alexruzenhack (approved):

__tests__/websocket/websocket.test.ts
```diff
@@ -46,9 +46,11 @@ describe('onMessage', () => {
 
   it('should receive messages after initializing the timeoutTimer', () => {
     const wsInstance = new GenericWebSocket(baseWsOptions);
+    const msgSpy = jest.spyOn(wsInstance, 'onMessage').mockImplementation(jest.fn());
 
     wsInstance.sendPing();
     wsInstance.onMessage({ data: JSON.stringify({ type: 'other' }) });
+    expect(msgSpy).toHaveBeenCalled();
```
andreabadesso, line 53: So this test wasn't doing anything?

andreabadesso (approved):

__tests__/integration/nanocontracts/bet.test.ts
```diff
@@ -312,6 +312,7 @@ describe('full cycle of bet nano contract', () => {
     expect(txHistory).toHaveLength(4);
   };
 
+  /* eslint-disable jest/expect-expect -- All assertions are on the helper function */
```
andreabadesso, line 315: _Comment unrelated to line number_

You can accomplish the same thing, without disabling the rule by adding `expect.hasAssertions` in `beforeEach` of a shared jest file and adding `additionalTestBlockFunctions` as a parameter to the `jest/expect-expect` rule in the config, e.g.:

```json
{
  "rules": {
    "jest/expect-expect": [
      "error",
      { "additionalTestBlockFunctions": ["executeTests", "testLockedUtxoMethods"] }
    ]
  }
}
```

andreabadesso (changes requested):

__tests__/websocket/websocket.test.ts
tuliomir: That's correct. It would throw if the code failed, so it increased our coverage, but had no assertions.

__tests__/integration/nanocontracts/bet.test.ts
tuliomir: The `additionalTestBlockFunctions` didn't work, but I managed to make the same improvement using `assertFunctionNames`.

Do you agree with this approach?

andreabadesso (approved):

alexruzenhack (approved):

PR #725, opened 2024-07-31
style: `@typescript-eslint/ban-ts-comment` rule

tuliomir:
### Acceptance Criteria
- Fixes the [`ban-ts-comment` rule](https://typescript-eslint.io/rules/ban-ts-comment/)

### Security Checklist

### DIFF
--- a/__tests__/wallet/api/swapService.test.ts
+++ b/__tests__/wallet/api/swapService.test.ts
@@ -24,14 +24,14 @@ describe('hashing and encrypting', () => {
   it('should reject calls with missing parameters', () => {
     const originalString = 'PartialTx|123123||';
 
-    // @ts-ignore
+    // @ts-expect-error -- Testing invalid inputs
     expect(() => encryptString()).toThrow('Missing encrypted string');
-    // @ts-ignore
+    // @ts-expect-error -- Testing invalid inputs
     expect(() => encryptString(originalString)).toThrow('Missing password');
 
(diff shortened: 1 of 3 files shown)


__tests__/wallet/api/swapService.test.ts
```diff
@@ -24,14 +24,14 @@ describe('hashing and encrypting', () => {
   it('should reject calls with missing parameters', () => {
     const originalString = 'PartialTx|123123||';
 
-    // @ts-ignore
+    // @ts-expect-error -- Testing invalid inputs
```
alexruzenhack, line 27: Why not set this directive rule for the file?

alexruzenhack (approved):

__tests__/wallet/api/swapService.test.ts
tuliomir: This would be a valid use of `ts-ignore`, as its scope is more ample. However, `ts-expect-error` is a directive valid for a single line and is aimed precisely at unit tests with invalid parameters.

You can learn more about it on [these release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) for TypeScript 3.9.

In short, `ts-expect-error` acts as an additional `expect()` clause, but instead of testing the runtime it asserts that the next line would throw an error through a TypeScript compiler.

andreabadesso (approved):

PR #726, opened 2024-07-31
style: `jest/no-mocks-import` rule

tuliomir:
### Acceptance Criteria
- Fixes the [`jest/no-mock-import` rule](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-mocks-import.md)

### Security Checklist


alexruzenhack (approved):

__tests__/pushNotification.test.ts
```diff
@@ -1,4 +1,4 @@
-import { mockAxiosAdapter } from './__mocks__/wallet.mock';
+import { mockAxiosAdapter } from './__mock_helpers/axios-adapter.mock';
```
andreabadesso, line 1: Why not

```suggestion
import { mockAxiosAdapter } from './__mock_helpers__/axios-adapter.mock';
```

?

andreabadesso (changes requested):

__tests__/pushNotification.test.ts
tuliomir: That's a folder that already existed, and I had only moved the helper to it. By changing its name I had to increase the scope of the PR a bit, but I think this is a good suggestion.
Implemented the rename on https://github.com/HathorNetwork/hathor-wallet-lib/pull/726/commits/46f0410662c258ce080fbe26edce37bf8524f374

andreabadesso (approved):

alexruzenhack (approved):

PR #728, opened 2024-08-01
feat: implement bigint support

glevco:
Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/797

### Motivation

This PR adds the necessary structures to support bigint fields, however it does _not_ yet change the `OutputValueType` to `bigint`.

### Acceptance Criteria

- Implement `bigIntToBytes()` and remove `signedIntToBytes()` support for 8 bytes.
- Implement `unpackToBigInt()` and remove `unpackToInt()` support for 8 bytes.
- Update `bytesToOutputValue()` to use `unpackToBigInt()`.
- Dependencies:
  - Added `zod`.
  - `level-transcoder` was added explicitly but was already an indirect dependency.
  - Removed `long`.
- Create `api/txApi` and `api/wallet` Zod schemas, updating requests to use `transformJsonBigIntResponse()`.
- Update `Output.valueToBytes()` to use `bigIntToBytes()` instead of `signedIntToBytes()`.
- Create Zod schemas for some necessary interfaces.
- Use Zod parsing in WS history txs.
- Update LevelDB sublevels to use `jsonBigIntEncoding()` where necessary.
- Implement `utils/bigint.ts`.
- Change WS handler to use `JSONBigInt.parse()` instead of `JSON.parse()`.
- Change TS target to `es2020` instead of `es2018`, adding support for `bigint`.

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,10 +22,11 @@
     "crypto-js": "4.2.0",
     "isomorphic-ws": "5.0.0",
     "level": "8.0.1",
+    "level-transcoder": "1.0.1",
     "lodash": "4.17.21",
-    "long": "5.2.3",
     "queue-microtask": "1.2.3",
-    "ws": "8.17.1"
+    "ws": "8.17.1",
+    "zod": "3.23.8"
   },
   "scripts": {
     "test": "jest --env=node --forceExit",
--- a/src/api/schemas/txApi.ts
+++ b/src/api/schemas/txApi.ts
@@ -1,0 +1,106 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { z } from 'zod';
+
+export const transactionSchema = z.discriminatedUnion('success', [
+  z
+    .object({
+      success: z.literal(true),
+      tx: z
+        .object({
+          hash: z.string(),
+          nonce: z.string(),
+          timestamp: z.number(),
+          version: z.number(),
+          weight: z.number(),
+          signal_bits: z.number(),
+          parents: z.string().array(),
+          nc_id: z.string().nullish(),
+          nc_method: z.string().nullish(),
+          nc_pubkey: z.string().nullish(),
+          nc_args: z.string().nullish(),
+          nc_blueprint_id: z.string().nullish(),
+          inputs: z
+            .object({
+              value: z.number(),
+              token_data: z.number(),
+              script: z.string(),
+              decoded: z
+                .object({
+                  type: z.string(),
+                  address: z.string(),
+                  timelock: z.number().nullish(),
+                  value: z.number(),
+                  token_data: z.number(),
+                })
+                .passthrough(),
+              tx_id: z.string(),
+              index: z.number(),
+              token: z.string().nullish(),
+              spent_by: z.string().nullish(),
+            })
+            .passthrough()
+            .array(),
+          outputs: z
+            .object({
+              value: z.number(),
+              token_data: z.number(),
+              script: z.string(),
+              decoded: z
+                .object({
+                  type: z.string(),
+                  address: z.string().optional(),
+                  timelock: z.number().nullish(),
+                  value: z.number(),
+                  token_data: z.number().optional(),
+                })
+                .passthrough(),
+              token: z.string().nullish(),
+              spent_by: z.string().nullish(),
+            })
+            .passthrough()
+            .array(),
+          tokens: z
+            .object({
+              uid: z.string(),
+              name: z.string().nullable(),
+              symbol: z.string().nullable(),
+            })
+            .passthrough()
+            .array(),
+          token_name: z.string().nullish(),
+          token_symbol: z.string().nullish(),
+          raw: z.string(),
+        })
+        .passthrough(),
+      meta: z
+        .object({
+          hash: z.string(),
+          spent_outputs: z.tuple([z.number(), z.string().array()]).array(),
+          received_by: z.string().array(),
+          children: z.string().array(),
+          conflict_with: z.string().array(),
+          voided_by: z.string().array(),
+          twins: z.string().array(),
+          accumulated_weight: z.number(),
+          score: z.number(),
+          height: z.number(),
+          min_height: z.number(),
+          feature_activation_bit_counts: z.number().array().nullable(),
+          first_block: z.string().nullish(),
+          validation: z.string().nullish(),
+          first_block_height: z.number().nullish(),
+        })
+        .passthrough(),
+      spent_outputs: z.record(z.coerce.number(), z.string()),
+    })
+    .passthrough(),
+  z.object({ success: z.literal(false), message: z.string().nullish() }).passthrough(),
+]);
+
+export type TransactionSchema = z.infer<typeof transactionSchema>;
--- a/src/api/schemas/wallet.ts
+++ b/src/api/schemas/wallet.ts
@@ -1,0 +1,48 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { z } from 'zod';
+import { IHistoryTxSchema } from '../../schemas';
+
+export const addressHistorySchema = z.discriminatedUnion('success', [
+  z
+    .object({
+      success: z.literal(true),
+      history: IHistoryTxSchema.array(),
+      has_more: z.boolean(),
+      first_hash: z.string().nullish(),
+      first_address: z.string().nullish(),
+    })
+    .passthrough(),
+  z.object({ success: z.literal(false), message: z.string() }).passthrough(),
+]);
+
+export type AddressHistorySchema = z.infer<typeof addressHistorySchema>;
+
+export const mintMeltUtxoSchema = z
+  .object({
+    tx_id: z.string(),
+    index: z.number(),
+  })
+  .passthrough();
+
+export const generalTokenInfoSchema = z.discriminatedUnion('success', [
+  z
+    .object({
+      success: z.literal(true),
+      name: z.string(),
+      symbol: z.string(),
+      mint: mintMeltUtxoSchema.array(),
+      melt: mintMeltUtxoSchema.array(),
+      total: z.number(),
+      transactions_count: z.number(),
+    })
+    .passthrough(),
+  z.object({ success: z.literal(false), message: z.string() }).passthrough(),
+]);
+
+export type GeneralTokenInfoSchema = z.infer<typeof generalTokenInfoSchema>;
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -49,6 +49,7 @@ import { MemoryStore, Storage } from '../storage';
 import { deriveAddressP2PKH, deriveAddressP2SH, getAddressFromPubkey } from '../utils/address';
 import NanoContractTransactionBuilder from '../nano_contracts/builder';
 import { prepareNanoSendTransaction } from '../nano_contracts/utils';
+import { IHistoryTxSchema } from '../schemas';
 import GLL from '../sync/gll';
 
 /**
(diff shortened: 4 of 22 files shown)


.github/workflows/integration-test.yml
```diff
@@ -5,7 +5,7 @@ on:
       - release
       - release-candidate
       - master
-      - refactor/output-value-type # TODO: REMOVE THIS
+      - refactor/typing-improvements # TODO: REMOVE THIS
```

.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ on:
     - release
     - release-candidate
     - master
-    - refactor/output-value-type # TODO: REMOVE THIS
+    - refactor/typing-improvements # TODO: REMOVE THIS
```

.github/workflows/integration-test.yml
glevco: Done

.github/workflows/main.yml
glevco: Done

src/utils/bigint.ts
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * Parse some `unknown` data with a Zod schema. If parsing fails, it logs the error and throws.
+ */
+export function parseSchema<T>(data: unknown, schema: ZodSchema<T>): T {
+  const result = schema.safeParse(data);
+
+  if (!result.success) {
+    // TODO: How to log correctly?
+    // eslint-disable-next-line no-console
+    console.error(`error: ${result.error.message}\ncaused by input: ${data}`);
```
glevco, line 105: What's the correct logger to use here? `getDefaultLogger()`?

.github/workflows/integration-test.yml
```diff
@@ -12,6 +12,7 @@ on:
       - release
       - release-candidate
       - master
+      - ci/node-v22 # TODO: Remove this
```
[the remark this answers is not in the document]
glevco: Done.

.github/workflows/lint.yml
```diff
@@ -12,6 +12,7 @@ on:
     - release
     - release-candidate
     - master
+    - ci/node-v22 # TODO: Remove this
```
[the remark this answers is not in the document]
glevco: Done.

.github/workflows/unit-test.yml
```diff
@@ -12,6 +12,7 @@ on:
     - release
     - release-candidate
     - master
+    - ci/node-v22 # TODO: Remove this
```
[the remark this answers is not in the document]
glevco: Done.

src/api/schemas/txApi.ts
```diff
@@ -0,0 +1,106 @@
...
+          score: z.number(),
+          height: z.number(),
+          min_height: z.number(),
+          feature_activation_bit_counts: z.number().array().nullable(),
+          first_block: z.string().nullish(),
+          validation: z.string().nullish(),
+          first_block_height: z.number().nullish(),
+        })
+        .passthrough(),
+      spent_outputs: z.record(z.coerce.number(), z.string()),
```
pedroferreira1, line 100: Why are you using `z.coerce.number` here?

src/utils/buffer.ts
```diff
@@ -43,7 +42,7 @@ export function intToBytes(value: number, bytes: number): Buffer {
  * @inner
  */
 export function signedIntToBytes(value: number, bytes: number): Buffer {
```
pedroferreira1, line 44: You have to keep this method only because nano contract still uses this to parse int arguments, right?

src/api/schemas/wallet.ts
```diff
@@ -0,0 +1,48 @@
...
+      history: IHistoryTxSchema.array(),
+      has_more: z.boolean(),
+      first_hash: z.string().nullish(),
+      first_address: z.string().nullish(),
+    })
+    .passthrough(),
+  z.object({ success: z.literal(false), message: z.string() }).passthrough(),
+]);
+
+export type AddressHistorySchema = z.infer<typeof addressHistorySchema>;
```
r4mmer, line 24: suggestion(minor): Change naming convention.

`AddressHistorySchema` and `addressHistorySchema` are too similar and may cause confution.
Some options I think can be more indicative:
- `AddressHistoryType` and `AddressHistorySchema`
- `AddressHistory` and `AddressHistorySchema`
- `TAddressHistory` and `AddressHistorySchema`
- `TAddressHistorySchema` and `AddressHistorySchema`

This is also relevant for the other zod schemas, but what do you think?

src/schemas.ts
```diff
@@ -0,0 +1,138 @@
...
+  IHistoryTx,
+  ILockedUtxo,
+  ITokenBalance,
+  ITokenMetadata,
+  IUtxo,
+  TxHistoryProcessingStatus,
+} from './types';
+import { ZodSchema } from './utils/bigint';
+
+export const ITokenBalanceSchema: ZodSchema<ITokenBalance> = z
```
r4mmer, line 25: `ITokenBalanceSchema` starts with `I` because it is an interface (just a naming convention).
Maybe we should change the names since we are using zod schemas for validation/typing

src/schemas.ts
```diff
@@ -0,0 +1,138 @@
...
+  .passthrough();
+
+export const IHistoryOutputSchema: ZodSchema<IHistoryOutput> = z
+  .object({
+    value: z.number(),
+    token_data: z.number(),
+    script: z.string(),
+    decoded: IHistoryOutputDecodedSchema,
+    token: z.string(),
+    spent_by: z.string().nullable(),
```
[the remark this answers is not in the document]
r4mmer: I checked this, the best explanation I found is:
```js
z.object({
    optional: z.string().optional(), // field not provided, or explicitly `undefined`
    nullable: z.string().nullable(), // field explicitly `null`
    nullish: z.string().nullish(), // field not provided, explicitly `null`, or explicitly `undefined`
});
```

Which makes `nullish().optional()` redundant, should we check this and refactor?

src/utils/storage.ts
```diff
@@ -410,14 +402,9 @@ export async function processHistory(
...
-    | {
-        success: true;
-        name: string;
-        symbol: string;
-      }
-    | { success: false; message: string }
-  > {
+  async function fetchTokenData(
+    uid: string
+  ): Promise<GeneralTokenInfoSchema | { success: true; name: string; symbol: string }> {
```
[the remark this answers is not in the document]
glevco: Yes, because below there's a return with this explicit type, which does not conform to `GeneralTokenInfoSchema`:

https://github.com/HathorNetwork/hathor-wallet-lib/blob/ed121c6a6a0a8f0db7abda2626b640a60a8f58df/src/utils/storage.ts#L412-L416

This function could definitely be improved, but I think this is a more general typing issue that should be addressed in separate PRs.

src/utils/bigint.ts
```diff
@@ -0,0 +1,169 @@
...
+          // '12345678901234567890.1', JS will parse it as an integer with loss of precision, `12345678901234567000`.
+          throw Error(`large float will lose precision! in "${text}"`);
+        }
+      }
+
+      const bigIntValue = BigInt(source);
+      if (bigIntValue !== BigInt(value)) {
+        // If the parsed value is an integer and its BigInt representation is a different value,
+        // it means we lost precision, so we return the BigInt.
+        return bigIntValue;
```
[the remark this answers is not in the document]
glevco: > Why don't we always return a `BigInt`?

Because this will be called for schemas/functions that expect `Number` too, not just `BigInt`.

> How are we going to handle operations if this is sometimes int and sometimes BigInt?

The way we differentiate between them is through Zod, which is exactly the reason why I had to use it. In other words, the custom JSON parser will only return a `BigInt` if the value is large enough, and Zod will take care of validating the correct types — for example, converting a small `Number` such as `123` to a `123n` when necessary.

src/api/schemas/txApi.ts
glevco: Because the API returns it as a number in a string, such as `"0"`, so I use Zod to convert it to a `Number`. Check this [API call example](https://node1.mainnet.hathor.network/v1a/transaction?id=0000000008bfd52ed4756fc2204d89f3dd3499b4df62113f6adc4949a9ba714c), where `"spent_outputs":{"0":"00000000066a2a62b3a8148e0b1afcf36c6c943a0ae00b6531f2c2c38d0fb509"}`.

In fact, this is part of the JSON specification, which only allows key names to be strings.

src/utils/buffer.ts
glevco: Yes!

src/api/schemas/wallet.ts
glevco: Just to clarify, the schema itself (`addressHistorySchema` in this example) is an object, not a type. So it shouldn't use PascalCase, right? I assumed the casing would be enough to avoid confusion: `addressHistorySchema` is the schema object and `AddressHistorySchema` is the schema type. And TypeScript will make sure they can't be used interchangeably.

But I would be fine with changing the name of the types to include `Type`, for example.

src/schemas.ts
glevco: I kept it purposefully to indicate that this is schema is tied to an existing interface, in this case, `ITokenBalance`. So this is `ITokenBalance`'s schema, or, `ITokenBalanceSchema`.

I do agree this could be improved and the `I` becomes unnecessary when we remove the interfaces and use `Zod.infer` instead. This is set to be changed in this issue: https://github.com/HathorNetwork/hathor-wallet-lib/issues/800, so I would rather do it as part of another PR. What do you think?

src/utils/bigint.ts
glevco: > Is this an unexpected error or it can happen sometimes?

It can happen sometimes, this function receives an untyped object and tries to parse it using the provided schema. If the object does not conform, the error will be thrown. This can happen, for example, if the full node returns an incorrect response schema in the API.

> And yes, I think the `getDefaultLogger()` is the best one to use

Changed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/728/commits/9ef97a7156a03be8b6da0a68f85009d6ec615d11

package.json
```diff
@@ -22,10 +22,11 @@
     "crypto-js": "4.2.0",
     "isomorphic-ws": "5.0.0",
     "level": "8.0.1",
+    "level-transcoder": "^1.0.1",
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-wallet-lib/pull/728/commits/9ef97a7156a03be8b6da0a68f85009d6ec615d11

package.json
```diff
@@ -22,10 +22,11 @@
...
     "lodash": "4.17.21",
-    "long": "5.2.3",
     "queue-microtask": "1.2.3",
-    "ws": "8.17.1"
+    "ws": "8.17.1",
+    "zod": "^3.23.8"
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-wallet-lib/pull/728/commits/9ef97a7156a03be8b6da0a68f85009d6ec615d11

src/schemas.ts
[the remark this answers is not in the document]
glevco: Carneiro's comment provides a correct summary. Indeed the `nullish().optional()` was redundant in `timelock`, so I fixed it in https://github.com/HathorNetwork/hathor-wallet-lib/pull/728/commits/9ef97a7156a03be8b6da0a68f85009d6ec615d11

src/new/wallet.js
```diff
@@ -1322,7 +1323,7 @@ class HathorWallet extends EventEmitter {
   }
 
   async onNewTx(wsData) {
-    const newTx = wsData.history;
+    const newTx = IHistoryTxSchema.parse(wsData.history);
```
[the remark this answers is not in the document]
glevco: Changed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/728/commits/9ef97a7156a03be8b6da0a68f85009d6ec615d11

src/schemas.ts
r4mmer: Great!
Just wanted to keep consistency to avoid any confusion later

src/api/schemas/wallet.ts
r4mmer: While I don't think we should use it here, we usually name a constant for later use with all caps snake case

I do think we should have a consistent naming convention since this validation schema will be used throughout the lib and wallets, so having inconsistent or non-descriptive names may increase overall confusion on the code.
But we can discuss this to refactor on another PR, what do you think?

r4mmer (approved):

PR #729, opened 2024-08-02
feat: handle nano custom types in the parser and builder classes

pedroferreira1:
### Context

There are nano contract method arguments and attributes that have specific types and we can improve user usability of the clients when considering them. For example, an address is decoded in bytes but if we know that the argument is an address we can validate it better and expect in the clients a base58 string instead of a hexadecimal of the decoded address.

In this PR we will consider the new types as following:

bytes: `TxOutputScript`, `TokenUid`, `VertexId`, and `Address`
int: `Timestamp` and `Amount`

Even though there are no differences when serializing/deserializing a Timestamp/Amount from a normal integer, we can use this information to create the UI in a way that is more user friendly.

### Acceptance Criteria
- Consider new custom types when serializing/deserializing data for nano contract.
- Accept address as a base58 string in the nano contract builder.
- Use new nano hathor core experimental image to validate the integration tests

### Security Checklist

### DIFF
--- a/src/nano_contracts/deserializer.ts
+++ b/src/nano_contracts/deserializer.ts
@@ -9,6 +9,7 @@ import { bufferToHex, unpackToFloat, unpackToInt } from '../utils/buffer';
 import helpersUtils from '../utils/helpers';
 import Network from '../models/network';
 import { NanoContractArgumentType } from './types';
+import { OutputValueType } from '../types';
 
 class Deserializer {
   network: Network;
@@ -51,8 +52,9 @@ class Deserializer {
         return this.toAddress(value);
       case 'int':
       case 'Timestamp':
-      case 'Amount':
         return this.toInt(value);
+      case 'Amount':
+        return this.toAmount(value);
       case 'float':
         return this.toFloat(value);
       case 'bool':
@@ -98,6 +100,18 @@ class Deserializer {
    */
   toInt(value: Buffer): number {
     return unpackToInt(4, true, value)[0];
+  }
+
+  /**
+   * Deserialize amount value
+   *
+   * @param {value} Value to deserialize
+   *
+   * @memberof Deserializer
+   * @inner
+   */
+  toAmount(value: Buffer): OutputValueType {
+    return this.toInt(value);
   }
 
   /**
--- a/src/nano_contracts/deserializer.ts
+++ b/src/nano_contracts/deserializer.ts
@@ -214,7 +214,9 @@ class Deserializer {
     const address = helpersUtils.encodeAddress(addressBytes, this.network);
     const decoded = address.decode();
     if (decoded[0] !== value[0]) {
-      throw new Error(`Asked to deserialize an address with version byte ${value[0]} but the network from the deserializer object has version byte ${decoded[0]}.`);
+      throw new Error(
+        `Asked to deserialize an address with version byte ${value[0]} but the network from the deserializer object has version byte ${decoded[0]}.`
+      );
     }
     return address.base58;
   }
--- a/src/nano_contracts/deserializer.ts
+++ b/src/nano_contracts/deserializer.ts
@@ -211,7 +211,12 @@ class Deserializer {
   toAddress(value: Buffer): string {
     // First we get the 20 bytes of the address without the version byte and checksum
     const addressBytes = value.slice(1, 21);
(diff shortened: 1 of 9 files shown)


src/nano_contracts/deserializer.ts
```diff
@@ -170,6 +185,20 @@ class Deserializer {
...
+   *
+   * @param {value} Value to deserialize
+   *
+   * @memberof Deserializer
+   * @inner
+   */
+  toAddress(value: Buffer): string {
+    // First we get the 20 bytes of the address without the version byte and checksum
+    const addressBytes = value.slice(1, 21);
+    return helpersUtils.encodeAddress(addressBytes, this.network).base58;
```
tuliomir, line 200: question: Since we're discarding the first byte of the input here, are we at risk of accepting an address from another network as valid?

Would it be worth it to add a test validating the behavior of having an address from another network being decoded?

src/nano_contracts/utils.ts
```diff
@@ -144,8 +151,18 @@ export const validateBlueprintMethodArgs = async (blueprintId, method, args): Pr
     throw new NanoContractTransactionError(`Blueprint does not have method ${method}.`);
   }
 
-  // Args may come as undefined
-  const argsLen = args ? args.length : 0;
+  // Args may come as undefined or null
+  if (args == null) {
+    if (methodArgs.length !== 0) {
+      throw new NanoContractTransactionError(
```
tuliomir, line 157: chore: We should add a test to validate this protection

src/nano_contracts/utils.ts
```diff
@@ -189,6 +215,29 @@ export const validateBlueprintMethodArgs = async (blueprintId, method, args): Pr
           );
         }
         break;
+      // Creating a block {} in the case below
+      // because we can't create a variable without it (linter - no-case-declarations)
+      case 'Address': {
+        const argValue = args[index];
+        if (typeof argValue !== 'string') {
+          throw new NanoContractTransactionError(
```
tuliomir, line 223: chore: We should add a test to validate those protections from invalid address inputs

src/nano_contracts/deserializer.ts
pedroferreira1: Makes sense! 

Done
https://github.com/HathorNetwork/hathor-wallet-lib/pull/729/commits/12200527419855290ae7ede7ebd0e35555c41039
https://github.com/HathorNetwork/hathor-wallet-lib/pull/729/commits/cd74c357d0f39b0c153651cc8394a27f7661c888

src/nano_contracts/utils.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/729/commits/3009dce76ecd371c1e3d1630a58e856b164afb80

src/nano_contracts/utils.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/729/commits/3009dce76ecd371c1e3d1630a58e856b164afb80

r4mmer (approved):

src/nano_contracts/deserializer.ts
```diff
@@ -85,6 +102,18 @@ class Deserializer {
...
+  /**
+   * Deserialize amount value
+   *
+   * @param {value} Value to deserialize
+   *
+   * @memberof Deserializer
+   * @inner
+   */
+  toAmount(value: Buffer): OutputValueType {
+    return this.toInt(value);
```
tuliomir, line 114: suggestion(non-blocking): It would be nice to have a simple test here, just to complete the coverage, although we know this will soon be made more complex by the increased limit of accepted values.

tuliomir (dismissed):

src/nano_contracts/deserializer.ts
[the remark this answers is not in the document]
pedroferreira1: No complains about it, the `npm run build` command runs without errors/warnings. Probably because the `OutputValueType` is exactly a `number`

pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/729/commits/02b94f0553285cef8c2ad3e222da8fb8d065ea9d

tuliomir (approved):

PR #731, opened 2024-08-08
chore: Tests linting on ci

tuliomir:
### Acceptance Criteria
- Adds the linting and formatting of test files to the CI
- Fixes `no-cond-assign` oversight from #721 
- Fixes all linting and formatting from the recent `stream` test file

### Security Checklist


package.json
```diff
@@ -36,8 +36,8 @@
     "test_network_integration": "jest --config jest-integration.config.js --runInBand --forceExit",
     "test_network_partial_down": "docker compose -f ./__tests__/integration/configuration/docker-compose.yml -p configuration stop cpuminer",
     "test_network_down": "docker compose -f ./__tests__/integration/configuration/docker-compose.yml down && rm ./tmp/wallets.json",
-    "lint": "eslint 'src/**/*.{js,ts}'",
-    "lint:fix": "eslint 'src/**/*.{js,ts}' --fix",
+    "lint": "eslint 'src/**/*.{js,ts}' '__tests__/**/*.{js,ts}'",
+    "lint:fix": "eslint 'src/**/*.{js,ts}' '__tests__/**/*.{js,ts}' --fix",
```
r4mmer, line 40: question(if-minor): Can the target files be configured on eslintrc.js?

If we move the configuration there the call will be simpler `eslint .` and  `eslint . --fix`

r4mmer (approved):

alexruzenhack (approved):

package.json
tuliomir: I tried finding some configuration for this [on the official docs](https://eslint.org/docs/latest/use/configure/) and on alternative guides, but it doesn't seem to be possible.

The scope of the eslint run is really meant to be defined in the command line at runtime, not in the configurations.

The parts of the config that deal with specific files are regarding to rule overriding. For example, on line `90` of the `eslintrc.js` file we specify that files within the `__tests__` folder should have an additional set of rules that regard Jest.

PR #745, opened 2024-08-28
chore: treat begin stream event

r4mmer:
### Acceptance Criteria

- Treat begin stream event with no action.

### Security Checklist

### DIFF
--- a/__tests__/stream.test.ts
+++ b/__tests__/stream.test.ts
@@ -50,10 +50,16 @@ const SERVER_MOCK_TYPE = {
    * This is meant to give time for the client to abort.
    */
   abort: 'abort',
+  /**
+   * Sends an event from the wrong stream to test how
+   * the client reacts to this error.
+   */
+  unknownId: 'unknown_id',
 };
 
 function makeServerMock(mockServer, mockType) {
   function streamHistoryForSocket(streamId, socket) {
+    // Begin event marks the start of a stream
     socket.send(
       JSON.stringify({
         id: streamId,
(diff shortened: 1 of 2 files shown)


src/sync/stream.ts
```diff
@@ -402,8 +412,10 @@ function buildListener(manager: StreamManager, resolve: () => void) {
       );
       return;
     }
-    // Vertex is a transaction in the history of the last address received
-    if (isStreamSyncHistoryVertex(wsData)) {
+    if (isStreamSyncHistoryBegin(wsData)) {
+      manager.logger.info('Begin stream event received.');
```
msbrogli, line 416: Does it log the stream id?

__tests__/stream.test.ts
```diff
@@ -226,4 +249,32 @@ describe('Websocket stream history sync', () => {
       mockServer.stop();
     }
   }, 10000);
+
+  it('should ignore unknown stream ids', async () => {
```
tuliomir, line 253: question: I'm under the impression that even if the wallet fails to handle the unknown stream id we would not be able to validate this in this test. Is there a more explicit way to check the results of these unknown stream ids?

src/sync/stream.ts
r4mmer: No, but we can include it if you want.

__tests__/stream.test.ts
tuliomir: As discussed, this test provides enough confidence on the results of the websocket with the balance validation.
Future possibilities can be discussed later.

tuliomir (approved):

PR #749, opened 2024-08-29
feat: move the history sort to the pre-process method

r4mmer:
### Acceptance Criteria
- Move the history sorting to before the processing

### Impact

It was tested loading a wallet with almost 400k transactions, the load time went from 40 minutes using the normal stream to 1 and a half minute.
This is because this change is only really impactful on wallets with over 20k transactions where each sorting becomes slow enough to impact the sync prrocess.

### Other changes

- Fix Websocket logger argument being undefined.
- When logging an unknown stream message, show the contents

### Security Checklist

### DIFF
--- a/src/new/connection.ts
+++ b/src/new/connection.ts
@@ -11,7 +11,7 @@ import helpers from '../utils/helpers';
 import BaseConnection, { ConnectionParams } from '../connection';
 import { ConnectionState } from '../wallet/types';
 import { handleSubscribeAddress, handleWsDashboard } from '../utils/connection';
-import { IStorage, ILogger } from '../types';
+import { IStorage, ILogger, getDefaultLogger } from '../types';
 
 const STREAM_ABORT_TIMEOUT = 10000; // 10s
 
@@ -66,7 +66,10 @@ class WalletConnection extends BaseConnection {
       connectionTimeout?: number;
       wsURL: string;
       logger: ILogger;
-    } = { wsURL: helpers.getWSServerURL(this.currentServer), logger: options.logger };
+    } = {
+      wsURL: helpers.getWSServerURL(this.currentServer),
+      logger: options.logger || getDefaultLogger(),
+    };
 
     if (options.connectionTimeout) {
       wsOptions.connectionTimeout = options.connectionTimeout;
--- a/src/storage/memory_store.ts
+++ b/src/storage/memory_store.ts
@@ -188,6 +188,13 @@ export class MemoryStore implements IStore {
     // This is a noop since the memory store always starts clean.
   }
 
+  /**
+   * Prepare the store for history processing.
+   */
+  async preProcess(): Promise<void> {
+    this.historyTs.sort();
+  }
+
   /** ADDRESSES */
 
   /**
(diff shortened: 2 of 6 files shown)


src/storage/storage.ts
```diff
@@ -354,6 +354,7 @@ export class Storage implements IStorage {
    * @returns {Promise<void>}
    */
   async processHistory(): Promise<void> {
+    await this.store.preProcess();
     await processHistory(this, { rewardLock: this.version?.reward_spend_min_blocks });
```
msbrogli, line 358: Is this a recursion? I got confused by the name of the method.

src/storage/memory_store.ts
```diff
@@ -188,6 +188,13 @@ export class MemoryStore implements IStore {
     // This is a noop since the memory store always starts clean.
   }
 
+  /**
+   * Prepare the store for history processing.
+   */
+  async preProcess(): Promise<void> {
```
msbrogli, line 194: Should we rename it to `preProcessHistory()`?

msbrogli (dismissed):

src/storage/storage.ts
r4mmer: The `processHistory` is the name of the method and an util from `src/utils/storage.ts`.
Since the implementation requires other imports having it here creates some circular dependencies.

This also makes it easier to test the history processing.

src/storage/memory_store.ts
r4mmer: Done

src/storage/storage.ts
msbrogli: Got it. I guess this is very error prone. Should we rename one of these? Maybe just rename during the import phase and call it `storageProcessHistory`?

msbrogli (approved):

src/storage/storage.ts
r4mmer: Renamed the util to `processHistoryUtil`

src/new/connection.ts
```diff
@@ -66,7 +66,10 @@ class WalletConnection extends BaseConnection {
       connectionTimeout?: number;
       wsURL: string;
       logger: ILogger;
-    } = { wsURL: helpers.getWSServerURL(this.currentServer), logger: options.logger };
+    } = {
+      wsURL: helpers.getWSServerURL(this.currentServer),
+      logger: options.logger || getDefaultLogger(),
+    };
```
r4mmer, line 72: _Unrelated to PR objective_
This is an error found during testing.
The default logger from the websocket class does not work when passing undefined logger here.
So if there is no logger, we should use the default.

src/sync/stream.ts
```diff
@@ -420,7 +420,7 @@ function buildListener(manager: StreamManager, resolve: () => void) {
       manager.logger.error(`Stream error: ${wsData.errmsg}`);
       manager.abortWithError(wsData.errmsg);
     } else {
-      manager.logger.error(`Unknown event type ${wsData}`);
+      manager.logger.error(`Unknown event type ${JSON.stringify(wsData)}`);
```
r4mmer, line 423: _Uunrelated to PR_
Just to log the content of the message

src/storage/storage.ts
msbrogli: Thanks!

msbrogli (approved):

jansegre (approved):

PR #750, opened 2024-08-30
feat: stream to queue implementation with flow control

r4mmer:
### Acceptance criteria

- Send history stream events to queue for processing.
- Process each item sequentially while allowing aborts.
- Implement the client side of the flow control https://github.com/HathorNetwork/hathor-core/pull/1126

### DIFF
--- a/src/new/connection.ts
+++ b/src/new/connection.ts
@@ -14,7 +14,7 @@ import { handleSubscribeAddress, handleWsDashboard } from '../utils/connection';
 import { IStorage, ILogger, getDefaultLogger } from '../types';
 
 const STREAM_ABORT_TIMEOUT = 10000; // 10s
-const STREAM_WINDOW_SIZE = 1000;
+const STREAM_WINDOW_SIZE: number | string = 'inf';
 
 /**
  * Event names for requesting stream from fullnode
@@ -59,7 +59,7 @@ class WalletConnection extends BaseConnection {
 
   streamController: StreamController | null = null;
 
-  streamWindowSize: number;
+  streamWindowSize: number | string;
 
   constructor(options: ConnectionParams & { streamWindowSize?: number }) {
     super(options);
--- a/src/new/connection.ts
+++ b/src/new/connection.ts
@@ -14,7 +14,6 @@ import { handleSubscribeAddress, handleWsDashboard } from '../utils/connection';
 import { IStorage, ILogger, getDefaultLogger } from '../types';
 
 const STREAM_ABORT_TIMEOUT = 10000; // 10s
-const STREAM_WINDOW_SIZE: number | string = 'inf';
 
 /**
  * Event names for requesting stream from fullnode
@@ -59,7 +58,7 @@ class WalletConnection extends BaseConnection {
 
   streamController: StreamController | null = null;
 
-  streamWindowSize: number | string;
+  streamWindowSize: number | undefined;
 
   constructor(options: ConnectionParams & { streamWindowSize?: number }) {
     super(options);
@@ -80,7 +79,7 @@ class WalletConnection extends BaseConnection {
       wsOptions.connectionTimeout = options.connectionTimeout;
     }
 
-    this.streamWindowSize = options.streamWindowSize || STREAM_WINDOW_SIZE;
+    this.streamWindowSize = options.streamWindowSize;
 
     this.websocket = new GenericWebSocket(wsOptions);
   }
@@ -170,8 +169,10 @@ class WalletConnection extends BaseConnection {
       type: StreamRequestEvent.REQUEST_HISTORY_XPUB,
       'first-index': firstIndex,
       'gap-limit': gapLimit,
-      window_size: this.streamWindowSize,
     });
+    if (this.streamWindowSize) {
+      data['window-size'] = this.streamWindowSize;
+    }
     this.websocket.sendMessage(data);
   }
 
@@ -196,8 +197,10 @@ class WalletConnection extends BaseConnection {
       type: StreamRequestEvent.REQUEST_HISTORY_MANUAL,
       'first-index': firstIndex,
       'gap-limit': gapLimit,
-      window_size: this.streamWindowSize,
     });
+    if (this.streamWindowSize) {
+      data['window-size'] = this.streamWindowSize;
+    }
     this.websocket.sendMessage(data);
   }
 
--- a/src/new/connection.ts
+++ b/src/new/connection.ts
@@ -170,7 +170,7 @@ class WalletConnection extends BaseConnection {
       type: StreamRequestEvent.REQUEST_HISTORY_XPUB,
       'first-index': firstIndex,
       'gap-limit': gapLimit,
-      'window_size': this.streamWindowSize,
+      window_size: this.streamWindowSize,
     });
     this.websocket.sendMessage(data);
   }
@@ -196,7 +196,7 @@ class WalletConnection extends BaseConnection {
       type: StreamRequestEvent.REQUEST_HISTORY_MANUAL,
       'first-index': firstIndex,
       'gap-limit': gapLimit,
-      'window_size': this.streamWindowSize,
+      window_size: this.streamWindowSize,
     });
     this.websocket.sendMessage(data);
   }
@@ -205,10 +205,7 @@ class WalletConnection extends BaseConnection {
    * Send an ACK message to the fullnode to confirm we received all events up to
    * the event of sequence number `ack`.
    */
-  sendStreamHistoryAck(
-    id: string,
-    ack: number,
-  ) {
+  sendStreamHistoryAck(id: string, ack: number) {
     if (this.streamController?.streamId !== id) {
       throw new Error('There is an on-going stream, cannot start a second one');
     }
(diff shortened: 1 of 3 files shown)


src/sync/stream.ts
```diff
@@ -51,6 +65,114 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+    this.ackCounter += 1;
+    this.logger.debug(
+      `[*] ACKed ${seq} with queue at ${this.q.size()}. Sent ACK ${this.ackCounter} times`
+    );
+  }
+
+  recv() {
+    if (!this.inTimer) {
+      this.inTimer = setTimeout(() => {
+        this.logger.debug(`[+] => in_rate: ${(1000 * this.recvCounter) / this.dt} items/s`);
```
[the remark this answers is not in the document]
r4mmer: `dt` is 2000ms, we convert to seconds by dividing by 1000

src/sync/stream.ts
```diff
@@ -297,27 +442,81 @@ export class StreamManager extends AbortController {
...
+      this.lastProcSeq = item.seq;
+      if (isStreamItemAddress(item)) {
+        const addr = item.address;
+        const alreadyExists = await this.storage.isAddressMine(addr.base58);
+        if (!alreadyExists) {
+          await this.storage.saveAddress(addr);
+        }
+      } else if (isStreamItemVertex(item)) {
+        await this.storage.addTx(item.vertex);
+      }
```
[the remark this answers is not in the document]
r4mmer: We enqueue items so having an unknown type would mean we put something wrong there.
The place we check fullnode messages is the listener (under `buildListener`) where we do log unknown events so we are only processing or acking addresses and vertices.

There is another issue of weather the vertex sent is actually a vertex but that would be a fullnode error since the "api contract" was broken.

src/sync/stream.ts
```diff
@@ -51,6 +65,114 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+  ack(seq: number) {
+    this.ackCounter += 1;
+    this.logger.debug(
+      `[*] ACKed ${seq} with queue at ${this.q.size()}. Sent ACK ${this.ackCounter} times`
+    );
+  }
+
+  recv() {
+    if (!this.inTimer) {
+      this.inTimer = setTimeout(() => {
```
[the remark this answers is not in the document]
r4mmer: For now this functionality only gives us some logs for statistics purpuses, testing logs seems inneficient.
In the future this will be using this for window maintenance and the tests will be quite relevant, but not now.

src/sync/stream.ts
```diff
@@ -1,19 +1,33 @@
...
+  IStorage,
+  IHistoryTx,
+  HistorySyncMode,
+  isGapLimitScanPolicy,
+  IAddressInfo,
+  ILogger,
+} from '../types';
 import Network from '../models/network';
+import Queue from '../models/queue';
+/* eslint max-classes-per-file: ["error", 2] */
```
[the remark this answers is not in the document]
r4mmer: Thanks!

src/sync/stream.ts
```diff
@@ -51,6 +65,114 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+
+  outTimer?: ReturnType<typeof setTimeout>;
+
+  qTimer?: ReturnType<typeof setTimeout>;
+
+  logger: ILogger;
+
+  q: Queue;
+
+  dt: number;
```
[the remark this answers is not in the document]
r4mmer: Done

tuliomir (approved):

src/new/connection.ts
```diff
@@ -56,7 +58,9 @@ class WalletConnection extends BaseConnection {
 
   streamController: StreamController | null = null;
 
-  constructor(options: ConnectionParams) {
+  streamWindowSize: number | undefined;
```
msbrogli, line 61: Should it be in the `streamController` object?

src/sync/stream.ts
```diff
@@ -307,27 +485,98 @@ export class StreamManager extends AbortController {
...
+      !this.signal.aborted &&
+      this.lastAcked <= this.lastProcSeq &&
+      this.itemQueue.size() <= MIN_QUEUE_SIZE_FOR_ACK
+    );
+  }
+
+  /**
+   * Ack the stream messages.
+   */
+  ack() {
```
msbrogli, line 504: I'd rather have two method `ackIfNeeded()` and `ack()`. The second one would simply send the ack unconditionally.

msbrogli (changes requested):

src/new/connection.ts
r4mmer: We could move this to the stream manager, but I think that this is a Connection level configuration.
What do you think?

src/sync/stream.ts
r4mmer: But then we would never call the `ack`, only the `ackIfNeeded` since we need to check if we are ready to send a message to the server with the ACK.

In this separation we have the `shouldAck` a method that controls the ACK strategy and the `ack` which is called every time we process an event.
The `ack` just checks with the strategy if we should send the message.
What do you think?

tuliomir (dismissed):

src/sync/stream.ts
```diff
@@ -61,6 +74,147 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+  /**
+   * Mark that an item has been received by the queue.
+   * This also manages the inTimer, which calculates the rate of received items.
+   */
+  recv() {
+    if (!this.inTimer) {
+      this.inTimer = setTimeout(() => {
+        this.logger.debug(`[+] => in_rate: ${(1000 * this.recvCounter) / this.dt} items/s`);
+        this.inTimer = undefined;
+        this.recvCounter = -1;
```
msbrogli, line 183: Why do you need to decrease the counter here?

src/sync/stream.ts
```diff
@@ -61,6 +74,147 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+  }
+
+  /**
+   * Mark that an item has been received by the queue.
+   * This also manages the inTimer, which calculates the rate of received items.
+   */
+  recv() {
+    if (!this.inTimer) {
+      this.inTimer = setTimeout(() => {
+        this.logger.debug(`[+] => in_rate: ${(1000 * this.recvCounter) / this.dt} items/s`);
```
msbrogli, line 181: The current method of calculating `in_rate` will not be really useful because it provides an average over the entire duration of the stream, from start to finish. Instead, we should calculate the average over the last N seconds (with N being between 5 and 10 seconds) to get a more accurate, real-time reflection of the rate. To further improve this, we can smooth the result using an Exponential Moving Average (EMA).

src/sync/stream.ts
```diff
@@ -61,6 +74,147 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+  }
+
+  /**
+   * Mark that an item has been processed from the queue.
+   * This also manages the outTimer, which calculates the rate of processed items.
+   */
+  proc() {
+    if (!this.outTimer) {
+      this.outTimer = setTimeout(() => {
+        this.logger.debug(`[+] <= out_rate: ${(1000 * this.procCounter) / this.dt} items/s`);
```
msbrogli, line 197: Same thing here.

msbrogli (approved):

src/sync/stream.ts
r4mmer: I'm not decreasing, this sets it to -1, so the next line `this.recv()` will increment to 0 and start the timer again.
This is so the `setTimeout` callback logs the rate of only the last `dt` inteval and not over all events.

src/sync/stream.ts
```diff
@@ -61,6 +74,147 @@ function isStreamSyncHistoryError(data: IStreamSyncHistoryData): data is IStream
...
+  dt: number;
+
+  constructor(queue: Queue, logger: ILogger) {
+    this.recvCounter = 0;
+    this.procCounter = 0;
+    this.ackCounter = 0;
+    this.emptyCounter = 0;
+
+    this.logger = logger;
+    this.dt = 2000;
```
msbrogli, line 141: So this is the sample rate for the `in_rate` and `out_rate`, right?

msbrogli (approved):

src/sync/stream.ts
msbrogli: You are already sampling every 2s. Thanks for clarifying.

src/sync/stream.ts
msbrogli: You are already sampling every 2s. Thanks for clarifying.

src/sync/stream.ts
msbrogli: So you're actually resetting the counter. Thanks for clarifying.

src/new/connection.ts
msbrogli: Got it. Ok. As discussed, I guess it might be worth refactoring it later. Thanks!

src/sync/stream.ts
r4mmer: Yes

msbrogli: Thanks for renaming it.

src/sync/stream.ts
msbrogli: Ok. We can keep it this way. Thanks for the explanation.

msbrogli (approved):

tuliomir (approved):

PR #751, opened 2024-09-03
feat: global load lock

r4mmer:
### Acceptance Criteria
- Create a global load lock (GLL).
- GLL should only allow 5 wallets to load concurrently at any time.
- Any wallets trying to sync the history should wait on the GLL for their turn.
- Log time wallets waited on GLL to start syncing.

### How

The GLL is an instance of a promise queue, meaning it runs the tasks on queue but only allow a number of concurrent runs.
For instance 5 concurrent tasks, meaning only 5 wallets can load at any given time.

The `syncHistory` method of the wallet now just uses the `addTask` to enqueue the actual history sync
```js
  async syncHistory(startIndex, count, shouldProcessHistory = false) {
    if (!(await getSupportedSyncMode(this.storage)).includes(this.historySyncMode)) {
      throw new Error('Trying to use an unsupported sync method for this wallet.');
    }
    const syncMethod = getHistorySyncMethod(this.historySyncMode);
    await addTask(async () => {
      await syncMethod(startIndex, count, this.storage, this.conn, shouldProcessHistory);
    }, this.logger);
  }
```

```js
wallet0.syncHistory(0, 20);
wallet1.syncHistory(0, 20);
wallet2.syncHistory(0, 20);
wallet3.syncHistory(0, 20);
wallet4.syncHistory(0, 20);
// This next one will only start when one of the previous finishes
wallet5.syncHistory(0, 20);
```

### Security Checklist

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -387,7 +387,7 @@ export class OracleParseError extends Error {
   errorCode: string = ErrorMessages.NANO_ORACLE_PARSE_ERROR;
 }
 
-export class GllTaskError extends Error {
+export class GlobalLoadLockTaskError extends Error {
   taskId: string;
 
   innerError: Error;
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -386,3 +386,15 @@ export class NanoContractTransactionParseError extends Error {
 export class OracleParseError extends Error {
   errorCode: string = ErrorMessages.NANO_ORACLE_PARSE_ERROR;
 }
+
+export class GllTaskError extends Error {
+  taskId: string;
+
+  innerError: Error;
+
+  constructor(taskId: string, innerError: Error) {
+    super(`${taskId} has failed with ${innerError}`);
+    this.taskId = taskId;
+    this.innerError = innerError;
+  }
+}
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -49,6 +49,7 @@ import { MemoryStore, Storage } from '../storage';
 import { deriveAddressP2PKH, deriveAddressP2SH, getAddressFromPubkey } from '../utils/address';
 import NanoContractTransactionBuilder from '../nano_contracts/builder';
 import { prepareNanoSendTransaction } from '../nano_contracts/utils';
+import { addTask } from '../sync/gll';
 
 const ERROR_MESSAGE_PIN_REQUIRED = 'Pin is required.';
 
@@ -2857,13 +2858,16 @@ class HathorWallet extends EventEmitter {
     if (!(await getSupportedSyncMode(this.storage)).includes(this.historySyncMode)) {
       throw new Error('Trying to use an unsupported sync method for this wallet.');
     }
-    await getHistorySyncMethod(this.historySyncMode)(
-      startIndex,
-      count,
-      this.storage,
-      this.conn,
-      shouldProcessHistory
-    );
+    const syncMethod = getHistorySyncMethod(this.historySyncMode);
+    await addTask(async () => {
+      await syncMethod(
+        startIndex,
+        count,
+        this.storage,
+        this.conn,
+        shouldProcessHistory
+      );
+    }, this.logger);
   }
 
   /**
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -2860,13 +2860,7 @@ class HathorWallet extends EventEmitter {
     }
     const syncMethod = getHistorySyncMethod(this.historySyncMode);
     await addTask(async () => {
-      await syncMethod(
-        startIndex,
-        count,
-        this.storage,
-        this.conn,
-        shouldProcessHistory
-      );
+      await syncMethod(startIndex, count, this.storage, this.conn, shouldProcessHistory);
     }, this.logger);
   }
 
--- a/src/sync/gll.ts
+++ b/src/sync/gll.ts
@@ -1,0 +1,67 @@
+import PQueue from 'queue-promise';
+import { ILogger } from '../types';
+
+let GLL = new PQueue({ concurrent: 2 });
+
+export function setConcurrency(value: number) {
+  GLL = new PQueue({ concurrent: value });
+}
+
+export function addTask(task: () => Promise<void>, logger: ILogger) {
+  const taskId = Math.random().toString(36).substring(2, 15);
+  const startTime = Date.now();
+  const promise = new Promise<void>((resolve, reject) => {
+    GLL.on('resolve', (data) => {
+      if (data === taskId) {
+        resolve();
+      }
+    });
+
+    GLL.on('reject', (data) => {
+      if (data.taskId === taskId) {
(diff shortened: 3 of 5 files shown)


src/errors.ts
```diff
@@ -386,3 +386,15 @@ export class NanoContractTransactionParseError extends Error {
 export class OracleParseError extends Error {
   errorCode: string = ErrorMessages.NANO_ORACLE_PARSE_ERROR;
 }
+
+export class GllTaskError extends Error {
```
msbrogli, line 390: I'd rename it to `GlobalLoadLockerTaskError`.

src/sync/gll.ts
```diff
@@ -0,0 +1,42 @@
+import PQueue from 'queue-promise';
+import { ILogger } from '../types';
+import { GllTaskError } from '../errors';
+
+let GLL = new PQueue({ concurrent: 5 });
+
+export function setConcurrency(value: number) {
+  GLL = new PQueue({ concurrent: value });
```
msbrogli, line 8: If this method is called during the node execution, then we could have two instances running simultaneously? I feel it is a bug.

src/sync/gll.ts
```diff
@@ -0,0 +1,42 @@
+import PQueue from 'queue-promise';
+import { ILogger } from '../types';
+import { GllTaskError } from '../errors';
+
+let GLL = new PQueue({ concurrent: 5 });
```
msbrogli, line 5: Move `5` to a constant, maybe `MAX_CONCURRENT_LOAD_TASKS`?

src/sync/gll.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+  });
+
+  GLL.enqueue(async () => {
+    try {
+      logger.info(`Task waited on queue for ${(Date.now() - startTime) / 1000}s`);
+      await task();
+      return taskId;
+    } catch (error: unknown) {
+      if (error instanceof Error) {
```
msbrogli, line 34: Why this logic?

src/sync/gll.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    GLL.on('reject', data => {
+      if (data.taskId === taskId) {
+        reject(data.innerError);
+      }
+    });
+  });
+
+  GLL.enqueue(async () => {
```
msbrogli, line 28: Missing docstring.

src/sync/gll.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ILogger } from '../types';
+import { GllTaskError } from '../errors';
+
+let GLL = new PQueue({ concurrent: 5 });
+
+export function setConcurrency(value: number) {
+  GLL = new PQueue({ concurrent: value });
+}
+
+export function addTask(task: () => Promise<void>, logger: ILogger) {
```
msbrogli, line 11: Missing docstring.

src/new/wallet.js
msbrogli, line 1: How are things being locked? Can you add an example to the acceptance criteria, please?

msbrogli (changes requested):
I added some threads with questions and suggestions.

src/sync/gll.ts
r4mmer: In typescript the catch argument is inherently `unknown`.
This is because anything can be thrown not just errors.

So we always need to make some type checks to ensure we are dealing with an error or something else.

src/sync/gll.ts
r4mmer: You're right, removed this method since there is no way to safely update the concurrency for now.

src/sync/gll.ts
r4mmer: Done

src/sync/gll.ts
r4mmer: Done

src/sync/gll.ts
r4mmer: Done

src/errors.ts
r4mmer: Done

src/new/wallet.js
r4mmer: Done

src/sync/gll.ts
msbrogli: Thanks!

src/sync/gll.ts
msbrogli: Thanks!

src/sync/gll.ts
msbrogli: Got it. Thanks!

src/sync/gll.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+  });
+
+  /**
+   * Add the task to the queue and return or rejects with the taskId
+   * So the promise above can detect any issues and reject correctly.
+   */
+  GLL.enqueue(async () => {
+    try {
+      logger.info(`Task waited on queue for ${(Date.now() - startTime) / 1000}s`);
```
msbrogli, line 38: Can you improve the log to add information about which wallet was waiting for that long? It might be useful for debugging.

msbrogli (approved):

src/new/wallet.js
msbrogli: Thanks!

src/sync/gll.ts
msbrogli: Thanks!

src/sync/gll.ts
msbrogli: Ok. Thanks!

src/errors.ts
msbrogli: Thanks!

src/sync/gll.ts
r4mmer: The wallet-ilb does not identify each wallet, they are just instances.
The idea behing the logger is that the wallet implementation can add a logger instance with more data.

For instance the wallet-headless will add the date and wallet-id as prefixes to the actual message.

msbrogli: Got it. I'm wondering where we can store the queue time information so the client can log it properly. Should the wallet-lib provide an object with sync information (e.g., queue time, type, downloading time, processing time, and total time)?

r4mmer: The wallet instance goes through the states:
- `CONNECTING`: Trying to establish a websocket connection with the fullnode.
- `SYNCING`: queue time + time to download the history.
- `PROCESSING`: Go through the history and process the balances and other metadata.
- `READY`: Finished processing

We could just keep an object that holds the timestamp when we last entered each state.
So knowning the queue time from this PR and the time between `SYNCING` and `PROCESSING` we can calculate how long it took to download the history.
From the time between `PROCESSING` and `READY` we know how long it took to process the history.
Since this is reacting to the state changes of the wallet we could just do this on the headless, what do you think?

Alternatively we could create a stats manager for the wallet like the stream stats manager from #750

tuliomir (approved):

PR #770, opened 2024-09-09
feat: treat streaming edge cases

r4mmer:
### Acceptance Criteria

- During history stream queue processing, await the end of JS tasks execution at the end of each loop
  - This will make IO tasks and other tasks free to run so the queue processing does not block the main loop.
- If a sync is aborted while it waits for execution, we should abort as soon as possible to avoid executing needlessly.
- Save the capabilities set from the fullnode and check if the fullnode is history-streaming capable.

### Security Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "level": "8.0.1",
         "lodash": "4.17.21",
         "long": "5.2.3",
+        "queue-microtask": "^1.2.3",
         "queue-promise": "^2.2.1",
         "ws": "8.17.1"
       },
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "level": "8.0.1",
         "lodash": "4.17.21",
         "long": "5.2.3",
-        "queue-microtask": "^1.2.3",
+        "queue-microtask": "1.2.3",
         "ws": "8.17.1"
       },
       "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "level": "8.0.1",
     "lodash": "4.17.21",
     "long": "5.2.3",
+    "queue-microtask": "^1.2.3",
     "queue-promise": "^2.2.1",
     "ws": "8.17.1"
   },
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "level": "8.0.1",
     "lodash": "4.17.21",
     "long": "5.2.3",
-    "queue-microtask": "^1.2.3",
+    "queue-microtask": "1.2.3",
     "ws": "8.17.1"
   },
   "scripts": {
--- a/src/new/connection.ts
+++ b/src/new/connection.ts
@@ -25,6 +25,8 @@ enum StreamRequestEvent {
 
 const STREAM_HISTORY_ACK_EVENT = 'request:history:ack';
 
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     "level": "8.0.1",
     "lodash": "4.17.21",
     "long": "5.2.3",
+    "queue-microtask": "^1.2.3",
```
r4mmer, line 27: This is not a new dependency, we already have this as an indirect dependency and i'm adding as a direct deendency.

src/new/connection.ts
```diff
@@ -60,6 +62,8 @@ class WalletConnection extends BaseConnection {
 
   streamWindowSize: number | undefined;
 
+  capabilities?: FullnodeCapability[];
```
r4mmer, line 65: When the websocket connects the fullnode sends a capabilities message which contains some feature flags of the features enabled on the fullnode instance.

Currently only the `history-streaming` flag exists.

luislhl (approved):

luislhl (dismissed):

package-lock.json
```diff
@@ -19,6 +19,7 @@
         "level": "8.0.1",
         "lodash": "4.17.21",
         "long": "5.2.3",
+        "queue-microtask": "^1.2.3",
```
tuliomir, line 22: chore: We should use a fixed version to match the pattern we use on the other dependencies.
```suggestion
        "queue-microtask": "1.2.3",
```

src/new/connection.ts
```diff
@@ -109,6 +114,42 @@ class WalletConnection extends BaseConnection {
...
+  }
+
+  /**
+   * If the fullnode has not sent the capabilities yet wait a while.
+   */
+  async waitCapabilities() {
+    if (this.capabilities === undefined) {
+      // Wait 2s so the fullnode has some time to send the capabilities envent
+      await new Promise<void>(resolve => {
+        setTimeout(resolve, 2000);
```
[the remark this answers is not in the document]
tuliomir: suggestion(non-blocking): If we keep this same implementation, I'd only recommend moving this `2000` to a file-wide variable or even a property in the constants/configs file, for easier configuration.

src/sync/stream.ts
```diff
@@ -426,6 +427,12 @@ export class StreamManager extends AbortController {
       throw new Error('No abort controller on connection');
     }
 
+    // The connection with the fullnode may have been lost during the time waiting to sync
+    if (signal.aborted) {
+      this.abort();
+      throw new Error('The connection already aborted this stream');
```
tuliomir, line 433: question(non-blocking): Does the scope of this PR allow adding a test for this abort condition?

tuliomir (approved):

src/sync/stream.ts
r4mmer: During the process to add this test I checked that this was unnecessary.
The idea was that the connection could be lost when waiting on the queue for execution, but the signal is created here.

I'll check how to create the abort controller before waiting on the queue, so if the connection is lost we can truly abort.

package-lock.json
r4mmer: Done

luislhl (approved):

tuliomir (approved):

PR #791, opened 2024-10-09
fix: error message for over available tokens to mint or melt

alexruzenhack:
### Motivation

When we try to mint or melt tokens with more than the available amount, it throws an error. The error message informs the requested amount to mint or melt and the **available** token amount. The bug here is that the available amount is always zero.

### Acceptance Criteria
- The available amount should be equal to the total available token, not zero

It closes https://github.com/HathorNetwork/internal-issues/issues/398

### Security Checklist

### DIFF
--- a/__tests__/integration/hathorwallet_facade.test.ts
+++ b/__tests__/integration/hathorwallet_facade.test.ts
@@ -1981,8 +1981,9 @@ describe('mintTokens', () => {
     const { hash: tokenUid } = await createTokenHelper(hWallet, 'Token to Mint', 'TMINT', 100);
 
     // Should not mint more tokens than the HTR funds allow
-    await expect(hWallet.mintTokens(tokenUid, 9000))
-      .rejects.toThrow(/^Not enough HTR tokens for deposit: 90 required, \d+ available$/);
+    await expect(hWallet.mintTokens(tokenUid, 9000)).rejects.toThrow(
+      /^Not enough HTR tokens for deposit: 90 required, \d+ available$/
+    );
 
     // Minting more of the tokens
     const mintAmount = getRandomInt(100, 50);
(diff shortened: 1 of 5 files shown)


__tests__/integration/hathorwallet_facade.test.ts
```diff
@@ -1981,7 +1981,8 @@ describe('mintTokens', () => {
     const { hash: tokenUid } = await createTokenHelper(hWallet, 'Token to Mint', 'TMINT', 100);
 
     // Should not mint more tokens than the HTR funds allow
-    await expect(hWallet.mintTokens(tokenUid, 9000)).rejects.toThrow('HTR tokens');
+    await expect(hWallet.mintTokens(tokenUid, 9000))
+      .rejects.toThrow(/^Not enough HTR tokens for deposit: 90 required, \d+ available$/);
```
alexruzenhack, line 1985: The available tokens can change for each test, therefore we match for any digit with `\d+`.

src/utils/tokens.ts
```diff
@@ -360,8 +360,9 @@ const tokens = {
     }
 
     if (foundAmount < depositAmount) {
+      const availableAmount = selectedUtxos.available;
       throw new InsufficientFundsError(
-        `Not enough HTR tokens for deposit: ${depositAmount} required, ${foundAmount} available`
+        `Not enough HTR tokens for deposit: ${depositAmount} required, ${availableAmount} available`
```
alexruzenhack, line 365: The error message is used as an error feedback message for users on Wallet Desktop. We could have removed the "available" part from the message as an easier solution, however as it is used in the frontend Carneiro and I discussed it would be better to change the "internals" rather then the message.

src/utils/tokens.ts
```diff
@@ -360,8 +360,9 @@ const tokens = {
     }
 
     if (foundAmount < depositAmount) {
+      const availableAmount = selectedUtxos.available;
```
r4mmer, line 363: This should never happen, but it is a fail-safe to avoid showing to the user `undefined available` tokens

```suggestion
      const availableAmount = selectedUtxos.available ?? 0;
```

src/utils/utxo.ts
```diff
@@ -49,7 +49,7 @@ export async function fastUtxoSelection(
   storage: IStorage,
   token: string,
   amount: OutputValueType
-): Promise<{ utxos: IUtxo[]; amount: OutputValueType }> {
+): Promise<{ utxos: IUtxo[]; amount: OutputValueType; available?: OutputValueType }> {
```
pedroferreira1, line 52: Missing this new return key in the docstring

src/utils/utxo.ts
```diff
@@ -93,7 +94,7 @@ export async function bestUtxoSelection(
...
```
pedroferreira1, line 97: Missing this new return key in the docstring

pedroferreira1 (approved):

pedroferreira1 (approved):

r4mmer (approved):

pedroferreira1 (approved):

PR #805, opened 2024-12-12
fix: bigint reviver

glevco:
### Motivation

During manual testing with the updated wallet-headless, I found a bug in the logic used to revive BigInts. We should check using JS `Number.MAX_SAFE_INTEGER`, instead.

### Acceptance Criteria

- Make `bigIntReviver()` a standalone function in the `JSONBigInt` object.
- Fix logic `bigIntReviver()` uses to determine when it should return a `BigInt`.
- Add more tests.

### Security Checklist

### DIFF
--- a/__tests__/utils/bigint.test.js.ts
+++ b/__tests__/utils/bigint.test.js.ts
@@ -54,18 +54,41 @@ const bigIntObjSchema = z.object({
 
 describe('test JSONBigInt', () => {
   test('should parse numbers', () => {
+    // Doubles should be parsed normally as JS Numbers.
     expect(JSONBigInt.parse('123')).toStrictEqual(123);
     expect(JSONBigInt.parse('123.456')).toStrictEqual(123.456);
     expect(JSONBigInt.parse('1.0')).toStrictEqual(1);
     expect(JSONBigInt.parse('1.000000000000')).toStrictEqual(1);
-    expect(JSONBigInt.parse('12345678901234567890')).toStrictEqual(12345678901234567890n);
-    expect(JSONBigInt.parse('12345678901234567890.000')).toStrictEqual(12345678901234567890n);
     expect(JSONBigInt.parse('1e2')).toStrictEqual(100);
     expect(JSONBigInt.parse('1E2')).toStrictEqual(100);
 
-    expect(() => JSONBigInt.parse('12345678901234567890.1')).toThrow(
-      Error('large float will lose precision! in "12345678901234567890.1"')
-    );
+    // This is 2**53-1 which is the MAX_SAFE_INTEGER, so it remains a Number, not a BigInt.
+    // And the analogous for MIN_SAFE_INTEGER.
+    expect(JSONBigInt.parse('9007199254740991')).toStrictEqual(9007199254740991);
+    expect(JSONBigInt.parse('-9007199254740991')).toStrictEqual(-9007199254740991);
+
+    // One more than the MAX_SAFE_INTEGER, so it becomes a BigInt. And the analogous for MIN_SAFE_INTEGER.
+    expect(JSONBigInt.parse('9007199254740992')).toStrictEqual(9007199254740992n);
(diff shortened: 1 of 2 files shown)


r4mmer (approved):

pedroferreira1 (approved):

__tests__/utils/bigint.test.js.ts
```diff
@@ -54,18 +54,41 @@ const bigIntObjSchema = z.object({
 
 describe('test JSONBigInt', () => {
   test('should parse numbers', () => {
+    // Doubles should be parsed normally as JS Numbers.
     expect(JSONBigInt.parse('123')).toStrictEqual(123);
     expect(JSONBigInt.parse('123.456')).toStrictEqual(123.456);
     expect(JSONBigInt.parse('1.0')).toStrictEqual(1);
     expect(JSONBigInt.parse('1.000000000000')).toStrictEqual(1);
-    expect(JSONBigInt.parse('12345678901234567890')).toStrictEqual(12345678901234567890n);
-    expect(JSONBigInt.parse('12345678901234567890.000')).toStrictEqual(12345678901234567890n);
```
pedroferreira1, line 62: Why did you remove this?

__tests__/utils/bigint.test.js.ts
```diff
@@ -54,18 +54,41 @@ const bigIntObjSchema = z.object({
...
     expect(JSONBigInt.parse('1e2')).toStrictEqual(100);
     expect(JSONBigInt.parse('1E2')).toStrictEqual(100);
 
-    expect(() => JSONBigInt.parse('12345678901234567890.1')).toThrow(
-      Error('large float will lose precision! in "12345678901234567890.1"')
-    );
```
pedroferreira1, line 68: Why did you remove this?

pedroferreira1 (approved):

__tests__/utils/bigint.test.js.ts
glevco: This was actually kind of incorrect. The JSON specification imposes no limitation on numbers in JSON files, that is, both very large integers and very large floats are allowed. So in theory we could have `"some_num": 12345678901234567890.1` in a JSON.

However, in practice, JS would try to parse this as a `Number` and Python would try to parse this as a `float`, and both are represented using the IEEE 754 double specification. This means this number is actually useless in all cases, as it's impossible to represent it precisely as a JS `Number` or Python `float`. Neither language would ever generate a JSON with it.

Check this out, in JS and Python:

```js
> JSON.parse('12345678901234567890.1') === 12345678901234567000.0
true
```

```python
>>> json.loads('12345678901234567890.1') == 12345678901234567000.0
True
```

Summing up, I was unnecessarily prohibiting this parse, while we should actually just let it be parsed normally as a double by JS and Python. In practice, this number will never even appear in a JSON generated by these languages.

__tests__/utils/bigint.test.js.ts
glevco: The first line was just moved below.

The second line was actually kind of incorrect. Before, I was using `Number.isInteger()` to check whether a parsed JSON value was an integer, and if it returned `true`, I would try the `BigInt` check. Because of that, another problem was created: JS returns `true` for floats ending in `.0`, so `Number.isInteger(123.0)` returns `true`. That's why I also had to add the check for zeroes in the previous code. It was convoluted and unnecessary, because I shouldn't be using `Number.isInteger()` in the first place.

Then, consider the explanation in https://github.com/HathorNetwork/hathor-wallet-lib/pull/805#discussion_r1884282621. The same theory applies in this case. in JS and Python:

```python
>>> json.loads('12345678901234567890.000') == 12345678901234567000.0
True
```

```js
> JSON.parse('12345678901234567890.000') == 12345678901234567000.0
true
```

Summing up, I was unnecessarily trying to parse this number as a `BigInt`, while we should actually just let it be parsed as a double by JS and Python. In practice, this number will never even appear in a JSON generated by these languages.

PR #808, opened 2024-12-13
feat: transaction template

r4mmer:
### Acceptance Criteria
- Implement the transaction template design ([reference](https://github.com/HathorNetwork/rfcs/blob/f83645343bf5888c83a2d9e10a80c07f1aaf7e91/projects/hathor-wallet-headless/0004-tx-transaction.md))

### Code organization

The transaction template code is organized in the following modules:
- `instructions`
  - Parsers and definitions of the user input and the formatted data that will be used as transaction templates.
- `executor`
  - Methods that will determine how each of the instructions work when running a template.
- `setvarcommands`
  - sub-routines to execute with the `action/setvar` instructions.
- `interpreter`
  - Wallet facade interpreter, acts as the bridge between the template and the wallet (fetching utxos, txs, balance.)
- `context`
  - State of the template execution and tracking of data (for instance, the balance of the transaction being built).
- `builder`
  - A helper class to build a transaction template.
- `types`
  - Typing to add abstraction.
  - This is required to split the modules in different files, without this we would introduce some import cycles.

>[!NOTE]
> The interpreter for the wallet-service is not part of this implementation.

>[!NOTE]
> There are some differences between the design and the instructions on the implementation but this is because of a "not-intuitive-behavior" on zod, specially when dealing with unions.

### Future possibilities

1. Code duplication in the executors

In the executors there is some code duplication (especially when creating inputs/outputs and updating the context balance), but I avoided abstracting too much in the initial implementation so we can have more freedom in how we implement each instruction, later I think we should come back to this and make the executor code cleaner.

2. Choosing change address

An issue in the usual transactions of the wallet facade is that if we get a wallet address to use and then a change address it can be the same address since the wallet does not have the context of other addresses that are being used in the transaction.
The template has this information and it could be used to avoid choosing the same address twice in the same tx.
Since this is not an issue with the wallet facade I did not implement in the tx template but we could easily add this.


src/new/wallet.js
```diff
@@ -2781,7 +2782,7 @@ class HathorWallet extends EventEmitter {
       const tokenIdx = tokenUtils.getTokenIndexFromData(token_data);
       const tokenUid = tokens[tokenIdx - 1]?.uid;
       if (!tokenUid) {
-        throw new Error(`Token ${tokenUid} not found in tokens list`);
```
r4mmer, line 2784: if `!tokenUid` means that `tokenUid` is `undefined` or empty string, which makes no sense adding to the error message.
This is why I changed to `token_data` instead, this makes understanding what went wrong easier.

src/template/transaction/builder.ts
```diff
@@ -0,0 +1,187 @@
...
+const SetVarInsArgs = SetVarInstruction.omit({ type: true });
+
+export class TransactionTemplateBuilder {
+  template: z.infer<typeof TransactionTemplate>;
+
+  constructor() {
+    this.template = [];
+  }
+
+  static from(instructions: z.input<typeof TransactionTemplate>): TransactionTemplateBuilder {
```
r4mmer, line 48: Obs: `z.input<T>` means that this accepts the "unparsed" instruction.
When the type is `z.infer<T>` it means the output of the parsed data.

src/template/transaction/instructions.ts
```diff
@@ -0,0 +1,229 @@
...
+export const SetVarGetWalletBalanceOpts = z.object({
+  method: z.literal('get_wallet_balance'),
+  token: TemplateRef.or(TokenSchema.default('00')),
+  authority: z.enum(['mint', 'melt']).optional(),
+});
+
+export const SetVarCallArgs = z.discriminatedUnion('method', [
+  SetVarGetWalletAddressOpts,
+  SetVarGetWalletBalanceOpts,
+]);
```
r4mmer, line 205: This was changed from the design because non discriminated unions with a lot of optionals were being difficult to parse correctly.

__tests__/template/transaction/executor.test.ts
```diff
@@ -0,0 +1,617 @@
...
+    type: 'action/shuffle',
+    target: 'all',
+  });
+  const arr: bigint[] = [];
+  for (let i = 1n; i < 10; i++) {
+    ctx.addOutput(-1, new Output(i, Buffer.alloc(1)));
+    ctx.addInput(-1, new Input(txId, Number(i)));
+    arr.push(i);
+  }
+  await executor(interpreter, ctx, ins);
```
tuliomir, line 546: suggestion(non-blocking): Add assertions that ensure the steps are the same with a method like `contains`, or check the final balance of the transaction to confirm its equivalence to the original one.

__tests__/integration/hathorwallet_facade.test.ts
```diff
@@ -34,11 +34,62 @@ import Network from '../../src/models/network';
...
+    const hWallet = await generateWalletHelper();
+    const address = await hWallet.getAddressAtIndex(1);
+
+    await GenesisWalletHelper.injectFunds(hWallet, address, 10n);
+
+    const template = new TransactionTemplateBuilder()
+      .addConfigAction({ tokenName: 'Tmpl Token', tokenSymbol: 'TT' })
+      .addSetVarAction({ name: 'addr', call: { method: 'get_wallet_address' } })
+      .addUtxoSelect({ fill: 1 })
+      .addTokenOutput({ address: '{addr}', amount: 100, useCreatedToken: true })
```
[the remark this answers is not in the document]
r4mmer: We already enforce this since the methods will parse these inputs and convert number to bigint

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+      // Creates an output with the value of the outstanding balance
+      const output = new Output(TOKEN_MINT_MASK, script, {
+        timelock,
+        tokenData: tokenData | TOKEN_AUTHORITY_MASK,
+      });
+      ctx.addOutput(-1, ...Array(count).fill(output));
+    }
+
+    if (balance.melt_authorities > 0) {
+      const count = balance.mint_authorities;
```
pedroferreira1, line 540:
```suggestion
      const count = balance.melt_authorities;
```

Maybe we should add all possibilities to the tests.

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+    tokensToCheck.push(token);
+  } else {
+    // Check HTR and all tokens on the transaction
+    tokensToCheck.push(NATIVE_TOKEN_UID);
+    ctx.tokens.forEach(tk => {
+      tokensToCheck.push(tk);
+    });
+  }
+
+  // const script = createOutputScriptFromAddress(address, interpreter.getNetwork());
```
pedroferreira1, line 601: Remove?

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+      });
+      ctx.addOutput(-1, ...Array(count).fill(output));
+    }
+  }
+}
+
+/**
+ * Execution for CompleteTxInstruction
+ */
+export async function execCompleteTxInstruction(
```
pedroferreira1, line 559: Maybe you could create a method to fill inputs and use the method above and the new one. I feel that you are duplicating exactly the same code as above only because this 'complete' method fills also inputs.

Maybe just reuse the methods above to select utxos for authority and tokens.

What do you think?

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+        ctx.log(`Found utxo with ${utxo.value} of ${utxo.tokenId}`);
+        ctx.log(`Create input ${utxo.index} / ${utxo.txId}`);
+        inputs.push(new Input(utxo.txId, utxo.index));
+      }
+
+      // Update balance
+      for (const input of inputs) {
+        const origTx = await interpreter.getTx(input.hash);
+        ctx.balance.addInput(origTx, input.index);
+      }
```
pedroferreira1, line 641: Why don't you do this in the same loop above?

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+  if (ins.call.method === 'get_wallet_balance') {
+    // Validate options and get token variable
+    const callArgs = SetVarGetWalletBalanceOpts.parse(ins.call);
+    const token = getVariable<string>(
+      callArgs.token,
+      ctx.vars,
+      SetVarGetWalletBalanceOpts.shape.token
+    );
+    callArgs.token = token;
+    const newOptions = { ...callArgs, token };
```
pedroferreira1, line 814: You have just set `callArgs.token = token;` above. Do you need this `{ token }`?

src/template/transaction/instructions.ts
r4mmer:
```ts
const SetVarOptions = z.union([SetVarGetWalletAddressOpts, SetVarGetWalletBalanceOpts]);

const SetVarInstruction = z.object({
  type: z.literal('action/setvar'),
  name: z.string().regex(/[\d\w]+/),
  value: z.any().optional(),
  action: z.enum(['get_wallet_address', 'get_wallet_balance']),
  options: SetVarOptions.optional(),
});
```

This was the original setvar instruction, the issue with this is that the action + options was not working because the options could not differentiate each "Opts" of the union.
To fix this we would need to create a discriminated union and move the "action" to inside the options, i.e. `{..., options: {type: action, ...}}`.
If you look at the instruction, this is exactly what I did, just changing some names because the old names did not seem to fit the new schema.

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+export class TxTemplateContext {
+  version: number;
+
+  signalBits: number;
+
+  inputs: Input[];
+
+  outputs: Output[];
+
+  tokens: string[]; // use token data?
```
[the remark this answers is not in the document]
r4mmer: Removed the comment.
It was left over from initial development

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  balance: TxBalance;
+
+  tokenName?: string;
+
+  tokenSymbol?: string;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  vars: Record<string, any>;
```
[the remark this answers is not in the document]
r4mmer: Done

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+  useCreateTxContext() {
+    if (this.tokens.length !== 0) {
+      throw new Error(
+        `Trying to build a create token tx with ${this.tokens.length} tokens on the array`
+      );
+    }
+    this.version = CREATE_TOKEN_TX_VERSION;
+  }
+
+  getTokenDataFor(token: string, useCreatedToken: boolean) {
```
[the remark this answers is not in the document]
r4mmer: Nice catch, this method was not being used.
Added an explanation on the `addToken` method which is used by the executors.

src/template/transaction/interpreter.ts
```diff
@@ -0,0 +1,163 @@
...
+      return this.txCache[txId];
+    }
+
+    const histtx = await this.wallet.getTx(txId);
+    if (histtx) {
+      this.txCache[txId] = histtx as IHistoryTx;
+      return this.txCache[txId];
+    }
+
+    function hidrateIOWithToken<T extends FullNodeInput | FullNodeOutput>(
```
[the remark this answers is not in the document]
r4mmer: Done

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (index > -1) {
+      // Token is already on the list.
+      return index + 1;
+    }
+    // Token is not on the list, adding now
+    this.tokens.push(token);
+    return this.tokens.length;
+  }
+
+  addInput(position: number, ...inputs: Input[]) {
```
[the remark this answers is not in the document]
r4mmer: done

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (position === -1) {
+      this.inputs.push(...inputs);
+      return;
+    }
+
+    this.inputs.splice(position, 0, ...inputs);
+  }
+
+  addOutput(position: number, ...outputs: Output[]) {
```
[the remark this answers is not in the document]
r4mmer: done

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+  );
+  ctx.log(`count(${count}), address(${address}), token(${token})`);
+
+  let authoritiesInt = 0n;
+  if (authority === 'mint') {
+    authoritiesInt += 1n;
+  }
+  if (authority === 'melt') {
+    authoritiesInt += 2n;
+  }
```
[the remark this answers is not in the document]
r4mmer: Done

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+      amount = TOKEN_MELT_MASK;
+      tokenData |= TOKEN_AUTHORITY_MASK;
+      break;
+    default:
+      break;
+  }
+  ctx.log(`amount(${amount})`);
+  if (!amount) {
+    throw new Error('Raw token output missing amount');
+  }
```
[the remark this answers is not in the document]
r4mmer: Added a validation closer to line 263, but I still have to leave this so the typing system understands that amount cannot be undefined.

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+    await waitForTxReceived(hWallet, tx.hash, null);
+    expect(tx.hash).toBeDefined();
+  });
+
+  it('should be able to complete change', async () => {
```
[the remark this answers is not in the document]
r4mmer: Changed the names of the tests

__tests__/template/transaction/executor.test.ts
r4mmer: Done

src/utils/transaction.ts
```diff
@@ -692,6 +693,54 @@ const transaction = {
     // If there is no match
     return 'Unknown';
   },
+
+  convertFullNodeTxToHistoryTx(tx: FullNodeTx) {
```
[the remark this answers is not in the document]
r4mmer: Added

src/utils/transaction.ts
```diff
@@ -692,6 +693,54 @@ const transaction = {
...
+    const resp = {} as IHistoryTx;
+    resp.tx_id = tx.hash;
+    resp.version = tx.version;
+    resp.weight = tx.weight;
+    resp.parents = tx.parents;
+    resp.timestamp = tx.timestamp;
+    resp.token_name = tx.token_name ?? undefined;
+    resp.token_symbol = tx.token_symbol ?? undefined;
+    // resp.nonce = tx.nonce; // number = string, should we convert?
+    // resp.height, resp.first_block, resp.nc_* missing.
```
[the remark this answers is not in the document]
r4mmer: The meta was not being passed to the function, made changes to the method so we can get all fields.

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (this.debug) {
+      this._logger.info(message);
+    }
+  }
+
+  get logArray(): string[] {
+    return this._logs;
+  }
+
+  useCreateTxContext() {
```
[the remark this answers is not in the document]
r4mmer: Renamed

src/template/transaction/context.ts
```diff
@@ -0,0 +1,221 @@
...
+import Output from '../../models/output';
+import transactionUtils from '../../utils/transaction';
+import { CREATE_TOKEN_TX_VERSION, DEFAULT_TX_VERSION, NATIVE_TOKEN_UID } from '../../constants';
+
+export interface TokenBalance {
+  tokens: OutputValueType;
+  mint_authorities: number;
+  melt_authorities: number;
+}
+
```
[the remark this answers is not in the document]
r4mmer: added

src/template/transaction/executor.ts
r4mmer: Done

src/template/transaction/executor.ts
r4mmer: Done

src/template/transaction/executor.ts
r4mmer: This is part of what I said about allowing some code duplication on the executors in this first PR.
Later we can abstract based on usage.

src/template/transaction/executor.ts
r4mmer: Done

src/template/transaction/executor.ts
r4mmer: Done

src/template/transaction/executor.ts
```diff
@@ -0,0 +1,822 @@
...
+      // Need to create a token output
+      // Add balance to the ctx.balance
+      ctx.balance.addOutput(value, tokenUid);
+
+      // Creates an output with the value of the outstanding balance
+      const output = new Output(value, script, { timelock, tokenData });
+      ctx.addOutput(-1, output);
+    }
+
+    if (balance.mint_authorities > 0) {
```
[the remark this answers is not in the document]
r4mmer: After some thought I think you are right, decided to remove mint changes.
Any authority should be explicitly stated as outputs so i removed these as per your suggestion.

src/new/wallet.js
```diff
@@ -1067,6 +1070,7 @@ class HathorWallet extends EventEmitter {
       token: NATIVE_TOKEN_UID,
       filter_address: null,
       ...options,
+      order_by_value: 'desc',
```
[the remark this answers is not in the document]
r4mmer: This issue has since been solved in another [PR](https://github.com/HathorNetwork/hathor-wallet-lib/pull/818).

src/template/transaction/instructions.ts
```diff
@@ -0,0 +1,229 @@
...
+  token: TemplateRef.or(CustomTokenSchema),
+  count: TemplateRef.or(CountSchema.default(1)),
+  address: TemplateRef.or(AddressSchema.optional()),
+});
+
+export const RawOutputInstruction = z.object({
+  type: z.literal('output/raw'),
+  position: z.number().default(-1),
+  amount: TemplateRef.or(AmountSchema.optional()),
+  script: TemplateRef.or(z.string().regex(/^([a-fA-F0-9]{2})+$/)),
```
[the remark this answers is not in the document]
r4mmer: As discussed, we are not limiting the size of the script on the lib.
The limit of the output script is configured in the fullnode ([reference](https://github.com/HathorNetwork/hathor-core/blob/dbcff5570956d08d878c42250653a5a7f5897e93/hathor/conf/settings.py#L327-L328))

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    expect(tx.outputs[0].value).toBe(1n);
+    expect(tx.outputs[1].tokenData).toBe(129);
+    expect(tx.outputs[1].value).toBe(2n);
+  });
+
+  it('should be able to complete a transaction inputs', async () => {
+    const address = await hWallet.getAddressAtIndex(25);
+    const template = new TransactionTemplateBuilder()
+      .addSetVarAction({ name: 'addr', value: address })
+      .addSetVarAction({ name: 'token', value: tokenUid })
```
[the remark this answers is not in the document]
r4mmer: Added some tests that do not use template variables

__tests__/integration/hathorwallet_facade.test.ts
```diff
@@ -34,11 +34,62 @@ import Network from '../../src/models/network';
...
+      .build();
+
+    const tx = await hWallet.buildTxTemplate(template, { signTx: true, pinCode: DEFAULT_PIN_CODE });
+    expect(tx.version).toEqual(2); // Create token transaction
+    expect(tx.inputs).toHaveLength(1);
+    expect(tx.inputs[0].data).not.toBeFalsy(); // Tx is signed
+    // Transaction is not mined yet
```
pedroferreira1, line 66: suggestion(non-blocking): why don't you mine and send this now to complete the test?

__tests__/integration/hathorwallet_facade.test.ts
```diff
@@ -34,11 +34,62 @@ import Network from '../../src/models/network';
...
+      .addUtxoSelect({ fill: 1 })
+      .addTokenOutput({ address: '{addr}', amount: 100, useCreatedToken: true })
+      .build();
+
+    const tx = await hWallet.runTxTemplate(template, DEFAULT_PIN_CODE);
+    expect(tx.version).toEqual(2); // Create token transaction
+    expect(tx.inputs).toHaveLength(1);
+    expect(tx.inputs[0].data).not.toBeFalsy(); // Tx is signed
+    // Transaction is mined and pushed
+    expect(tx.hash).not.toBeNull();
```
pedroferreira1, line 89: suggestion: validate that the tx has two outputs, the mined tokens with 100n amount and the change output.

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      .build();
+
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+    if (tx.hash === null) {
+      throw new Error('Transaction does not have a hash');
+    }
```
pedroferreira1, line 53: suggestion: isn't it better to `expect`?

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      .build();
+
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+    await waitForTxReceived(hWallet, tx.hash, null);
+
+    expect(tx.outputs).toHaveLength(0);
```
pedroferreira1, line 145: suggestion: validate that the 8 inputs are there with the authorities

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      .addShuffleAction({ target: 'all' })
+      .build();
+
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+    await waitForTxReceived(hWallet, tx.hash, null);
+    expect(tx.hash).toBeDefined();
```
pedroferreira1, line 217: suggestion: validate that there are 3 outputs.

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      .addCompleteAction({})
+      .build();
+
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+    await waitForTxReceived(hWallet, tx.hash, null);
+    expect(tx.hash).toBeDefined();
```
pedroferreira1, line 242: suggestion: validate that there are 4 outputs

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      .addChangeAction({})
+      .build();
+
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+    await waitForTxReceived(hWallet, tx.hash, null);
+    expect(tx.hash).toBeDefined();
```
pedroferreira1, line 262: suggestion: validate there are 3 outputs.

__tests__/integration/template/transaction/template.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    const tx = await interpreter.build(template, DEBUG);
+    await transactionUtils.signTransaction(tx, hWallet.storage, DEFAULT_PIN_CODE);
+    tx.prepareToSend();
+    const sendTx = new SendTransaction({ storage: hWallet.storage, transaction: tx });
+    await sendTx.runFromMining();
+
+    expect(tx.hash).toBeDefined();
+    if (!tx.hash) {
+      throw new Error('tx hash should be defined');
+    }
```
pedroferreira1, line 287: suggestion: move to `expect`

pedroferreira1 (dismissed):

__tests__/integration/template/transaction/template.test.ts
r4mmer: Added expect, but need to keep the if statement to ensure typing.

__tests__/integration/template/transaction/template.test.ts
r4mmer: We already have an expect, but the expect does not ensure the typing so we also need the if statement.

__tests__/integration/template/transaction/template.test.ts
r4mmer: Done

__tests__/integration/template/transaction/template.test.ts
r4mmer: Done

__tests__/integration/template/transaction/template.test.ts
r4mmer: Done

__tests__/integration/template/transaction/template.test.ts
r4mmer: Done

__tests__/integration/hathorwallet_facade.test.ts
r4mmer: Done

__tests__/integration/hathorwallet_facade.test.ts
r4mmer: Added the expect, but the if statement is required to ensure typing consistency

tuliomir (approved):

PR #810, opened 2025-01-10
feat: process txs as they arrive

r4mmer:
### Context

Nowadays we run the method to process the wallet full history every time a new tx arrives in the ws. Wallets with many transactions/addresses usually take a long time (0.5s) to process their entire history. There are cases where the wallet returns and error because of insufficient funds because it's processing the history of a recent transaction that has just arrived when the user wants to send a new tx.

### Acceptance Criteria
- Process only the single tx that arrives in the ws (including delete the utxos that are being spent in this tx).
- If this tx is an update and the voided state has changed, then we run the full process history, just to make this easier.
- While this processing is being done we change the wallet state to PROCESSING, so users know that they must wait until creating the next tx.

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { get } from 'lodash';
+import { cloneDeep, get } from 'lodash';
 import bitcore, { HDPrivateKey } from 'bitcore-lib';
 import EventEmitter from 'events';
 import { NATIVE_TOKEN_UID, P2SH_ACCT_PATH, P2PKH_ACCT_PATH } from '../constants';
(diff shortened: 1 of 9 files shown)


src/new/wallet.js
```diff
@@ -1334,12 +1337,30 @@ class HathorWallet extends EventEmitter {
...
+      // voidness alone, without processing everything
+      // but for now it's an isolated case and it's easier
+      await this.storage.processHistory();
+    } else if (isNewTx) {
+      // Process this single transaction.
+      // Handling new metadatas and deleting utxos that are not available anymore
+      await this.storage.processNewTx(newTx);
+    }
+    // restore previous state
+    this.state = previousState;
```
andreabadesso, line 1363: _Comment semi-related to line number_

Code looks ok, but should we add tests for this new mechanism?

andreabadesso (approved):

src/new/wallet.js
pedroferreira1: Added the test https://github.com/HathorNetwork/hathor-wallet-lib/pull/810/commits/54c51ceead6667c3c29427d793bcfb71f71dd62f

pedroferreira1 (approved):

andreabadesso (approved):

PR #812, opened 2025-02-03
tests: check tx processing on metadata updates

r4mmer:
### Acceptance Criteria
- Check that when a metadata update event is received the history is processed correctly.

### Security Checklist

### DIFF
--- a/__tests__/integration/hathorwallet_others.test.ts
+++ b/__tests__/integration/hathorwallet_others.test.ts
@@ -1319,7 +1319,7 @@ describe('processing transaction metadata changes', () => {
         token: expect.objectContaining({ id: NATIVE_TOKEN_UID }),
         transactions: 0,
         balance: { unlocked: 0n, locked: 0n },
-      })
+      }),
     ]);
 
     const injectedTx = await GenesisWalletHelper.injectFunds(hWallet, addr0, 10n);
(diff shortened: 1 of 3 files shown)


__tests__/integration/hathorwallet_others.test.ts
```diff
@@ -18,9 +19,88 @@ import dateFormatter from '../../src/utils/date';
...
+      }),
+    ]);
+
+    expect(wsSpy).toHaveBeenCalled();
+
+    const { lastCall } = wsSpy.mock;
+    expect(lastCall).toBeDefined();
+    if (!lastCall) {
+      throw new Error('Unexpected error');
+    }
```
tuliomir, line 81: suggestion(non-blocking): Lines 79 to 81 are redundant: if `!lastCall`, jest will have already finished the test with a failure.

tuliomir (dismissed):

__tests__/integration/hathorwallet_others.test.ts
r4mmer: Yes, but typescript type system does not know this.
The `expect` statement was to make sure the test fails gracefully and the if statement was to ensure the typing.

glevco (approved):

tuliomir (approved):

PR #813, opened 2025-02-04
feat: process tx after push

r4mmer:
### Acceptance Criteria
- Add to storage and process after push-tx on SendTransaction facade

### Security Checklist

### DIFF
--- a/src/new/sendTransaction.ts
+++ b/src/new/sendTransaction.ts
@@ -422,6 +422,18 @@ export default class SendTransaction extends EventEmitter {
               throw new WalletError(ErrorMessages.TRANSACTION_IS_NULL);
             }
             this.transaction.updateHash();
+            if (this.storage) {
+              // Add transaction to storage and process storage
+              (async () => {
+                if (!(this.storage && this.transaction)) {
+                  return;
+                }
+                const historyTx = await transactionUtils.convertTransactionToHistoryTx(this.transaction, this.storage);
+                await this.storage.addTx(historyTx)
+                // scan addresses?
+                await this.storage.processNewTx(historyTx);
+              })();
+            }
             this.emit('send-tx-success', this.transaction);
             resolve(this.transaction);
           } else {
--- a/src/utils/transaction.ts
+++ b/src/utils/transaction.ts
@@ -39,6 +39,7 @@ import {
   IInputSignature,
   ITxSignatureData,
   OutputValueType,
+  IHistoryInput,
 } from '../types';
 import Address from '../models/address';
 import P2PKH from '../models/p2pkh';
@@ -566,6 +567,71 @@ const transaction = {
       return new Transaction(inputs, outputs, options);
     }
     throw new ParseError('Invalid transaction version.');
+  },
+
+  async convertTransactionToHistoryTx(tx: Transaction|CreateTokenTransaction|NanoContract, storage: IStorage): Promise<IHistoryTx> {
+    if (!tx.hash) {
+      throw new Error('To be a history tx a calculated hash is required');
+    }
+
+    const inputs: IHistoryInput[] = [];
+    const outputs: IHistoryOutput[] = [];
+
+    for (const input of tx.inputs) {
+      let spentTx = await storage.getTx(input.hash);
+      if (!spentTx) {
+        // Get from API?
+        spentTx = {} as IHistoryTx;
+      }
+      const spentOut = spentTx.outputs[input.index];
+      inputs.push({
+        tx_id: input.hash,
+        index: input.index,
+        script: spentOut.script,
+        decoded: spentOut.decoded,
+        token_data: spentOut.token_data,
+        token: spentOut.token,
+        value: spentOut.value,
+      });
+    }
+    for (const output of tx.outputs) {
+      const script = output.parseScript(storage.config.getNetwork());
+      outputs.push({
+        value: output.value,
+        token_data: output.tokenData,
+        script: output.script.toString('hex'),
+        decoded: script?.toData() ?? {},
+        token: output.isTokenHTR() ? NATIVE_TOKEN_UID : tx.tokens[output.getTokenIndex()],
+        spent_by: null, // Cannot reconstruct this field
+      });
+    }
+    const histTx: IHistoryTx = {
+      tx_id: tx.hash,
+      signalBits: tx.signalBits,
+      version: tx.version,
+      weight: tx.weight,
+      timestamp: tx.timestamp ?? 0,
+      is_voided: false, // tx.voided?
+      nonce: tx.nonce,
+      inputs,
+      outputs,
+      parents: tx.parents,
+      tokens: tx.tokens,
+    };
+
+    if (tx.version === CREATE_TOKEN_TX_VERSION) {
+      histTx.token_name = (tx as CreateTokenTransaction).name;
+      histTx.token_symbol = (tx as CreateTokenTransaction).symbol;
+    }
+
+    if (tx.version === NANO_CONTRACTS_VERSION) {
+      histTx.nc_id = (tx as NanoContract).id;
+      histTx.nc_args = (tx as NanoContract).args.map(a => a.toString('hex')).join('');
+      histTx.nc_pubkey = (tx as NanoContract).pubkey.toString('hex');
+      // histTx.nc_blueprint_id // is this required?
+    }
+
+    return histTx;
   },
 
   /**
--- a/src/models/p2pkh.ts
+++ b/src/models/p2pkh.ts
@@ -16,6 +16,7 @@ import {
 import { intToBytes } from '../utils/buffer';
 import helpers from '../utils/helpers';
 import Address from './address';
+import { IHistoryOutputDecoded } from '../types';
 
 type optionsType = {
   timelock?: number | null | undefined;
@@ -80,6 +81,17 @@ class P2PKH {
     arr.push(OP_EQUALVERIFY);
     arr.push(OP_CHECKSIG);
     return util.buffer.concat(arr);
+  }
+
+  /**
+   * Build the original decoded script
+   */
+  toData(): IHistoryOutputDecoded {
+    return {
+      type: this.getType(),
+      address: this.address.base58,
+      timelock: this.timelock,
+    }
   }
 
   /**
--- a/src/models/p2sh.ts
+++ b/src/models/p2sh.ts
@@ -10,6 +10,7 @@ import { OP_GREATERTHAN_TIMESTAMP, OP_HASH160, OP_EQUAL } from '../opcodes';
 import helpers from '../utils/helpers';
 import { intToBytes } from '../utils/buffer';
 import Address from './address';
+import { IHistoryOutputDecoded } from '../types';
 
 type optionsType = {
   timelock?: number | null | undefined;
@@ -47,6 +48,17 @@ class P2SH {
   // eslint-disable-next-line class-methods-use-this -- This method returns a hardcoded constant
   getType(): 'p2sh' {
     return 'p2sh';
+  }
+
+  /**
+   * Build the original decoded script
+   */
+  toData(): IHistoryOutputDecoded {
+    return {
+      type: this.getType(),
+      address: this.address.base58,
+      timelock: this.timelock,
+    }
   }
 
   /**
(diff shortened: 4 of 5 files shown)


__tests__/integration/helpers/wallet.helper.ts
```diff
@@ -234,7 +234,7 @@ export async function stopAllWallets() {
  *
  * @return {Promise<CreateNewTokenResponse>}
  */
-export async function createTokenHelper(hWallet, name, symbol, amount, options) {
+export async function createTokenHelper(hWallet, name, symbol, amount, options = {}) {
```
r4mmer, line 237: _unrelated to line_
All changes in this file are related to typing, some jsdoc types that were not converted into ts types.
This removes some typing alerts from the tests.
Obs: We don't have alerts on the CI because we exclude tests from tsc

src/new/sendTransaction.ts
```diff
@@ -422,6 +424,23 @@ export default class SendTransaction extends EventEmitter {
               throw new WalletError(ErrorMessages.TRANSACTION_IS_NULL);
             }
             this.transaction.updateHash();
+            if (this.storage) {
```
r4mmer, line 427: This is the core of the PR.

When we push a transaction we need to convert it from a `Transaction` instance to a `IHistoryTx` (which is what is received via websocket) and add it to the storage with normal processing.

If we are creating a token we should also add it to the tokens here because the fullnode may not know that this token exists when we ask for the name and symbol for this UID. It also saves us from making an extraneous API call.

src/utils/transaction.ts
```diff
@@ -568,6 +570,137 @@ const transaction = {
...
+              // XXX: should we cover nano and other types?
+
+              return resolve(outRet);
+            })
+            .catch(err => reject(err));
+        });
+
+        // If we could not get tx from api or any other error happened
+        if (!spentOut) {
+          throw new Error(`Could not fetch tx ${tx.hash} from fullnode`);
```
tuliomir, line 635: issue: I think this is important to test, as it's handling errors for a pain point for both users and developers.

src/utils/transaction.ts
```diff
@@ -568,6 +570,137 @@ const transaction = {
...
+    for (const output of tx.outputs) {
+      const script = output.parseScript(storage.config.getNetwork());
+      let token = NATIVE_TOKEN_UID;
+      if (!output.isTokenHTR()) {
+        if (tx.version === CREATE_TOKEN_TX_VERSION) {
+          token = tx.hash;
+        } else {
+          const index = output.getTokenIndex();
+          if (tx.tokens.length <= index) {
+            throw new Error('Fetching a token that is not l');
```
tuliomir, line 659: chore: I believe this message is incomplete

src/utils/transaction.ts
```diff
@@ -568,6 +570,137 @@ const transaction = {
...
+      });
+    }
+
+    const histTx: IHistoryTx = {
+      tx_id: tx.hash,
+      signalBits: tx.signalBits,
+      version: tx.version,
+      weight: tx.weight,
+      timestamp: tx.timestamp ?? 0,
+      is_voided: false, // tx.voided?
```
tuliomir, line 681: question(non-blocking): Should we leave this comment here?

r4mmer: Done

src/utils/transaction.ts
r4mmer: Fixed

src/utils/storage.ts
```diff
@@ -890,3 +897,24 @@ export async function processUtxoUnlock(
...
+export async function addCreatedTokenFromTx(
+  tx: CreateTokenTransaction,
+  storage: IStorage
+): Promise<void> {
+  if (tx.version !== CREATE_TOKEN_TX_VERSION) {
+    return;
+  }
+
+  if (!tx.hash) {
+    throw new Error('Cannot infer UID from transaction without hash');
```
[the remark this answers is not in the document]
r4mmer: Added unit tests for these conditions

src/utils/transaction.ts
r4mmer: Added unit tests for these conditions

__tests__/utils/transaction.test.ts
```diff
@@ -464,3 +468,164 @@ test('getTxType', () => {
...
+  try {
+    tx.hash = 'resolve-fail-case';
+    tx.inputs = [new Input('resolve-fail', 0)];
+    await expect(transaction.convertTransactionToHistoryTx(tx, storage)).rejects.toThrow(
+      'failed api call'
+    );
+    tx.hash = 'no-outputs-case';
+    tx.inputs = [new Input('no-outputs', 0)];
+    await expect(transaction.convertTransactionToHistoryTx(tx, storage)).rejects.toThrow(
+      'Index outsite of tx output array bounds'
```
tuliomir, line 569: fix: minor typo
```suggestion
      'Index outside of tx output array bounds'
```

tuliomir (approved):

__tests__/utils/transaction.test.ts
r4mmer: Done

src/utils/storage.ts
```diff
@@ -555,8 +558,12 @@ export async function _updateTokensData(storage: IStorage, tokens: Set<string>):
...
       const { name, symbol } = response;
       const tokenData = { uid, name, symbol };
-      // saveToken will ignore the meta and save only the token config
-      await store.saveToken(tokenData);
+
+      const storeToken = await store.getToken(uid);
+      if (storeToken === null) {
+        // saveToken will ignore the meta and save only the token config
+        await store.saveToken(tokenData);
+      }
```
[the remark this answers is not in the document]
r4mmer: Yes, changed to use the storage method

src/storage/storage.ts
```diff
@@ -120,8 +120,12 @@ export class Storage implements IStorage {
    * Set the native token config on the store
    */
   async saveNativeToken(): Promise<void> {
-    if ((await this.store.getToken(NATIVE_TOKEN_UID)) === null) {
-      await this.store.saveToken(this.getNativeTokenData());
+    await this.addToken(this.getNativeTokenData());
+  }
+
+  async addToken(data: ITokenData): Promise<void> {
```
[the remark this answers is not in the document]
r4mmer: Added docstring

tuliomir (approved):

src/new/sendTransaction.ts
```diff
@@ -422,6 +424,23 @@ export default class SendTransaction extends EventEmitter {
...
+                  transaction,
+                  storage
+                );
+                // Add transaction to storage
+                await storage.addTx(historyTx);
+                // Add token from a create token transaction to the storage
+                // This just returns if the transaction is not a CREATE_TOKEN_TX
+                await addCreatedTokenFromTx(transaction as CreateTokenTransaction, storage);
+                // Process new transaction
+                await storage.processNewTx(historyTx);
```
pedroferreira1, line 441: Is it possible to call the `wallet.onNewTx` method instead of this? I think you are missing a call to `await this.scanAddressesToLoad();`. My main concern here is that we add things in that method and forget to add here as well, like this scan addresses to load.

src/utils/transaction.ts
```diff
@@ -572,6 +574,124 @@ const transaction = {
...
+        );
+      }
+
+      const spentOut = spentTx.outputs[input.index];
+      inputs.push({
+        tx_id: input.hash,
+        index: input.index,
+        script: spentOut.script,
+        decoded: spentOut.decoded,
+        token_data: spentOut.token_data,
```
pedroferreira1, line 639: I think this might be wrong. The tokens array of the spent tx might be different than the tokens array of the tx, so the token data might be different.

src/utils/transaction.ts
```diff
@@ -572,6 +574,124 @@ const transaction = {
...
+      };
+      outputs.push(this.hydrateIOWithToken(out, tokensArray));
+    }
+
+    const histTx: IHistoryTx = {
+      tx_id: tx.hash,
+      signalBits: tx.signalBits,
+      version: tx.version,
+      weight: tx.weight,
+      timestamp: tx.timestamp ?? 0,
```
pedroferreira1, line 666: Is it possible not to have a timestamp?

src/utils/transaction.ts
```diff
@@ -572,6 +574,124 @@ const transaction = {
...
+      histTx.token_name = (tx as CreateTokenTransaction).name;
+      histTx.token_symbol = (tx as CreateTokenTransaction).symbol;
+    }
+
+    if (tx.version === NANO_CONTRACTS_VERSION) {
+      histTx.nc_id = (tx as NanoContract).id;
+      histTx.nc_method = (tx as NanoContract).method;
+      histTx.nc_args = (tx as NanoContract).args.map(a => a.toString('hex')).join('');
+      histTx.nc_pubkey = (tx as NanoContract).pubkey.toString('hex');
+      // XXX: is the nc_blueprint_id required?
```
pedroferreira1, line 688: Did you test what comes from the ws in case of a nano contract initialize method?

src/new/sendTransaction.ts
r4mmer: The issue is that the SendTransaction does not have the wallet instance.
We can change this to use the wallet instance instead of the storage instance, but it would require some breaking changes.

While the `scanAddressesToLoad` is missing from this logic it will run as soon as the tx is received via websocket.

src/utils/transaction.ts
r4mmer: Yes, but the `token_data` of the input is the `token_data` on the transaction being spent (checked on the hathor-code repo).

The token array is only for the tokens on the outputs, for instance, if we have a tx that is melting tokens and not creating a melt authority the token array will be empty, meaning that the token_data for the token being melted cannot be constructed with the array from the current tx.

src/utils/transaction.ts
r4mmer: It shouldn't, but the types are:
- `Transaction.timestamp: number | null` 
- `IHistoryTx.timestamp: number`

So i'm required to change the "possible" `null` case into a valid number, hence `?? 0`

src/utils/transaction.ts
r4mmer: The integration tests should cover this.
But the question is because for the purpose of this PR this field is not used by the history processing method.
So it should not change the behavior when scanning for utxos, but i don't know if it changes anything related to nano.

If it does we should change this bit to make sure the data will be exact.

src/new/sendTransaction.ts
```diff
@@ -100,13 +105,15 @@ export default class SendTransaction extends EventEmitter {
    * @param {IStorage|null} [options.network=null] Network object
    */
   constructor({
+    wallet = null,
```
pedroferreira1, line 108: Missing docstring for the new parameter

pedroferreira1 (dismissed):

pedroferreira1 (dismissed):

src/new/sendTransaction.ts
```diff
@@ -92,6 +95,7 @@ export default class SendTransaction extends EventEmitter {
 
   /**
    *
+   * @param {HathorWallet} wallet Wallet instance
    * @param {IStorage} storage Storage object
```
tuliomir, line 99: suggestion(non-blocking): Make it clear that this storage will be discarded if the wallet parameter is present
```suggestion
   * @param {IStorage} storage Storage object, superseded by `wallet.storage` if wallet is present
```

src/utils/transaction.ts
```diff
@@ -685,8 +685,7 @@ const transaction = {
       histTx.nc_method = (tx as NanoContract).method;
       histTx.nc_args = (tx as NanoContract).args.map(a => a.toString('hex')).join('');
       histTx.nc_pubkey = (tx as NanoContract).pubkey.toString('hex');
-      // XXX: is the nc_blueprint_id required?
-      // histTx.nc_blueprint_id
+      // Cannot fetch histTx.the nc_blueprint_id with the current data
```
tuliomir, line 688: typo(non-blocking): Is the `the` word correct here?
```suggestion
      // Cannot fetch histTx.nc_blueprint_id with the current data
```

tuliomir (approved):

pedroferreira1 (approved):

PR #816, opened 2025-02-17
feat: void txs index

r4mmer:
### Acceptance Criteria

- Create a voided txs index

### Security Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -12,6 +12,7 @@ on:
       - release
       - release-candidate
       - master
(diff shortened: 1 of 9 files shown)


.github/workflows/integration-test.yml
```diff
@@ -12,6 +12,7 @@ on:
       - release
       - release-candidate
       - master
+      - test/void-tx-utxos
```
r4mmer, line 15: These are to allow the CI to run.
Will be rolledback after PR is approved

PR #818, opened 2025-02-18
fix: utxo selection for getUtxosForAmount

r4mmer:
### Motivation

`selectUtxos` util in `src/utils/transaction.ts` requires the utxo array to be sorted for the selection to be optimal.
The `getUtxosForAmount` method of the wallet facade was not following this and in some cases the utxos chosen by this method were not the best selection possible, maybe even failing some transactions due to going over the input limit.

To fix this we need to get the sorted utxo list.

### Acceptance Criteria

- on `getUtxosForAmount` use the option to retrieve utxos sorted by amount.
- test a case where utxo selection would choose a list of utxos instead of a single utxo that could meet the requirements.

### Security Checklist

### DIFF
--- a/__tests__/new/hathorwallet.test.ts
+++ b/__tests__/new/hathorwallet.test.ts
@@ -22,6 +22,7 @@ import Queue from '../../src/models/queue';
 import { WalletType } from '../../src/types';
 import txApi from '../../src/api/txApi';
 import * as addressUtils from '../../src/utils/address';
+import walletUtils from '../../src/utils/wallet';
 import versionApi from '../../src/api/version';
 import { decryptData, verifyMessage } from '../../src/utils/crypto';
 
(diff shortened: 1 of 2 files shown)


pedroferreira1 (approved):

__tests__/new/hathorwallet.test.ts
```diff
@@ -1371,3 +1372,43 @@ test('setExternalTxSigningMethod', async () => {
   hwallet.setExternalTxSigningMethod(async () => {});
   expect(hwallet.isSignedExternally).toBe(true);
 });
+
+test('getUtxosForAmount - should always get the best utxos', async () => {
+  const seed =
+    'upon tennis increase embark dismiss diamond monitor face magnet jungle scout salute rural master shoulder cry juice jeans radar present close meat antenna mind';
```
tuliomir, line 1378: question(non-blocking): Should we have this seed extracted to a constant somewhere?

tuliomir (approved):

PR #819, opened 2025-02-26
tests: use rocksdb in the integration tests

pedroferreira1:
### Motivation

The next full node version will not support memory storage anymore and one of the nano contract indexes already doesn't support it. The main problem with using rocksdb is that the full node service takes more time to become healthy (with status port ready to receive requests), so the tx mining service and the test itself were breaking. 

Then I created a healthcheck to validate if the node is already receiving http requests. I tried using `curl` or `wget` but the docker container doesn't have them installed, that's why it's using a python script to validate.

### Acceptance Criteria
- Use rocksdb in the fullnode that runs in the integration tests

### Security Checklist

### DIFF
--- a/__tests__/integration/configuration/docker-compose.yml
+++ b/__tests__/integration/configuration/docker-compose.yml
@@ -30,7 +30,7 @@ services:
     networks:
       - hathor-privnet
     healthcheck:
-      test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:8080/v1a/status')"]
+      test: ["CMD", "python", "-c", "import urllib.request; import json; r = urllib.request.urlopen('http://localhost:8080/v1a/status'); body = json.loads(r.read()); assert body['server']['state'] == 'READY'"]
       interval: 5s
       timeout: 10s
       retries: 10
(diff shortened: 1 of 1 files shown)


tuliomir (dismissed):

__tests__/integration/configuration/docker-compose.yml
```diff
@@ -29,12 +29,18 @@ services:
         target: /privnet/conf
     networks:
       - hathor-privnet
+    healthcheck:
+      test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:8080/v1a/status')"]
```
r4mmer, line 33: suggestion(non-blocking): Maybe we should check the server state inside the status.

```suggestion
      test: ["CMD", "python", "-c", "import urllib.request; import json; r = urllib.request.urlopen('http://localhost:8080/v1a/status'); body = json.loads(r.read()); assert body['server']['state'] == 'READY'"]
```

This would make the healthcheck fail if the status is not a JSON object, invalid or if the fullnode reports a state other than `READY`.
What do you think?

r4mmer (dismissed):

__tests__/integration/configuration/docker-compose.yml
pedroferreira1: Makes sense, even better. Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/819/commits/de702e9590617113d676699850d1f4dcb7e6de85

r4mmer (approved):

tuliomir (approved):

PR #825, opened 2025-03-13
tests: improve how we wait for a tx to be confirmed

pedroferreira1:
### Motivation

The test wants to wait until the nano contract tx has its method executed and I was using `waitNextBlock`. This method was causing some intermittent errors because it's an async call. The best way to guarantee that the nano method will have already been executed is to use `waitTxConfirmed`, then the tx will have a `first_block`.

### Acceptance Criteria
- Use `waitTxConfirmed` to guarantee a tx has first block.

### Security Checklist

### DIFF
--- a/__tests__/integration/nanocontracts/bet.test.ts
+++ b/__tests__/integration/nanocontracts/bet.test.ts
@@ -478,7 +478,7 @@ describe('full cycle of bet nano contract', () => {
 
     jest.spyOn(wallet.storage, 'processHistory');
     expect(wallet.storage.processHistory.mock.calls.length).toBe(0);
-    await waitNextBlock(wallet.storage);
+    await waitTxConfirmed(wallet, txWithdrawal2.hash);
     const txWithdrawal2Data = await wallet.getFullTxById(txWithdrawal2.hash);
 
     // The tx became voided after the block because of the nano execution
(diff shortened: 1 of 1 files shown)


r4mmer (approved):

__tests__/integration/nanocontracts/bet.test.ts
```diff
@@ -478,7 +478,7 @@ describe('full cycle of bet nano contract', () => {
 
     jest.spyOn(wallet.storage, 'processHistory');
     expect(wallet.storage.processHistory.mock.calls.length).toBe(0);
-    await waitNextBlock(wallet.storage);
```
tuliomir, line 481: chore: We should just remove the `waitNextBlock` import from the file, as it's not being used anymore.

tuliomir (approved):

__tests__/integration/nanocontracts/bet.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/825/commits/a4827652e34169f0a1ecd2b7d284cde02dbb8bf3

tuliomir (approved):

r4mmer (approved):

PR #830, opened 2025-03-20
feat: parsing big int values that come from http requests

andreabadesso:
### Acceptance Criteria
- We should be able to parse numbers from wallet-service requests as big ints
- We should use zod to validate all responses from the wallet-service lambdas
- We should fix all current types that are different from the actual responses 

### Security Checklist


src/wallet/types.ts
```diff
@@ -124,8 +124,6 @@ export interface TxProposalCreateResponseData {
   success: boolean;
   txProposalId: string; // Id of the tx proposal
   inputs: TxProposalInputs[]; // Inputs data of the tx proposal
-  outputs: TxProposalOutputs[]; // Outputs data of the tx proposal
-  tokens: string[];
```
andreabadesso, line 128: These are not coming from the wallet-service:

https://github.com/HathorNetwork/hathor-wallet-service/blob/239b5a769b04a09714ca7fa673e99e064e91f471/packages/wallet-service/src/api/txProposalCreate.ts#L138-L145

src/wallet/types.ts
```diff
@@ -449,14 +454,16 @@ export interface TxByIdTokenData {
   voided: boolean;
   height?: number | null;
   weight: number;
-  balance: Balance;
```
andreabadesso, line 452: Type is wrong, it never came as a Balance object from the wallet-service:

https://github.com/HathorNetwork/hathor-wallet-service/blob/239b5a769b04a09714ca7fa673e99e064e91f471/packages/wallet-service/src/db/index.ts#L2963

src/wallet/types.ts
```diff
@@ -449,14 +454,16 @@ export interface TxByIdTokenData {
...
+  balance: bigint;
   tokenId: string;
-  walletId: string;
```
andreabadesso, line 454: Same, no walletId coming from the wallet-service lambdas

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,513 @@
...
+const baseResponseSchema = z.object({
+  success: z.boolean(),
+});
+
+/**
+ * Schema for individual address information.
+ * Represents a single address in the wallet with its derivation index and transaction count.
+ */
+export const getAddressesObjectSchema = z.object({
+  address: z.string(), // Address in base58
```
r4mmer, line 24: We can use a more specific address schema ([reference](https://github.com/HathorNetwork/hathor-wallet-lib/blob/fix/big-int-http-requests/src/template/transaction/instructions.ts#L71))

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,513 @@
...
+export const addressesResponseSchema = baseResponseSchema.extend({
+  addresses: z.array(getAddressesObjectSchema),
+});
+
+/**
+ * Response schema for checking if addresses belong to the wallet.
+ * Maps addresses to boolean values indicating ownership.
+ */
+export const checkAddressesMineResponseSchema = baseResponseSchema.extend({
+  addresses: z.record(z.boolean()), // WalletAddressMap
```
r4mmer, line 41: We should make sure the keys are addresses right?

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,513 @@
...
+});
+
+/**
+ * Schema for address information used in new address generation.
+ */
+export const addressInfoObjectSchema = z
+  .object({
+    address: z.string(),
+    index: z.number(),
+    addressPath: z.string(),
```
r4mmer, line 51:
```suggestion
    addressPath: z.string().regex(/^m(\/\d+\'?){1,5}$/),
```

Makes sense?

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,513 @@
...
+ * Response schema for token details.
+ * Contains information about a token's name, symbol, total supply, and authorities.
+ */
+export const tokenDetailsResponseSchema = baseResponseSchema.extend({
+  details: z.object({
+    tokenInfo: z.object({
+      id: z.string(),
+      name: z.string(),
+      symbol: z.string(),
+    }),
```
r4mmer, line 73: Use the tokenInfoSchema below for this?

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,513 @@
...
+      melt: z.boolean(),
+    }),
+  }),
+});
+
+/**
+ * Schema for token information.
+ */
+export const tokenInfoSchema = z.object({
+  id: z.string(),
```
r4mmer, line 87: Check for 64 length and hex characters?
I included the native token case as well.

```suggestion
  id: z.string().regex(/^[a-fA-F0-9]{64}$|^00$/),
```

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,513 @@
...
+  network: z.string(),
+  minWeight: z.number(),
+  minTxWeight: z.number(),
+  minTxWeightCoefficient: z.number(),
+  minTxWeightK: z.number(),
+  tokenDepositPercentage: z.number(),
+  rewardSpendMinBlocks: z.number(),
+  maxNumberInputs: z.number(),
+  maxNumberOutputs: z.number(),
+});
```
r4mmer, line 191: Currently any new keys will be ignored, but we can pass them into the final data without validation (so that new keys from the fullnode do not break the api)

```suggestion
}).passthrough();
```

src/wallet/api/schemas/walletApi.ts
andreabadesso: Yes we should, done!

src/wallet/api/schemas/walletApi.ts
andreabadesso: Done! Thanks!

src/wallet/api/schemas/walletApi.ts
andreabadesso: Great, thanks!

src/wallet/api/schemas/walletApi.ts
andreabadesso: Done!

Also added a test to test for new attributes

src/wallet/api/schemas/walletApi.ts
andreabadesso: Done, thanks

src/wallet/types.ts
```diff
@@ -378,14 +376,17 @@ export interface DecodedOutput {
 
 export interface TxOutput {
   value: OutputValueType;
-  script: string;
+  script: {
+    type: 'Buffer';
+    data: number[];
+  };
```
r4mmer, line 382: If the wallet-service is sending this "Buffer" type maybe we should change how the wallet-service is serializing the script

src/wallet/connection.ts
```diff
@@ -72,8 +74,14 @@ export default class WalletServiceConnection extends BaseConnection {
 
     this.websocket = new WalletServiceWebSocket(wsOptions);
     this.websocket.on('is_online', online => this.onConnectionChange(online));
-    this.websocket.on('new-tx', payload => this.emit('new-tx', payload.data as WsTransaction));
-    this.websocket.on('update-tx', payload => this.emit('update-tx', payload.data));
+    this.websocket.on('new-tx', payload => {
+      const validatedTx = parseSchema(payload.data, wsTransactionSchema);
```
r4mmer, line 78: Maybe we should put a try..catch around this or use `safeParse` directly.
Or do you think its better to have an unhandled error than a silent fail (only log an error) on new-tx?

andreabadesso: You are right, we shouldn't throw an error here

Fixed

src/wallet/types.ts
andreabadesso: Hey, we actually receive a simpler `TxInput` and `TxOutput` in websocket transactions, e.g.:

```json
{
  "height": 0,
  "inputs": [
    {
      "address": "H79QdDV5i6KuW3dNJEahkUNe8a9axtHuCu",
      "timelock": null,
      "type": "P2PKH"
    }
  ],
  "nonce": 16746194,
  "outputs": [
    {
      "address": "HDY5QHiE33goHcLxFyJmgxTErwgGHDrLdy",
      "timelock": null,
      "type": "P2PKH"
    }
  ],
  "parents": [],
  "signal_bits": 0,
  "timestamp": 1743779693,
  "token_name": null,
  "token_symbol": null,
  "tx_id": "000009532ec33ba944c5d23b8dc642063539055ca31b4f6d4f534d33651ce8a3",
  "version": 1,
  "voided": false,
  "weight": 16.818364568681208
}
```

Fixed the type and the schema validation

src/wallet/types.ts
```diff
@@ -395,7 +396,10 @@ export interface TxInput {
   value: OutputValueType;
   // eslint-disable-next-line camelcase
   token_data: number;
-  script: string;
+  script: {
+    type: 'Buffer';
+    data: number[];
+  };
```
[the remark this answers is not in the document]
andreabadesso: Same, can you please check?

src/wallet/wallet.ts
```diff
@@ -539,7 +539,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     let shouldGetNewAddresses = false;
 
     for (const output of outputs) {
-      if (this.newAddresses.find(newAddress => newAddress.address === output.decoded.address)) {
```
andreabadesso, line 542: This was a bug, we were not generating new addresses when a new tx came from the websocket.

r4mmer (approved):

r4mmer (approved):

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,508 @@
...
+    version: z.string(),
+    network: z.string(),
+    minWeight: z.number(),
+    minTxWeight: z.number(),
+    minTxWeightCoefficient: z.number(),
+    minTxWeightK: z.number(),
+    tokenDepositPercentage: z.number(),
+    rewardSpendMinBlocks: z.number(),
+    maxNumberInputs: z.number(),
+    maxNumberOutputs: z.number(),
```
pedroferreira1, line 208: Does it make sense to add as optional fields the genesis hashes and decimal places that will be available in the next full node version?

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,508 @@
...
+  decoded: z.object({
+    type: z.string(),
+    address: z.string(),
+    timelock: z.number().nullable().optional(),
+    value: bigIntCoercibleSchema,
+    token_data: z.number(),
+  }),
+  tx_id: z.string(),
+  index: z.number(),
+  token: z.string().nullable().optional(),
```
pedroferreira1, line 229: Can it be `null`? We have a token_id defined as `z.string().regex(/^[a-fA-F0-9]{64}$|^00$/)`. I think we could export this as another type and use here.

(the same below in the output)

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,508 @@
...
+  token_data: z.number(),
+  script: z.string(),
+  decoded: z.object({
+    type: z.string(),
+    address: z.string(),
+    timelock: z.number().nullable().optional(),
+    value: bigIntCoercibleSchema,
+    token_data: z.number(),
+  }),
+  address: z.string(),
```
pedroferreira1, line 248: Is it required even for data outputs? In that case, should you use the `AddressSchema`?

src/wallet/api/schemas/walletApi.ts
```diff
@@ -0,0 +1,508 @@
...
+});
+
+/**
+ * Response schema for transaction history.
+ * Contains an array of transaction information.
+ */
+export const historyResponseSchema = baseResponseSchema.extend({
+  history: z.array(
+    z.object({
+      txId: z.string(),
```
pedroferreira1, line 358: suggestion(non-blocking): we have many places where we use a tx hash. We could create a regex schema for it, instead of just string.

src/wallet/connection.ts
```diff
@@ -72,8 +77,26 @@ export default class WalletServiceConnection extends BaseConnection {
...
-    this.websocket.on('update-tx', payload => this.emit('update-tx', payload.data));
+    this.websocket.on('new-tx', payload => {
+      try {
+        const validatedTx = parseSchema(payload.data, wsTransactionSchema);
+        this.emit('new-tx', validatedTx as WsTransaction);
+      } catch (e) {
+        // parseSchema already logs the validation error, so no need to log it
+        // again here.
+        logger.error('Received a new transaction but schema validation failed.');
+      }
```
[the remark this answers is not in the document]
andreabadesso: Good idea, thanks. Done

src/wallet/api/schemas/walletApi.ts
andreabadesso: I don't see why not!

Done

src/wallet/api/schemas/walletApi.ts
andreabadesso: It actually comes as `null` when it's HTR, e.g.:

```json
{
  "type": "EVENT",
  "peer_id": "df392f0cd2eaeb2a4c76b0c6e8d2f47d13cb8dfd03213f1fac02a084593b54a9",
  "network": "mainnet",
  "event": {
    "id": 73,
    "timestamp": 1708011606.9498572,
    "type": "NEW_VERTEX_ACCEPTED",
    "data": {
      "hash": "0000000004fc870e3c08b0cfa5496633b8c739ef02c380c5c3c8d9be8f897ed6",
      "nonce": 94429542,
      "timestamp": 1578077989,
      "signal_bits": null,
      "version": 0,
      "weight": 34.55483669195596,
      "inputs": [],
      "outputs": [
        {
          "value": 6400,
          "token_data": 0,
          "script": "dqkUym0SWcUWwA1Du+i9fiZl4MbEfwWIrA==",
          "decoded": {
            "type": "P2PKH",
            "address": "HQyTYXYS9Z5gARuhsHqav86bV3BW6x9rBe",
            "timelock": null
          }
        }
      ],
      "parents": [
        "0000000008e7e27868a2b4b8f3811075893dc7c49359d9457d6335044b3401c0",
        "0002d4d2a15def7604688e1878ab681142a7b155cbe52a6b4e031250ae96db0a",
        "0002ad8d1519daaddc8e1a37b14aac0b045129c01832281fb1c02d873c7abbf9"
      ],
      "tokens": [],
      "token_name": null,
      "token_symbol": null,
      "metadata": {
        "hash": "0000000004fc870e3c08b0cfa5496633b8c739ef02c380c5c3c8d9be8f897ed6",
        "spent_outputs": [
          {
            "index": 0,
            "tx_ids": [
              "00000bf244d0826c7ec90d20187e7165acbb30101d8bc76fc1fba9a5dfc8891b"
            ]
          }
        ],
        "conflict_with": [],
        "voided_by": [],
        "received_by": [],
        "children": [
          "0000000009d17098052105ae9abab3c7d2178ff2ed7b750dfb60572bc1f63e5a"
        ],
        "twins": [],
        "accumulated_weight": 34.55483669195596,
        "score": 42.48518641026811,
        "first_block": null,
        "height": 70,
        "validation": "full"
      },
      "aux_pow": null
    },
    "group_id": null
  },
  "latest_event_id": 10479176,
  "stream_id": "95402327-8dde-4511-8bba-75a687cad6ea"
}
```

Changed all tokenIds to use the correct schema too

src/wallet/api/schemas/walletApi.ts
andreabadesso: It's not. Fixed and refactored all types to use `AddressSchema`

src/wallet/api/schemas/walletApi.ts
andreabadesso: Done!

src/wallet/connection.ts
```diff
@@ -59,11 +59,25 @@ export default class WalletServiceConnection extends BaseConnection {
...
+  onWsMessage(type: string, payload) {
     const logger = getDefaultLogger();
 
+    try {
+      const validatedTx = parseSchema(payload.data, wsTransactionSchema);
+      this.emit(type, validatedTx as WsTransaction);
+    } catch (e) {
+      // parseSchema already logs the validation error, so no need to log it
+      // again here.
+      logger.error('Received a new transaction update but schema validation failed.');
```
pedroferreira1, line 73: Maybe add the type in the log here?

pedroferreira1 (approved):

src/wallet/connection.ts
andreabadesso: Done, thanks

pedroferreira1 (approved):

PR #843, opened 2025-04-17
feat: varint type for nano contract argument

r4mmer:
### Motivation

https://github.com/HathorNetwork/nano-hathor-core/pull/167

### Acceptance Criteria

- Add the new VarInt to the nano contract serializers

### Security Checklist

### DIFF
--- a/src/nano_contracts/serializer.ts
+++ b/src/nano_contracts/serializer.ts
@@ -12,6 +12,7 @@ import {
   signedIntToBytes,
   bigIntToBytes,
 } from '../utils/buffer';
+import { encodeSigned as leb128EncodeSigned } from '../utils/leb128';
 import { NanoContractArgumentType } from './types';
 import { OutputValueType } from '../types';
 
@@ -76,6 +77,8 @@ class Serializer {
         return this.fromFloat(value as number);
       case 'bool':
         return this.fromBool(value as boolean);
+      case 'VarInt':
+        return this.fromVarInt(value as bigint);
       default:
         throw new Error(`Invalid type. ${type}.`);
     }
@@ -250,6 +253,18 @@ class Serializer {
 
     ret.push(this.fromBytes(inputData));
     return Buffer.concat(ret);
+  }
+
+  /**
+   * Serialize a bigint value as a variable length integer.
+   * The serialization will use leb128.
+   *
+   * @param {bigint} value
+   *
+   * @memberof Serializer
+   */
+  fromVarInt(value: bigint): Buffer {
+    return leb128EncodeSigned(value);
   }
 }
 /* eslint-disable class-methods-use-this */
--- a/src/nano_contracts/types.ts
+++ b/src/nano_contracts/types.ts
@@ -11,7 +11,13 @@ export enum NanoContractActionType {
   WITHDRAWAL = 'withdrawal',
 }
 
-export type NanoContractArgumentApiInputType = string | number | bigint | OutputValueType | boolean | null;
+export type NanoContractArgumentApiInputType =
+  | string
+  | number
+  | bigint
+  | OutputValueType
+  | boolean
+  | null;
 export type NanoContractArgumentType = NanoContractArgumentApiInputType | Buffer;
 
 export interface NanoContractAction {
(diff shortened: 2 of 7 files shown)


src/utils/leb128.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (byte === undefined) {
+      throw new Error('Buffer is not valid leb128, cannot read from empty buffer');
+    }
+    result |= (byte & 0b0111_1111n) << shift;
+    shift += 7n;
+    // assert shift % 7n === 0
+    if (shift % 7n !== 0n)
+      throw new Error(`AssertionError: shift is ${shift} and is not divisible by 7`);
+
+    if (maxBytes !== null && shift / 7n > maxBytes) {
```
pedroferreira1, line 55: In the python implementation it's using `shift // 7`. Are there any reasons not to do the integer division?

src/utils/leb128.ts
```diff
@@ -0,0 +1,80 @@
...
+      (val === 0n && (byte & 0b0100_0000n) === 0n) ||
+      (val === -1n && (byte & 0b0100_0000n) !== 0n)
+    ) {
+      result.push(byte);
+      if (maxBytes !== null && result.length > maxBytes) {
+        throw new Error(`Cannot encode more than ${maxBytes} bytes`);
+      }
+      // Need to convert the bigint values of `result` into Number to use with Buffer.
+      // This is a safe operation since the value of the elements can only go up to 0xFF
+      return Buffer.from(result.map(b => Number(b)));
```
pedroferreira1, line 29: Why wouldn't this be a problem? Can't we have problems converting bigint values into Number?

src/nano_contracts/types.ts
```diff
@@ -11,7 +11,13 @@ export enum NanoContractActionType {
   WITHDRAWAL = 'withdrawal',
 }
 
-export type NanoContractArgumentApiInputType = string | number | OutputValueType | boolean | null;
+export type NanoContractArgumentApiInputType =
+  | string
+  | number
+  | bigint
```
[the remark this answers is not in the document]
r4mmer: No, but instead of having an implicit bigint through `OutputValueType` we now can have an explicit bigint for the `VarInt`

src/utils/leb128.ts
r4mmer: Yes, bigint values higher than [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) cannot be converted safely.
Since `MAX_SAFE_INTEGER` is `2^53 -1` it can happen with 8 byte integers (like our amount) but we can safely convert if the value cannot go up that high.

src/nano_contracts/serializer.ts
```diff
@@ -251,6 +254,18 @@ class Serializer {
...
+  /**
+   * Serialize a bigint value as a variable length integer.
+   * The serialization will use leb128.
+   *
+   * @param {bigint} value
+   *
+   * @memberof Serializer
+   */
+  fromVarInt(value: bigint): Buffer {
+    return leb128EncodeSigned(value);
```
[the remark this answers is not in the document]
r4mmer: Done

src/utils/leb128.ts
pedroferreira1: Ahh ok! I think you should just leave a comment in the code that this is already an integer division because it's a bigint.

pedroferreira1 (approved):

src/utils/leb128.ts
```diff
@@ -0,0 +1,80 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export interface Leb128DecodeResult {
+  value: bigint;
+  rest: Buffer;
+}
+
+export function encodeSigned(value: bigint | number, maxBytes: number | null = null): Buffer {
+  let val = BigInt(value);
```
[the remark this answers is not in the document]
r4mmer: Since this is a safe conversion I thought it would be better to make the util usable with both types.
Specially since we will commonly encode numbers as leb128 on nano contracts (e.g. length of args, number of args).

PR #845, opened 2025-04-24
feat: nc args new serialization

r4mmer:
### Motivation

All nano contract arguments would have 2 bytes inserted at the beginning for length when serialized as a tuple (from the call args).
This is slightly wasteful since some have fixed length and most can be expressed with 1 byte for length.

To address this the 2 bytes for length was removed and the "length" of the argument should be read on the arg parser.
The leb128 unsigned encoding should be used for types with variable length due to its flexibility and space usage.

### Acceptance Criteria

To avoid repetition, `leb128U_len` will be used instead of "leb128 unsigned integer with the value of the length of the following argument".

Update serialization of nano contract argument as follows:

- `str`
  - encode as: `leb128U_len` + `string encoded as utf-8`
- `bytes`, `BlueprintId`, `ContractId`, `TokenUid`, `TxOutputScript`, `VertexId`, `Address`
  - encode as: `leb128U_len` + `raw byte sequence`
- `int`, `Timestamp`
  - encode as: `32bit integer`
  - int serialization is bound to change for a variable size int as default, this PR will not address this at the moment since there is no implementation on the fullnode
- `Amount`
  - Serialization is unchanged
- `bool`
  - Serialization is unchanged
- `VarInt`
  - Serialization is unchanged (already uses leb128 signed)
  
 
> [!IMPORTANT]  
> Float nano contract argument type was removed, so we are removing support on the lib as well.

Should also update how we parse these arguments following the same rules.

### Security Checklist


src/nano_contracts/deserializer.ts
```diff
@@ -24,45 +31,58 @@ class Deserializer {
...
-      return this.toOptional(value, optionalType);
-    }
-
-    if (type.startsWith('SignedData[')) {
-      return this.toSigned(value, type);
+  deserializeFromType(
+    buf: Buffer,
+    type: string
+  ): DeserializeResult<NanoContractArgumentType | null> {
+    const isContainerType = getContainerType(type) !== null;
```
pedroferreira1, line 44: This method calls `getContainerInternalType`, which has the code below

```
  const match = type.match(/^(.*?)\[(.*)\]/);
  const containerType = match ? match[1] : null;
  const internalType = match ? match[2] : null;
  if ((!internalType) || (!containerType)) {
    throw new Error('Unable to extract type');
  }
```

Won't this throw the error when `type` is `str`, for example?

src/nano_contracts/deserializer.ts
```diff
@@ -73,91 +93,129 @@ class Deserializer {
...
+    }
+    // If lengthBN is lower than 64 KiB than its safe to convert to Number
+    const length = Number(lengthBN);
+    if (rest.length < length) {
+      throw new Error('Do not have enough bytes to read the expected length');
+    }
+    return {
+      value: rest.subarray(0, length),
+      bytesRead: length + bytesReadForLength,
+    };
```
pedroferreira1, line 152: This method is the same as `toString`, except for the `toString('utf-8')`. Should we use a single util method for it?

src/nano_contracts/deserializer.ts
```diff
@@ -192,53 +257,85 @@ class Deserializer {
...
-    // Get signed data type inside []
-    const match = type.match(/\[(.*?)\]/);
-    const valueType = match ? match[1] : null;
-    if (!valueType) {
+  toSigned(signedData: Buffer, type: string): DeserializeResult<string> {
+    const [containerType, internalType] = getContainerInternalType(type);
+    if (containerType !== 'SignedData') {
+      throw new Error('Type is not SignedData');
+    }
+    if (!internalType) {
```
pedroferreira1, line 271: This will throw before, when calling `getContainerInternalType`. But I don't mind keeping this throw here as well.

src/nano_contracts/deserializer.ts
```diff
@@ -24,45 +31,58 @@ class Deserializer {
...
-        return this.toAddress(value);
+        return this.toAddress(buf);
       case 'int':
       case 'Timestamp':
-        return this.toInt(value);
+        return this.toInt(buf);
       case 'Amount':
-        return this.toAmount(value);
-      case 'float':
-        return this.toFloat(value);
```
tuliomir, line 61: chore: We should inform in the PR description that the `float` type is being removed: this is a breaking change, even though it only affects the alpha nano testnet.

src/nano_contracts/deserializer.ts
```diff
@@ -73,91 +93,129 @@ class Deserializer {
...
-   * @param {value} Value to deserialize
+   * @param {Buffer} buf Value to deserialize
    *
    * @memberof Deserializer
    * @inner
    */
-  toString(value: Buffer): string {
-    return value.toString('utf8');
+  toString(buf: Buffer): DeserializeResult<string> {
+    // INFO: maxBytes is set to 3 becuase the max allowed length in bytes for a string is
```
tuliomir, line 102: chore: typo
```suggestion
    // INFO: maxBytes is set to 3 because the max allowed length in bytes for a string is
```

src/nano_contracts/deserializer.ts
```diff
@@ -192,53 +257,85 @@ class Deserializer {
...
+      throw new Error('Address should be 25 bytes long');
+    }
     // First we get the 20 bytes of the address without the version byte and checksum
-    const addressBytes = value.slice(1, 21);
+    const addressBytes = buf.subarray(2, 22);
     const address = helpersUtils.encodeAddress(addressBytes, this.network);
+    address.validateAddress();
     const decoded = address.decode();
-    if (decoded[0] !== value[0]) {
+    if (decoded[0] !== buf[1]) {
```
tuliomir, line 324: question: Why do we compare bytes at different indexes in the decoded and in the original Buffer? We should have a comment explaining this.

Also, the error message does not show the same index for the `buf` Buffer.

src/nano_contracts/deserializer.ts
```diff
@@ -192,53 +257,85 @@ class Deserializer {
...
+      throw new Error(
+        `Asked to deserialize an address with version byte ${buf[0]} but the network from the deserializer object has version byte ${decoded[0]}.`
+      );
+    }
+    if (decoded.subarray(21, 25).toString('hex') !== buf.subarray(22, 26).toString('hex')) {
+      // Checksum value generated does not match value from fullnode
       throw new Error(
-        `Asked to deserialize an address with version byte ${value[0]} but the network from the deserializer object has version byte ${decoded[0]}.`
+        `When parsing and Address(${address.base58}) we calculated checksum(${decoded.subarray(21, 25).toString('hex')}) but it does not match the checksum it came with ${buf.subarray(22, 26).toString('hex')}.`
```
tuliomir, line 332: suggestion(non-blocking): Create variables for storing both checksums, so that it's harder to accidentally add an issue in the error message like the one that happened above.

src/nano_contracts/deserializer.ts
tuliomir: I believe we should have tests that ensure this. I couldn't find any tests for `Tuple` or `RawSignedData`, and CodeCov seems to have not found this lack of tests too. :thinking:

src/nano_contracts/deserializer.ts
r4mmer: Done

src/nano_contracts/utils.ts
```diff
@@ -22,9 +22,52 @@ import { NanoContractTransactionError, OracleParseError, WalletFromXPubGuard } f
...
+      throw new Error('Not a ContainerType');
+  }
+}
+
+export function getContainerType(type: string): NanoContractArgumentContainerType | null {
+  try {
+    const [containerType, _internalType] = getContainerInternalType(type);
+    return containerType;
+  } catch (err: unknown) {
+    if (err instanceof Error && err.message === 'Not a ContainerType') {
```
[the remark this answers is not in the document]
r4mmer: This is an issue addresses on the other PR, so I think we should make any necessary changes there so it will be merged back here later.

src/nano_contracts/deserializer.ts
r4mmer: Done

src/nano_contracts/deserializer.ts
r4mmer: Done

src/nano_contracts/deserializer.ts
r4mmer: Done

src/nano_contracts/parser.ts
```diff
@@ -84,18 +84,31 @@ class NanoContractTransactionParser {
...
+        size = parseResult.bytesRead;
+      } catch (err: unknown) {
+        console.error(err);
+        throw new NanoContractTransactionParseError(`Failed to deserialize argument ${arg.type}.`);
       }
       parsedArgs.push({ ...arg, parsed });
-      argsBuffer = argsBuffer.slice(size);
+      argsBuffer = argsBuffer.subarray(size);
     }
+    // XXX: argsBuffer should be empty since we parsed all arguments
```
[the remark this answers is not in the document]
r4mmer: Done

src/nano_contracts/deserializer.ts
```diff
@@ -24,45 +31,58 @@ class Deserializer {
    * We receive these types from the full node, so we
    * use the python syntax
    *
-   * @param {value} Value to deserialize
-   * @param {type} Type of the value to be deserialized
+   * @param {Buffer} buf Value to deserialize
+   * @param {string} type Type of the value to be deserialized
```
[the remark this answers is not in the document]
r4mmer: Done

src/nano_contracts/deserializer.ts
r4mmer: This method was fixed on the other PR, maybe we should address any necessary changes there.

src/nano_contracts/serializer.ts
```diff
@@ -5,22 +5,24 @@
...
-import { encodeSigned as leb128EncodeSigned } from '../utils/leb128';
+import Address from '../models/address';
+import Network from '../models/network';
+import { hexToBuffer, intToBytes, signedIntToBytes, bigIntToBytes } from '../utils/buffer';
 import { NanoContractArgumentType } from './types';
 import { OutputValueType } from '../types';
+import leb128Util from '../utils/leb128';
 
 // Number of bytes used to serialize the size of the value
 const SERIALIZATION_SIZE_LEN = 2;
```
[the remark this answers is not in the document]
r4mmer: This was addresses on the other PR.

src/nano_contracts/deserializer.ts
```diff
@@ -316,20 +307,27 @@ class Deserializer {
...
-        `Asked to deserialize an address with version byte ${buf[0]} but the network from the deserializer object has version byte ${decoded[0]}.`
+        `Asked to deserialize an address with version byte ${addressBytes[0]} but the network from the deserializer object has version byte ${decoded[0]}.`
       );
     }
-    if (decoded.subarray(21, 25).toString('hex') !== buf.subarray(22, 26).toString('hex')) {
+    // Check checksum bytes
+    if (!decoded.subarray(21, 25).equals(addressBytes.subarray(21, 25))) {
       // Checksum value generated does not match value from fullnode
+      const calcChecksum = decoded.subarray(21, 25).toString('hex');
+      const recvChecksum = addressBytes.subarray(21, 25).toString('hex');
```
tuliomir, line 328: suggestion(non-blocking): The `calcChecksum` and `recvChecksum` could be declared before the `if`, so that they're used on both the conditional and the error message.

tuliomir (approved):

pedroferreira1 (dismissed):

pedroferreira1 (dismissed):

tuliomir (dismissed):

pedroferreira1 (approved):

PR #854, opened 2025-05-02
fix: SendTransaction should enqueue call to onNewTx

r4mmer:
### Motivation

When running a localnet with streaming sync, users wallet's would fail after sending a transaction due to having requests to start multiple streams at the same time.

### Acceptance Criteria

All calls to `onNewTx` should be enqueued so 2 calls to `onNewTx` cannot run at the same time.

### Security Checklist

### DIFF
--- a/src/new/wallet.js
+++ b/src/new/wallet.js
@@ -675,7 +675,7 @@ class HathorWallet extends EventEmitter {
         // So we will enqueue this message to be processed later
         this.wsTxQueue.enqueue(wsData);
       } else {
-        this.newTxPromise = this.newTxPromise.then(() => this.onNewTx(wsData));
+        this.enqueueOnNewTx(wsData);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/new/wallet.js
```diff
@@ -1330,6 +1330,14 @@ class HathorWallet extends EventEmitter {
     this.emit('state', state);
   }
 
+  /**
+   * Enqueue the call for onNewTx with the given data.
+   * @param {{ history: import('../types').IHistoryTx }} wsData
+   */
+  enqueueOnNewTx(wsData) {
+    this.newTxPromise = this.newTxPromise.then(() => this.onNewTx(wsData));
```
tuliomir, line 1338: question: For most cases I believe this should be perfectly fine. However, if I understand correctly, every new transaction will preserve the last promise and add a new one.

Would this be a significant memory leak for wallets with a large number of transactions?

tuliomir (approved):

PR #860, opened 2025-05-08
feat: add support for creating a create token transaction with nano header

pedroferreira1:
### Acceptance Criteria

- Add method in the wallet facade to create a Create Token Transaction with nano header.
- Add new flag to the mint/create token utils methods to skip deposit fee management (so the contract can handle it)
- Refactor NanoContractBuilder build method in different methods (verify, serializeArgs, buildInputsOutputs, buildTransaction)
- Add support in NanoContractBuilder to create a Transaction or CreateTokenTransaction with nano header and with possibility for the contract's withdrawal action to pay for the token creation deposit fee.

### DIFF
--- a/src/nano_contracts/builder.ts
+++ b/src/nano_contracts/builder.ts
@@ -5,13 +5,17 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { concat, get } from 'lodash';
-import Output from '../models/output';
-import Input from '../models/input';
+import { concat, get, uniq } from 'lodash';
 import Transaction from '../models/transaction';
-import { createOutputScriptFromAddress } from '../utils/address';
+import CreateTokenTransaction from '../models/create_token_transaction';
+import { getAddressType } from '../utils/address';
 import tokensUtils from '../utils/tokens';
-import { NATIVE_TOKEN_UID, NANO_CONTRACTS_INITIALIZE_METHOD } from '../constants';
+import transactionUtils from '../utils/transaction';
+import {
+  DEFAULT_TX_VERSION,
+  NATIVE_TOKEN_UID,
+  NANO_CONTRACTS_INITIALIZE_METHOD,
+} from '../constants';
 import Serializer from './serializer';
 import HathorWallet from '../new/wallet';
 import { NanoContractTransactionError } from '../errors';
@@ -23,8 +27,10 @@ import {
   MethodArgInfo,
   NanoContractArgumentApiInputType,
   NanoContractArgumentType,
+  NanoContractBuilderCreateTokenOptions,
+  NanoContractVertexType,
 } from './types';
-import { ITokenData } from '../types';
+import { IDataInput, IDataOutput, ITokenData } from '../types';
 import ncApi from '../api/nano';
 import { validateAndUpdateBlueprintMethodArgs } from './utils';
 import NanoContractHeader from './header';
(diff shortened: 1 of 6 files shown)


src/nano_contracts/builder.ts
```diff
@@ -190,28 +226,33 @@ class NanoContractTransactionBuilder {
       utxoOptions.filter_address = action.address;
     }
     const utxosData = await this.wallet.getUtxosForAmount(action.amount, utxoOptions);
-    const inputs: Input[] = [];
+    // XXX What if I don't have enough funds? Validate it!
```
pedroferreira1, line 229: This will be done in another PR

src/nano_contracts/builder.ts
```diff
@@ -324,47 +441,104 @@ class NanoContractTransactionBuilder {
...
+      );
+    }
 
-      for (const [index, arg] of methodArgs.entries()) {
-        const serialized = serializer.serializeFromType(this.args[index], arg.type);
-        serializedArgs.push(serialized);
-      }
+    if (this.createTokenOptions === null) {
+      throw new NanoContractTransactionError(
+        "Create token options can't be null when creating a create token transaction."
```
andreabadesso, line 485:
```suggestion
        'Create token options can't be null when creating a create token transaction.'
```

src/nano_contracts/builder.ts
```diff
@@ -190,28 +226,33 @@ class NanoContractTransactionBuilder {
...
-      // This will throw AddressError in case the adress is invalid
-      // this handles p2pkh and p2sh scripts
-      const outputScript = createOutputScriptFromAddress(changeAddressStr, network);
-      const tokenIndex =
-        action.token === NATIVE_TOKEN_UID
-          ? 0
-          : tokens.findIndex(token => token === action.token) + 1;
-      const outputObj = new Output(utxosData.changeAmount, outputScript, {
-        tokenData: tokenIndex,
+      outputs.push({
```
andreabadesso, line 247: _Comment unrelated to line number_

Will this still throw AddressError?

src/nano_contracts/builder.ts
```diff
@@ -324,47 +441,104 @@ class NanoContractTransactionBuilder {
...
+    if (this.createTokenOptions === null) {
+      throw new NanoContractTransactionError(
+        "Create token options can't be null when creating a create token transaction."
+      );
     }
 
+    // It's a token creation transaction
+    // then we get the token creation data from the utils method
+    // and concatenate the nano actions inputs/outputs/tokens
+    const data = await tokensUtils.prepareCreateTokenData(
```
r4mmer, line 492: Maybe we should have this inside an if and not the "default".
If the vertex type is not `TRANSACTION` or `CREATE_TOKEN_TRANSACTION` we throw an error.
What do you think?

src/nano_contracts/builder.ts
```diff
@@ -324,47 +441,104 @@ class NanoContractTransactionBuilder {
...
+   * @throws {NanoContractTransactionError} In case the arguments to build the tx are invalid
+   *
+   * @memberof NanoContractTransactionBuilder
+   * @inner
+   */
+  async build(): Promise<Transaction> {
+    await this.verify();
+
+    // Transform actions into inputs and outputs
+    const [inputs, outputs, tokens] = await this.buildInputsOutputs();
```
r4mmer, line 529: Maybe this method could also build the `actions` array, so the `if (this.actions)` below would not be needed

pedroferreira1: I can't do this because I need the full array of tokens to get the correct `tokenIndex`, which may change in the `buildTransaction` method.

Actually this an interesting test, having an action for a token that is not HTR and a new token being created in the same tx.

src/nano_contracts/builder.ts
pedroferreira1: Makes sense! https://github.com/HathorNetwork/hathor-wallet-lib/pull/860/commits/c98a7eb481d8ea3f7dff158a0b65d85e307ecf95

src/nano_contracts/builder.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/860/commits/c98a7eb481d8ea3f7dff158a0b65d85e307ecf95

andreabadesso (approved):

andreabadesso (approved):

r4mmer (approved):

PR #862, opened 2025-05-12
feat: new SignedData and method argument parser

r4mmer:
#### SignedData and RawSignedData
The new `SignedData` is meant to improve protection, the data should be `<signature, value, type>`.

In the original `SignedData` (now called `RawSignedData`) the oracle signs the value (serialized as the type) to generate the signature.
For the new `SignedData` the oracle signs the tuple `[ncId, value]` serialized as  `Tuple[ContractId, type]` to generate the signature.
`ncId` is the nano contract id being invoked.

The idea is to avoid replay attacks where we get the signed value from one contract to use on another contract.
The serialization itself  does not change since the `ncId` is not part of the data sent, it is only used when generating the signature and when validating it.

#### MethodArgument parser

There is a distinction from:
- "user inputs" meant to be used as nano contract arguments
- Serialized arguments from fullnode API
- Validated and correctly typed nano contract arguments

Each kind of data should not be treated in the same way, user inputs may not follow the correct typing (due to user errors) so they should be parsed and validated, for this we should use `zod`.
Fullnode data should also be validated and deserialized into the same type as the user input.

Once the data is valid we can store in a class that makes it easier to (de)serialize and use the data.
To achieve this we need to create a `NanoContractMethodArgument` class that can be used to (de)serialize arguments to/from the bytes values and from user input with zod validation.

## Acceptance Criteria

- Add NanoContractMethodArgument class to parse and validate arguments
- Add support for the new SignedData
- Add support for Tuple fields
- Prepare support for (de)serializing other container types

### Security Checklist

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -6,6 +6,7 @@ on:
       - release
       - release-candidate
       - feat/nc-args-serialization
+      - feat/nc-args-class
     tags:
       - v*
   pull_request:
@@ -14,6 +15,7 @@ on:
       - release-candidate
       - master
       - feat/nc-args-serialization
+      - feat/nc-args-class
 
 env:
   TEST_WALLET_START_TIMEOUT: '180000' # 3 minutes
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -5,6 +5,7 @@ on:
       - master
       - release
       - release-candidate
+      - feat/nc-args-serialization
     tags:
       - v*
   pull_request:
@@ -12,6 +13,7 @@ on:
       - release
       - release-candidate
       - master
+      - feat/nc-args-serialization
 
 env:
   TEST_WALLET_START_TIMEOUT: '180000' # 3 minutes
--- a/__tests__/integration/nanocontracts/bet.test.ts
+++ b/__tests__/integration/nanocontracts/bet.test.ts
@@ -327,7 +327,7 @@ describe('full cycle of bet nano contract', () => {
     const inputData = await getOracleInputData(oracleData, resultSerialized, wallet);
     const txSetResult = await wallet.createAndSendNanoContractTransaction('set_result', address1, {
       ncId: tx1.hash,
-      args: [`${bufferToHex(inputData)},${result},str`],
+      args: [`${bufferToHex(inputData)},${bufferToHex(tx1.hash)},${result},str`],
     });
     await checkTxValid(wallet, txSetResult);
     txIds.push(txSetResult.hash);
--- a/__tests__/integration/nanocontracts/bet.test.ts
+++ b/__tests__/integration/nanocontracts/bet.test.ts
@@ -95,7 +95,6 @@ describe('full cycle of bet nano contract', () => {
 
     // Create NC
     const oracleData = getOracleBuffer(address1, network);
-    console.debug('Call to initialize');
     const tx1 = await wallet.createAndSendNanoContractTransaction(
       NANO_CONTRACTS_INITIALIZE_METHOD,
       address0,
@@ -178,7 +177,6 @@ describe('full cycle of bet nano contract', () => {
     ).rejects.toThrow(NanoContractTransactionError);
 
     // Bet 100 to address 2
-    console.debug('First call to bet');
     const txBet = await wallet.createAndSendNanoContractTransaction('bet', address2, {
       ncId: tx1.hash,
       args: [address2, '1x0'],
@@ -234,7 +232,6 @@ describe('full cycle of bet nano contract', () => {
 
     // Bet 200 to address 3
     const address3 = await wallet.getAddressAtIndex(3);
-    console.debug('Second call to bet');
     const txBet2 = await wallet.createAndSendNanoContractTransaction('bet', address3, {
       ncId: tx1.hash,
       args: [address3, '2x0'],
@@ -328,7 +325,6 @@ describe('full cycle of bet nano contract', () => {
     const result = '1x0';
     const resultSerialized = nanoSerializer.serializeFromType(result, 'str');
     const inputData = await getOracleInputData(oracleData, tx1.hash, resultSerialized, wallet);
-    console.debug('Call to set_result');
     const txSetResult = await wallet.createAndSendNanoContractTransaction('set_result', address1, {
       ncId: tx1.hash,
       args: [`${bufferToHex(inputData)},${result},str`],
@@ -369,7 +365,6 @@ describe('full cycle of bet nano contract', () => {
     expect((txSetResultParser.parsedArgs[0].value as NanoContractSignedData).value).toEqual(result);
 
     // Try to withdraw to address 2, success
-    console.debug('Call to withdraw');
     const txWithdrawal = await wallet.createAndSendNanoContractTransaction('withdraw', address2, {
       ncId: tx1.hash,
       actions: [
--- a/__tests__/integration/nanocontracts/bet.test.ts
+++ b/__tests__/integration/nanocontracts/bet.test.ts
@@ -126,13 +126,20 @@ describe('full cycle of bet nano contract', () => {
       throw new Error('Could not parse args');
     }
     expect(tx1Parser.parsedArgs).toHaveLength(3);
-    expect(tx1Parser.parsedArgs[0]).toMatchObject({ name: 'oracle_script', type: 'TxOutputScript'});
-    // @ts-expect-error
+    expect(tx1Parser.parsedArgs[0]).toMatchObject({
+      name: 'oracle_script',
+      type: 'TxOutputScript',
+    });
+    // @ts-expect-error toMatchBuffer is defined in setupTests.js
     expect(tx1Parser.parsedArgs[0].value).toMatchBuffer(oracleData);
-    expect(tx1Parser.parsedArgs[1]).toMatchObject({ name: 'token_uid', type: 'TokenUid'});
-    // @ts-expect-error
+    expect(tx1Parser.parsedArgs[1]).toMatchObject({ name: 'token_uid', type: 'TokenUid' });
+    // @ts-expect-error toMatchBuffer is defined in setupTests.js
     expect(tx1Parser.parsedArgs[1].value).toMatchBuffer(Buffer.from(NATIVE_TOKEN_UID, 'hex'));
-    expect(tx1Parser.parsedArgs[2]).toMatchObject({ name: 'date_last_bet', type: 'Timestamp', value: dateLastBet});
+    expect(tx1Parser.parsedArgs[2]).toMatchObject({
+      name: 'date_last_bet',
+      type: 'Timestamp',
+      value: dateLastBet,
+    });
 
     // First validate some bet arguments error handling
     const address2 = await wallet.getAddressAtIndex(2);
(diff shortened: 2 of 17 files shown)


.github/workflows/integration-test.yml
```diff
@@ -5,13 +5,17 @@ on:
       - master
       - release
       - release-candidate
+      - feat/nc-args-serialization
+      - feat/nc-args-class
```
r4mmer, line 9: These will be removed before merging the PR

src/nano_contracts/deserializer.ts
```diff
@@ -75,14 +72,17 @@ class Deserializer {
...
     type: string
-  ): DeserializeResult<NanoContractArgumentType | null> {
+  ): BufferROExtract<NanoContractArgumentType | null> {
     const [containerType, internalType] = getContainerInternalType(type);
 
     switch (containerType) {
       case 'Optional':
         return this.toOptional(buf, internalType);
+      case 'RawSignedData':
       case 'SignedData':
```
r4mmer, line 82: Both serialize to the same bytes.
The only difference is that the signature of `SignedData` also includes the contract id, which is not part of the serialized field.

src/nano_contracts/deserializer.ts
```diff
@@ -263,78 +299,54 @@ class Deserializer {
...
-      throw new Error('Type is not SignedData');
-    }
-    if (!internalType) {
-      throw new Error('Unable to extract type');
-    }
-    // Should we check that the valueType is valid?
-
-    // SignData[T] is serialized as Serialize(T)+Serialize(sign(T)) where sign() returns a byte str
-    // Which means we can parse the T argument, then read the bytes after.
+  toSignedData(signedData: Buffer, type: string): BufferROExtract<NanoContractSignedData> {
```
r4mmer, line 302: The `type` argument changed from being the whole `SignedData[innerType]` to just the inner type

src/nano_contracts/methodArg.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+    // value is not null, should transform based on the type
+    return refineSingleValue(ctx, value[1], value[0]);
+  });
+
+/**
+ * Type and value validation for SignedData types.
+ * returns an instance of NanoContractSignedData
+ */
+const SignedDataApiInputScheme = z
```
r4mmer, line 166: This is meant to receive the signed data string `<signature>,<value>,<type>`, validate that `value` is of type `type`, `signature` is a hex string and transform to the internal type `NanoContractSignedData`

src/nano_contracts/types.ts
```diff
@@ -4,18 +4,94 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
+import { z } from 'zod';
```
r4mmer, line 7: Changing the types to be derived from a zod schema was meant to allow parsing and checking the types more easily where needed.

src/nano_contracts/utils.ts
```diff
@@ -212,88 +220,18 @@ export const validateAndUpdateBlueprintMethodArgs = async (
...
-  for (const [index, arg] of methodArgs.entries()) {
-    let typeToCheck = arg.type;
-    if (typeToCheck.startsWith('SignedData')) {
-      // Signed data will always be an hexadecimal with the
-      // signature len, signature, and the data itself
-      typeToCheck = 'str';
+  try {
+    const parsedArgs: NanoContractMethodArgument[] = [];
+    for (const [index, arg] of methodArgs.entries()) {
+      const parsedArg = NanoContractMethodArgument.fromApiInput(arg.name, arg.type, args[index]);
```
r4mmer, line 226: `fromApiInput` parses, validates and converts the data.

src/nano_contracts/utils.ts
```diff
@@ -41,15 +44,16 @@ export function getContainerInternalType(
...
     throw new Error('Unable to extract type');
   }
   // Only some values are allowed for containerType
   switch (containerType) {
     case 'Tuple':
     case 'SignedData':
     case 'RawSignedData':
-      return [containerType, internalType]
+    case 'Optional':
+      return [containerType, internalType];
```
pedroferreira1, line 56: Why do you need this? You already handle the Optional in the beginning of the method.

__tests__/integration/nanocontracts/bet.test.ts
```diff
@@ -351,14 +393,14 @@ describe('full cycle of bet nano contract', () => {
 
     const txWithdrawalParser = new NanoContractTransactionParser(
       blueprintId,
-      'set_result',
+      'withdraw',
```
pedroferreira1, line 396: Ops 😅

src/nano_contracts/parser.ts
```diff
@@ -94,18 +95,26 @@ class NanoContractTransactionParser {
...
+          deserializer
+        );
         parsed = parseResult.value;
         size = parseResult.bytesRead;
       } catch (err: unknown) {
-        console.error(err);
         throw new NanoContractTransactionParseError(`Failed to deserialize argument ${arg.type}.`);
       }
-      parsedArgs.push({ ...arg, parsed });
+      parsedArgs.push(parsed);
```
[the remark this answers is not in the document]
pedroferreira1: Ah, from the tests I saw that you still have name/type in the parsed data, is that correct?

setupTests-integration.js
```diff
@@ -50,3 +50,25 @@ afterAll(async () => {
   // Storing data about used precalculated wallets for the next test suites
   await precalculationHelpers.test.storeDbIntoWalletsFile();
 });
+
+expect.extend({
+  toMatchBuffer(received, expected) {
```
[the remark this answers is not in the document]
r4mmer: Checking buffer A is the same as buffer B is very common and there is no native way to do this.
We have this extension on the unit tests and I think we could add it here instead of having to check `buffer.toString('hex')` on all buffer values.

src/nano_contracts/methodArg.ts
```diff
@@ -0,0 +1,297 @@
...
+    });
+  }
+
+  // Meant to keep the typing correct
+  return z.NEVER;
+}
+
+/**
+ * Type and value validation for non-container types.
+ * Returns the internal TS type for the argument given.
```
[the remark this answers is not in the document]
r4mmer: TS as in TypeScript type.
Changed to be "parsed value"

src/nano_contracts/methodArg.ts
```diff
@@ -0,0 +1,297 @@
...
+    z.string(), // Inner type
+    NanoContractArgumentApiInputSchema, // value
+  ])
+  .transform((value, ctx) => {
+    const parse = z.null().safeParse(value[1]);
+    if (parse.success) {
+      return parse.data;
+    }
+    // value is not null, should transform based on the type
+    return refineSingleValue(ctx, value[1], value[0]);
```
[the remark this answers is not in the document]
r4mmer: This is why this is the `refineSingleValue`, the `Single` means only single native types (str, bytes, Address, ContractId, int, VarInt) not container types (Optional, SignedValue, Dict).

The `str?` would be identified as `Optional[str]` and the `OptionalApiInputScheme` would call this with `str`.

src/nano_contracts/types.ts
```diff
@@ -4,18 +4,94 @@
...
+export type NanoContractArgumentSingleType = z.output<typeof NanoContractArgumentSingleSchema>;
+
+/**
+ * NanoContract SignedData method argument type
+ */
+export const NanoContractSignedDataSchema = z.object({
+  type: z.string(),
+  signature: z.instanceof(Buffer),
+  value: NanoContractArgumentSingleSchema,
+  ncId: z.instanceof(Buffer).nullish(),
```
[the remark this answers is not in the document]
r4mmer: `nullish` means that the value can be null, undefined or not be present.
It will only check that it is an instance of Buffer if the value is present and is not null/undefined

src/nano_contracts/types.ts
```diff
@@ -4,18 +4,94 @@
...
+});
+export type NanoContractSignedData = z.output<typeof NanoContractSignedDataSchema>;
+
+/**
+ * Intermediate schema for all possible Nano contract argument type
```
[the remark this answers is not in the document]
r4mmer: It is not the type names but type of the values.
`NanoContractArgumentApiInputSchema` already adds `null` as a possibility so the `Optional` type will work

src/nano_contracts/parser.ts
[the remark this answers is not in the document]
r4mmer: Yes, the `parsedArgs` will be an instance that has all the required argument information

src/nano_contracts/utils.ts
r4mmer: This is something I was looking forward to ask, there are a lot of ways to make a value optional.

```py
class Foo(Blueprint):
  @public
  def foo(self, ctx: Context, val: Union[str, None]):
      noop
  @public
  def bar(self, ctx: Context, val: Optional[str]):
      noop
  @public
  def baz(self, ctx: Context, val: str | None):
      noop
```

Can we be sure all of these will evaluate to `str?` on the API?

src/nano_contracts/methodArg.ts
```diff
@@ -0,0 +1,297 @@
...
+    return new NanoContractMethodArgument(name, type, data);
+  }
+
+  toApiInput(): NanoContractParsedArgument {
+    function prepSingleValue(type: string, value: NanoContractArgumentSingleType) {
+      if (type === 'bool') {
+        return (value as boolean) ? 'true' : 'false';
+      }
+      if (
+        ['bytes', 'BlueprintId', 'ContractId', 'TokenUid', 'TxOutputScript', 'VertexId'].includes(
```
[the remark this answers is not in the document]
r4mmer: Done

__tests__/integration/nanocontracts/bet.test.ts
```diff
@@ -94,6 +95,7 @@ describe('full cycle of bet nano contract', () => {
 
     // Create NC
     const oracleData = getOracleBuffer(address1, network);
+    console.debug('Call to initialize');
```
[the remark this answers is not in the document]
r4mmer: Done

src/nano_contracts/utils.ts
tuliomir: This seems to me an important suite of tests to be implemented outside of this PR, which is already pretty large.

That way we can optimize as needed in the future, certain that there'll be no breaking changes.

tuliomir (approved):

src/nano_contracts/deserializer.ts
```diff
@@ -248,91 +297,56 @@ class Deserializer {
...
-      // Checksum value generated does not match value from fullnode
-      throw new Error(
-        `When parsing and Address(${address.base58}) we calculated checksum(${calcChecksum.toString('hex')}) but it does not match the checksum it came with ${recvChecksum.toString('hex')}.`
-      );
+  toTuple(buf: Buffer, type: string): BufferROExtract<NanoContractArgumentSingleType[]> {
+    const typeArr = type.split(',').map(s => s.trim());
+    const tupleValues: NanoContractArgumentSingleType[] = [];
+    let bytesReadTotal = 0;
+    let tupleBuf = buf.subarray();
+    for (const t of typeArr) {
```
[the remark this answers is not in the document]
r4mmer: We read 1 value for each type in `typeArr`, so the length of values read will be equal to the length of tuple subtypes.
Checking that the buffer is empty should not be done here since there are more values to read in the arguments.

Consider the case `def Foo(balance: Tuple[TokenUid, Amount], dest: Address)`, when we finish reading the `Tuple[TokenUid, Amount]` the buffer will still have the address bytes left so it will not be empty.
The `deserializeFromType` ensures the type returned and the number of subtypes is ensures as well.

src/nano_contracts/serializer.ts
```diff
@@ -23,41 +21,21 @@ class Serializer {
...
    *
    * @memberof Serializer
    * @inner
    */
   serializeFromType(value: NanoContractArgumentType, type: string): Buffer {
-    if (type.endsWith('?')) {
-      // This is an optional
-      const optionalType = type.slice(0, -1);
-      return this.fromOptional(value, optionalType);
-    }
```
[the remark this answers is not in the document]
r4mmer: Not at all, the `Optional` type is categorized as a "container type", meaning it contains another type/value.
So at the start of the function we check if it is a container type and use the `serializeContainerType` method which will serialize all container types, including `Optional`

src/nano_contracts/deserializer.ts
r4mmer: Done

src/nano_contracts/utils.ts
r4mmer: We already have tests for all Optional fields (`str?`, `int?`, `VarInt?`, etc.) and SignedData fields.
We will implement other container types and expand these tests, but as you said this is not in the scope of this PR.

pedroferreira1 (approved):

PR #866, opened 2025-05-21
chore: prepare nc argument to api input

r4mmer:
### Motivation

The old nano contract argument parsed values would convert to either number or string (with formats based on type), we improved this system to use typescript native typing and left a `NanoContractMethodArgument.toApiInput` method to convert to the old value if needed but since there was no use on the lib it stayed untested.

The explorer was using the old parsed values to show a human-readable version of the values so we now have a reason to improve the `toApiInput` method, making it complete (convert all supported types) and checking the types and values being converted so we do not silently show an invalid type on the explorer.

### Acceptance Criteria
- Make the `NanoContractMethodArgument.toApiInput` complete and more strictly validated.

### Security Checklist

### DIFF
--- a/src/nano_contracts/deserializer.ts
+++ b/src/nano_contracts/deserializer.ts
@@ -14,6 +14,8 @@ import {
   BufferROExtract,
   NanoContractSignedData,
   NanoContractArgumentSingleType,
+  NanoContractArgumentSingleTypeName,
+  NanoContractArgumentSingleTypeNameSchema,
 } from './types';
 import { OutputValueType } from '../types';
 import { NC_ARGS_MAX_BYTES_LENGTH } from '../constants';
@@ -267,7 +269,7 @@ class Deserializer {
    * @memberof Deserializer
    * @inner
    */
-  toOptional(buf: Buffer, type: string): BufferROExtract<NanoContractArgumentType> {
+  toOptional(buf: Buffer, type: NanoContractArgumentSingleTypeName): BufferROExtract<NanoContractArgumentType> {
     if (buf[0] === 0) {
       // It's an empty optional
       return {
@@ -297,7 +299,7 @@ class Deserializer {
    * @memberof Deserializer
    * @inner
    */
-  toSignedData(signedData: Buffer, type: string): BufferROExtract<NanoContractSignedData> {
+  toSignedData(signedData: Buffer, type: NanoContractArgumentSingleTypeName): BufferROExtract<NanoContractSignedData> {
     // The SignedData is serialized as `data+Signature`
 
     // Reading argument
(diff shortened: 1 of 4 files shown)


src/nano_contracts/types.ts
```diff
@@ -7,6 +7,64 @@
 import { z } from 'zod';
 import { IHistoryTx, OutputValueType } from '../types';
 
+export const NanoContractArgumentByteTypes = z.enum([
```
r4mmer, line 10: Moved the names up in the file so they can be used on other schemas

src/nano_contracts/deserializer.ts
```diff
@@ -328,13 +339,15 @@ class Deserializer {
    * It does not support chained container types, meaning Tuple[Dict[str,str]] should not happen.
    *
    * @param buf Value to deserialize
-   * @param type Comma separated types, e.g. `str,int,VarInt`
+   * @param typeArr Comma separated types, e.g. `str,int,VarInt`
```
tuliomir, line 342: question: If I understand correctly, this is an actual array, not a comma separated string value. If so, the docstring should reflect this.

src/nano_contracts/serializer.ts
```diff
@@ -234,7 +245,7 @@ class Serializer {
    * Serialize a tuple of values
    *
    * @param value List of values to serialize
-   * @param typeStr Comma separated list of types e.g. `str,int,VarInt`
+   * @param typeArr Comma separated list of types e.g. `str,int,VarInt`
```
tuliomir, line 248: chore: Same here. The typescript type specifies an array, but the description says it's a comma separated string.

src/nano_contracts/deserializer.ts
r4mmer: Fixed

src/nano_contracts/serializer.ts
r4mmer: Fixed

tuliomir (approved):

src/nano_contracts/utils.ts
```diff
@@ -27,17 +27,23 @@ import {
...
-): [NanoContractArgumentContainerType, string] {
+): [
+  NanoContractArgumentContainerType,
+  NanoContractArgumentSingleTypeName | NanoContractArgumentSingleTypeName[],
+] {
   if (type.endsWith('?')) {
     // Optional value
-    return ['Optional', type.slice(0, -1)];
+    const innerType = type.slice(0, -1);
+    return ['Optional', NanoContractArgumentSingleTypeNameSchema.parse(innerType)];
```
pedroferreira1, line 46: Will it work if it's an optional tuple?

pedroferreira1 (approved):

PR #877, opened 2025-05-28
fix: use single utxo selection for deposit action and token deposit fee

pedroferreira1:
### Motivation

There were two bugs:

(i) the utxo selection for deposit action and token creation deposit fee happens in different phases, so we must mark the utxos as selected after the first phase, so we don't select the same one in the second phase.
(ii) we were selecting two different utxos when we could have only one for an HTR deposit action and token creation deposit fee.

### Acceptance Criteria
- Mark utxos selected in actions as used.
- When selecting an HTR utxo for deposit action, it already selects also for the token creation deposit, if there are any creation.

Note: I decided to change the logic a bit for the `contractPaysForTokenDeposit` flag. Now it will pay for any data output fee in the token creation as well. The nano contract method code will validate any logic like this.

### Security Checklist

### DIFF
--- a/__tests__/integration/nanocontracts/authority.test.ts
+++ b/__tests__/integration/nanocontracts/authority.test.ts
@@ -481,6 +481,79 @@ describe('Authority actions blueprint test', () => {
     expect(initializeTokenCreateData.tx.outputs[3].token_data).toBe(0);
     expect(initializeTokenCreateData.tx.outputs[4].value).toBe(889n);
     expect(initializeTokenCreateData.tx.outputs[4].token_data).toBe(0);
+
+    // Create a token that will need two utxos
+    // Use a new wallet so the utxos don't get mixed with previous change utxos
+    const newWallet = await generateWalletHelper();
+    const newAddress0 = await newWallet.getAddressAtIndex(0);
+    await GenesisWalletHelper.injectFunds(newWallet, newAddress0, 100n);
+    await GenesisWalletHelper.injectFunds(newWallet, newAddress0, 100n);
+
+    // Create NC with deposit of HTR and token creation
+    const twoUtxosInitializeData = {
+      blueprintId,
+      actions: [
+        {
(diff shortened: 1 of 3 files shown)


src/utils/tokens.ts
```diff
@@ -356,14 +356,11 @@ const tokens = {
     // We might have transactions where the nano contract will pay for deposit fees
     // so we must consider the skipDepositFee flag to skip the utxo selection
     if (!skipDepositFee) {
```
r4mmer, line 358:
```suggestion
    if (!skipDepositFee) {
      depositAmount += this.getTransactionHTRDeposit(amount, data?.length ?? 0, storage);
```

And we can remove the rest of the if, right?

__tests__/integration/nanocontracts/authority.test.ts
```diff
@@ -415,6 +415,72 @@ describe('Authority actions blueprint test', () => {
     expect(tokenDetail5.totalSupply).toBe(1000n);
     expect(tokenDetail5.authorities.mint).toBe(true);
     expect(tokenDetail5.authorities.melt).toBe(true);
+
+    // Create a new nano contract of this blueprint with token creation and deposit
```
tuliomir, line 419: question: Reading these integration tests, I'm under the impression that we don't have a test for a case when more than one utxo is needed to execute this transaction.

Should we add this case? Maybe there's some condition in there that may fail when, for example, the data outputs can be covered by a set of utxos but a large amount of token mints would need another set.

pedroferreira1: It shouldn't change anything in the logic but I added the test: https://github.com/HathorNetwork/hathor-wallet-lib/pull/877/commits/064742fcb35da2fe55b6d1b577eb9c1ab995a058

tuliomir (approved):

r4mmer (approved):

PR #901, opened 2025-08-06
refactor: send nano contract tx working on the wallet-service

andreabadesso:
## Acceptance Criteria
- Update schemas for the proxy methods with nano contract responses
- Added missing `createAndSendNanoContractCreateTokenTransaction` method
- Created a `storage` proxy, to override util methods so they're compatible the the wallet-service

### DIFF
--- a/src/wallet/api/schemas/walletApi.ts
+++ b/src/wallet/api/schemas/walletApi.ts
@@ -275,8 +275,8 @@ export const fullNodeOutputSchema = z.object({
   }),
   address: AddressSchema.nullable().optional(),
   token: tokenIdSchema.nullable().optional(),
-  authorities: bigIntCoercibleSchema,
-  timelock: z.number().nullable(),
+  authorities: bigIntCoercibleSchema.optional(),
+  timelock: z.number().nullable().optional(),
 });
 
 /**
@@ -304,8 +304,8 @@ export const fullNodeTxSchema = z.object({
   inputs: z.array(fullNodeInputSchema),
   outputs: z.array(fullNodeOutputSchema),
   tokens: z.array(fullNodeTokenSchema),
-  token_name: z.string().nullable(),
-  token_symbol: z.string().nullable(),
+  token_name: z.string().nullable().optional(),
+  token_symbol: z.string().nullable().optional(),
   raw: z.string(),
 });
 
@@ -322,9 +322,9 @@ export const fullNodeMetaSchema = z.object({
   height: z.number(),
   voided_by: z.array(z.string()),
   spent_outputs: z.array(z.tuple([z.number(), z.array(z.string())])),
-  received_timestamp: z.number().nullable(),
-  is_voided: z.boolean(),
-  verification_status: z.string(),
+  received_timestamp: z.number().nullable().optional(),
+  is_voided: z.boolean().optional(),
+  verification_status: z.string().optional(),
   twins: z.array(z.string()),
   accumulated_weight: z.number(),
   score: z.number(),
--- a/src/wallet/api/schemas/walletApi.ts
+++ b/src/wallet/api/schemas/walletApi.ts
@@ -282,12 +282,15 @@ export const fullNodeOutputSchema = z.object({
 /**
  * Schema for full node token information.
  * Represents token details as seen by the full node.
+ * Note: amount is optional because this schema is reused across different APIs:
+ * - Regular transaction APIs include amount field
+ * - Nano contract token creation APIs only include uid, name, and symbol
  */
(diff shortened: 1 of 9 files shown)


src/wallet/wallet.ts
```diff
@@ -1698,6 +1704,54 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      filter_address: null,
+      ...options,
+    };
+
+    return this._getAuthorityTxOutput({
+      tokenId: tokenUid,
+      authority: authorityValue,
+      skipSpent: newOptions.only_available_utxos,
+      maxOutputs: newOptions.many ? undefined : 1,
+    });
```
tuliomir, line 1752: question: The `filter_address` parameter is not being used. Is this by design?

src/wallet/wallet.ts
```diff
@@ -2645,9 +2745,24 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+    if (createTokenOptions.name === undefined || createTokenOptions.name === null) {
+      throw new Error('createTokenOptions.name is required and cannot be null or undefined');
+    }
+    if (createTokenOptions.symbol === undefined || createTokenOptions.symbol === null) {
+      throw new Error('createTokenOptions.symbol is required and cannot be null or undefined');
+    }
+    // If mintAddress is not provided, use the address from the nano contract caller
+    if (!createTokenOptions.mintAddress) {
+      createTokenOptions.mintAddress = address;
+    }
```
tuliomir, line 2761: suggestion(code-style): I suspect this validation could be done internally by zod if the `fullNodeTokenSchema` ( on file `api/schemas/walletApi.ts` ) was implemented using `z.discriminatedUnion` ( see [src/nano_contract/types](https://github.com/HathorNetwork/hathor-wallet-lib/blob/3475450c2444b1c91bc8328bc774e8782bcaee5c/src/nano_contracts/types.ts#L96-L101) ).

src/wallet/api/schemas/walletApi.ts
```diff
@@ -275,19 +275,41 @@ export const fullNodeOutputSchema = z.object({
...
-  amount: bigIntCoercibleSchema,
+  amount: bigIntCoercibleSchema.optional(),
+});
+
+/**
+ * Schema for nano contract context actions.
+ */
+export const ncActionSchema = z.object({
+  type: z.string(),
+  token_uid: z.string().optional(),
```
pedroferreira1, line 301: Why is this optional?

src/wallet/wallet.ts
```diff
@@ -1316,17 +1318,22 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      const addressDetails = await this.getAddressDetails(address);
+      const addressIndex = addressDetails?.index;
 
-    // Return undefined if the address is not found
-    return undefined;
+      return addressIndex;
+    } catch (_e) {
+      // Return undefined if the address is not found
+      return undefined;
+    }
```
pedroferreira1, line 1336: If the address is not from this wallet, what's the return? Maybe we should handle this error specifically and a possible timeout / client error in another way. What do you think?

src/wallet/walletServiceStorageProxy.ts
```diff
@@ -0,0 +1,186 @@
...
+   * Creates a proxy that wraps the original storage with additional methods
+   * needed for nano contract transaction signing.
+   */
+  createProxy(): IStorage {
+    return new Proxy(this.originalStorage, {
+      get: this.proxyHandler.bind(this),
+    });
+  }
+
+  private proxyHandler(target: IStorage, prop: string | symbol, receiver: IStorage): unknown {
```
pedroferreira1, line 43: What's `symbol` and this `receiver`?

src/wallet/walletServiceStorageProxy.ts
```diff
@@ -0,0 +1,186 @@
...
+   */
+  private async *getSpentTxs(inputs: any[]) {
+    for (let index = 0; index < inputs.length; index++) {
+      const input = inputs[index];
+      const tx = await this.getTx(input.hash);
+      if (tx) {
+        yield { tx, input, index };
+      }
+    }
+  }
```
pedroferreira1, line 120: Not sure if it's worth implementing bug we could have more than one input of the same tx, so we could have a cache of the requested txs, so we don't request the same one twice.

src/wallet/walletServiceStorageProxy.ts
tuliomir: Maybe a docstring could help explain those parameters.

PR #902, opened 2025-08-11
refactor: create token options

andreabadesso:
### Acceptance Criteria
- We must have the same signatures for the `prepareCreateNewToken` method in both facades
- We must fix the `ncContext` schema validation from `address` to `caller_id`
- We must fix `getAddressIndex` to be async and request the index from the wallet-service instead of the local `newAddresses` cache (which is a bug)

### Security Checklist

### DIFF
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -1540,8 +1540,6 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
 
     // Create outputs
     const outputsObj: Output[] = [];
-
-    // Add data outputs if specified - for NFT creation, these go first
     const dataOutputs: Output[] = [];
     if (newOptions.data && newOptions.data.length > 0) {
       for (const dataString of newOptions.data) {
@@ -1549,7 +1547,6 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
       }
     }
 
-    // For NFT creation, data outputs go first
     if (newOptions.isCreateNFT && dataOutputs.length > 0) {
       outputsObj.push(...dataOutputs);
     }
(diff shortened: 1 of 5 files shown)


src/wallet/wallet.ts
```diff
@@ -1498,15 +1500,16 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      newOptions.isCreateNFT || (newOptions.data !== null && newOptions.data.length > 0);
 
     const depositPercent = this.storage.getTokenDepositPercentage();
     // 1. Calculate HTR deposit needed
     let deposit = tokens.getDepositAmount(amount, depositPercent);
 
-    if (isNFT) {
-      // For NFT we have a fee of 0.01 HTR, then the deposit utxo query must get an additional 1
-      deposit += 1n;
+    if (isNFT && newOptions.data && newOptions.data.length > 0) {
```
tuliomir, line 1510: suggestion: The definition of `isNFT` above already includes `newOptions.data?.length > 0`. I guess we can remove this redundant validation here.

src/wallet/wallet.ts
```diff
@@ -1529,10 +1532,17 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
 
     // Create outputs
     const outputsObj: Output[] = [];
-    // NFT transactions must have the first output as the script data
-    if (isNFT) {
-      outputsObj.push(helpers.createNFTOutput(newOptions.nftData!));
+    const dataOutputs: Output[] = [];
+    if (newOptions.data && newOptions.data.length > 0) {
+      for (const dataString of newOptions.data) {
+        dataOutputs.push(helpers.createNFTOutput(dataString));
```
tuliomir, line 1538: question: It seems to me that this helper will create any data output, not only NFT related ones. 

Would it be a valid suggestion AND in the scope of this PR to refactor this helper method to `createDataOutput()`?

andreabadesso: Refactored this so it is exactly like the fullnode facade, can you please check again?

src/wallet/wallet.ts
```diff
@@ -2700,7 +2714,7 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     });
 
     const mergedCreateTokenOptions: NanoContractBuilderCreateTokenOptions = {
-      mintAddress: null,
+      mintAddress: createTokenOptions.mintAddress,
```
pedroferreira1, line 2717: The `mintAddress` does not belong to `createTokenOptions`?

PR #905, opened 2025-08-19
fix: nano ctx schema and update integration tests full node version

pedroferreira1:
### Motivation

The new hathor core version has some breaking changes in the APIs and blueprint sdk. This PR will update the code for the new docker image.

Note: there's a small fix for the template module

### Acceptance Criteria
- Update `address` to `caller_id` in the `nc_context` APIs.
- Update hathor core docker image, cli parameters, and configuration that was changed in the latest version.
- Fix template servarcommands option with index as `0`.
- Fix nano template test to stop revoking melt authority that doesn't exist.

### Security Checklist


src/template/transaction/setvarcommands.ts
```diff
@@ -23,7 +23,7 @@ export async function getWalletAddress(
   _ctx: TxTemplateContext,
   options: z.infer<typeof SetVarGetWalletAddressOpts>
 ): Promise<string> {
-  if (options.index) {
+  if (options.index != null) {
```
pedroferreira1, line 26: This is the template small fix. The index might be 0, which wasn't entering the `if`

PR #908, opened 2025-08-21
feat: add getServerUrl to wallet service facade

pedroferreira1:
### Acceptance Criteria
- Create method to get server URL in the wallet service facade

Note: we already have this method in the old facade

### Security Checklist

### DIFF
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -269,6 +269,16 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
   }
 
   /**
+   * Get server URL from config object
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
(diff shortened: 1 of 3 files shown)


src/wallet/wallet.ts
```diff
@@ -268,6 +268,17 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     };
   }
 
+  /**
+   * Get server URL from config object
+   *
+   * @memberof HathorWalletServiceWallet
+   * @inner
+   */
+  // eslint-disable-next-line class-methods-use-this
```
pedroferreira1, line 277: It's complaining that the method doesn't use `this`, so it can't be a class method. But I need it to be like this, so both facades are equal, then I'm just disabling the rule for this method

andreabadesso (approved):

PR #912, opened 2025-08-28
feat: complete IHathorWallet interface with new methods for typing

pedroferreira1:
### Acceptance Criteria
- Add new methods to the `IHathorWallet` facade
- Export types from `wallet/types.ts`

### Security Checklist

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -121,3 +121,4 @@ export {
 
 export * from './nano_contracts/types';
 export * from './models/types';
+export type { IHathorWallet, OutputRequestObj, DataScriptOutputRequestObj } from './wallet/types';
--- a/src/wallet/sendTransactionWalletService.ts
+++ b/src/wallet/sendTransactionWalletService.ts
@@ -52,13 +52,13 @@ class SendTransactionWalletService extends EventEmitter implements ISendTransact
   private changeAddress: string | null;
 
   // Transaction object to be used after it's already prepared
-  private transaction: Transaction | null;
+  public transaction: Transaction | null;
 
   // MineTransaction object
   private mineTransaction: MineTransaction | null;
 
   // PIN to load the seed from memory
-  private pin: string | null;
+  public pin: string | null;
 
   // Data for the transaction after it's prepared
   public fullTxData: IDataTx | null;
--- a/src/wallet/sendTransactionWalletService.ts
+++ b/src/wallet/sendTransactionWalletService.ts
@@ -58,7 +58,7 @@ class SendTransactionWalletService extends EventEmitter implements ISendTransact
   private mineTransaction: MineTransaction | null;
 
   // PIN to load the seed from memory
-  public pin: string | null;
+  private pin: string | null;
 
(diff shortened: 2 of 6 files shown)


src/lib.ts
```diff
@@ -121,3 +121,4 @@ export {
 
 export * from './nano_contracts/types';
 export * from './models/types';
+export type { IHathorWallet, OutputRequestObj, DataScriptOutputRequestObj } from './wallet/types';
```
pedroferreira1, line 124: In `models/types` we have types with the same name as in the `wallet/types`, so I can't do `export * from './wallet/types';`, I get the following error doing it

```
src/lib.ts:124:1 - error TS2308: Module './models/types' has already exported a member named 'Authority'. Consider explicitly re-exporting to resolve the ambiguity.

124 export * from './wallet/types';
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/lib.ts:124:1 - error TS2308: Module './models/types' has already exported a member named 'Balance'. Consider explicitly re-exporting to resolve the ambiguity.

124 export * from './wallet/types';
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/lib.ts:124
```

We should normalize these types, so we can export all.

src/wallet/sendTransactionWalletService.ts
```diff
@@ -52,13 +52,13 @@ class SendTransactionWalletService extends EventEmitter implements ISendTransact
...
   // Transaction object to be used after it's already prepared
-  private transaction: Transaction | null;
+  public transaction: Transaction | null;
 
   // MineTransaction object
   private mineTransaction: MineTransaction | null;
 
   // PIN to load the seed from memory
-  private pin: string | null;
+  public pin: string | null;
```
pedroferreira1, line 61: We access `transaction` and `pin` in the RPC handler methods (which we should stop) but for now we need to make them public.

tuliomir: Since this is security-related, should we create an issue in the RPC Lib to make sure this is fixed?

src/wallet/types.ts
```diff
@@ -402,6 +404,62 @@ export interface IHathorWallet {
...
+  ): Promise<Transaction>;
+  isAddressMine(address: string): Promise<boolean>;
+  getUtxos(options?: {
+    token?: string;
+    authorities?: number;
+    max_utxos?: number;
+    filter_address?: string;
+    amount_smaller_than?: number;
+    amount_bigger_than?: number;
+    max_amount?: number;
```
tuliomir, line 447: question: Shouldn't we use `bigint` values here?

src/wallet/sendTransactionWalletService.ts
pedroferreira1: I added a new commit to add some new parameters to util methods and now I don't need to make pin public anymore.

tuliomir (approved):

andreabadesso (approved):

PR #938, opened 2025-09-22
refactor: move get blueprintId method to a util

pedroferreira1:
### Acceptance Criteria
- Move `getBlueprintId` to a util method

### Security Checklist

### DIFF
--- a/src/nano_contracts/utils.ts
+++ b/src/nano_contracts/utils.ts
@@ -20,7 +20,7 @@ import P2SH from '../models/p2sh';
 import Address from '../models/address';
 import Transaction from '../models/transaction';
 import { NanoContractTransactionError, OracleParseError, WalletFromXPubGuard } from '../errors';
-import { OutputType, IHathorWallet } from '../wallet/types';
+import { FullNodeTxResponse, OutputType, IHathorWallet } from '../wallet/types';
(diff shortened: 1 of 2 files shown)


src/nano_contracts/utils.ts
```diff
@@ -315,3 +315,39 @@ export const getResultHelper = async <T>(
...
+  if (!ncId) {
+    throw new Error('Nano contract ID is not defined');
+  }
+
+  const [txError, txResponse] = (await getResultHelper(() =>
+    wallet.getFullTxById(ncId)
+  )) as [Error | null, FullNodeTxResponse | null];
+  if (!txError && txResponse?.tx?.nc_id && txResponse.tx.nc_blueprint_id) {
+    return txResponse.tx.nc_blueprint_id;
+  }
```
r4mmer, line 341: Since we use the wallet only for this method and this method only calls the `txApi.getTransaction`, why not remove the wallet from the arguments and call the api directly?

r4mmer (changes requested):

src/nano_contracts/utils.ts
pedroferreira1: The idea is that we could use this method with the wallet service facade as well, which calls a different API for this. For now we still depend on the state API call but we should trust in the facade methods when possible.

r4mmer (approved):

PR #951, opened 2025-10-15
feat: fee header

raul-oliveira:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1412 and https://github.com/HathorNetwork/hathor-core/pull/1415

Fee Header implementation following the protocol changes in the master branch

### Acceptance Criteria
- Should add the fee header without any usage, only with the tests.

### Security Checklist

### DIFF
--- a/src/headers/fee.ts
+++ b/src/headers/fee.ts
@@ -1,0 +1,236 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { getVertexHeaderIdBuffer, getVertexHeaderIdFromBuffer, VertexHeaderId } from './types';
+import Header from './base';
+import Network from '../models/network';
+import { IFeeEntry, OutputValueType } from '../types';
+import { bytesToOutputValue, intToBytes, outputValueToBytes, unpackToInt } from '../utils/buffer';
+import { MAX_FEE_HEADER_ENTRIES } from '../constants';
+
+/**
+ * FeeHeader represents the fee payment information in a transaction.
+ * It explicitly indicates which tokens and amounts are being used to pay transaction fees.
+ *
+ * @class FeeHeader
+ * @extends {Header}
+ */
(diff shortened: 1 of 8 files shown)


src/headers/types.ts
```diff
@@ -12,6 +12,7 @@
  */
 export const enum VertexHeaderId {
   NANO_HEADER = '10',
+  FEE_HEADER = '20',
```
r4mmer: _Unrelated to PR_

Is there a rule to decide new header ids or is it arbitrary?

src/headers/fee.ts
```diff
@@ -0,0 +1,245 @@
...
+   * @inner
+   */
+  static getHeadersFromTx(tx: { headers: Header[] }): FeeHeader | null {
+    for (const header of tx.headers) {
+      if (header instanceof FeeHeader) {
+        return header;
+      }
+    }
+    return null;
+  }
```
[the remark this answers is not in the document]
raul-oliveira: Yep. The full node will reject more than 1 fee header.

src/headers/types.ts
raul-oliveira: In this case it's arbitrary and I copied from the fullnode fee header id.

src/headers/fee.ts
```diff
@@ -0,0 +1,240 @@
...
+    // Number of entries
+    array.push(intToBytes(this.entries.length, 1));
+
+    // Serialize each entry
+    for (const entry of this.entries) {
+      // Token Index (1 byte)
+      array.push(intToBytes(entry.tokenIndex, 1));
+
+      // Amount (variable length, LEB128 unsigned)
+      array.push(leb128Util.encodeUnsigned(entry.amount));
```
pedroferreira1: I'm not sure we are using leb128 for serialization of output values. In the core it's using the util method `output_value_to_bytes`, similar to what we are using in the nano header for the action amount `outputValueToBytes`

src/headers/types.ts
raul-oliveira: Tks @pedroferreira1. It's fixed.

pedroferreira1 (approved):

PR #858, opened 2025-05-05
feat: token info version upgrade

raul-oliveira:
# Choose a pull request template

- [feature PR](?template=feature.md&quick_pull=1)
- [Release process](https://github.com/HathorNetwork/hathor-wallet-lib/compare/release...master?template=release.md&quick_pull=1)

PR #964, opened 2025-10-28
feat: fee token in tx template

raul-oliveira:
Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/858

### Motivation
Add the fee, and fee completion api to the transaction template. It will enable developers to create custom transactions without having to add new methods or args to the wallet facade

### Acceptance Criteria
- Add tests for tx template
- Add the `addFee` instruction, and executor

### Security Checklist

### DIFF
--- a/__tests__/integration/walletservice_facade.test.ts
+++ b/__tests__/integration/walletservice_facade.test.ts
@@ -626,6 +626,41 @@ describe('basic transaction methods', () => {
           tokenData: 0,
         })
       );
+
+      await pollForTx(gWallet, sendTransaction.hash!);
+    });
+
+    it('should send a transaction to a P2SH (multisig) address', async () => {
+      // Use a P2SH address from the multisig wallet constants
+      const p2shAddress = WALLET_CONSTANTS.multisig.addresses[0];
+
+      const sendTransaction = await gWallet.sendTransaction(p2shAddress, 5n, {
+        pinCode,
+      });
+
+      // Validate the transaction was created successfully
+      expect(sendTransaction).toEqual(
+        expect.objectContaining({
+          hash: expect.any(String),
+          inputs: expect.any(Array),
+          outputs: expect.any(Array),
+        })
+      );
+
+      // Wait for transaction to be confirmed and verify it on the full node
+      await pollForTx(gWallet, sendTransaction.hash!);
+
+      // Get the full transaction from the network to verify the P2SH output
+      const fullTx = await gWallet.getFullTxById(sendTransaction.hash!);
+      expect(fullTx.success).toBe(true);
+
+      // Find the output with the P2SH address
+      const p2shOutput = fullTx.tx.outputs.find(output => output.decoded?.address === p2shAddress);
+
+      expect(p2shOutput).toBeDefined();
+      expect(p2shOutput!.value).toBe(5n);
+      expect(p2shOutput!.decoded.type).toBe('MultiSig');
+      expect(p2shOutput!.decoded.address).toBe(p2shAddress);
     });
 
     it('should send a transaction with a set changeAddress', async () => {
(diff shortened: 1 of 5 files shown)


src/template/transaction/context.ts
```diff
@@ -6,56 +6,102 @@
...
+    this.tokenVersion = tokenVersion;
+  }
+
+  needsToAddInputs() {
+    return this.tokens > 0;
+  }
+
+  needsToAddOutputs() {
+    return this.tokens < 0;
+  }
```
r4mmer: These are not being used?

raul-oliveira: removed

src/template/transaction/context.ts
```diff
@@ -145,6 +207,37 @@ export class TxBalance {
...
+    return fee;
+  }
+
+  private async getTokenVersion(token: string): Promise<TokenVersion> {
+    if (token === NATIVE_TOKEN_UID) {
+      return TokenVersion.NATIVE;
+    }
+    const tokenDetails = await this._interpreter.getTokenDetails(token);
+    return tokenDetails.tokenInfo.version;
+  }
```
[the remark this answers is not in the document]
raul-oliveira: Done. I placed a comment to warn about the total supply and total transaction, they may be stale while using cached values.

src/template/transaction/utils.ts
```diff
@@ -47,7 +49,7 @@ export async function selectTokens(
   if (autoChange && changeAmount) {
     ctx.log(`Creating change for address: ${changeAddress}`);
     // Token should only be on the array if present on the outputs
-    const tokenData = ctx.addToken(token);
+    const tokenData = await ctx.addToken(interpreter, token);
```
pedroferreira1: Why do you need to add it here, if we already added in the beginning of the method? If it's because of the `tokenData`, then we can get it in the first call

r4mmer (dismissed):

pedroferreira1 (dismissed):

src/template/transaction/context.ts
```diff
@@ -11,51 +11,92 @@ import { TokenVersion, IHistoryTx, ILogger, OutputValueType, getDefaultLogger }
...
+}
+
+/**
+ * Calculate the fee for a single token balance.
+ */
+function calculateTokenFee(balance: TokenBalance): bigint {
+  let fee = 0n;
+  if (balance.chargeableOutputs > 0) {
+    fee += BigInt(balance.chargeableOutputs) * FEE_PER_OUTPUT;
+  } else if (balance.chargeableInputs > 0) {
```
tuliomir: question: Isn't it possible to have both chargeable inputs _and_ outputs in the same transaction?

src/template/transaction/interpreter.ts
```diff
@@ -153,6 +155,17 @@ export class WalletTxTemplateInterpreter implements ITxTemplateInterpreter {
     );
   }
 
+  private static buildFeeHeader(ctx: TxTemplateContext): FeeHeader {
+    const entries: IFeeEntry[] = Array.from(ctx.fees.entries()).map(([tokenUid, amount]) => ({
+      tokenIndex: tokensUtils.getTokenIndex(
+        ctx.tokens.map(t => ({ uid: t })),
```
tuliomir: question: There is a warning that this header section wasn't tested. Isn't this a critical part of the resulting transaction, or can we skip this testing to a later stage?

tuliomir (dismissed):

src/template/transaction/interpreter.ts
raul-oliveira: Nice catch. We are using the fee header in the integration tests. I'll see why it's nove flagging as covered :)

pedroferreira1 (approved):

pedroferreira1 (approved):

PR #979, opened 2025-12-05
feat: fee token creation in nano

raul-oliveira:
Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/964

### Motivation
Allow users to use the nano builder to deal with fee based tokens (deposit, withdrawal). 

### Security Checklist
 Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

### DIFF
--- a/src/nano_contracts/builder.ts
+++ b/src/nano_contracts/builder.ts
@@ -916,6 +916,8 @@ class NanoContractTransactionBuilder {
       }
 
       // When contractPaysFees is true, deduct fee from HTR withdrawal output
+      // We assume that we have only one HTR output when contractPaysFees, otherwise the user must use the tx template
+      // to build a more complex tx.
       if (fee > 0n && this.contractPaysFees) {
         const htrOutputIndex = outputs.findIndex(
           output => 'token' in output && output.token === NATIVE_TOKEN_UID && !output.authorities
--- a/src/utils/fee.ts
+++ b/src/utils/fee.ts
@@ -26,6 +26,7 @@ export class Fee {
    * For nano contract actions:
    * - Deposit actions behave like outputs (tokens going INTO contract)
    * - Withdraw actions behave like inputs (tokens coming FROM contract)
+   * - Grant/Acquire authority actions behave like authority outputs/inputs (mint/melt authorities), and have no fee impact
    *
    * @param inputs the inputs of the transaction
    * @param outputs the outputs of the transaction


__tests__/integration/template/transaction/fee.test.ts
```diff
@@ -0,0 +1,315 @@
...
+            changeAddress: address0,
+          },
+        ],
+      }
+    );
+    await checkTxValid(hWallet, tx);
+
+    const ncState = await ncApi.getNanoContractState(contractId, ['fbt_uid'], [NATIVE_TOKEN_UID], []);
+    expect(ncState.fields.fbt_uid.value).toBeDefined();
+    fbtUid = ncState.fields.fbt_uid.value;
```
r4mmer: how do we check that this is a fbt?

__tests__/integration/template/transaction/fee.test.ts
```diff
@@ -0,0 +1,315 @@
...
+  stopAllWallets,
+  waitForTxReceived,
+  waitTxConfirmed,
+} from '../../helpers/wallet.helper';
+
+import ncApi from '../../../../src/api/nano';
+import HathorWallet from '../../../../src/new/wallet';
+import { NATIVE_TOKEN_UID, NANO_CONTRACTS_INITIALIZE_METHOD } from '../../../../src/constants';
+
+describe('FeeBlueprint Template execution', () => {
```
r4mmer: Maybe add a test where the contract pays the fee, i.e. feeAmount 0

raul-oliveira: Could you explain it a little bit more?

In the `should move FBT tokens from nc1 to nc2` and `should get FBT tokens back from nc2 to nc1` you can see the actions charging the contract balance (the contract here is acting as the caller).

Everytime you need to deposit/withdraw from to/from a contract, you need to pay the fee because in the operation perspective the tx is a caller.

__tests__/integration/template/transaction/fee.test.ts
raul-oliveira: By calling the `getTokenDetails` api. I added it to the test.

src/new/wallet.ts
```diff
@@ -3104,6 +3104,11 @@ class HathorWallet extends EventEmitter {
       .setArgs(data.args)
       .setVertexType(NanoContractVertexType.TRANSACTION);
 
+    // Set fee amount if declared
+    if (data.feeAmount) {
+      builder.setFeeAmount(BigInt(data.feeAmount));
+    }
+
```
pedroferreira1: I didn't understand this flow. The facade call has the fee amount as parameter? Why do we require the user to know how much fee regarding fee based tokens he will need? Do we have a design discussion about this?

Isn't it better to calculate the fee amount depending on the actions executions and get the utxos from it?

raul-oliveira: After some discussions we changed the api to have a max fee instead of setting the value directly

src/new/wallet.ts
```diff
@@ -3104,6 +3104,11 @@ class HathorWallet extends EventEmitter {
       .setArgs(data.args)
       .setVertexType(NanoContractVertexType.TRANSACTION);
 
+    // Set fee amount if declared
```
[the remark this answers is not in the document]
raul-oliveira: We changed to achieve the `maxFee` and I applied the same behavior of deposit tokens to the fee tokens in the `createNanoContractCreateTokenTransaction`.

src/nano_contracts/builder.ts
```diff
@@ -679,13 +852,17 @@ class NanoContractTransactionBuilder {
...
+            this.tokenFeeAddedInDeposit ||
+            this.createTokenOptions.contractPaysFees,
+          tokenVersion: this.createTokenOptions.tokenVersion,
         }
       );
 
       data.inputs = concat(data.inputs, inputs);
       data.outputs = concat(data.outputs, outputs);
       data.tokens = uniq(concat(data.tokens, tokens));
+      data.headers = concat(data.headers ?? [], headers);
```
pedroferreira1: Why did you add this here? Do we need the headers in this `transactionUtils.createTransactionFromData` method? It seems we are already adding the headers in the other method.

If we do, shouldn't we add in the first `if` above when `this.vertexType === NanoContractVertexType.TRANSACTION`?

src/nano_contracts/builder.ts
```diff
@@ -197,6 +237,62 @@ class NanoContractTransactionBuilder {
...
+
+    // Add fee for deposit actions (tokens going into contracts count as outputs)
+    this.actions?.forEach(action => {
+      if (action.type === NanoContractActionType.DEPOSIT && action.token) {
+        const tokenData = tokensMap.get(action.token);
+        if (tokenData && tokenData.version === TokenVersion.FEE) {
+          fee += FEE_PER_OUTPUT;
+        }
+      }
+    });
```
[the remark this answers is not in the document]
raul-oliveira: No, it doesn't. We only charge non-authority outputs. In the fee calculation rule this deposit action will count as an authority output, and we shouldn't consider it. 

I'll add a test for it.

src/nano_contracts/builder.ts
```diff
@@ -197,6 +237,62 @@ class NanoContractTransactionBuilder {
...
+          version: tokenInfo.version,
+        };
+      }
+      tokensMap.set(uid, tokenData);
+    }
+
+    // Calculate fee from transaction outputs
+    let fee = await Fee.calculate(inputs, outputs as IDataOutputWithToken[], tokensMap);
+
+    if (this.isCreatingFeeToken) {
```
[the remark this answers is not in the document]
raul-oliveira: I'll check it. But if I'm not wrong, in this case you only pay 1 htr. Because the HTR payment of the token creation is based in the outputs. And by the sum it will count as 1 htr (deposit counts as outputs)

src/nano_contracts/builder.ts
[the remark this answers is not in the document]
raul-oliveira: Done in https://github.com/HathorNetwork/hathor-wallet-lib/pull/979/commits/77502a9c9f01f1d47f85effa2ffab1066fc67511

src/nano_contracts/builder.ts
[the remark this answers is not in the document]
raul-oliveira: This was definitely a good catch. It's done in https://github.com/HathorNetwork/hathor-wallet-lib/pull/979/commits/2216a96d8cd70aa78b31cdad4ed9ab2649b177d2

src/nano_contracts/builder.ts
```diff
@@ -197,6 +237,62 @@ class NanoContractTransactionBuilder {
...
+        tokenData = {
+          uid,
+          name: tokenInfo.name,
+          symbol: tokenInfo.symbol,
+          version: tokenInfo.version,
+        };
+      }
+      tokensMap.set(uid, tokenData);
+    }
+
```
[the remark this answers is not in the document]
raul-oliveira: Done in [7cbc4d8](https://github.com/HathorNetwork/hathor-wallet-lib/pull/979/commits/7cbc4d8964eaad02fab74ab2ca89108f83e17d83)

src/new/types.ts
```diff
@@ -243,6 +243,8 @@ export interface WalletWebSocketData {
  */
 export interface CreateNanoTxOptions {
   pinCode?: string | null;
+  /** Optional maximum fee in HTR. If not set, fee is auto-calculated without limit. */
+  maxFee?: OutputValueType;
 }
```
r4mmer: Why is this duplicated here? and why is it missing a property?

src/nano_contracts/types.ts
```diff
@@ -225,6 +227,14 @@ export type CreateNanoTxData = {
   args?: unknown[];
 };
 
+export type CreateNanoTxOptions = {
```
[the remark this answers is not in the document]
raul-oliveira: it's using in the `HathorWalletServiceWallet` in the methods:
`createNanoContractTransaction`
`createAndSendNanoContractTransaction`
`createAndSendNanoContractCreateTokenTransaction`
`createNanoContractCreateTokenTransaction`

r4mmer (dismissed):

PR #986, opened 2025-12-30
Test: Shared facades

tuliomir:
> [!IMPORTANT]
> This PR is not supposed to be merged directly to `master`.

This PR and branch are used to help development of the shared tests for the Fullnode and Wallet Service facades. Instead, it will be split among many PRs for review process, which would otherwise be unfeasible.

It depends on #982 and builds upon concepts from the PoC on #947

List of PRs that implement this task:
- #1017
- #1032 
- #1047 
- #1051 
- #1052 
- #1053
- #1076 
- #1077 
- #1078 
- #1079
- #1120 
- #1121
- #1126
- #1128
- #1140 
- #1146 
- #1150


tuliomir:
## Pending review feedback across the shared-facades PRs (draft — to become a tracked issue later)

While going back over the review threads of every PR under this umbrella, I collected the reviewer asks that were **deferred, declined-but-standing, or left unresolved** — i.e. never implemented in their own PR. Posting here informally so we can later turn it into a properly tracked issue and split into follow-up PRs.

Of the 14 implementation PRs, PRs 1052, 1120 and 1126 closed clean with nothing pending. Everything below is grouped by theme, with the PR and reviewer that raised it.

### 1. Facade inconsistency / adapter design (the big theme — mostly RFC 100 territory)
- **PR 1051** — `getBalance` — @pedroferreira1: make the wallet-service `getBalance` accept a `tokenUid` arg to match the fullnode signature (a breaking change deliberately kept out to limit PR scope).
- **PR 1053** — `service.adapter.ts` — @raul-oliveira: `getTx()` only exists as an adapter workaround; it needs a real Wallet Service implementation.
- **PR 1032** — `start.test.ts` — @r4mmer: move the `try/catch` into the adapter rather than the shared test (follow-up suggestion never answered).
- **PR 1076** — `adapters/types.ts` — @raul-oliveira: use a union of the two facade types instead of redeclaring types in `adapters/types` (the "single pass moving every type to its right place" was promised but not scheduled).

### 2. Type-safety / unsafe casts / dead code
- **PR 1078** — `utxos.test.ts` — @coderabbitai: drop the `{ authorities: 3n } as unknown as Parameters<…>` cast — either surface `authorities` in the fullnode options type or assert at the API layer. _(Tracked in the open issue 1107.)_
- **PR 1076** — `create-token.test.ts` — @r4mmer: remove the nullish handling of `addresses` — it can never be nullish (dead conservative path).
- **PR 1079** — service-specific teardown — @coderabbitai: `(e as Error).message` yields `undefined` for non-Error throws; normalize with `e instanceof Error ? e.message : String(e)`.

### 3. Error handling / silent failures
- **PR 1079** — `src/wallet/wallet.ts` (via `addresses.test.ts`) — @r4mmer: make wallet `stop()` idempotent / throw a typed "already stopped" error so teardown can stop swallowing every error. This is a lib-level change, not test-only.
- **PR 1017** — `genesis-wallet.helper.ts` — @raul-oliveira: guard the null/undefined logger prop that currently masks the underlying error (suggested as a higher-level guard for all logger instances).
- **PR 1051** — service-specific `getBalance` — @raul-oliveira: assert the error is specifically the wallet-stopping error, not just any error.
- **PR 1077** — `hathorwallet_facade.test.ts` — @r4mmer: `await` several async methods that are currently fired without awaiting.
- **PR 1047** — `internal.test.ts` — @coderabbitai: the axios `/version` catch returns `{}`, then the test asserts `.status === 200` on that empty object — let it throw instead.

### 4. Code duplication / shared-helper extraction
- **PR 1121** — `mint-tokens.test.ts` — @raul-oliveira: a helper already exists for this; extract a shared util instead of re-implementing.
- **PR 1121** — `token-authority.test.ts` — @raul-oliveira: reuse `tokenBalance()` for `htrBalance` rather than duplicating.
- **PR 1077** — `hathorwallet_facade.test.ts` — @r4mmer: replace the local `validateFeeAmount` helper with the shared `expectFeeAmount` util.
- **PR 1128** — `metadata-voiding.test.ts` — @coderabbitai: extract the repeated per-test setup (`setupVoidingScenario`); ~20 duplicated lines across four `it` blocks.

### 5. Test isolation / lifecycle / flakiness
- **PR 1121** — `token-authority.test.ts` — @raul-oliveira: move the `stopWallet` call into an `afterEach` hook.
- **PR 1032** — `start.test.ts` — @coderabbitai: replace the `delay(1000)` before `stop()` with a deterministic tx-confirmation wait (flakiness risk).

### 6. Naming
- **PR 1121** — `token-authority.test.ts` — @raul-oliveira: rename the `auth*` variables to `authorityAddress` / `authorityAtAddress` — they read as authn/authz. _(This is the one thing the reviewer explicitly wanted addressed before merge, so best fixed in the PR itself.)_
- **PR 1032** — `fullnode.adapter.ts` — @coderabbitai: `startedWallets` actually holds non-started wallets — rename to `trackedWallets`.

### 7. Test coverage gaps & skipped tests awaiting upstream
- **PR 1053** — `send-transaction-tokens.test.ts` — @r4mmer: skipped "custom token after HTR" test was parked on issue 1057. **That issue is now closed/fixed — the test can likely be unskipped now.**
- **PR 1051** — service-specific `getBalance` — @raul-oliveira / @pedroferreira1: unskip the empty-wallet / specific-token balance tests once the wallet-service returns a zero-balance entry instead of an empty array (relates to the open issue 397). A tracking issue was requested and is still pending.
- **PR 1051** — fullnode `getBalance` — @pedroferreira1: promote the "unknown token balance" test to the shared suite.
- **PR 1121** — `service.adapter.ts` — @raul-oliveira: also assert the balance stays unchanged after destroying the authority (non-blocker).
- **PR 1047** — `internal.test.ts` — @coderabbitai: add `timestamp` to the `rawToFacade` mapping so the `/version` timestamp is verified too.
- **PR 1053** — fee-header validation — @raul-oliveira: handle fee headers carrying multiple tokens with varying amounts (scoped out for future multi-fee-token support).
- **PR 1078** — `service.adapter.ts` — @coderabbitai: broaden filter-parameter coverage and guard the `bigint`→`number` narrowing. _(The concrete fix landed; the broader coverage is tracked in the open issue 1107.)_

### 8. Test-environment coupling / config
- **PR 1047** — `fullnode.adapter.ts` — @raul-oliveira: don't couple the adapter to the public testnet URL.
- **PR 1032** — `start.test.ts` — @coderabbitai: network mismatch — keys derived with `privatenet` while the wallet connects to `testnet`; align them (possible correctness bug).

### 9. Tracking-issue debt (cross-cutting)
Several deferrals were promised a follow-up issue that was never confirmed created — the "will be forgotten" pile this doc is meant to absorb:
- **PR 1076** — "do you have an issue to catalog these type redeclarations?" — unanswered.
- **PR 1077** — "create an issue to track the un-awaited async methods" — unconfirmed.
- **PR 1077 / 1079** — follow-ups for the `expectFeeAmount` refactor and the idempotent `stop()` — promised, unconfirmed.
- **PR 1051** — "open an issue for the skipped balance tests" — pending.

---

### Rough grouping for the eventual follow-up PRs
1. **Facade type unification** — PRs 1076, 1078, 1051 (`tokenUid`), 1053 (`getTx`); the RFC 100 cluster, biggest and most breaking.
2. **Lib-level error semantics** — idempotent `stop()` (1079), logger guard (1017), un-awaited async (1077); source changes, not test-only.
3. **Test-helper de-duplication & isolation** — 1121, 1128, 1077, plus the `afterEach` and deterministic-wait asks; pure test refactors, low risk.
4. **Re-enable skipped tests** — 1053 (issue 1057 now closed), 1051 empty-wallet tests; quick coverage wins.
5. **Naming + nitpick sweep** — the 1121 authority names ideally land in that PR before merge, not here.

tuliomir:
## Status checkpoint — what is left before this PR can be closed

`#1146` merged today (`1bef9d61`), which empties the address-info block out of `hathorwallet_others.test.ts`. Recording here what the project still owes, so the closing criteria for this umbrella PR are explicit.

### 1. Migrations still in flight

| PR | Scope | State |
| --- | --- | --- |
| #1140 | token creation — `create token with data outputs` + `createNFT` + `getToken methods` → **shared** | open, awaiting review |
| #1150 | `getFullTxById` + `getTxConfirmationData` + `graphvizNeighborsQuery` → **fullnode-specific** | open, awaiting review |

### 2. Migrations not yet started

Block map against `origin/master` @ `1bef9d61` (line numbers from that revision):

| File | Remaining blocks | Bucket guess (unverified) |
| --- | --- | --- |
| `hathorwallet_facade.test.ts` (989 lines) | `meltTokens` — plain HTR (293) | shared — WS melt exists since #1077; deletion range is adjacent to both #1140 and #1150, so open it once either lands |
| | `template methods` (25), `getWalletInputInfo` (95), `signTx` (770), `getTxHistory` (811), `storage methods` (952) | shared (verify each) — small scattered blocks, a mop-up wave |
| `hathorwallet_others.test.ts` (44 lines) | `index-limit address scanning policy` (8) | likely fullnode-specific (scanning policy) |
| `walletservice_facade.test.ts` (193 lines) | `single-address mode` (89) | shared — verify fullnode support; fold in the single-address fixture hardening (inline seed → precalculated wallets, self-contained tests) |
| | `describe.skip('websocket events', () => {})` (83) | an empty skipped block — decide whether to implement or delete, don't migrate it as-is |

Standing caveat: every "shared" guess above is **unverified**. Bucket guesses have flipped repeatedly on close inspection (history/query, metadata/voiding and 2 of 3 address-info blocks all went fullnode-specific; #1150 kept its bucket but on entirely different evidence than predicted). Evidence-check each against actual wallet-service support — is the method implemented there, and does it return a matching shape — before assuming it lands shared. Bucketing can also be per-*behavior* rather than per-block: `checkAddressesMine` was split in #1146 because only its malformed-input case is fullnode-only.

### 3. Review follow-ups inherited from already-merged PRs

These merged with review threads left open. They are not blockers for the migrations above, but they should be closed out before this umbrella PR is.

**#1146 — 7 open threads, all from @raul-oliveira, all the same theme: comment density.** Worth one small follow-up PR rather than seven scattered fixes:

- `adapters/types.ts` — the `checkAddressesMine` docblock explains *how each facade resolves the call*, which is exactly what the adapter abstracts away, and pins the file to `src/` internals ([r3666189990](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666189990))
- `adapters/types.ts` — 9-line docblock on the one-line `CheckAddressesMineResult` alias; its second paragraph argues for a decision `= WalletAddressMap` already states ([r3666190025](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190025))
- `fullnode-specific/address-info.test.ts:116` — comment restates the call below it; the `injectFunds` wait behaviour is the part actually worth keeping ([r3666190040](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190040))
- `fullnode-specific/address-info.test.ts:144` — same note repeated; stating it once is enough ([r3666190051](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190051))
- `fullnode-specific/address-info.test.ts:276` — base58 reasoning already spelled out in the file header ~150 lines up ([r3666190062](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190062))
- `shared/addresses.test.ts:200` — `foreignAddress` + `WALLET_CONSTANTS.genesis` already say "a valid address from another wallet" ([r3666190071](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190071))
- `shared/addresses.test.ts:215` — why the test lives in the shared suite is the file header's job, and the short-circuit/loop detail ties the test to `src/` internals ([r3666190093](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190093))

There is an eighth open thread on `hathorwallet_others.test.ts` ([r3666190105](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1146#discussion_r3666190105)) suggesting the migration list in the header be collapsed to one line. That one was answered — the file is meant to disappear entirely, so keeping it reduced to a list of pointers is a useful progress marker — it just needs resolving, no code change.

**Earlier PRs — unanswered review threads still open:**

- #1121 — 5 unanswered, all @raul-oliveira, all actionable: a helper in `fullnode-specific/mint-tokens.test.ts:41` that may duplicate an existing one; assert the balance is unchanged after `destroyAuthority` (`service.adapter.ts:588`); reuse `tokenBalance` for the HTR balance helper (`shared/token-authority.test.ts:44`); move a `stopWallet` into `afterEach` (`:82`); rename `authorityAtAddr` → `authorityAtAddress` (`:96`)
- #1051 — assert that the raised error is actually the wallet-stop error, so it can't hide something else (`service-specific/get-balance.test.ts:40`)

Everything else still showing as unresolved on the merged PRs (#1017, #1032, #1047, #1052, #1053, #1076, #1077, #1079, #1128) was answered in-thread and simply never marked resolved — no work owed, just housekeeping.

### 4. Definition of done for closing this PR

1. Every block in section 2 either migrated or explicitly ruled out in writing.
2. The three monolithic suites deleted — `hathorwallet_others.test.ts` is already down to a single block, and removing all three is the actual end goal of this project.
3. Section 3's follow-ups landed and the threads resolved.
4. A full integration run green against **both** adapters.

PR #990, opened 2026-01-13
fix: create output script correctly depending on the address type

pedroferreira1:
### Acceptance Criteria
- Wallet service should create the correct output script depending on the address type

### Security Checklist

### DIFF
--- a/__tests__/integration/walletservice_facade.test.ts
+++ b/__tests__/integration/walletservice_facade.test.ts
@@ -626,6 +626,8 @@ describe('basic transaction methods', () => {
           tokenData: 0,
         })
       );
+
+      await pollForTx(gWallet, sendTransaction.hash!);
     });
 
     it('should send a transaction to a P2SH (multisig) address', async () => {


__tests__/integration/walletservice_facade.test.ts
```diff
@@ -626,6 +626,41 @@ describe('basic transaction methods', () => {
...
+      await pollForTx(gWallet, sendTransaction.hash!);
+    });
+
+    it('should send a transaction to a P2SH (multisig) address', async () => {
+      // Use a P2SH address from the multisig wallet constants
+      const p2shAddress = WALLET_CONSTANTS.multisig.addresses[0];
+
+      const sendTransaction = await gWallet.sendTransaction(p2shAddress, 5n, {
+        pinCode,
+      });
```
tuliomir: Curious: I would consider a test built on the Genesis wallet instead of a dedicated one to be an incorrect test, but I can´t think of a simpler way to implement this one.

PR #1013, opened 2026-02-09
docs: Fullnode Typings [11] - Connection and Websockets

tuliomir:
### Acceptance Criteria
- Adds typings to the Fullnode Connection methods
- Adds typings to the Start and Stop methods
- Adds typings to the Fullnode Websocket handling methods
- Update the unit and integration tests with the new types
- Adds typings to SendTransaction ( previously done at #1014 )

### Security Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved type safety across wallet APIs (transaction queue, connection state, start/stop options) and added a connection-state accessor.
  * Start/stop flows now use structured option objects; PIN/password validation messages clarified.
  * Signing method can now be cleared to allow fallback signing behavior.

* **Tests**
  * Strengthened websocket and connection typings in tests; mocks updated for network handling.

* **Chores**
  * Adjusted coverage thresholds for wallet implementation.

### DIFF
--- a/src/new/wallet.ts
+++ b/src/new/wallet.ts
@@ -3296,9 +3296,7 @@ class HathorWallet extends EventEmitter {
    */
   setExternalTxSigningMethod(method: EcdsaTxSign | null): void {
     this.isSignedExternally = !!method;
-    if (method) {
-      this.storage.setTxSignatureMethod(method);
-    }
+    this.storage.setTxSignatureMethod(method);
   }
 
   /**
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -161,9 +161,9 @@ export class Storage implements IStorage {
 
   /**
    * Set the tx signing function
-   * @param {EcdsaTxSign} txSign The signing function
+   * @param txSign The signing function, or a null value to clear it
    */
-  setTxSignatureMethod(txSign: EcdsaTxSign): void {
+  setTxSignatureMethod(txSign: EcdsaTxSign | null): void {
     this.txSignFunc = txSign;
   }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -587,7 +587,7 @@ export interface IStorage {
   setLogger(logger: ILogger): void;
 
   hasTxSignatureMethod(): boolean;
-  setTxSignatureMethod(txSign: EcdsaTxSign): void;
+  setTxSignatureMethod(txSign: EcdsaTxSign | null): void;
   getTxSignatures(tx: Transaction, pinCode: string): Promise<ITxSignatureData>;
 
   // Address methods


src/new/wallet.ts
```diff
@@ -1645,14 +1647,23 @@ class HathorWallet extends EventEmitter {
     let accessData = await this.storage.getAccessData();
     if (!accessData) {
       if (this.seed) {
+        if (!pinCode) {
+          throw new Error(ERROR_MESSAGE_PIN_REQUIRED);
+        }
+        if (!password) {
+          throw new Error(ERROR_MESSAGE_PASSWORD_REQUIRED);
+        }
```
tuliomir: Guarding the types of the variables that are used below

jest-integration.config.js
```diff
@@ -25,10 +25,10 @@ module.exports = {
     },
     // We need a high coverage for the HathorWallet class
     './src/new/wallet.ts': {
-      statements: 92,
-      branches: 85,
+      statements: 90,
+      branches: 83,
       functions: 90,
-      lines: 92,
+      lines: 90,
```
r4mmer: I understand that most of the code changed did not have tests initially, but I don't think we should lower the standards without an approval from @pedroferreira1 

Maybe we can add some tests for the new error cases to increase the line coverage?

src/new/wallet.ts
r4mmer: It is a type guard but also a new error case.

We should check that the wallets don't expect some special case where pin or password are not passed (shouldn't exist this case, but we should check anyway).

src/new/wallet.ts
```diff
@@ -1379,14 +1377,19 @@ class HathorWallet extends EventEmitter {
...
+    let wsData = this.wsTxQueue.dequeue();
 
-    while (wsData !== undefined) {
+    // local type guard to narrow to WalletWebSocketData
+    const isWalletWebSocketData = (obj: unknown): obj is WalletWebSocketData =>
+      obj !== undefined &&
+      obj !== null &&
+      typeof obj === 'object' &&
+      'type' in obj &&
+      typeof (obj as { type: unknown }).type === 'string';
```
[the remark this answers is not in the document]
tuliomir: Thanks for the idea! Implemented on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1013/commits/75f91968fd679543c6b84508a9e96a1944c730de , with some unit test changes to match the stricter typing.

src/new/wallet.ts
tuliomir: `★ Insight ─────────────────────────────────────`
The headless wallet uses a **"default credentials" pattern** — both `getReadonlyWalletConfig` and `getWalletConfigFromSeed` always inject `DEFAULT_PASSWORD` and `DEFAULT_PIN` into the config object (lines 72-73 and 105-106). These are passed to the `HathorWallet` constructor (not to `.start()`), so when `wallet.start()` is called with no args, the `start()` method falls back to `this.pinCode` / `this.password` which were set during construction.
`─────────────────────────────────────────────────`

Here's the full trace for all three controllers:

## Index Controller (`index.controller.js`)

Two paths, both safe:

1. **xpub path** (line 165): calls `getReadonlyWalletConfig({ xpub, ... })` → sets `pinCode: DEFAULT_PIN`, `password: DEFAULT_PASSWORD`
2. **seed/seedKey path** (line 200): calls `getWalletConfigFromSeed({ seed, ... })` → sets `pinCode: DEFAULT_PIN`, `password: DEFAULT_PASSWORD`

Both configs are passed to `new HathorWallet(hydratedWalletConfig)` at construction. Then `wallet.start()` on line 140 passes no args, but `start()` falls back to `this.pinCode` / `this.password` from the constructor. **No crack.**

## HSM Controller (`hsm.controller.js`)

- Line 119: calls `getReadonlyWalletConfig({ xpub: xPub })` → sets `pinCode: DEFAULT_PIN`, `password: DEFAULT_PASSWORD`
- Line 125: calls `startWallet(walletId, walletConfig, config, { hsmKeyName })`

Same pattern — credentials are in the constructor config. **No crack.**

## Fireblocks Controller (`fireblocks.controller.js`)

- Line 107: calls `getReadonlyWalletConfig({ xpub })` → sets `pinCode: DEFAULT_PIN`, `password: DEFAULT_PASSWORD`
- Line 110: calls `startWallet(walletId, walletConfig, config)`

Same pattern. **No crack.**

## The Safety Net: `wallet.helper.js`

Both helper functions (`getReadonlyWalletConfig` at line 63, `getWalletConfigFromSeed` at line 81) **always** hardcode credentials:

```js
// wallet.helper.js:70-73
const walletConfig = {
  xpub,
  password: DEFAULT_PASSWORD,  // always set
  pinCode: DEFAULT_PIN,        // always set
  ...
};
```

With a comment explicitly noting: *"Previous versions of the lib would have password and pin default as '123'. We currently need something to be defined, otherwise we get an error when starting the wallet."* ( [link to code](https://github.com/HathorNetwork/hathor-wallet-headless/blob/feb35f3d9a13c5ee30a338726c93c2f1ae139444/src/helpers/wallet.helper.js#L68-L69) )

## Verdict

**All three controllers are safe.** Every path through the headless wallet passes `pinCode` and `password` at constructor time via the wallet helper functions. The `wallet.start()` call with no args works fine because it falls back to `this.pinCode` and `this.password`. Your error handling changes in `start()` won't break any headless wallet flows.

__tests__/integration/hathorwallet_facade.test.ts
```diff
@@ -362,7 +363,12 @@ describe('start', () => {
       () =>
         new HathorWallet({
           seed: walletData.words,
-          connection: { state: ConnectionState.CONNECTED },
+          connection: {
+            state: ConnectionState.CONNECTED,
+            getState(): ConnectionState {
```
raul-oliveira: since you already added the getState in the connectionFile. What about use it here?

__tests__/new/hathorwallet.test.ts
```diff
@@ -389,8 +394,11 @@ test('handleWebsocketMsg', async () => {
   });
 
   // wsTxQueue is not part of the prototype so it won't be faked on FakeHathorWallet
-  hWallet.wsTxQueue = new Queue();
-  hWallet.wsTxQueue.enqueue({ type: 'wallet:address_history', history: [1] });
+  hWallet.wsTxQueue = new Queue<WalletWebSocketData>();
+  hWallet.wsTxQueue.enqueue({
+    type: 'wallet:address_history',
+    history: [1] as unknown as IHistoryTx,
+  });
```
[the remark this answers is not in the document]
tuliomir: I'm aware of this. Please keep raising this point in other PRs, but the context of this specific PR is only to change the typings, and not the code or the tests.

src/new/wallet.ts
```diff
@@ -3281,32 +3291,41 @@ class HathorWallet extends EventEmitter {
...
+   *
+   * @param method The external transaction signing method, or null to clear
    */
-  setExternalTxSigningMethod(method: any): any {
+  setExternalTxSigningMethod(method: EcdsaTxSign | null): void {
     this.isSignedExternally = !!method;
-    this.storage.setTxSignatureMethod(method);
+    if (method) {
+      this.storage.setTxSignatureMethod(method);
+    }
```
r4mmer: Why do we only write to storage if the method is present?
This makes it so we cannot "clean" the tx sign method.

I know this is a very unlikely case, but we should at least make it possible, right?

tuliomir: Good catch. The problem was that the storage method itself was incorrectly typed. I had to change two more files to fix it, and then the logic I broke was fixed as well. Done on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1013/commits/985d170c92e774d86a6cb941b3122735c91bc46e

PR #1026, opened 2026-03-05
docs: Fullnode Typings [15] - Strict typing rules

tuliomir:
### Acceptance Criteria
- Sets typescript rules on the `src/new/wallet.ts` to be strict
- Finishes the round of fullnode typings
- Do not change any code or test: only map the types to the current state of the application, not the desired future one

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Storage stop/cleanup can now optionally include token cleanup.
  * Wallet API now exposes a read-only auth token retrieval capability.

* **Improvements**
  * Token minting and transaction outputs accept omitted fields and now auto-derive missing values.
  * Wallet API typings strengthened for clearer readiness signaling and connection handling.

* **Documentation**
  * Minor JSDoc and inline comment clarifications.


src/new/wallet.ts
```diff
@@ -3093,7 +3080,7 @@ class HathorWallet extends EventEmitter {
 
     const nc = await builder.build();
     if (newOptions.signTx !== false) {
-      return prepareNanoSendTransaction(nc, pin, this.storage);
+      return prepareNanoSendTransaction(nc, pin!, this.storage);
```
tuliomir: Line 3048 validates the pin is available, but it would take an additional guard to make Typescript identify this. I found it better to just use `!` for now, instead of creating more lines of code and hence lowering the code coverage for this class even further.

PR #1038, opened 2026-03-18
feat: single address policy

r4mmer:
### Acceptance Criteria

- Create a single address scanning policy
- Enable both facades to use the scanning policy

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Single-address scanning mode added; wallets can start in, enable/disable, and report address mode at runtime.
  * Public APIs to enable/disable single/multi address modes and to query current address mode.

* **Bug Fixes**
  * Address index no longer advances when single-address mode is active.
  * Enabling single-address mode is prevented when transactions exist outside the first address (explicit error reported).

* **Tests**
  * Comprehensive tests covering single-address behavior across storage, utils, types, and wallet flows.


src/wallet/wallet.ts
```diff
@@ -1474,6 +1492,10 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
    * @inner
    */
   getCurrentAddress({ markAsUsed = false } = {}): AddressInfoObject {
+    if (this.singleAddress) {
+      return this.newAddresses[0];
```
pedroferreira1: Why don't you return `this.firstAddress`? Then we don't even need to call `getNewAddresses`

src/new/wallet.ts
```diff
@@ -755,12 +776,23 @@ class HathorWallet extends EventEmitter {
...
+    for await (const gotTx of loadAddressHistory(addresses, this.storage, false)) {
+      if (gotTx) {
+        // This will signal we have found a transaction when syncing the history
+        foundAnyTx = true;
+        break;
       }
     }
-    return false;
+
+    return foundAnyTx;
```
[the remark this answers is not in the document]
r4mmer: Since the most predominant address scanning policy is gap limit with 20 addresses it means that if there is no address in the first 20 addresses then no more addresses will load, which means that to even have a tx on the 21 address we need some tx in one of the first 20 so checking the first 20 is safe enough.

__tests__/wallet/enableSingleAddressMode.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+  });
+
+  it('should throw if wallet is not ready', async () => {
+    await expect(wallet.enableSingleAddressMode()).rejects.toThrow('Wallet not ready');
+  });
+
+  it('should throw if there are transactions outside first address', async () => {
+    // Force wallet to ready state
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
r4mmer: This is just to get around the "setting a private variable" protection.

src/wallet/wallet.ts
```diff
@@ -1237,6 +1263,9 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
     // Derive walletId from xpub
     const walletId = HathorWalletServiceWallet.getWalletIdFromXPub(this.xpub);
     this.walletId = walletId;
+    if (this.singleAddress) {
+      await this.storage.setScanningPolicyData({ policy: SCANNING_POLICY.SINGLE_ADDRESS });
```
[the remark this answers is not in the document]
r4mmer: Since the wallet-service facade does not use the storage scanning policy and this is just to avoid drifting the storage and facade states I think this is unnecessary

__tests__/integration/hathorwallet_facade.test.ts
tuliomir: question(non-blocking): I think this test was already implemented on the shared suite and could be removed from here, doesn't it?

tuliomir (dismissed):

src/new/wallet.ts
```diff
@@ -615,6 +623,19 @@ class HathorWallet extends EventEmitter {
...
+          const scanPolicy = await this.storage.getScanningPolicy();
+          if (scanPolicy === SCANNING_POLICY.SINGLE_ADDRESS) {
+            try {
+              await this.enableSingleAddressMode();
+            } catch (err) {
+              if (err instanceof HasTxOutsideFirstAddressError) {
+                this.scanPolicy = { policy: SCANNING_POLICY.GAP_LIMIT, gapLimit: 20 };
+                await this.storage.setScanningPolicyData(this.scanPolicy);
+              } else {
+                throw err;
```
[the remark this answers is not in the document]
r4mmer: This branch is untested but it is a throw inside a try..catch which is tested, so I think this is safe

__tests__/integration/hathorwallet_facade.test.ts
r4mmer: Done

src/new/wallet.ts
```diff
@@ -780,7 +813,6 @@ class HathorWallet extends EventEmitter {
       } else {
         address = await deriveAddressP2SH(index, this.storage);
       }
-      await this.storage.saveAddress(address);
```
pedroferreira1, line 783: Why did you remove this? I understand that for single address policy this is not necessary but for the others we also don't need it?

src/new/wallet.ts
```diff
@@ -3481,6 +3513,81 @@ class HathorWallet extends EventEmitter {
...
+      return;
+    }
+
+    // Switch policy in storage
+    await this.storage.setScanningPolicyData({
+      policy: SCANNING_POLICY.SINGLE_ADDRESS,
+    });
+
+    // Load new addresses
+    await this.reloadStorage();
```
pedroferreira1, line 3589: You call the method `enableSingleAddress` when loading the wallet for the first time. Does it make sense to call this method also? If it's not necessary, maybe we could just have a parameter here

src/wallet/wallet.ts
```diff
@@ -141,12 +145,6 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
   // State of the wallet. One of the walletState enum options
   private state: string;
 
-  // Variable to prevent start sending more than one tx concurrently
-  private isSendingTx: boolean;
-
-  // ID of tx proposal
-  private txProposalId: string | null;
-
```
pedroferreira1, line 149: Aren't they in use?

src/wallet/wallet.ts
pedroferreira1: `firstAddress` is returned by internal methods. If it's easy to transform it in address info, would be good. Otherwise, it's just moving one API call from a place to another, so it's fine

src/new/wallet.ts
r4mmer: Well, when I implemented this the idea was that if we run `wallet.getAddressAtIndex(10000)` and send a tx there we would like to see it arrive in the wallet, correct?
But this change does not persist through a restart of the wallet since the address scanning policies dictate which addresses to check.
So imagine having 10 HTR in your wallet, then resetting and it is now empty... it is there, just not where you can find it.

So ultimately this is not "needed" by any policy and now it also breaks the single address policy so I decided to remove it.

src/new/wallet.ts
r4mmer: It is specially required to call this method when starting for the first time.
For normal use-cases it does not make sense, but the pre-calculated addresses feature we use in tests makes the storage come with unwanted addresses loaded which we need to delete so we can start in true single address mode.

src/wallet/wallet.ts
r4mmer: Could not find a single use for them, and since they are private they cannot be accessed by an outside caller.
So it does not appear that they are in use.

src/wallet/wallet.ts
r4mmer: Basically this is what we do in `prepareSingleAddressMode`

```ts
    this.newAddresses = [
      {
        address: this.firstAddress,
        index: 0,
        addressPath: await this.getAddressPathForIndex(0),
      },
    ];
    this.indexToUse = 0;
```

We transform the `firstAddress` into address info and save it on `newAddresses`, this way we can use the normal internal usage of the wallet facade to make the single address mode work just like the multi address mode.

It's just that instead of doing it on the `getCurrentAddress` call we call it when the address loads then just use the saved value.

pedroferreira1: I think we can keep as it is

pedroferreira1 (dismissed):

pedroferreira1 (approved):

tuliomir (approved):

PR #1045, opened 2026-03-21
fix(utils): make bigIntReviver engine-agnostic by checking SyntaxError

### LINKED ISSUE #1018: `bigIntReviver` crashes on non-V8 engines (Bun/JSC) due to engine-specific error message checks
`JSONBigInt.bigIntReviver` in `lib/utils/bigint.js` relies on V8-specific `SyntaxError` messages to detect when `BigInt()` fails on float/scientific-notation numbers (e.g. `1.5`, `1e2`). This breaks on Bun (JavaScriptCore) and potentially other JS engines.

**Current behavior (on Bun):**

Any websocket message or JSON response containing a float crashes the process:

```
unexpected error in bigIntReviver: SyntaxError: Failed to parse String to BigInt
SyntaxError: Failed to parse String to BigInt
    at bigIntReviver (node_modules/@hathor/wallet-lib/lib/utils/bigint.js:58:13)
    at onMessage (node_modules/@hathor/wallet-lib/lib/websocket/index.js:41:40)
```

**Root cause:**

The catch block at [lib/utils/bigint.js:50](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/utils/bigint.ts) checks for two specific error messages:

```js
e.message === `Cannot convert ${context.source} to a BigInt`  // V8 (Node/Chrome)
e.message === `invalid BigInt syntax`                          // V8 variant
```

But different engines use different messages for the same error:

| Engine | Runtime | Error message |
|--------|---------|---------------|
| V8 | Node.js, Chrome | `Cannot convert 1.5 to a BigInt` |
| V8 (variant) | Node.js | `invalid BigInt syntax` |
| JavaScriptCore | Bun, Safari | `Failed to parse String to BigInt` |
| SpiderMonkey | Firefox, Deno (future?) | `can't convert 1.5 to BigInt` |

**Suggested fix:**

Instead of matching engine-specific error messages, simply catch any `SyntaxError` from `BigInt()`. The only reason `BigInt(context.source)` throws a `SyntaxError` is when the source is a float or scientific notation — which is exactly the case being handled. There's no need to distinguish between error messages:

```js
} catch (e) {
  if (e instanceof SyntaxError) {
    // Float or scientific notation (e.g. '1.5', '1e2') — keep as number.
    return value;
  }
  throw e;
}
```

This is engine-agnostic and handles the exact same set of cases.

**Environment:**

- `@hathor/wallet-lib`: 2.14.0
- Runtime: Bun 1.3.9 (JavaScriptCore)
- OS: Linux x64

Sigmabrogz:
Fixes #1018

### Description
`JSONBigInt.bigIntReviver` currently relies on matching specific string messages for `SyntaxError`s when parsing floats (like `1.5` or `1e2`). Since V8, JavaScriptCore (Bun/Safari), and SpiderMonkey throw different error messages, this breaks the reviver on non-Node/Chrome runtimes (like Bun).

By simply catching `e instanceof SyntaxError`, the code becomes engine-agnostic while still preserving the original logic of keeping the value as a number if it cannot be converted to a `BigInt` (i.e. if it's a float/scientific notation).

### Changes
- Replaced the string-matching `e.message === ...` logic with a simple `e instanceof SyntaxError` check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of BigInt conversion error handling to work consistently across all JavaScript engines, ensuring decimal and exponential number formats are handled correctly.

### DIFF
--- a/src/utils/bigint.ts
+++ b/src/utils/bigint.ts
@@ -42,13 +42,10 @@ export const JSONBigInt = {
       // Otherwise, we can keep it as a Number.
       return value;
     } catch (e) {
-      if (
-        e instanceof SyntaxError &&
-        (e.message === `Cannot convert ${context.source} to a BigInt` ||
-          e.message === `invalid BigInt syntax`)
-      ) {
+      if (e instanceof SyntaxError) {
         // When this error happens, it means the number cannot be converted to a BigInt,
         // so it's a double, for example '123.456' or '1e2'.
+        // We catch any SyntaxError to be engine-agnostic (V8 vs JavaScriptCore vs SpiderMonkey).
         return value;
       }
       // This should never happen, any other error thrown by BigInt() is unexpected.


tuliomir (approved):

glevco (changes requested):

glevco (approved):

tuliomir:
Superseded by #1130

PR #1047, opened 2026-03-26
test: Shared tests for internal wallet methods

tuliomir:
### Acceptance Criteria
- Moves the internal methods integration tests to the shared tests structure

## Summary

Second step of the integration test file splitting: extracts the "internal methods" block from `hathorwallet_others.test.ts` and the "wallet public methods" block from `walletservice_facade.test.ts` into dedicated, specifically-scoped test files.

- Shared network query tests (`getServerUrl`, `getNetwork`, `getNetworkObject`, `getVersionData`) now run against both facades via `describe.each`
- Shared `changeServer` contract test verifies both facades accept URL changes without throwing
- Fullnode-specific tests for debug mode, storage reload, and server change with `getVersionData` validation
- Adds `originalServerUrl` to `IWalletTestAdapter` for safe server revert across facades
- Includes direct fullnode `/version` cross-check to verify both facades return consistent data

## Breaking change

`HathorWallet.changeServer(url)` signature changed from `void` to `Promise<void>`, matching the service facade. `changeServer` was also added to the `IHathorWallet` interface.

**Migration:** Add `await` to any `changeServer()` calls. Callers already using `await` (like the service facade) are unaffected.

## New file structure

```
__tests__/integration/
├── shared/
│   ├── start.test.ts          # (PR #1032)
│   ├── internal.test.ts       # NEW — network query methods
│   └── server_changes.test.ts # NEW — portable changeServer contract
├── fullnode-specific/
│   ├── start.test.ts          # (PR #1032)
│   ├── internal.test.ts       # NEW — debug mode, storage reload
│   └── server_changes.test.ts # NEW — changeServer + getVersionData
└── service-specific/
    └── start.test.ts          # (PR #1032)
```

## Summary by CodeRabbit

* **New Features**
  * Wallet changeServer is now asynchronous and returns a Promise for safer async handling.

* **Tests**
  * Added/expanded integration suites for server-switching and internal wallet behaviors across adapters.
  * Test adapters now expose an original server URL so tests reliably revert server changes.
  * Several legacy wallet test blocks were removed or reorganized into the new adapter-driven suites.


__tests__/integration/hathorwallet_others.test.ts
```diff
@@ -1617,94 +1611,6 @@ describe('getAuthorityUtxos', () => {
   });
 });
 
-// This section tests methods that have side effects impacting the whole wallet. Executing it last.
```
raul-oliveira: <3

raul-oliveira (approved):

src/new/wallet.ts
```diff
@@ -470,7 +470,7 @@ class HathorWallet extends EventEmitter {
    * @memberof HathorWallet
    * @inner
    * */
-  changeServer(newServer: string): void {
+  async changeServer(newServer: string): Promise<void> {
```
pedroferreira1, line 473: Is this the first step of the breaking changes planned in the RFC?

pedroferreira1 (approved):

src/new/wallet.ts
tuliomir: Actually, it's an unplanned change that appeared during the shared tests building.

The `changeServer()` wasn't discovered during the initial research of RFC 100, but this shared test suite identified this need. It's in line with all the changes proposed by the RFC, though: unifying the behaviors of both facades as much as possible.

PR #1051, opened 2026-03-31
Test: Shared getBalance tests for both facades

tuliomir:
## Summary

- Extract `getBalance` tests from `hathorwallet_facade.test.ts` and `walletservice_facade.test.ts` into the shared test framework
- Create `shared/get-balance.test.ts` with adapter-driven tests that run against both facades
- Move fullnode-only tests (custom token balance, mandatory tokenUid, internal transfer) to `fullnode-specific/get-balance.test.ts`
- Move service-only tests (no-arg getBalance, not-ready error, skipped empty-wallet bugs) to `service-specific/get-balance.test.ts`

## Acceptance Criteria

- Shared balance tests pass for both fullnode and wallet-service adapters
- No `getBalance`/`balances` describe blocks remain in the monolithic facade files

## Test plan

- [ ] `npm run test` passes
- [ ] `npm run test_integration` passes (shared tests run against both facades)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Reorganized balance tests: consolidated shared cases and added fullnode- and service-specific integration suites.
  * Added assertions for: requiring a token id for token-specific queries; genesis wallet visibility for newly created tokens; wallet readiness errors when not started; native balance after injecting funds; internal transfers not changing unlocked balance.
  * Removed duplicated legacy balance suites and added a few skipped regression markers for known behavior differences.


__tests__/integration/fullnode-specific/get-balance.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+describe('[Fullnode] getBalance', () => {
+  afterEach(async () => {
+    await stopAllWallets();
+  });
+
+  it('should reject when tokenUid is not provided', async () => {
+    const hWallet = await generateWalletHelper();
+    await expect(hWallet.getBalance()).rejects.toThrow();
+  });
```
pedroferreira1, line 31: This is different in both facades? Wallet service accepts no `tokenUid`?

__tests__/integration/fullnode-specific/get-balance.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const { hWallet: gWallet } = await GenesisWalletHelper.getSingleton();
+    const genesisTknBalance = await gWallet.getBalance(tokenUid);
+    expect(genesisTknBalance).toHaveLength(1);
+    expect(genesisTknBalance[0]).toMatchObject({
+      token: { id: tokenUid },
+      balance: { unlocked: 0n, locked: 0n },
+      transactions: 0,
+    });
+  });
+});
```
pedroferreira1, line 66: Why isn't it in the shared tests?

__tests__/integration/service-specific/get-balance.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const htrBalance = balances.find(b => b.token.id === NATIVE_TOKEN_UID);
+    expect(htrBalance).toBeDefined();
+    expect(typeof htrBalance?.balance).toBe('object');
+  });
+
+  // FIXME(wallet-service): getBalance() on an empty wallet should return a single
+  // entry with 0 balance for the native token, but currently returns an empty array.
+  // Ref: https://github.com/HathorNetwork/hathor-wallet-lib/issues/397
+  it.skip('should return balance array for empty wallet', async () => {
```
pedroferreira1, line 65: We had fixed that and the next one, not sure why it's not working

__tests__/integration/fullnode-specific/get-balance.test.ts
tuliomir: Yes, it's different.

Wallet Service: https://github.com/HathorNetwork/hathor-wallet-lib/blob/c6636205d8e08bf4d7f7a86cb4911ae5205b6ed6/src/wallet/wallet.ts#L895-L897

Fullnode: https://github.com/HathorNetwork/hathor-wallet-lib/blob/a69b16cb964b643a57d236656d056e3b284b3cbb/src/new/wallet.ts#L865-L870

This is one breaking change worth making, but I decided not to make this PR more complex than it already is.

__tests__/integration/fullnode-specific/get-balance.test.ts
tuliomir: **TL;DR:** It's blocked by a wallet-service inconsistency (issue #397). Once the wallet-service returns a zero-balance entry instead of an empty array for unknown tokens, this test can be promoted to the shared suite.

--

The fullnode facade returns a **zero-balance entry** (line 60-64 of the test) when you ask for the balance of a token the wallet has never interacted with — the fullnode always synthesizes a response. The wallet-service facade, on the other hand, has a known bug where it returns an **empty array** for tokens the wallet doesn't hold (see the two `FIXME(wallet-service)` comments at lines 62-64 and 79-80 of `service-specific/get-balance.test.ts`, referencing issue #397).

Since the two facades don't agree on the return shape for "token not on this wallet," the test can't be written with a single shared assertion. Moving it to the shared suite would either require:
1. Different expectations per adapter (defeating the purpose of shared tests), or
2. The wallet-service bug to be fixed first so both facades return a consistent zero-balance entry.

The test was intentionally kept fullnode-specific until the wallet-service behavior converges.

__tests__/integration/service-specific/get-balance.test.ts
tuliomir: I tried removing all the `skip`s but the tests just failed. Here's a small report on it:

--

1. should return balance array for empty wallet — wallet-service still returns an empty array (`length: 0`) instead of one entry with zero balance (`length: 1`)

2. should return balance for specific token when token parameter is provided — similarly, the assertion at line `83` expects `balances.length` to be `1`, but the first test at line `69` already shows it's `0`.

tuliomir: Because of this investigation above I ended up identifying another specific test that could to go Shared.

It's not perfect since it cannot validate the value, but it's better than a skipped specific test anyway. Done on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1051/commits/08f37efb77dcf07a3b5572f187ea5f50f544cfff

__tests__/integration/service-specific/get-balance.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+describe('[Service] getBalance', () => {
+  let wallet: HathorWalletServiceWallet;
+
+  afterEach(async () => {
+    if (wallet) {
+      try {
+        await wallet.stop({ cleanStorage: true });
+      } catch {
+        // Wallet may already be stopped
```
raul-oliveira, line 40: it's good to see if the error that is being raised is related to the wallet stopping. Otherwise it could hide something.

__tests__/integration/service-specific/get-balance.test.ts
raul-oliveira: @tuliomir did you run with a updated image of wallet-sevice? This is a known issue of fixes not being applied to the integration tests due to the usage of stale container images.

__tests__/integration/fullnode-specific/get-balance.test.ts
pedroferreira1: Did you test it? We had this problem and Abadesso fixed it in the wallet service. It was implemented December last year

image

__tests__/integration/service-specific/get-balance.test.ts
tuliomir: I just re-created the images locally with the HEAD at https://github.com/HathorNetwork/hathor-wallet-service/commit/947e90f31ba66240eae7713b0dc7014190a66ebd . This has still failed.

The wallet service still does not synthesize a "zero balance" record for HTR when there are no transactions.

Could we solve this in a future PR so that we can advance with the Shared Tests suite construction? Future improvements will be easier to implement when we have that in place.

raul-oliveira: Yep. Open an issue for it pls

raul-oliveira (approved):

pedroferreira1 (approved):

PR #1058, opened 2026-04-07
fix: Auth token race conditions in `start()`

tuliomir:
## Summary

Eliminates race conditions in the wallet-service `start()` auth token flow that caused flaky 403 errors in CI integration tests. Replaces the dual fire-and-forget promise pattern with interceptor-driven on-demand auth, adds timeout bounds and transient error tolerance to polling, and fixes a logical contradiction in `startReadOnly()`.

## Acceptance Criteria

- Wallet-service integration tests (`start.test.ts`) pass consistently without 403 flakes
- `pollForWalletStatus` times out after 60s instead of hanging indefinitely

# Why we're refactoring the auth token flow in `start()`

## The problem in one sentence

The wallet service `start()` method launches auth token renewal as background promises that race with wallet status polling, causing flaky 403 errors in CI.

## What's happening today

When `HathorWalletServiceWallet.start()` runs, it needs to do two things:

1. **Create the wallet** on the wallet-service backend (`POST /wallet/init`)
2. **Get an auth token** so subsequent API calls are authenticated

These two operations have an ordering constraint: the token endpoint requires the wallet to already exist. For a brand-new wallet, the first token request will always fail because the wallet hasn't been created yet.

The current code tries to be clever about this by launching token renewal **in the background** while the wallet is being created, hoping it will finish by the time we need it. Here's the simplified flow:

```
                          start()
                            |
            +---------------+---------------+
            |                               |
  renewPromise = renewToken()     await createWallet()
  (fire-and-forget, may fail)        |
            |                        |
            +-------+----------------+
                    |
              await renewPromise
              (if it failed...)
                    |
            renewPromise2 = renewToken()   <-- another fire-and-forget
                    |
              handleCreate(status, renewPromise2)
                    |
              if status == 'creating':
                  await pollForWalletStatus()  <-- needs auth token!
                  |                               but renewPromise2
                  |                               hasn't been awaited yet
              await renewPromise2              <-- too late, polling already needed it
                  |
              onWalletReady()
```

### The race condition

`pollForWalletStatus()` fires a `setInterval` that calls `getWalletStatus()` every second. Each call needs an auth token. But `renewPromise2` (the token renewal) is running concurrently in the background. There are three possible outcomes:

1. **renewPromise2 finishes first** -- polling finds a valid token. Things work.
2. **Polling fires before renewPromise2 completes** -- `validateAndRenewAuthToken()` is called again from inside the polling tick, racing with `renewPromise2`. Both try to set `this.authToken`. Usually one wins and things work.
3. **renewPromise2 fails silently** -- `renewAuthToken()` has a catch block that sets `this.authToken = null` without throwing. If this happens between another operation reading the token and using it, the request goes out with no token. The wallet-service returns **403 Forbidden**.

Outcome 3 is the flaky failure we see in CI. It depends on exact timing between the polling interval, the token endpoint response time, and JavaScript's microtask scheduling. It's not reproducible on demand, but it happens often enough to block CI.

### Additional problems

**`pollForWalletStatus` can run forever.** It uses `setInterval` with no maximum attempt count. If the wallet-service never returns 'ready', the test (or app) hangs indefinitely.

**`setInterval` callbacks can stack.** If `getWalletStatus()` takes longer than the 1-second interval, multiple callbacks run concurrently, each trying to renew the auth token simultaneously.

**`setInterval` async callbacks swallow errors.** When `getWalletStatus()` throws inside the `setInterval` callback, the rejection is unhandled — the wrapping Promise never settles, and in Node.js 15+ this can terminate the process.

**`startReadOnly()` has a logical contradiction.** When the read-only token request fails, the catch block calls `getWalletStatus()` -- which requires an auth token we just failed to obtain.

**`walletId` is overwritten 4+ times** during startup (lines 455, 476, 526, 487), creating a window where concurrent token renewals might sign with a stale value.

### The new `start()` flow

Replace the concurrent fire-and-forget pattern with interceptor-driven on-demand auth:

```
                    start()
                      |
              generate keys (unchanged)
                      |
              await createWallet()          <-- no auth needed
                      |
              this.walletId = response.walletId   (set ONCE)
              this.authPrivKey = derived key       (set for interceptor)
                      |
              if (!waitReady) return        (early exit, unchanged)
                      |
              if status == 'creating':
                  await pollForWalletStatus()
                  |    └─ getWalletStatus() goes through axios interceptor
                  |       └─ interceptor calls validateAndRenewAuthToken()
                  |          └─ finds authPrivKey in memory, obtains token
                  |       └─ sets Authorization header, fires request
              else if status != 'ready':
                  throw error
                      |
              await onWalletReady()         <-- also goes through interceptor
                      |
              clearSensitiveData()          <-- NOW clears authPrivKey
```

**Why this is better:**

- **No race conditions.** Auth token is obtained on-demand by the axios interceptor (`walletServiceAxios.ts`) before each authenticated request. By the time `pollForWalletStatus` runs, both `walletId` and `authPrivKey` are in memory, so the interceptor can obtain a token without the PIN.
- **No fire-and-forget promises.** Every async operation is awaited.
- **No closure-captured error variables.** The flow is linear; errors propagate naturally through try/catch.
- **`walletId` set once.** No redundant overwrites.
- **Token renewal is self-healing.** Each poll call goes through the interceptor, which checks `validateJWTExpireDate()`. If the token is near expiry, it gets renewed before the request fires — no special handling needed for long polling windows.

**"But won't this be slower?"** The preemptive renewal (the first fire-and-forget) always failed for new wallets because the wallet didn't exist yet. We were paying the cost of a failing HTTP request for zero benefit. For existing wallets (the `WALLET_ALREADY_LOADED` path), the interceptor obtains the token on the first API call, which adds one sequential round-trip. This is a small, acceptable cost for correctness.

### The new `pollForWalletStatus()`

Replace `setInterval` with a bounded `for` loop with transient error tolerance:

```typescript
async pollForWalletStatus(): Promise<void> {
  let lastError: Error | null = null;
  for (let attempt = 0; attempt < MAX_POLL_ATTEMPTS; attempt++) {
    let data;
    try {
      data = await walletApi.getWalletStatus(this);
    } catch (err) {
      if (!(err instanceof WalletRequestError)) throw err;
      // Transient server/auth error — retry
      lastError = err;
      await delay(POLLING_INTERVAL);
      continue;
    }

    lastError = null;
    if (data.status.status === 'ready') return;
    if (data.status.status !== 'creating') throw new WalletRequestError(...);

    await delay(POLLING_INTERVAL);
  }
  throw new WalletRequestError('Wallet status polling timed out', { cause: lastError });
}
```

- **Sequential:** only one `getWalletStatus` in-flight at a time.
- **Bounded:** times out after 60 seconds (configurable via constant).
- **No leaked intervals:** no `setInterval` to clean up.
- **Transient error tolerance:** `WalletRequestError` (server responded but said no) is retried; non-`WalletRequestError` (programming errors) fails fast. Permanent wallet error states ('error') also fail fast — the status check is outside the try-catch.

### The new `startReadOnly()`

Replace the catch block's `getWalletStatus()` (which needs auth) with a retry loop on `getReadOnlyAuthToken()` (which doesn't need auth), with error classification:

```typescript
async startReadOnly() {
  // ...
  let lastError: Error | null = null;
  for (let attempt = 0; attempt < MAX_POLL_ATTEMPTS; attempt++) {
    try {
      await this.getReadOnlyAuthToken();
      lastError = null;
      break;
    } catch (err) {
      if (!(err instanceof WalletRequestError)) throw err;
      // Only retry on HTTP 400 (wallet still creating).
      // Other status codes (401, 403, etc.) are permanent failures.
      const cause = err.cause as { status?: number } | undefined;
      if (cause?.status && cause.status !== 400) throw err;
      lastError = err;
      await delay(POLLING_INTERVAL);
    }
  }
  if (lastError) throw new WalletRequestError('Read-only wallet startup timed out', { cause: lastError });
  await this.onWalletReady(skipAddressFetch);
}
```

The RO token endpoint returns 400 when the wallet is still 'creating'. We retry only that case. Non-400 HTTP errors (401, 403, 404) fail fast as permanent failures. Non-`WalletRequestError` (network/transport) also fails fast.

### `renewAuthToken` now throws on failure

The old `renewAuthToken()` silently caught all errors and set `this.authToken = null`. This was the root cause of the 403 race condition — a fire-and-forget renewal could null the token between another operation reading it and using it. The new code removes the try-catch entirely, letting errors propagate to the caller. All callers now `await` the result, so the non-throwing contract is no longer needed.

### Await fire-and-forget in `validateAndRenewAuthToken`

The `else if (usePassword)` branch at line 1191 calls `renewAuthToken()` without `await`. This is a fire-and-forget that can silently null a valid token if the renewal fails. We add `await` to make it blocking. The latency cost is negligible since this branch only fires when the token is already valid (opportunistic refresh).

## What we're NOT changing

- **Public API is unchanged.** `start()`, `startReadOnly()`, `pollForWalletStatus()`, `validateAndRenewAuthToken()` keep the same signatures and return types.
- **`axiosInstance` interceptor stays.** The interceptor in `walletServiceAxios.ts` that calls `validateAndRenewAuthToken()` before authenticated requests is the mechanism we now rely on for startup auth — it was already doing this for all other API calls.

## Test coverage

We're adding unit tests for every edge case in the refactored flow:

| Scenario | What we verify |
|----------|----------------|
| New wallet, status 'ready' | Token obtained before onWalletReady |
| New wallet, status 'creating' | Polling completes, wallet reaches ready |
| Status 'error' | WalletRequestError thrown immediately (not retried) |
| Token renewal fails | Error propagates (not swallowed) |
| waitReady=false | No token renewal, no polling |
| start() does not call validateAndRenewAuthToken | Auth delegated to interceptor |
| Poll returns 'ready' first try | Resolves immediately |
| Poll returns 'creating' then 'ready' | Resolves after retry |
| Poll exceeds max attempts | Timeout error |
| Poll callbacks don't stack | Only one getWalletStatus in-flight |
| Poll transient error then recovery | WalletRequestError retried, then succeeds |
| Poll non-WalletRequestError | Propagates immediately (fail-fast) |
| Poll persistent transient errors | Timeout with last error as cause |
| startReadOnly with ready wallet | Token on first try |
| startReadOnly with creating wallet | Token after retries |
| startReadOnly timeout | WalletRequestError with cause chain |
| startReadOnly permanent error (non-400) | Fails fast, no retry |
| startReadOnly never calls getWalletStatus | Authenticated endpoint not used |

## Summary by CodeRabbit

* **Tests**
  * Added an integration test verifying wallet initialization, readiness, auth token, and address availability.
  * Expanded wallet tests to cover status polling, auth-token renewal, retry semantics, timeout/error gating, sequential polling, and read-only startup retry/timeouts.
  * Added dedicated tests for the retry helper validating retry/count/delay/timeout behaviors.

* **Refactor**
  * Startup and read-only flows now use bounded retries, clearer timeout behavior, and stricter error propagation for more deterministic startup.


__tests__/integration/service-specific/start.test.ts
```diff
@@ -122,6 +122,24 @@ describe('[Service-specific] start', () => {
     expect(currentAddress.address).toEqual(emptyWallet.addresses[currentAddress.index]);
   });
 
+  it('should complete start() on a brand-new wallet that goes through creating→ready', async () => {
+    // This test exercises the full sequential auth flow: createWallet returns
+    // 'creating', then validateAndRenewAuthToken obtains a token, then
+    // pollForWalletStatus polls until 'ready'. This is the exact path where
+    // the old fire-and-forget pattern caused flaky 403 errors.
+    ({ wallet } = buildWalletInstance());
```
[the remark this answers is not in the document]
tuliomir: The precalculated wallets are unknown to both the fullnode and the wallet service, which ensures it will be the first time the entire blockchain sees it.

tuliomir:
## Performance Impact: Wallet Startup Time Benchmark

### TL;DR

**No meaningful performance regression.** Total wallet startup time is statistically equivalent between this branch and `master` for both tested wallets. The auth token work shifts from the API phase to the Ready phase, but the total time is unchanged.

---

### Methodology

A standalone benchmark script measures end-to-end `wallet.start()` time (from call to `state === 'Ready'`) against the **public testnet wallet service**, with sub-phase instrumentation via API-level monkey-patching (no source code changes needed).

- **10 measured iterations** per wallet, per branch (+ 1 warmup iteration excluded from stats)
- **2-second cooldown** between iterations to avoid rate-limiting
- **WebSocket disabled** to isolate the startup path
- **Sub-phases measured**: Setup (crypto key derivation), API (`POST /wallet/init`), Poll (`GET /wallet/status` loop), Ready (address fetch + auth token on-demand)

Tested against: `https://wallet-service.testnet.hathor.network/`

### Results

#### Wallet 1 (moderate history)

| Branch | Total (median) | Setup | API | Poll | Ready | Stddev |
|--------|:-:|:-:|:-:|:-:|:-:|:-:|
| `fix/auth-token-race-conditions` | **2570ms** | 1460ms | 238ms | 0ms | 818ms | 141ms |
| `master` | **2558ms** | 1624ms | 714ms | 0ms | 254ms | 176ms |
| **Delta** | **+12ms (+0.5%)** | -164ms | -476ms | 0ms | +564ms | |

#### Wallet 2 (heavy history -- triggers polling on cold start)

| Branch | Total (median) | Setup | API | Poll | Ready | Stddev |
|--------|:-:|:-:|:-:|:-:|:-:|:-:|
| `fix/auth-token-race-conditions` | **2620ms** | 1455ms | 247ms | 0ms | 818ms | 171ms |
| `master` | **2451ms** | 1564ms | 703ms | 0ms | 279ms | 127ms |
| **Delta** | **+169ms (+6.9%)** | -109ms | -456ms | 0ms | +539ms | |

### Interpretation

Both wallets show the same pattern: **the branch is faster at API but slower at Ready, with totals within noise.**

The reason is the auth token timing shift:

- **`master`**: Auth token renewal fires in parallel (fire-and-forget) during `createWallet()`. The token is usually resolved by the time `getNewAddresses()` runs. Hence: **slow API** (token renewal overlaps), **fast Ready** (token already cached).

- **This branch**: Auth token renewal is deferred to on-demand via the axios interceptor. `createWallet()` completes without auth overhead. The first authenticated call (`getNewAddresses()` in the Ready phase) triggers token renewal. Hence: **fast API** (no auth overhead), **slow Ready** (token obtained here).

**The total work is the same -- it just moved.** The +169ms delta on the heavy wallet is within 1 standard deviation (stddev=127-171ms) and is not statistically significant.

### Polling path

The heavy wallet triggered the polling path during warmup (cold start): **9022ms total, 4 polling attempts over ~5 seconds**. This exercises the new bounded `for`-loop polling introduced by this branch. On subsequent iterations (warm start), the wallet service returns `status: 'ready'` immediately (0 polling), which is the common case for production wallets.

The `master` warmup for the same wallet showed no polling (2765ms) -- because the wallet was already known to the Wallet Service from the branch's prior run.

### Conclusion

| Metric | Result |
|--------|--------|
| Total startup time regression | **None** (within noise) |
| Auth token overhead | **Redistributed**, not added |
| Polling path | **Exercised successfully** on cold start |
| Bounded timeout | **Working** (60s max, vs. unbounded on `master`) |

The refactored auth flow trades a negligible, non-statistically-significant timing shift for deterministic, race-condition-free startup behavior.

src/wallet/api/walletApi.ts
```diff
@@ -309,7 +313,9 @@ const walletApi = {
       return parseSchema(response.data, authTokenResponseSchema);
     }
 
-    throw new WalletRequestError('Error requesting read-only auth token.');
+    throw new WalletRequestError('Error requesting read-only auth token.', {
+      cause: { status: response.status, data: response.data },
```
andreabadesso, line 317: Nice

src/wallet/wallet.ts
```diff
@@ -1207,16 +1171,9 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
-
-      this.authToken = data.token;
-    } catch (err) {
-      // We should not throw here since this method is called in a fire-and-forget manner
-      // TODO: When this wallet has a logger, we should log this error to help with debugging
-      this.authToken = null;
-    }
+    const sign = this.signMessage(privKey, timestamp, this.walletId);
+    const data = await walletApi.createAuthToken(this, timestamp, privKey.xpubkey, sign);
+    this.authToken = data.token;
```
andreabadesso, line 1188: Why did you remove the try/catch? Are you capturing this anywhere up?

tuliomir: The errors that are raised here propagate naturally to `checkAndRenewAuthToken`, which is already inside a `try/catch` in every caller (e.g. the polling loop and the wallet start flow).

The old `try/catch` was there because renewAuthToken was called fire-and-forget (notice the old call site had no `await`). The catch swallowed the error and silently set `this.authToken = null`, which was actually the root cause of several issues in this PR — callers had no idea the renewal failed, so they'd proceed with a null token and hit confusing downstream errors.

andreabadesso (dismissed):

src/wallet/wallet.ts
```diff
@@ -1207,16 +1171,9 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
 
-    try {
-      const sign = this.signMessage(privKey, timestamp, this.walletId);
-      const data = await walletApi.createAuthToken(this, timestamp, privKey.xpubkey, sign);
-
-      this.authToken = data.token;
-    } catch (err) {
-      // We should not throw here since this method is called in a fire-and-forget manner
-      // TODO: When this wallet has a logger, we should log this error to help with debugging
-      this.authToken = null;
```
r4mmer, line 1218: question: why are we removing the try..catch?

The comment being removed indicates that throwing here will lead to uncaught exceptions in un-awaited async methods, this can lead to any number of undefined behaviors on wallets.

src/wallet/wallet.ts
```diff
@@ -1180,21 +1147,18 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
-      // If we have received the user PIN, we should renew the token anyway
-      // without blocking this method's promise
-
+      // Token is valid but the user provided a PIN — renew proactively.
       const privKey = bitcore.HDPrivateKey.fromString(
         await this.storage.getAuthPrivKey(usePassword)
       );
 
-      this.renewAuthToken(privKey, timestampNow);
+      await this.renewAuthToken(privKey, timestampNow);
```
r4mmer: Why are we awaiting here?

src/wallet/wallet.ts
tuliomir: The silent catch only existed to support fire-and-forget, and its side effect — `this.authToken = null` on failure — is exactly the root cause of the 403 race this PR fixes: a background renewal could null a token that a concurrent request was about to use.

This PR removes every fire-and-forget call site:
- `start()`: the two `validateAndRenewAuthToken(...).catch(...)` calls are gone. Auth now flows on-demand via the axios interceptor (`walletServiceAxios.ts:56`, awaited).
- `validateAndRenewAuthToken` line 1155: the `usePassword` branch now awaits (your other comment).

With no fire-and-forget callers left in `src/`, the non-throwing contract is obsolete — keeping the silent catch would mask bugs rather than prevent them.

Errors thrown here will be correctly caught like any other network or server error would, by the `walletServiceAxios` interceptor. It's the right place to catch them, since they'd indicate the Wallet Service integration has failed for this request.

src/wallet/wallet.ts
tuliomir: This is the counterpart to removing the try/catch in `renewAuthToken`. Previously this branch (token still valid + `usePassword`) did an opportunistic fire-and-forget refresh; if that refresh failed, the silent catch would null `this.authToken`, and the caller's next request would 403 despite the token being valid.

Awaiting adds one HTTP round-trip on a narrow path in exchange for error propagation instead of silent token corruption. If the latency here matters, a cleaner follow-up would be to drop this branch entirely (token is already valid — why renew?) rather than restore the fire-and-forget.

r4mmer (dismissed):

tuliomir:
## Follow-up: CI integration-test failure investigation and fix

The itest job on run [24842212984](https://github.com/HathorNetwork/hathor-wallet-lib/actions/runs/24842212984/job/72732666470) reported 3 failing tests in `service-specific/get-balance.test.ts`, all with the same root error:

```
Error requesting auth token.
  at createAuthToken (walletApi.ts:298)
  at renewAuthToken (wallet.ts:1175)
  at validateAndRenewAuthToken (wallet.ts:1148)
  at axios interceptor (walletServiceAxios.ts:56)
  at getNewAddresses
  at onWalletReady (wallet.ts:799)
  at start (wallet.ts:501)
  at GenesisWalletServiceHelper.start
  at ServiceWalletTestAdapter.suiteSetup
```

(A fourth failure in `hathorwallet_facade.test.ts` is the pre-existing ECDSA leading-zero flake — `s must be 32 bytes` — will be solved by #1033 .)

### Why this failure surfaced on this PR specifically

The same test file passes on master and on other branches (e.g. run [24807903816](https://github.com/HathorNetwork/hathor-wallet-lib/actions/runs/24807903816) on `fix/service-adapter-poll-token-after-create` ran the same `service-specific/get-balance.test.ts` green in 16.3s). So the test itself is not flaky — this PR's changes exposed a latent race.

Walking through the stack on this PR's branch for a pre-existing wallet:

1. `suiteSetup` reuses the genesis wallet singleton. The wallet-service backend already has this wallet — `createWallet` returns status `'ready'` immediately.
2. Because status is already `'ready'`, `pollForWalletStatus` is skipped. **No delay between create and first authenticated call.**
3. `onWalletReady` fires → calls `getNewAddresses` → axios interceptor invokes `validateAndRenewAuthToken` → `renewAuthToken` → `createAuthToken`.
4. The `/auth/token` endpoint has a brief post-creation settling window where it can return a non-2xx. The request lands inside that window.
5. `createAuthToken` throws `WalletRequestError`. The new contract lets it propagate all the way up and fail the test.

### How the old code masked this

In the pre-PR code path, every run of `start()` effectively had **three** chances to obtain an auth token:

1. `start()` fired a preemptive `validateAndRenewAuthToken(...).catch(...)` in the background. Failure was caught and set `this.authToken = null`.
2. After `createWallet` returned, `handleCreate` fired a second `validateAndRenewAuthToken(...).catch(...)`, also silently catching and nulling on failure.
3. When the interceptor finally fired during `onWalletReady → getNewAddresses`, it called `validateAndRenewAuthToken` for the third time. By then, enough wall-clock had elapsed for the wallet-service to settle, and this third attempt usually succeeded.

This PR removes every fire-and-forget, awaits every call, and makes `renewAuthToken` throw on failure. That's the correct design — the silent nulling was the root cause of the 403 race this PR fixes. But collapsing three implicit tries into one synchronous try removed the accidental settling delay, which is what the CI failure exposed.

### How this branch makes the error *easier* to debug and fix

Previously this bug was invisible: the silent catch ate the evidence, `authToken` was silently nulled, and the next 403 looked like "something elsewhere broke." The stack trace in the failing run now points **directly** at the real call site: `renewAuthToken → createAuthToken`. A human or an agent reading the CI log can see exactly which HTTP call failed and where the retry budget lives.

Commit https://github.com/HathorNetwork/hathor-wallet-lib/pull/1058/commits/c9ceeca4783bb2923cfbca3bc9362bec6ec423db (in this PR) also added `{ status, data }` to the `WalletRequestError.cause` — so once the fix lands we'll see the actual HTTP status from the wallet-service in future failures.

### The fix (two commits added to this branch)

**Commit https://github.com/HathorNetwork/hathor-wallet-lib/pull/1058/commits/ed107dd01749a65917cadb896c49c5b9b1ea637e ** — `test: seal renewAuthToken throw contract`
Adds five regression guards that lock in the new throw-on-failure contract: success path, transient-error propagation, existing-token preservation on failure, generic-error propagation, and `walletId` pre-check. If a future refactor reintroduces the silent catch, CI fails loudly.

**Commit https://github.com/HathorNetwork/hathor-wallet-lib/pull/1058/commits/6b033239234588116607dbc5864c7d8a7adcaedf ** — `feat: bounded retry for transient auth-token renewal`
Extracts a generic helper `retryOnTransientWalletError(operation, { maxAttempts, intervalMs })` and wraps both `renewAuthToken` call sites inside `validateAndRenewAuthToken` with it. Retry budget is 3 attempts with 1s delay (same interval as `pollForWalletStatus`). Mirrors the pattern already used in `pollForWalletStatus` and `startReadOnly`:

- **Transient** (`WalletRequestError`: server responded with non-2xx) → retried within budget.
- **Non-transient** (network failures, programming errors) → propagated immediately with no retry.

The helper lives in its own module (`src/wallet/walletServiceRetry.ts`) and is covered by six unit tests in isolation. It's small and generic enough that `pollForWalletStatus` and `startReadOnly` could be refactored to use it in a future PR, unifying the retry-on-transient-WalletRequestError pattern.

### Why this is the right shape

- **No silent swallowing.** The helper retries within a bounded budget and re-throws the *last* transient error, preserving the full cause chain. Permanent errors still fail fast.
- **Addresses [r4mmer's concern directly.](#discussion_r3111484148)** Removing the try/catch in `renewAuthToken` was safe *because* every caller awaits — but it also removed the implicit settling tolerance that the try/catch-plus-null-plus-interceptor-retry dance used to provide. The explicit retry helper makes that tolerance visible, bounded, and testable, instead of depending on a silent side effect.
- **Integration tests tell the truth.** If the wallet-service's auth endpoint regresses to a longer settling window or fails permanently, CI will now show the actual `{ status, data }` in the error cause, not a "403 somewhere downstream" mystery.

src/wallet/wallet.ts
```diff
@@ -775,21 +719,23 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      if (data.status.status !== WS_STATUS_CREATING) {
+        throw new WalletRequestError('Error getting wallet status.', { cause: data.status });
+      }
+
+      await new Promise(resolve => {
+        setTimeout(resolve, WALLET_STATUS_POLLING_INTERVAL);
+      });
+    }
+    throw new WalletRequestError('Wallet status polling timed out.');
   }
```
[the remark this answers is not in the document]
tuliomir: If the loop takes so long it means the Wallet Service is both responding and extremely slow on doing so. It won't matter much the time for this loop, as the entire Wallet Service will be slow.

src/wallet/walletServiceRetry.ts
```diff
@@ -0,0 +1,93 @@
...
+        // Non-transient: propagate immediately without consuming retry budget.
+        throw err;
+      }
+      lastTransientError = err;
+      const isLastAttempt = attempt === opts.maxAttempts;
+      if (isLastAttempt) break;
+      // eslint-disable-next-line no-await-in-loop -- sequential by design
+      await new Promise<void>(resolve => {
+        setTimeout(resolve, opts.intervalMs);
+      });
```
r4mmer, line 82: Why not use the new helper here?

src/wallet/walletServiceRetry.ts
```diff
@@ -0,0 +1,93 @@
...
+  for (let attempt = 1; attempt <= opts.maxAttempts; attempt++) {
+    try {
+      // eslint-disable-next-line no-await-in-loop -- sequential by design
+      return await operation();
+    } catch (err) {
+      if (!(err instanceof WalletRequestError)) {
+        // Non-transient: propagate immediately without consuming retry budget.
+        throw err;
+      }
+      lastTransientError = err;
```
r4mmer, line 76: Should we log the error?
Just so we dont miss an error that is happenning every once in a while.

r4mmer (approved):

src/wallet/walletServiceRetry.ts
tuliomir: I'm afraid we can pollute the logs too much by writing those, but it sure is a possibility.

Another path I actually prefer would be to use the Wallet Service's new Open Telemetry to get this information instead, directly from the source instead of from the clients using the lib.

src/wallet/walletServiceRetry.ts
tuliomir: There are many other places I wanted to also use the helper ( it's actually from 2021! ), but didn't want to touch the diff on this PR.

But this one should have been implmented using it. If I have to change any other code in this PR, I'll fix this here. Otherwise, this PR is pretty important for the next test suite implementations, so I'd rather merge it as is and fix later.

tuliomir:
## Ultrareview follow-up — deferred to a future PR

An automated review surfaced one nit-level asymmetry in `validateAndRenewAuthToken`. Flagging here for posterity so it isn't lost; **not addressing in this PR** to keep the diff scoped to the `start()` race conditions this PR targets.

### The asymmetry

`validateAndRenewAuthToken` (`src/wallet/wallet.ts`) has three renewal branches, but only two are wrapped in the new `retryOnTransientWalletError` helper:

| Branch | Line | Wrapped? |
|---|---|---|
| Read-only (`getReadOnlyAuthToken`) | 1146 | ❌ |
| Authenticated (`renewAuthToken`) | 1160 | ✅ |
| Proactive `usePassword` (`renewAuthToken`) | 1170 | ✅ |

### Why it matters (eventually)

The axios interceptor at `src/wallet/api/walletServiceAxios.ts:56` invokes `validateAndRenewAuthToken()` on every authenticated request — including read-only wallets. When a read-only wallet's cached JWT expires post-startup and the wallet-service is briefly unhappy (e.g. mid-restart), the renewal fails immediately on that path, whereas full-wallet callers retry up to `MAX_AUTH_TOKEN_RENEW_ATTEMPTS` times.

The `startReadOnly` 60-attempt poll (lines 1256–1276) **does not** cover this — it only runs once at startup and is narrower by design (HTTP 400 only, since "wallet still being created" is a 400-specific signal).

### Why this is low-risk to defer

- Post-startup read-only renewal is rare in practice (long JWT lifetime × low collision window with a transient).
- The read-only startup path is already fully protected by its own retry loop.
- The buggy failure mode this PR replaces (silent catch + nulled token) is **already gone** from the read-only branch — it just doesn't have the new retry resilience.

### Suggested fix for the follow-up

One-liner wrap mirroring the other two branches:

```ts
await retryOnTransientWalletError(() => this.getReadOnlyAuthToken(), {
  maxAttempts: MAX_AUTH_TOKEN_RENEW_ATTEMPTS,
  intervalMs: WALLET_STATUS_POLLING_INTERVAL,
});
```

`getReadOnlyAuthToken` (lines 1204–1218) is idempotent (no state mutation beyond `this.walletId` derivation and `this.authToken` assignment), so it's a clean fit for the helper.

src/wallet/api/walletApi.ts
```diff
@@ -68,7 +68,9 @@ const walletApi = {
     if (response.status === 200 && data.success) {
       return parseSchema(data, walletStatusResponseSchema);
     }
-    throw new WalletRequestError('Error getting wallet status.');
+    throw new WalletRequestError('Error getting wallet status.', {
+      cause: { status: response.status, data: response.data },
```
raul-oliveira, line 72: Do we have an interface for this error body? It would be good to have a pattern for us to follow

src/wallet/api/walletApi.ts
```diff
@@ -293,7 +295,9 @@ const walletApi = {
       return parseSchema(response.data, authTokenResponseSchema);
     }
 
-    throw new WalletRequestError('Error requesting auth token.');
+    throw new WalletRequestError('Error requesting auth token.', {
+      cause: { status: response.status, data: response.data },
```
raul-oliveira, line 299: Given the context of this PR (well written btw), add a test to cover this line.

src/wallet/wallet.ts
```diff
@@ -530,29 +490,25 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
-        renewPromise2Error = renew2Err;
+    if (data.status.status === WS_STATUS_CREATING) {
+      // If the wallet status is creating, we should wait until it is ready
+      // before continuing
+      await this.pollForWalletStatus();
+    } else if (data.status.status !== WS_STATUS_READY) {
+      // At this stage, if the wallet is not 'ready' or 'creating' we should
+      // throw an error as there are only three states: 'ready', 'creating' or 'error'
+      throw new WalletRequestError(ErrorMessages.WALLET_STATUS_ERROR, {
+        cause: data.status,
```
raul-oliveira, line 501: I'm missing the .data field in the error return. Do we have any data that we could return here?

src/wallet/wallet.ts
```diff
@@ -775,20 +731,37 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+      }
+
+      lastError = null;
+      if (data.status.status === WS_STATUS_READY) {
+        return;
+      }
+      // Only possible states are 'ready', 'creating' and 'error'. If status
+      // is not ready or creating, we should throw an error.
+      if (data.status.status !== WS_STATUS_CREATING) {
+        throw new WalletRequestError('Error getting wallet status.', { cause: data.status });
```
raul-oliveira, line 758: Same here

src/wallet/wallet.ts
```diff
@@ -1207,16 +1183,9 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
       throw new Error('Wallet not ready yet.');
     }
 
-    try {
-      const sign = this.signMessage(privKey, timestamp, this.walletId);
```
raul-oliveira, line 1211: this removal is really cool!

src/wallet/walletServiceRetry.ts
```diff
@@ -0,0 +1,93 @@
...
+      `retryOnTransientWalletError: maxAttempts must be >= 1, got ${opts.maxAttempts}`
+    );
+  }
+
+  let lastTransientError: WalletRequestError | undefined;
+
+  for (let attempt = 1; attempt <= opts.maxAttempts; attempt++) {
+    try {
+      // eslint-disable-next-line no-await-in-loop -- sequential by design
+      return await operation();
```
raul-oliveira, line 70: I would change this piece to be a "for await" loop instead of disabling the next line here. What are your thoughts on it?

src/wallet/walletServiceRetry.ts
```diff
@@ -0,0 +1,93 @@
...
+    } catch (err) {
+      if (!(err instanceof WalletRequestError)) {
+        // Non-transient: propagate immediately without consuming retry budget.
+        throw err;
+      }
+      lastTransientError = err;
+      const isLastAttempt = attempt === opts.maxAttempts;
+      if (isLastAttempt) break;
+      // eslint-disable-next-line no-await-in-loop -- sequential by design
```
raul-oliveira, line 79: Same about the for await I already mentioned

src/wallet/walletServiceRetry.ts
raul-oliveira: Do you have a implementation/issue to address this?

src/wallet/api/walletApi.ts
tuliomir: Addressed in #1074 — `WalletRequestError.cause` is now a discriminated union with three mutually exclusive variants: `{ status: number; data: unknown }` (HTTP failure), `{ state: unknown }` (app-level rejection), and `{ source: WalletRequestError }` (chained failure). Each existing throw site picks the correct variant.

src/wallet/api/walletApi.ts
tuliomir: Added in #1074 — see the new `createAuthToken attaches { status, data } to cause on non-2xx` test in `__tests__/wallet/api/walletApi.test.ts`. Uses `toEqual` on the captured error so future shape extensions stay backward-compatible.

src/wallet/wallet.ts
tuliomir: Addressed in #1074 — this throw now uses the `{ state: data.status }` variant of the new typed cause. The wallet-status object is preserved under `state` (rather than the HTTP-shaped `{ status, data }` variant, since the HTTP call itself succeeded — the failure is app-level).

src/wallet/wallet.ts
tuliomir: Addressed in #1074 — same as the sibling site in `verifyWalletState`, this throw now uses the `{ state: data.status }` variant of the typed cause.

src/wallet/walletServiceRetry.ts
tuliomir: Tracked in #1075. The helper currently has no status-code filter, while `pollForWalletStatus` and `startReadOnly` retry on different conditions (the read-only loop selectively retries only on HTTP 400). Either extending the helper or adopting it as-is is its own design call, so deferring to a separate PR.

src/wallet/walletServiceRetry.ts
tuliomir: Thanks for the suggestion — I considered it but I don't think `for await…of` fits this helper.

`for await…of` consumes an async iterable, which is a good match for *iteration over an async producer* (paginated results, a stream of events). What this loop does is *retry semantics*: invoke `operation`, on a transient error sleep and invoke it again. There is no producer to iterate.

To use `for await` here I'd have to construct a synthetic generator that yields `operation()` results — which doesn't simplify the code, doesn't remove the in-loop `await` (the sleep on line 89 is still inside the body), and makes the intent harder to read. The narrowly-scoped `eslint-disable-next-line no-await-in-loop -- sequential by design` is the correct escape hatch: sequential awaiting is precisely what the helper provides.

Happy to revisit if you have a concrete shape in mind that I'm missing.

src/wallet/walletServiceRetry.ts
tuliomir: Same response as on the previous thread — see https://github.com/HathorNetwork/hathor-wallet-lib/pull/1058#discussion_r3163300898. The `await` on this line is the inter-attempt sleep, which is sequential by design; reshaping the loop into `for await…of` would not eliminate it.

raul-oliveira (approved):

PR #1053, opened 2026-03-31
Test: Shared sendTransaction tests for both facades

tuliomir:
## Summary

- Extract `sendTransaction` and `sendManyOutputsTransaction` tests from `hathorwallet_facade.test.ts` and `walletservice_facade.test.ts` into shared and facade-specific files
- Add `sendTransaction` adapter method to `IWalletTestAdapter` — handles pinCode injection for service facade
- Create `shared/send-transaction.test.ts` with adapter-driven tests (internal transfer balance invariant, external transfer balance decrease)
- Move fullnode-only tests (address tracking, custom/fee token sends, multisig, sendManyOutputsTransaction) to `fullnode-specific/send-transaction.test.ts`
- Move service-only tests (full structure validation, P2SH target, changeAddress with getUtxoFromId) to `service-specific/send-transaction.test.ts`

## Acceptance Criteria

- Shared sendTransaction tests pass for both fullnode and wallet-service adapters
- No `sendTransaction` or `sendManyOutputsTransaction` describe blocks remain in the monolithic facade files

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive transaction test suites covering native/token sends, multisig flows, multi-output sends, fee validation, address-usage tracking, and full-transaction retrieval across adapters.
  * Parameterized shared tests to run against multiple wallet backends for consistent behavior verification.

* **Chores**
  * Reorganized and cleaned up test structure and fixtures for clearer coverage and maintenance.

PR #1063, opened 2026-04-13
fix: address path ordering with mixed inputs

### LINKED ISSUE #1057: fix: SendTransactionWalletService signs inputs with wrong address paths when mixing token types
## Summary

`SendTransactionWalletService.prepareTx()` produces an incorrect `utxosAddressPath` ordering when the user provides manual inputs that mix HTR and custom token UTXOs. This causes invalid signatures, and the wallet service API rejects the transaction with "Error sending tx proposal."

## Root cause

When inputs are user-provided, `prepareTx()` calls `validateUtxos` in two passes:

1. **Custom tokens first** (`ignoreNative: true`) — collects address paths for non-HTR inputs
2. **HTR second** (`onlyNative: true`) — collects address paths for HTR inputs

The resulting `utxosAddressPath` is concatenated as `[token_paths..., htr_paths...]`, regardless of the order in `this.inputs`. During signing (`signTx`), input `i` is signed with `utxosAddressPath[i]`, causing a mismatch when the user provides HTR inputs before token inputs.

### Example

User provides `inputs: [htr_utxo, token_utxo]`:
- `utxosAddressPath` becomes `[token_addressPath, htr_addressPath]` (wrong order)
- Input 0 (HTR) gets signed with the token's address path → invalid signature

## Affected code

`src/wallet/sendTransactionWalletService.ts`, inside `prepareTx()` (the `else` branch at ~line 433 where `this.inputs.length > 0`).

## Suggested fix

Use a `Map<string, string>` keyed by `${txId}:${index}` to collect address paths during both `validateUtxos` calls, then rebuild `utxosAddressPath` in `this.inputs` order:

```typescript
const addressPathMap = new Map<string, string>();

// Pass the map to both calls (requires adding an optional param to validateUtxos)
await this.validateUtxos(tokensWithoutHtr, { ignoreNative: true, addressPathMap });
// ... HTR amount calculation ...
await this.validateUtxos(htrTokenAmount, { onlyNative: true, addressPathMap });

// Rebuild in input order
utxosAddressPath = this.inputs.map(input => {
  const path = addressPathMap.get(`${input.txId}:${input.index}`);
  if (!path) throw new SendTxError(`Address path not found for input ${input.txId}:${input.index}.`);
  return path;
});
```

`validateUtxos` needs an optional `addressPathMap` parameter — when provided, it populates the map alongside the existing array return.

## Workaround

Provide inputs with custom tokens before HTR (matching the internal processing order). This avoids the mismatch without code changes.

## How to verify

The skipped integration test `should send fee token with manually provided HTR input (no HTR output)` in `shared/send-transaction-tokens.test.ts` exercises this exact scenario. Unskip it after the fix.

tuliomir:
## Summary

Fixes the `utxosAddressPath` ordering bug in `SendTransactionWalletService.prepareTx()` when user-provided inputs mix HTR and custom token UTXOs.

Closes #1057

## Problem

When users provide manual inputs, `prepareTx()` validates them in two passes:
1. Custom tokens first (`ignoreNative: true`)
2. HTR second (`onlyNative: true`)

The resulting `utxosAddressPath` is concatenated as `[token_paths..., htr_paths...]`, regardless of the original input order. During signing, input `i` is signed with `utxosAddressPath[i]`, causing mismatched signatures when HTR inputs appear before token inputs.

`prepareTxData` is **not** affected: it processes all user inputs in a single in-order loop, so its `utxosAddressPath` already follows `this.inputs` order.

## Why two passes are necessary

The two-pass structure is load-bearing and cannot be collapsed into a single pass. Fee-token change outputs are **discovered** during the first pass (custom token validation), and each one increases `_feeAmount` by `FEE_PER_OUTPUT`. The correct HTR amount can only be computed after all fee-token changes are accounted for. Processing everything in one pass would validate HTR against an incomplete fee total.

## Fix

Adds a required `addressPathMap: AddressPathMap` parameter to `validateUtxos`. Both passes populate this map keyed by `${txId}:${index}`, then `prepareTx` rebuilds `utxosAddressPath` in `this.inputs` order. This preserves the two-pass fee calculation while decoupling address path collection from array concatenation order.

## Acceptance criteria

- Transactions with manually provided inputs mixing HTR and custom tokens produce valid signatures regardless of input ordering
- The previously failing integration test scenario (fee token with HTR input first) passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure address-paths for transaction inputs are collected and reconstructed in the exact input order; detect and error when an input lacks a recorded path during transaction preparation.
  * Provide a clearer error when a native-token input is supplied but not required by outputs.

* **Tests**
  * Added tests covering native-token validation, address-path ordering and overwrite semantics, map population, fee/change behavior, and related error cases.


src/wallet/sendTransactionWalletService.ts
```diff
@@ -445,9 +451,21 @@ class SendTransactionWalletService extends EventEmitter implements ISendTransact
...
           },
         };
-        const htrAddressPath = await this.validateUtxos(htrTokenAmount, { onlyNative: true });
-        utxosAddressPath.push(...htrAddressPath);
+        await this.validateUtxos(htrTokenAmount, { onlyNative: true, addressPathMap });
       }
+
+      // Rebuild utxosAddressPath in this.inputs order
+      utxosAddressPath = this.inputs.map(input => {
+        const path = addressPathMap.get(`${input.txId}:${input.index}`);
```
raul-oliveira: This key access logic should be together with address path map. keeping the map, key access, and key retrieval logic together. Something like: 
```
AddressPathMap extends Map { 
  // other stuff 
  set(input: input){}
  get(input: input) {}
}


```

__tests__/wallet/sendTransactionWalletService.test.ts
```diff
@@ -1080,6 +1080,115 @@ describe('prepareTxData', () => {
     expect(changeOutput).toBeDefined();
     expect(changeOutput.address).toBe(customChangeAddress);
   });
+
+  it('should reject an unnecessary HTR input when no HTR is needed', async () => {
+    // When an HTR input is provided but outputs only use custom tokens (no HTR
+    // needed), validateUtxos rejects it because NATIVE_TOKEN_UID is not in the
+    // tokenAmountMap. This validates the existing guard in prepareTxData.
```
raul-oliveira: Do we have an integration test that validates this scenario?

raul-oliveira (changes requested):

__tests__/wallet/sendTransactionWalletService.test.ts
tuliomir: Yes we do, in #1053 . The idea is that this PR can be merged before or after that other one with no conflicts.

I decided not to create integration tests here because they would have to be migrated to the new Shared Tests folder, an unavoidable conflict with 1053.

src/wallet/sendTransactionWalletService.ts
tuliomir: Good call — refactored to a small `AddressPathMap` wrapper in cdd8dd02. Went with composition over `extends Map` to keep the surface minimal (just `get` / `set` taking the input object directly), and kept the class file-local since it's a private helper to this flow with no other consumers. Key construction now lives in exactly one place, so the read and write sides can't drift out of sync.

raul-oliveira: tks

__tests__/wallet/sendTransactionWalletService.test.ts
raul-oliveira: tks

raul-oliveira (dismissed):

src/wallet/sendTransactionWalletService.ts
```diff
@@ -534,9 +574,13 @@ class SendTransactionWalletService extends EventEmitter implements ISendTransact
    */
   async validateUtxos(
```
r4mmer, line 593: suggestion: Make `addressPathMap` required.

`validateUtxos` basically is only meant to be used here and the `addressPathMap` is now the "intended" way to use it, correct?

So I think we should change the method to require an addressPathMap and not return a `string[]`.
What do you think?

tuliomir: Agreed — the return value was already dead at both call sites in `prepareTx` (we rebuild `utxosAddressPath` from the map immediately after the two passes), so `Promise<string[]>` was lying about the contract. Pushed as 0bd4bf4:

- Promoted `addressPathMap` to a required **positional** parameter on `validateUtxos` (kept outside `options` so the type system enforces it at every call site).
- Return type is now `Promise<void>`; removed the internal `string[]` accumulator and the `if (addressPathMap)` guard.
- Updated both call sites in `prepareTx` accordingly.
- Added a new `describe('validateUtxos', …)` block — the method previously had **zero** direct unit tests (only exercised indirectly through `prepareTxData`/`prepareTx`), so the post-refactor contract is now pinned (happy paths for both skip modes, change-output side effect, fee-token `_feeAmount` increment, and all four error branches).

Grep confirmed no callers outside this class, so the signature change is safe. Left the method `public` for this PR to keep the diff scoped — happy to flip it to `private` in a follow-up if you'd like.

tuliomir:
@coderabbitai regarding the outside-diff concern about `htrAmount === 0n` with a manually provided HTR input — verified against the current code: this case is already handled. The rebuild block in `prepareTx` throws a `SendTxError` with a user-facing message:

> Input <txId>:<index> was not processed by any validation pass. This usually means an HTR input was provided but no HTR is required (no HTR outputs and no fee-token fees).

and `__tests__/wallet/sendTransactionWalletService.test.ts` ("should show a helpful error when HTR input is unneeded via prepareTx path") pins that behavior.

The alternative (tracking `hasUserHtrInput` and running the HTR pass when it's true) would actually regress UX: with `htrAmount === 0n`, `validateUtxos` would accept the HTR input and then emit a spurious HTR change output equal to the input's value, because `amountInputMap[NATIVE_TOKEN_UID] > 0n` would trigger the change-output branch. Rejecting with a user-actionable error is the correct behavior — the user shouldn't have included the HTR input in the first place, and silently creating change masks the mistake.

Resolving this thread.

r4mmer (dismissed):

raul-oliveira (dismissed):

raul-oliveira (approved):

r4mmer (approved):

PR #1079, opened 2026-04-29
test: shared address-method tests for both facades

tuliomir:
## Summary
Migrates duplicated address-method coverage from the monolithic facade files into the shared-tests structure.

- New shared cases: \`__tests__/integration/shared/addresses.test.ts\` (6 tests: getAllAddresses ordering, getCurrentAddress consistency, markAsUsed advance, getNextAddress advance, getAddressIndex correct/unknown, getAddressAtIndex).
- Fullnode-specific kept: privkey operations, multisig addresses, runtime address derivation.
- Service-specific kept: \`GAP_LIMIT_REACHED\` behavior, \`getAddressDetails\`.
- Adapter additions: \`getAllAddresses\`, \`getCurrentAddress\`, \`getNextAddress\`, \`getAddressIndex\`, \`getAddressAtIndex\` on \`IWalletTestAdapter\` + both implementations; new \`AdapterAddress\` named type \`{ address, index, addressPath }\`.
- Removed the "addresses methods" / "empty wallet address methods" / "address management methods" describe blocks from both monolithic files.

\`getAddressIndex\` is normalized at the adapter boundary to return \`undefined\` for unknown addresses (underlying facades return \`null\`); aligns with the project's \`undefined\`-over-\`null\` preference.

## Acceptance Criteria
- Shared tests exercise the 6 cross-facade address-method cases.
- Facade-specific tests cover only behaviors that genuinely differ (privkey/multisig on fullnode, gap-limit/details on service).

## Summary by CodeRabbit

* **Tests**
  * Expanded integration coverage for wallet address behaviors, including enumeration, current/next navigation, index lookup, address-at-index retrieval, deterministic ordering, and address-path verification.
  * Added fullnode- and service-specific address test suites covering derivation, private-key/address consistency, message signing, multisig derivation, and gap-limit boundary handling.
  * Reorganized address-related facade tests into shared and service/fullnode-specific modules, and improved service/public test suite structure.
* **Chores**
  * Improved test teardown by emitting a warning when wallet stop/cleanup fails.


__tests__/integration/service-specific/addresses.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        // Wallet may already be stopped
+      }
+    }
+  });
+
+  it('getAddressPathForIndex returns the BIP32 path for known indices', async () => {
+    for (let i = 0; i < knownAddresses.length; i++) {
+      const path = await wallet.getAddressPathForIndex(i);
+      expect(path.endsWith(`/${i}`)).toBe(true);
+      expect(path).toMatch(/m\/44'\/280'\/0'\/0\/[0-9]+/);
```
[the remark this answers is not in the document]
tuliomir: Fixed on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1079/commits/801615b96b0782f71288127862e0ebc5182eb586

__tests__/integration/adapters/service.adapter.ts
```diff
@@ -394,4 +395,48 @@ export class ServiceWalletTestAdapter implements IWalletTestAdapter {
...
+  }
+
+  async getNextAddress(wallet: FuzzyWalletType): Promise<AdapterAddress> {
+    const next = this.concrete(wallet).getNextAddress();
+    return {
+      address: next.address,
+      index: next.index,
+      addressPath: next.addressPath,
+    };
+  }
```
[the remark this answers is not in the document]
tuliomir: These address methods must not return `null`. If they do, the correct behavior would be to break the entire test suite unhandled.

__tests__/integration/service-specific/addresses.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    allPrecalculated = created.addresses;
+    knownAddresses = created.addresses.slice(0, SERVICE_GAP_LIMIT);
+  });
+
+  afterEach(async () => {
+    if (wallet) {
+      try {
+        await wallet.stop({ cleanStorage: true });
+      } catch {
+        // Wallet may already be stopped
```
raul-oliveira: Is there any scenario that the wallet may not stop? If so we should at least log the error.

raul-oliveira (dismissed):

__tests__/integration/service-specific/addresses.test.ts
```diff
@@ -0,0 +1,226 @@
...
+      }
```
r4mmer, line 88: suggestion (blocking): silent catch may hide real teardown failures

This catch discards all errors, network failures, storage corruption, bugs in `wallet.stop()` itself. A surprising failure here lets the next test inherit leaked state with no signal. Consider logging unexpected errors or only suppressing the "already stopped" case:

```suggestion
      try {
        await wallet.stop({ cleanStorage: true });
      } catch {
        // swallow only "already stopped" — everything else is unexpected
        if (!(e instanceof Error && e.message.includes('stopped'))) {
        console.warn('wallet.stop() failed unexpectedly:', e);
      }
```

Don't recall the exact error throw when a wallet is stopped, but we should at least `console.warn` any other errors.

r4mmer (changes requested):

__tests__/integration/service-specific/addresses.test.ts
tuliomir: Good call. The wallet can fail to stop on transient teardown conditions (network blip, storage cleanup, an error thrown deeper in `stop()`), so I've replaced the silent catch with the same best-effort logging the rest of the suite uses:

```js
} catch (e) {
  loggers.test!.warn('Failed to stop wallet during cleanup', {
    error: (e as Error).message,
  });
}
```

Teardown stays non-fatal (one wallet failing to stop shouldn't abort the run), but a real failure now leaves a trace instead of vanishing. Applied to the sibling `get-balance.test.ts` too, which had the same silent catch.

__tests__/integration/service-specific/addresses.test.ts
tuliomir: Agreed on the core problem — swallowing everything hides real teardown failures. I've switched both service-specific teardowns to log-and-continue:

```js
} catch (e) {
  loggers.test!.warn('Failed to stop wallet during cleanup', {
    error: (e as Error).message,
  });
}
```

On the "suppress only the already-stopped case" part — I went with logging *all* errors rather than matching the stopped case, because there's nothing concrete to match on. `HathorWalletServiceWallet.stop()` (`src/wallet/wallet.ts:1477`) doesn't guard on state or throw a typed "already stopped" error — it just resets its fields and calls `conn.stop()`. There's no `already stopped` message or error class anywhere in `src/` to key off (`grep -rni "already stopped\|WalletStopped" src/` comes back empty), so `e.message.includes('stopped')` would never actually fire.

This also keeps the teardown consistent with how the rest of the suite already handles it — the shared `WalletTracker.stopAll()` (used by both adapters), the fullnode helper's `stopAllWallets()` (`wallet.helper.ts:217`, which logs `e.stack`), and `shared/start.test.ts` all log-and-continue without type-discriminating. No teardown in the suite filters by error type today.

If we later want a clean "already stopped" signal, the right fix is to make `stop()` idempotent / throw a typed error at the source and then have all the teardowns match on it — but that's a lib-level change I'd rather not slip into this test-only PR. Happy to open a follow-up for it.

raul-oliveira (dismissed):

r4mmer (dismissed):

raul-oliveira (approved):

r4mmer (approved):

Issue #1075, opened 2026-04-29
Make WalletRequestError.cause mandatory by sweeping bare throws


tuliomir:
## Context

Follow-up to HathorNetwork/hathor-wallet-lib#1074 (PR-A: typed `WalletRequestError.cause`). This issue tracks PR-B of the original two-PR split that emerged from the review threads on HathorNetwork/hathor-wallet-lib#1058.

PR-A introduced a discriminated union for `WalletRequestError.cause` with three mutually exclusive variants (`{ status, data }`, `{ state }`, `{ source }`), and swept every site that *already* passed a cause. To keep PR-A reviewable, the field was kept optional (`cause: WalletRequestErrorCause | undefined`) so that the ~15 currently-bare throws in `src/wallet/api/walletApi.ts` continued to compile.

## Goal

**Make `cause` mandatory at the type level.** A \"lazy dev\" should not be able to land `new WalletRequestError('msg')` without diagnostic context — the type system should reject it.

## Scope

Two coupled changes:

1. **Populate the 15 bare throw sites** in `src/wallet/api/walletApi.ts` with `{ cause: { status: response.status, data: response.data } }`. Each is inside an axios response handler with `response` in scope, so the conversion is mechanical. Methods affected: `getVersionData`, `createWallet`, `getAddresses`, `getAddressDetails`, `checkAddressesMine`, `getNewAddresses`, `getTokenDetails`, `getBalances`, `getTokens`, `getHistory`, `getTxOutputs`, `createTxProposal`, `updateTxProposal`, `deleteTxProposal`, `getHasTxOutsideFirstAddress`.

2. **Narrow the constructor signature** in `src/errors.ts`: drop the `| undefined` and the default. After step 1, the build passing is the proof every site complies.

A self-contained execution plan (worktree setup, tasks, verification, PR metadata) is checked into `~/code/wallet-lib/PR-B-PLAN.md` on @tuliomir's machine.

## Out of scope (other follow-ups)

- Adopt `retryOnTransientWalletError` at additional call sites (`pollForWalletStatus`, `startReadOnly`) — see HathorNetwork/hathor-wallet-lib#1058 (comment) and HathorNetwork/hathor-wallet-lib#1058 (comment). The helper currently has no status-code filter; `startReadOnly` selectively retries only when `cause.status === 400`.
[...]

PR #1074, opened 2026-04-29
feat(errors): type WalletRequestError.cause as a discriminated union

tuliomir:
## Summary

Builds on HathorNetwork/hathor-wallet-lib#1058 (currently the base branch — will retarget to `master` once the parent PR lands). Addresses three of the four review threads on the parent PR: the typed cause shape, the missing `data` field on two `wallet.ts` throws, and the missing `createAuthToken` cause-shape test.

This change defines the type contract for `WalletRequestError.cause` and sweeps every site that *already* passes a cause to fit it. A follow-up PR (tracked by HathorNetwork/hathor-wallet-lib#1075) will populate the remaining ~15 currently-bare `walletApi.ts` throw sites and narrow `cause` from optional to mandatory at the type level.

### What changed
- New discriminated union `WalletRequestErrorCause` in `src/errors.ts` with three mutually exclusive variants:
  - `{ status: number; data: unknown }` — HTTP failure (axios response)
  - `{ state: unknown }` — app-level rejection (HTTP succeeded, wallet reported a non-success state)
  - `{ source: WalletRequestError }` — chained failure wrapping a prior error (e.g. polling timeouts)
- Every existing cause-passing site swept to fit one of the variants: `walletApi.ts` (6 sites), `pushNotification.ts` (3 sites), `wallet.ts` (4 sites).
- `pollForWalletStatus` now tracks `lastState` alongside `lastError` so the timeout site always has either a `source` or a `state` to attach (no silent `cause: undefined` on timeout).
- Cleaned up a redundant `as { status?: number }` cast in `startReadOnly` — `'status' in cause` now narrows correctly via the new union.
- New `createAuthToken` unit test asserts the cause shape on a 400 response.
- Two pre-existing tests (`wallet.test.ts` polling-timeout, `readOnlyWallet.test.ts` startup-timeout) updated to match the new `{ source }` shape.

### Next steps (deferred to a follow-up PR)
- Populate the remaining ~15 bare `throw new WalletRequestError(...)` sites in `walletApi.ts` with appropriate `{ status, data }` causes.
- Narrow `WalletRequestError.cause` from `WalletRequestErrorCause | undefined` to `WalletRequestErrorCause` (mandatory) once every throw site has been covered.

### Breaking changes

[...]

PR #1081, opened 2026-04-30
feat(errors): type WalletRequestError.cause as a discriminated union

tuliomir:
## Summary

Re-opens the changes from HathorNetwork/hathor-wallet-lib#1074 against `master`. The original was stacked on top of HathorNetwork/hathor-wallet-lib#1058 (the auth-token race-condition fix) and was closed in an invalid state because GitHub did not retarget its base branch after `#1058` merged.

The diff is unchanged from the closed PR — just rebased onto current master. Addresses three of the four review threads on `#1058`: the typed cause shape, the missing `data` field on two `wallet.ts` throws, and the missing `createAuthToken` cause-shape test.

This change defines the type contract for `WalletRequestError.cause` and sweeps every site that *already* passes a cause to fit it. A follow-up PR (tracked by HathorNetwork/hathor-wallet-lib#1075) will populate the remaining ~15 currently-bare `walletApi.ts` throw sites and narrow `cause` from optional to mandatory at the type level.

### What changed
- New discriminated union `WalletRequestErrorCause` in `src/errors.ts` with three mutually exclusive variants:
  - `{ status: number; data: unknown }` — HTTP failure (axios response)
  - `{ state: unknown }` — app-level rejection (HTTP succeeded, wallet reported a non-success state)
  - `{ source: WalletRequestError }` — chained failure wrapping a prior error (e.g. polling timeouts)
- Every existing cause-passing site swept to fit one of the variants: `walletApi.ts` (6 sites), `pushNotification.ts` (3 sites), `wallet.ts` (4 sites).
- `pollForWalletStatus` now tracks `lastState` alongside `lastError` so the timeout site always has either a `source` or a `state` to attach (no silent `cause: undefined` on timeout).
- Cleaned up a redundant `as { status?: number }` cast in `startReadOnly` — `'status' in cause` now narrows correctly via the new union.
- New `createAuthToken` unit test asserts the cause shape on a 400 response.
- Two pre-existing tests (`wallet.test.ts` polling-timeout, `readOnlyWallet.test.ts` startup-timeout) updated to match the new `{ source }` shape.

### Next steps (deferred to a follow-up PR)
- Populate the remaining ~15 bare `throw new WalletRequestError(...)` sites in `walletApi.ts` with appropriate `{ status, data }` causes.
- Narrow `WalletRequestError.cause` from `WalletRequestErrorCause | undefined` to `WalletRequestErrorCause` (mandatory) once every throw site has been covered.

### Breaking changes

`WalletRequestError.cause` is no longer typed as `unknown`. It is now `WalletRequestErrorCause | undefined`, a discriminated union with three mutually exclusive variants. Consumer code that reads fields off `err.cause` will not compile and/or will read `undefined` at runtime until updated.

**Migration required for any code that catches `WalletRequestError` and inspects `err.cause`.**

1. **Narrow on a discriminant before reading variant fields.** TypeScript will no longer let you read `err.cause.status`, `err.cause.data`, or `err.cause.state` without first checking which variant you have.

   ```ts
   // Before
   if (err instanceof WalletRequestError) {
     const status = (err.cause as any)?.status;
     const body = (err.cause as any)?.data;
   }

   // After
   if (err instanceof WalletRequestError && err.cause) {
     if ('status' in err.cause) {
       const status = err.cause.status;        // number
       const body = err.cause.data;            // unknown
     } else if ('state' in err.cause) {
       const state = err.cause.state;          // unknown — wallet-reported failure state
     } else if ('source' in err.cause) {
       const inner = err.cause.source;         // WalletRequestError — chained failure
     }
   }
   ```

2. **Chained errors are no longer the inner error directly — they are wrapped in `{ source }`.** Previously, sites like `pollForWalletStatus` timeouts assigned the prior `WalletRequestError` straight into `cause`. They now wrap it.

   ```ts
   // Before
   const inner = err.cause as WalletRequestError;
   console.log(inner.message);

   // After
   if (err.cause && 'source' in err.cause) {
     console.log(err.cause.source.message);
   }
   ```

3. **`cause` may still be `undefined`** until the follow-up PR (HathorNetwork/hathor-wallet-lib#1075) covers the remaining bare throw sites in `walletApi.ts`. Always guard with a truthiness check before narrowing.

No runtime behavior changed beyond the polling-timeout site, which now consistently attaches either `{ source }` (last seen error) or `{ state }` (last seen wallet state) instead of occasionally producing `cause: undefined`.

## Acceptance criteria
- `WalletRequestError.cause` is typed; every site that already passed a cause now uses one of three variants.
- New unit test asserts `createAuthToken` attaches `{ status, data }` on non-2xx responses.

## Test plan
- [x] `npm run format`
- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm test` (855 passing, 0 failing)
- [ ] Integration tests — happy path is behaviorally unchanged; will run if requested before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet error diagnostics with structured causes that include HTTP status/body, wallet state, or chained errors—leading to clearer failure context.
  * Better tracking and propagation of errors during wallet status polling and read-only startup.
  * Push notification failures now surface structured error context.

* **Tests**
  * Updated tests to assert the new structured error-cause format across wallet API, polling, retry, and timeout scenarios.


__tests__/wallet/wallet.test.ts
```diff
@@ -3698,8 +3698,9 @@ describe('pollForWalletStatus', () => {
       const error = await caught;
       expect(error).toBeInstanceOf(WalletRequestError);
       expect(error.message).toContain('Wallet status polling timed out');
-      expect(error.cause).toBeInstanceOf(WalletRequestError);
-      expect((error.cause as Error).message).toBe('Persistent server error');
+      const cause = (error as WalletRequestError).cause as { source: WalletRequestError };
+      expect(cause.source).toBeInstanceOf(WalletRequestError);
+      expect(cause.source.message).toBe('Persistent server error');
```
r4mmer, line 3703: This is testing line 768 which introduces a new path the `cause: { state: lastState }` which is untested.
Can we add a test for this path?

src/errors.ts
```diff
@@ -176,15 +176,42 @@ export class WalletError extends Error {
...
+ * Error thrown when executing wallet requests.
+ *
+ * The {@link cause} is currently optional to permit incremental adoption: not
+ * every throw site populates it yet. A follow-up PR will make `cause`
+ * mandatory once every throw site has been swept.
  *
  * @memberof Errors
  * @inner
  */
 export class WalletRequestError extends WalletError {
```
r4mmer, line 208: Very minor issue, I searched where WalletRequestError is used and found `__tests__/wallet/walletServiceRetry.test.ts` which creates some mocked errors to catch them.
Since there is no typecheck on the tests I think this was missed but maybe we should add a mocked cause on that test as well.

__tests__/wallet/wallet.test.ts
tuliomir: Added a parallel test that mocks `getWalletStatus` to always resolve with a `creating` status, so the timeout is reached without `lastError` ever being populated and the ternary picks `{ state: lastState }`. The new test asserts `cause.state` matches the resolved status object.

Implemented on https://github.com/HathorNetwork/hathor-wallet-lib/commit/bec0692b8f25a7d87fc14a6ccba0a85f6f90e19e

src/errors.ts
tuliomir: Not a minor issue in my view: this exposes that the tests are not being properly exposed to the Typescript types, in that they do not break if the types are incorrect. This will demand a dedicated PR ( or more ) to fix. Thanks for bringing this up!

For now, updated the four `{ cause: { status: 400 } }` mocks in `walletServiceRetry.test.ts` to include `data: { success: false }` so they satisfy the new `{ status; data }` variant. Tests do not read `cause.data`, so the value is documentary — minimum-realistic shape.

Implemented on https://github.com/HathorNetwork/hathor-wallet-lib/commit/bec0692b8f25a7d87fc14a6ccba0a85f6f90e19e

raul-oliveira (approved):

r4mmer (approved):

PR #1082, opened 2026-05-06
feat: calculateWeight expects optional weight constants from network

pedroferreira1:
### Acceptance Criteria
- Tx weight is now computed from network-reported constants. Transaction.prepareToSend consults storage.version.{min_tx_weight, min_tx_weight_coefficient, min_tx_weight_k} (cached from the fullnode's /version response) and uses those values to compute tx.weight, instead of always using the hardcoded mainnet defaults (14, 1.6, 100).
- Transaction.calculateWeight now accepts an optional weight-constants override. Calling it with no argument keeps the previous behavior. External callers (downstream apps that construct Transaction instances themselves) can now opt into network-aware weight computation.

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Transactions now dynamically use network-specific weight constants retrieved from the node instead of hardcoded defaults, improving compatibility across different network configurations.

* **Tests**
  * Added integration and unit tests validating weight constants retrieval from storage and transaction calculations with custom constants.

### DIFF
--- a/__tests__/integration/shared/calculate-weight.test.ts
+++ b/__tests__/integration/shared/calculate-weight.test.ts
@@ -6,27 +6,23 @@
  */
 
 /**
- * Verifies that the wallet threads the network-reported
- * `min_tx_weight*` values (via
- * `transactionUtils.getWeightConstantsFromStorage`) into every
- * `Transaction.prepareToSend` invocation, instead of letting it fall
- * back to the hardcoded {@link TX_WEIGHT_CONSTANTS} mainnet defaults.
+ * Verifies that, after a real wallet `start()`, the storage's version
+ * data is populated AND `transactionUtils.getWeightConstantsFromStorage`
+ * resolves to the privnet's reported `(min_tx_weight,
+ * min_tx_weight_coefficient, min_tx_weight_k)` rather than the
+ * hardcoded {@link TX_WEIGHT_CONSTANTS} mainnet defaults.
  *
- * The privnet used by the integration suite reports
- * `min_tx_weight=1, min_tx_weight_coefficient=0, min_tx_weight_k=0`
- * (see `__tests__/integration/configuration/privnet.yml`), which
- * diverges from the defaults (14 / 1.6 / 100) — so checking that
- * `getWeightConstantsFromStorage(storage)` resolves to the privnet
- * triple proves the override is well-formed.
+ * This is the only weight-constants assertion that genuinely exercises
+ * the wallet/fullnode integration: without `start()` actually fetching
+ * /version and writing it to storage, the helper would return
+ * undefined and the wallet would silently fall back to the hardcoded
+ * defaults — the behaviour we're protecting against. Both facades go
+ * through this path: `HathorWallet.start()` (src/new/wallet.ts:1712)
+ * and `HathorWalletServiceWallet.start()` (src/wallet/wallet.ts).
  *
- * The post-send `tx.weight` cannot be inspected directly: the
- * miner's response weight overwrites whatever the wallet computed
- * (see `runFromMining` in `src/new/sendTransaction.ts`). We instead
- * spy on `Transaction.prototype.prepareToSend` to capture the
- * arguments it actually received during a real send.
- *
- * Both wallet facades (fullnode + wallet-service) consume the same
- * `prepareToSend` path, so this is a shared test.
+ * Pure-unit assertions about `getWeightConstantsFromStorage` and
+ * `prepareToSend` argument threading live in
+ * __tests__/utils/transaction.test.ts.
  */
 
 import type { FuzzyWalletType, IWalletTestAdapter } from '../adapters/types';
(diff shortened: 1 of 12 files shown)


__tests__/integration/shared/calculate-weight.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    const networkConstants = transactionUtils.getWeightConstantsFromStorage(storage);
+    expect(networkConstants).toEqual({
+      txMinWeight: versionData.minTxWeight,
+      txWeightCoefficient: versionData.minTxWeightCoefficient,
+      txMinWeightK: versionData.minTxWeightK,
+    });
+  });
+
+  it('returns undefined when the storage version data is not yet populated', () => {
```
tuliomir: chore: Move this test to the unit tests folder, as it has no integration with the blockchain

__tests__/integration/shared/calculate-weight.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    // The helper must tolerate a fresh storage where the fullnode's
+    // /version response hasn't landed yet — callers fall back to
+    // TX_WEIGHT_CONSTANTS in that case.
+    const emptyStorage = { version: null } as unknown as IStorage;
+    expect(transactionUtils.getWeightConstantsFromStorage(emptyStorage)).toBeUndefined();
+  });
+
+  it('threads the network constants through prepareToSend when sending', async () => {
```
tuliomir: chore: Move mocked tests to the unit tests folder, as it does not properly test the actual integration with the other services.

raul-oliveira (commented):
Am I missing something, or is this a no-op for `HathorWalletServiceWallet`?

`storage.setApiVersion(...)` is only called from `HathorWallet.start()` (`src/new/wallet.ts:1712`). I don't see an equivalent in `HathorWalletServiceWallet.start()`, and `walletApi.getVersionData()` doesn't write to storage. So `storage.version` stays `null`, `getWeightConstantsFromStorage(...)` returns `undefined`, and the new `prepareToSend(...)` calls fall back to `TX_WEIGHT_CONSTANTS`.

If that's right, `calculate-weight.test.ts` should also fail on the service adapter at lines 75 and 114.

Would it make sense to call `setApiVersion(await walletApi.getVersionData(this))` in `HathorWalletServiceWallet.start()`?

src/wallet/wallet.ts
```diff
@@ -502,6 +504,41 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
...
+        max_number_inputs: v.maxNumberInputs,
+        max_number_outputs: v.maxNumberOutputs,
+        decimal_places: DECIMAL_PLACES,
+        native_token: null,
+      };
+      this.storage.setApiVersion(apiVersion);
+    } catch (_e) {
+      // Non-fatal: if /version fails the wallet still works, prepareToSend
+      // just falls back to TX_WEIGHT_CONSTANTS as it did before this change.
+    }
```
tuliomir: suggestion(non-blocking): I'd recommend logging an error here using the Wallet Lib logger util.  For short-lived wallets like the Mobile or even a Desktop wallet this may not be a big issue as a reconnection is sure to happen shortly.

However if some long-lived client is generating transactions via the Wallet Service, this difference in calculated weighs may end up becoming an issue to whoever is managing the mining.

__tests__/utils/transaction.test.ts
```diff
@@ -733,3 +735,128 @@ test('convertTransactionToHistoryTx', async () => {
...
+      min_tx_weight: 1,
+      min_tx_weight_coefficient: 0,
+      min_tx_weight_k: 0,
+      token_deposit_percentage: 0.01,
+      reward_spend_min_blocks: 1,
+      max_number_inputs: 255,
+      max_number_outputs: 255,
+      decimal_places: 2,
+      native_token: null,
+    };
```
tuliomir: suggestion(non-blocking): Make this a reusable sample object, so that other tests ( like `'prepareTransaction threads weight constants from storage'` ) may benefit from it.

tuliomir (dismissed):

__tests__/integration/shared/calculate-weight.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1082/commits/62f974d52e54b6563ddae177c45b152a9bc53f46

__tests__/integration/shared/calculate-weight.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1082/commits/62f974d52e54b6563ddae177c45b152a9bc53f46

__tests__/utils/transaction.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1082/commits/853f36f10494ae834bb9763c26a2c7972b1af563

src/wallet/wallet.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1082/commits/853f36f10494ae834bb9763c26a2c7972b1af563

raul-oliveira (approved):

tuliomir (approved):

PR #1085, opened 2026-05-14
feat(shielded): crypto provider interface and send-side primitives

pedroferreira1:
## Summary

Self-contained shielded crypto module — the Node-side provider wrapping `@hathor/ct-crypto-node`, a browser stub that throws on non-verify methods, and the send-side primitive that builds shielded outputs with correct blinding-factor balancing. **PR 2 of 7** in the decomposition of [PR #1059](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1059).

```
master ─ PR1 ─ PR2 ◀ PR3 ─ PR4 ─ PR5 ─ PR6 ─ PR7
```

Base: `shielded/pr-1-types-and-model`.

Crypto provider package: https://github.com/HathorNetwork/hathor-ct-crypto/tree/master/packages/ct-crypto-provider

## Acceptance Criteria

- Wallet-lib can instantiate a Node-side shielded crypto provider that wraps the `@hathor/ct-crypto-node` native addon.
- `@hathor/ct-crypto-node` is NOT a declared dependency of wallet-lib (consumers install it themselves).
- The native-addon `require()` is lazy (no module-init load); calls from transparent-only flows never touch it.
- When the addon is missing at runtime, the wallet throws a clear remediation message instead of Node's raw `MODULE_NOT_FOUND`.
- A browser-side provider is available that exposes verifier-only primitives; non-verify methods throw with a descriptive error.
- `createShieldedOutputs` builds N shielded outputs whose blinding factors satisfy the homomorphic balance equation (N-1 random VBFs + 1 computed balancing factor).
- Surjection-proof domain construction is correct for transparent / AmountShielded / FullShielded inputs (uses `ZERO_TWEAK` where appropriate).
- Pubkey and token-UID sizes are validated up front before any crypto operation.
- No existing wallet code path imports the new modules — dead code at runtime until follow-up PRs wire them up.
- Transparent-only wallets see byte-identical behavior to master.

## ⚠️ Note on `@hathor/ct-crypto-node`

`@hathor/ct-crypto-node` is intentionally **NOT** a declared dependency of wallet-lib (the package is still WIP and not on npm). Consumers that need shielded crypto install it themselves; consumers that only use transparent flows are unaffected. The `require()` is **lazy** (inside the factory, not at module-import time) and wrapped in `try/catch` — if the package is missing when a client reaches into shielded crypto, the wallet-lib throws a clear remediation message instead of the raw Node `MODULE_NOT_FOUND`:

> _Shielded crypto support requires the @hathor/ct-crypto-node native addon. Install it in the consuming application (it is not a declared dependency of wallet-lib while the package is still WIP). Underlying error: ..._

## What's in this PR

- `src/shielded/provider.ts` — `createDefaultShieldedCryptoProvider`. Lazy `require('@hathor/ct-crypto-node')`, async-wraps the native addon's synchronous methods so every consumer can `await` uniformly. Clear error if the addon isn't installed.
- `src/shielded/provider.browser.ts` — `createBrowserShieldedCryptoProvider`. Every non-verify method throws (verify works against headerful txs from raw bytes; spending / decryption requires Node).
- `src/shielded/creation.ts` — `createShieldedOutputs`. Implements the homomorphic balance equation: N-1 outputs use random VBFs; the last is computed via `computeBalancingBlindingFactor` so all blinding factors sum to zero. Validates pubkey / token sizes up front, builds surjection-proof domain from input generators (with `ZERO_TWEAK` for transparent / AmountShielded inputs).
- `src/types.ts` — `IDataShieldedOutput` interface (intermediary repr during transaction building). Additive only.
- `src/shielded/index.ts` — barrel extended with provider factory exports.
- `__tests__/shielded/{provider,creation}.test.ts` — unit coverage. The provider test gracefully skips when the native addon isn't present.

## Behavior preservation

Nothing in the existing wallet code path imports any of these modules. Dead code from a runtime standpoint until follow-up PRs wire it up.

## Plan deviation

Receive-side processing (`processShieldedOutputs`) is **deferred to PR 5** because it needs the `IStorage` shielded extensions (`getScanXPrivKey`, etc.) which live with the storage-layer changes.

## Summary by CodeRabbit

* **New Features**
  * Added createShieldedOutputs to build Amount-shielded and Fully-shielded transaction outputs.
  * Shielded output types exposed through the main API for consumer use.

* **Tests**
  * Added comprehensive tests covering creation, validation guards, error propagation, balancing blinding behavior, and surjection-proof generation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-lib/pull/1085?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/shielded/index.ts
```diff
@@ -17,3 +17,11 @@ export type {
...
   IRewoundFullShieldedOutput,
 } from './types';
+
+// Provider implementations live with their respective ct-crypto packages:
+//   - Node:    @hathor/ct-crypto-node    → createDefaultShieldedCryptoProvider()
+//   - Browser: @hathor/ct-crypto-wasm    → createBrowserShieldedCryptoProvider()
+//   - Mobile:  built into the app target (RN native module)
+//
+// Install whichever crypto package matches your runtime, then wire it in via
+// `wallet.setShieldedCryptoProvider(createXxxShieldedCryptoProvider())`.
```
pedroferreira1, line 27: I thought about removing this comment added by claudin but I liked it in the end.

pedroferreira1:
@coderabbitai resolve the previous comments and run a full review again. Lot has changed

pedroferreira1:
@coderabbitai all your comments were addressed

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+  inputGenerators: InputGeneratorInfo[] = [],
+  blindedInputs: ShieldedInputBlinding[] = []
+): Promise<IDataShieldedOutput[]> {
+  if (proposals.length === 0) return [];
+
+  // Validate inputs upfront before expensive crypto work
+  const hasFullShielded = proposals.some(p => p.shieldedMode === ShieldedOutputMode.FULLY_SHIELDED);
+  for (const [idx, proposal] of proposals.entries()) {
+    const pubkeyBuf = Buffer.from(proposal.scanPubkey, 'hex');
+    if (pubkeyBuf.length !== 33) {
```
r4mmer: nitpick(non-blocking): Use constants instead of magic numbers.

33 as compressed pubkey length in bytes and 32 as token length are known so this is less of a problem but i like having the variable name explain what is being tested.

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * Create shielded outputs with cryptographic commitments and proofs.
+ *
+ * The homomorphic balance equation requires blinding factors to sum to zero.
+ * For transparent-only inputs, all input blinding factors are zero.
+ * For shielded inputs, their blinding factors are passed via `blindedInputs`.
+ * N-1 outputs use random blinding factors; the last output's blinding factor
+ * is computed via computeBalancingBlindingFactor to satisfy the constraint.
+ */
+export async function createShieldedOutputs(
```
r4mmer, line 256: suggestion(if-minor): Instead of having 1 big function that is kind of hard to read, maybe create 2 smaller ones. 

The current function is too big with 4 branches (amount/fully shielded and isLast or not) maybe you should create a `_createAmountShieldedOutput` and `_createFullyShieldedOutput` and call them in this function.
Way better to read and understant what each part is doin.

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+    const tokenUidBuf = Buffer.from(
+      proposal.token === NATIVE_TOKEN_UID ? NATIVE_TOKEN_UID_HEX : proposal.token,
+      'hex'
+    );
+
+    let cryptoResult;
+    const isLast = i === proposals.length - 1;
+
+    try {
+      if (isLast && createdOutputs.length > 0) {
```
r4mmer: The balancing blinding factor is only computed in this branch, but what about the 1 shielded output case?
`isLast` is true but `createdOutputs.length` is 0 so it would not enter here and would not compute the balancing blinding factor?

Is this correct?

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+        blindingFactor: cryptoResult.blindingFactor,
+        assetCommitment: cryptoResult.assetCommitment,
+        assetBlindingFactor: cryptoResult.assetBlindingFactor,
+        surjectionProof,
+        script: scriptBuf.toString('hex'),
+      });
+    } catch (e) {
+      const mode = fullyShielded ? 'FullShielded' : 'AmountShielded';
+      throw new Error(
+        `Failed to create shielded output ${i}/${proposals.length} (mode=${mode}, token=${proposal.token}): ${e}`
```
r4mmer: suggestion(nitpick): avoid swallowing the stack trace

```suggestion
        `Failed to create shielded output ${i}/${proposals.length} (mode=${mode}, token=${proposal.token}): ${e}`, { cause: e}
```

This is a small point just to help debug errors in the future

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+import Network from '../models/network';
+import {
+  IBlindingEntry,
+  IDataShieldedOutput,
+  IShieldedCryptoProvider,
+  ISurjectionDomainEntry,
+  ShieldedOutputMode,
+} from './types';
+
+interface ShieldedOutputProposal {
```
r4mmer: question: Why is this not exported?

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+  shieldedMode: ShieldedOutputMode;
+  timelock?: number;
+}
+
+/**
+ * Re-export of `IBlindingEntry` under wallet-lib's local name. The contract
+ * lives in `@hathor/ct-crypto-provider`; alias kept for back-compat with the
+ * old `ShieldedInputBlinding` type that callers may still import.
+ */
+export type ShieldedInputBlinding = IBlindingEntry;
```
r4mmer: back-compatibility with what? everything here is new.

src/shielded/types.ts
```diff
@@ -241,3 +79,31 @@ export interface IProcessedShieldedOutput {
...
+/**
+ * Intermediary representation of a shielded output during transaction
+ * building — used as the return shape of `createShieldedOutputs()`.
+ *
+ * The non-optional prefix is the pre-crypto definition (mirrors what
+ * `ShieldedOutputProposal` carries in). Everything after `shieldedMode` is
+ * populated by the crypto provider in a single pass; the fields are
+ * marked optional only because `assetCommitment`/`assetBlindingFactor`/
+ * `surjectionProof` are FullShielded-only — for AmountShielded outputs
+ * they remain undefined.
```
tuliomir: suggestion(non-blocking): Use zod schemas for this kind of very specific type union.

If we have an `AmountShieldedOutputSchema` and a `FullShieldedOutputSchema` we can specify which fields are mandatory for each type of output and then join them into a final type.

src/shielded/creation.ts
tuliomir: I agree. Having just read the RFC, it would be nice to have a code implementation that also separates the privacy tiers just like the RFC does.

It could also help with the zod typings mentioned in a comment above, since each function would need to have its own schema definitions of inputs and outputs.

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+      // For FullShielded inputs: use the blinded generator (asset_commitment) reconstructed
+      // from the input's asset blinding factor — the fullnode verifies against this.
+      let surjectionProof: Buffer | undefined;
+      if (fullyShielded) {
+        // assetBlindingFactor is guaranteed non-null inside the FullShielded
+        // branches above (we throw if the provider doesn't supply it), but
+        // TypeScript can't carry that narrowing across the cryptoResult
+        // reassignment in different if-branches — assert here.
+        if (!cryptoResult.assetBlindingFactor) {
+          throw new Error(
```
tuliomir: chore: This is a non-trivial error handling case, so we should have tests for it.

src/shielded/creation.ts
```diff
@@ -0,0 +1,287 @@
...
+        commitment: cryptoResult.commitment,
+        rangeProof: cryptoResult.rangeProof,
+        blindingFactor: cryptoResult.blindingFactor,
+        assetCommitment: cryptoResult.assetCommitment,
+        assetBlindingFactor: cryptoResult.assetBlindingFactor,
+        surjectionProof,
+        script: scriptBuf.toString('hex'),
+      });
+    } catch (e) {
+      const mode = fullyShielded ? 'FullShielded' : 'AmountShielded';
```
tuliomir: suggestion(non-blocking): Create an `enum` for this mode.

src/shielded/creation.ts
tuliomir: I think this case is being handled by the `'should handle single output (no balancing needed)'` test.

But if I'm mistaken I think this is a valuable edge case to add to the test suite: a tx with one shielded utxo as input, a transparent output and a shielded output. It would be a nice mirror of the RFC explanation of the rules exceptions.

src/shielded/creation.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/b1d9aa1704cad134e0342758169d89f04fb5385e

src/shielded/creation.ts
pedroferreira1: The protocol (hathor-core) does not allow a transaction with a single shielded outputs, so the wallet-lib also has a protection for it.

Added a unit test and another check in this method: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/0175202a673dabc35118c635298427fe8b04c578

src/shielded/creation.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/24bfa74fb4caf4e5d9025654f6ff3db64ffbfa37

had to add to the tsconfig as well to allow the error constructor to support this

src/shielded/creation.ts
pedroferreira1: It should've been exported. Actually I moved the types in this file to the shielded types file. https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/113a198a4481e59ad085cbab8b429984030fdc59

src/shielded/creation.ts
pedroferreira1: Correct, removed the alias, no need for backward compatibility: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/c32fca62562d7f561c00b726804e7fc0771b92bc

src/shielded/creation.ts
pedroferreira1: Used a reverse lookup from the string we have in the lib, so we don't need to create a new enum only for this error string

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/a3b0581a5fbf6f45d7ce5be6633dd40635c6400c

src/shielded/creation.ts
pedroferreira1: We do have many (many) tests in the PR7 of the stack of PRs. It just needs the full files to run integration tests

src/shielded/types.ts
pedroferreira1: Did this change but had to also push specific outputs now that the types are differents: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/bfb4a0568b824114c7fbc9e6941c7101ddc98a40

src/shielded/creation.ts
pedroferreira1: Much better with this refactor, thanks: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/3fb088397deaf180dfbfb489693b0af4c70c7b4c?diff=split

src/shielded/creation.ts
```diff
@@ -0,0 +1,368 @@
...
+  // assetBlindingFactor and assetCommitment for any FullShielded output.
+  // Silently falling back to ZERO_TWEAK or skipping the surjection proof
+  // would produce a malformed FullShielded output the fullnode (or wallet
+  // decryption) rejects later with a confusing error; fail loud here at
+  // the contract boundary instead.
+  if (!cryptoResult.assetBlindingFactor) {
+    throw new Error('Crypto provider returned no assetBlindingFactor for FullShielded output');
+  }
+  if (!cryptoResult.assetCommitment) {
+    throw new Error('Crypto provider returned no assetCommitment for FullShielded output');
```
tuliomir, line 208: question(non-blocking): I assume we'll have a unit test for that situation in a future PR?

It actually looks like we should have it right now, similar as the one with the `assetBlindingFactor` that is rejected on `'rejects when provider returns FullShielded result without assetBlindingFactor (last)'`, but maybe we'll have a more thorough test later.

src/shielded/creation.ts
```diff
@@ -0,0 +1,368 @@
...
+  network: Network,
+  inputGenerators: InputGeneratorInfo[] = [],
+  blindedInputs: IBlindingEntry[] = []
+): Promise<IDataShieldedOutput[]> {
+  if (proposals.length === 0) return [];
+  if (proposals.length === 1) {
+    throw new Error(
+      'At least 2 shielded outputs are required (hathor-core trivial-commitment rule)'
+    );
+  }
```
tuliomir, line 268: question: We have an exception on the RFC, which is a case where we have a transparent output alongside a shielded one, when the input is also shielded.

https://github.com/HathorNetwork/rfcs/blob/9e9006b72208c1ac0831a37e516e021a4491b3e3/text/0000-shielded-outputs.md?plain=1#L518

Is this accounted for in this logic here? It looks to me it would reject this edge case.

tuliomir (dismissed):

src/shielded/creation.ts
pedroferreira1: The verification function in the hathor-core is implemented differently:                                                            
                                                                                                      
```
  hathor/verification/transaction_verifier.py:1219-1227:                                              
                                                                                                      
  def verify_trivial_commitment_protection(self, tx: Transaction) -> None:                            
      """Without storage, conservatively require >= 2 shielded outputs always."""                     
      if not tx.shielded_outputs:                                                                     
          return                                                                                      
      if len(tx.shielded_outputs) < 2:                                                                
          raise TrivialCommitmentError(                                                               
              'at least 2 shielded outputs are required '                                             
              f'to prevent trivial commitment matching (got {len(tx.shielded_outputs)})' 
```

It raises error regardless of the amount of inputs, it just checkes if it has at least 2 outputs.

src/shielded/creation.ts
pedroferreira1: Added http://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/commits/3d472ff0223aafdf8b027975dcb75030e5263e42

tuliomir (dismissed):

__tests__/shielded/creation.test.ts
```diff
@@ -0,0 +1,289 @@
...
+
+import { createShieldedOutputs } from '../../src/shielded/creation';
+import { IShieldedCryptoProvider, ShieldedOutputMode } from '../../src/shielded/types';
+import Network from '../../src/models/network';
+
+function makeMockProvider(
+  overrides: Partial<IShieldedCryptoProvider> = {}
+): IShieldedCryptoProvider {
+  return {
+    generateRandomBlindingFactor: jest.fn().mockReturnValue(Buffer.alloc(32, 0x01)),
```
r4mmer: nitpick (non-blocking): Switch `mockReturnValue` → `mockResolvedValue` to match the async provider contract.

CodeRabbit flagged this but should we use async mocks to correctly reflect the implementation?
Using `mockResolvedValue(...)` or `mockRejectedValue(...)` (for success and errors) or simply using an async function on `mockImplementation` would work.

src/shielded/creation.ts
```diff
@@ -0,0 +1,368 @@
...
+      createdOutputs
+    );
+    cryptoResult = await cryptoProvider.createAmountShieldedOutput(
+      proposal.value,
+      recipientPubkeyBuf,
+      tokenUidBuf,
+      balancingBf
+    );
+  } else {
+    // Non-last output (or single-call with no siblings): random vbf.
```
r4mmer: is this up-to-date?
Correct me if i'm wrong but I think that a single-call with no siblings would not read here.

r4mmer (dismissed):

src/shielded/creation.ts
pedroferreira1: You are right, it's only the non-last output. Updated the comment: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/commits/7d844110a65a84dc51cf22676f9c93f501e8ab39

__tests__/shielded/creation.test.ts
pedroferreira1: It makes sense, updated it https://github.com/HathorNetwork/hathor-wallet-lib/pull/1085/changes/e5e96945c8b47d81f8aa89b027b4898d3e76ead7

r4mmer (approved):

tuliomir (approved):

PR #1087, opened 2026-05-14
feat(shielded): address derivation, network version byte, key chains

pedroferreira1:
## Summary

Establishes the shielded address format (71-byte: 1B version + 33B scan + 33B spend + 4B checksum) and the BIP32 key-chain layout that backs it. **PR 4 of 7** in the decomposition of [PR #1059](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1059).

```
master ─ PR1 ─ PR2 ─ PR3 ─ PR4 ◀ PR5 ─ PR6 ─ PR7
```

Base: `shielded/pr-3-headers`.

## Acceptance Criteria

- Wallet-lib can encode a 71-byte shielded address (`encodeShieldedAddress`) and decode it back to its scan + spend pubkeys.
- Wallet-lib can derive a shielded address pair at a given BIP32 index from scan/spend xpubs (`deriveShieldedAddress`).
- The `Address` model recognizes the 71-byte shielded format and exposes `isShielded()`, `getScanPubkey()`, `getSpendPubkey()`, `getSpendAddress()`, `getScript()`.
- Mainnet shielded version byte is `0x3c`, testnet/privatenet is `0x5d` — matches the protocol spec.
- `Network.isVersionByteValid` accepts the new byte alongside the existing P2PKH/P2SH ones.
- BIP32 paths `m/44'/280'/1'/0` (scan) and `m/44'/280'/2'/0` (spend) match the spec.
- Fresh seed-derived and xpriv-derived wallets land with shielded scan + spend keys in their access-data record automatically.
- Existing wallets persisted before shielded support can be upgraded via `migrateShieldedAccessData`, which is idempotent and safely no-ops for xpub-only / xpriv-imported wallets.
- PIN changes via `changePinOnAccessData` correctly re-encrypt the shielded keys when present.
- The transparent send pipeline rejects shielded addresses with a clear error (`getAddressType` throws — shielded routing lands in PR 6).
- `createOutputScriptFromAddress` correctly emits a P2PKH script derived from the spend pubkey for a shielded address.
- Malformed shielded addresses are rejected even with valid checksums: length must match the version-byte family, and both embedded pubkeys must be points on secp256k1.
- No behavior change for transparent-only wallets — existing serialized access-data, addresses, and tx flows remain identical.

## Behavior preservation

- `Network.versionBytes.shielded` is read only by `Address.getType()` when the first byte matches. Existing addresses never satisfy.
- `Address(71-byte string)` now constructs successfully (previously threw). To preserve the transparent send pipeline, `sendTransaction.ts` now routes through `getAddressType()` which throws on shielded input.
- New shielded keys land in `accessData` for **fresh** seed-derived / xpriv-derived wallets only; existing serialized `accessData` is untouched until `migrateShieldedAccessData` is called (no caller invokes it yet — that wiring is in PR 6).
- `migrateShieldedAccessData` is itself idempotent and returns `false` (no-op) on xpub-only / xpriv-imported / already-migrated records.

## Plan deviation

Storage-coupled derivation (`deriveShieldedAddressFromStorage`) is **deferred to PR 5** along with the `IStorage` shielded extensions (`getScanXPubKey`, `getSpendXPubKey`) that it depends on. That keeps PR 4 focused on the primitives + key chains and PR 5 focused on storage.

## Summary by CodeRabbit

* **New Features**
  * Added shielded address support: detection, validation, on-chain spend-address derivation, and P2PKH output-script handling; networks now recognize shielded version bytes; wallet access data can generate, migrate, and re-encrypt shielded scan/spend key material.

* **Tests**
  * Added comprehensive tests covering shielded encoding/decoding, curve/key validation, cross-network checks, deterministic derivation, spend-address/script derivation, wallet migration, and PIN/passphrase edge cases.

### DIFF
--- a/__tests__/utils/wallet.test.ts
+++ b/__tests__/utils/wallet.test.ts
@@ -687,3 +687,138 @@ test('getXprivFromData', () => {
   );
   expect(xpriv).toEqual(hdPrivKey.xprivkey);
 });
+
+describe('migrateShieldedAccessData', () => {
+  const seed =
+    'upon tennis increase embark dismiss diamond monitor face magnet jungle scout salute rural master shoulder cry juice jeans radar present close meat antenna mind';
+
+  function makePreShieldedRecord() {
+    // Freshly-generated record, then strip the four shielded fields to
+    // simulate a persisted record from a pre-shielded wallet-lib version.
+    const full = wallet.generateAccessDataFromSeed(seed, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    const { scanXpubkey, scanMainKey, spendXpubkey, spendMainKey, ...preShielded } = full;
+    // Ensure the stripping actually matched the current schema so this test
+    // fails loudly if the fields are renamed.
+    expect(scanXpubkey).toBeDefined();
+    expect(scanMainKey).toBeDefined();
+    expect(spendXpubkey).toBeDefined();
+    expect(spendMainKey).toBeDefined();
+    return preShielded;
+  }
+
+  test('M.1 — no-op when record already has all four shielded fields', () => {
+    const full = wallet.generateAccessDataFromSeed(seed, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    const before = { ...full };
+    const migrated = wallet.migrateShieldedAccessData(full, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    expect(migrated).toBe(false);
+    // Fields must be byte-identical — nothing should have been touched.
+    expect(full.scanXpubkey).toBe(before.scanXpubkey);
+    expect(full.scanMainKey).toEqual(before.scanMainKey);
+    expect(full.spendXpubkey).toBe(before.spendXpubkey);
+    expect(full.spendMainKey).toEqual(before.spendMainKey);
+  });
+
+  test('M.2 — no-op when xpub-only (no words to derive from)', () => {
+    // xpub-only records don't carry `words`, so migration has nothing to
+    // decrypt. This includes read-only view wallets that never had a seed.
+    const preShielded = makePreShieldedRecord();
+    delete preShielded.words;
+    const migrated = wallet.migrateShieldedAccessData(preShielded, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    expect(migrated).toBe(false);
+    expect(preShielded.scanXpubkey).toBeUndefined();
+    expect(preShielded.scanMainKey).toBeUndefined();
+    expect(preShielded.spendXpubkey).toBeUndefined();
+    expect(preShielded.spendMainKey).toBeUndefined();
+  });
+
+  test('M.3 — migrates pre-shielded record with the same derivation as fresh-create', () => {
+    const preShielded = makePreShieldedRecord();
+    const migrated = wallet.migrateShieldedAccessData(preShielded, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    expect(migrated).toBe(true);
+
+    const fresh = wallet.generateAccessDataFromSeed(seed, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+
+    // xpubs are deterministic (no random salt) — must match byte-for-byte so
+    // a migrated wallet and a freshly-created wallet derive identical
+    // shielded addresses at every index.
+    expect(preShielded.scanXpubkey).toBe(fresh.scanXpubkey);
+    expect(preShielded.spendXpubkey).toBe(fresh.spendXpubkey);
+
+    // mainKeys use a random salt — they will not be byte-identical, but
+    // they must be valid encrypted blobs that decrypt with the same PIN.
+    expect(preShielded.scanMainKey).toBeDefined();
+    expect(preShielded.spendMainKey).toBeDefined();
+    expect(checkPassword(preShielded.scanMainKey!, '123')).toBe(true);
+    expect(checkPassword(preShielded.spendMainKey!, '123')).toBe(true);
+  });
+
+  test('M.4 — idempotent: calling again after migration is a no-op', () => {
+    const record = makePreShieldedRecord();
+    wallet.migrateShieldedAccessData(record, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    const afterFirst = { ...record };
+    const second = wallet.migrateShieldedAccessData(record, {
+      pin: '123',
+      password: '456',
+      networkName: 'testnet',
+    });
+    expect(second).toBe(false);
+    expect(record.scanXpubkey).toBe(afterFirst.scanXpubkey);
+    expect(record.scanMainKey).toEqual(afterFirst.scanMainKey);
+    expect(record.spendXpubkey).toBe(afterFirst.spendXpubkey);
+    expect(record.spendMainKey).toEqual(afterFirst.spendMainKey);
+  });
+
+  test('M.5 — wrong password throws and leaves the record untouched', () => {
+    const record = makePreShieldedRecord();
+    let caught: unknown;
+    try {
+      wallet.migrateShieldedAccessData(record, {
+        pin: '123',
+        password: 'wrong-password',
+        networkName: 'testnet',
+      });
+    } catch (e) {
+      caught = e;
+    }
+    // We wrap decryptData's failure in a specific message so operators
+    // know to provide the wallet password (not PIN). The original
+    // InvalidPasswdError stays available via .cause.
+    expect(caught).toBeInstanceOf(Error);
+    expect((caught as Error).message).toMatch(/wallet password \(not PIN\)/);
+    expect((caught as Error & { cause?: unknown }).cause).toBeInstanceOf(InvalidPasswdError);
+    // Critical: no fields should have been assigned before the throw. If
+    // decryptData throws, we must not leave a half-migrated record.
+    expect(record.scanXpubkey).toBeUndefined();
+    expect(record.scanMainKey).toBeUndefined();
+    expect(record.spendXpubkey).toBeUndefined();
+    expect(record.spendMainKey).toBeUndefined();
+  });
+});
--- a/__tests__/utils/wallet.test.ts
+++ b/__tests__/utils/wallet.test.ts
@@ -15,10 +15,16 @@ import {
   InvalidPasswdError,
 } from '../../src/errors';
 import Network from '../../src/models/network';
-import { HD_WALLET_ENTROPY, HATHOR_BIP44_CODE, P2SH_ACCT_PATH } from '../../src/constants';
+import {
+  HD_WALLET_ENTROPY,
+  HATHOR_BIP44_CODE,
+  P2SH_ACCT_PATH,
+  SHIELDED_SCAN_ACCT_PATH,
+  SHIELDED_SPEND_ACCT_PATH,
+} from '../../src/constants';
 import { hexToBuffer } from '../../src/utils/buffer';
 import { WalletType, WALLET_FLAGS } from '../../src/types';
-import { checkPassword } from '../../src/utils/crypto';
+import { checkPassword, decryptData } from '../../src/utils/crypto';
 
 test('Words', () => {
   const words = wallet.generateWalletWords();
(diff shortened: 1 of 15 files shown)


pedroferreira1:
@coderabbitai all your comments were addressed. Do a re-review

pedroferreira1:
@coderabbitai all new comments addressed as well. Please do a re-review

src/new/sendTransaction.ts
```diff
@@ -199,7 +198,7 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
           timelock: output.timelock ? output.timelock : null,
           authorities: 0n,
           token: output.token,
-          type: addressObj.getType(),
```
pedroferreira1, line 202: This will throw for shielded addresses. Support for sending tx to shielded addresses will be added in the next PRs. For now, we are just refactoring to use the util method

src/models/address.ts
```diff
@@ -152,6 +286,12 @@ class Address {
    */
   getScript(): Buffer {
     const addressType = this.getType();
+    if (addressType === 'shielded') {
+      // For shielded addresses, derive P2PKH script from spend_pubkey
+      const spendAddress = this.getSpendAddress();
+      const p2pkh = new P2PKH(spendAddress);
+      return p2pkh.createScript();
+    }
```
pedroferreira1, line 337: For now, we accept only p2pkh scripts for shielded outputs

src/types.ts
```diff
@@ -96,11 +96,24 @@ export type HistorySyncFunction = (
   shouldProcessHistory?: boolean
 ) => Promise<void>;
 
+/**
+ * Valid address types, as returned by `Address.getType()`:
+ * the two legacy script types plus the 71-byte shielded address format.
+ */
+export type AddressType = 'p2pkh' | 'p2sh' | 'shielded';
```
tuliomir, line 103: thought: This would usually be served by an `enum`, but you need to extend it on `IAddressInfo` so I'm not really sure an enum would be the best fit here.

Bringing this up just as a consistency flag, but maybe we'll not be able to change this.

src/utils/wallet.ts
```diff
@@ -699,9 +723,95 @@ const wallet = {
...
+    const scanAcctXpriv = rootXpriv.deriveNonCompliantChild(SHIELDED_SCAN_ACCT_PATH);
+    const scanXpriv = scanAcctXpriv.deriveNonCompliantChild(0);
+    const spendAcctXpriv = rootXpriv.deriveNonCompliantChild(SHIELDED_SPEND_ACCT_PATH);
+    const spendXpriv = spendAcctXpriv.deriveNonCompliantChild(0);
+
+    /* eslint-disable no-param-reassign */
+    accessData.scanXpubkey = scanXpriv.xpubkey;
+    accessData.scanMainKey = encryptData(scanXpriv.xprivkey, pin);
+    accessData.spendXpubkey = spendXpriv.xpubkey;
+    accessData.spendMainKey = encryptData(spendXpriv.xprivkey, pin);
```
tuliomir, line 837: question: Investigating this code with Claude , it raised a possible issue that this rmutation could break with a PIN change, and we would need additional testing to ensure this doesn't happen. Can you please look into it?

> ### `changeEncryptionPin` re-encrypts scan/spend keys, but no test asserts it
> `src/utils/wallet.ts` adds re-encryption of `scanMainKey`/`spendMainKey`. The existing `change pin and password` test (`__tests__/utils/wallet.test.ts:631`) feeds it a `generateAccessDataFromSeed` record, so **these new lines run** and show as covered — but the test only asserts `mainKey`/`authKey`/`words` afterward (lines 653-655). It never checks `checkPassword(scanMainKey, newPin)` / `spendMainKey`.
> 
> Failure that slips through: swapping `oldPin`/`newPin`, copy-pasting `scanKey` into the spend branch, or dropping one block entirely — all stay green. After such a regression, a user who changes their PIN can never decrypt their shielded keys.
*Add:* `expect(checkPassword(pinChangedAccessData.scanMainKey!, '321')).toBe(true)` (+ spend) to that test.

src/new/sendTransaction.ts
tuliomir: chore: Let's add a test asserting that this throw actually happens for shielded addresses.

Even though we want to merge the next PRs in a matter of days, having this gap well covered with tests is an additional security.

src/utils/wallet.ts
tuliomir: Another approach to this potential issue, by another perspective:

> No negative test for a PIN that doesn't match the wallet's existing key material. The PIN round-trip is covered (M.3), but unlike the password (validated via decryptData(words, …), tested in M.5), the migration never cross-checks the PIN against mainKey, so a mismatched PIN silently yields scan/spend blobs encrypted under a different PIN than the rest of the record. Optional hardening: decrypt mainKey with pin first to fail loud, plus a test.
> Low priority — requires a caller error and the unlock PIN is normally pre-validated.

__tests__/utils/wallet.test.ts
```diff
@@ -687,3 +687,138 @@ test('getXprivFromData', () => {
...
 });
+
+describe('migrateShieldedAccessData', () => {
+  const seed =
+    'upon tennis increase embark dismiss diamond monitor face magnet jungle scout salute rural master shoulder cry juice jeans radar present close meat antenna mind';
+
+  function makePreShieldedRecord() {
+    // Freshly-generated record, then strip the four shielded fields to
+    // simulate a persisted record from a pre-shielded wallet-lib version.
+    const full = wallet.generateAccessDataFromSeed(seed, {
```
tuliomir, line 771: suggestion: In the same theme of the password/pin validation, Claude suggests adding a test for a seed+passphrase to ensure the encryption will work accordingly.

> ### 4. Migration is untested with a non-empty BIP39 passphrase — Medium
> Every migration test uses the default `passphrase = ''`. The shielded root is derived via `code.toHDPrivateKey(passphrase, …)`. A wallet originally created **with** a passphrase has its legacy keys rooted at seed+passphrase; if migration runs without it (or with the wrong one), the scan/spend keys are derived from a *different root* than the wallet's transparent keys — so shielded and transparent addresses no longer share a seed root, with no error raised. M.3 only proves consistency *within* the empty-passphrase case (migrated == fresh-create, both passphrase-less).
> *Add:* a test that creates with a passphrase, migrates with the same passphrase, and asserts the shielded xpubs match a fresh-create — and ideally one showing a wrong passphrase produces detectably different keys.

src/utils/wallet.ts
```diff
@@ -619,6 +621,21 @@ const wallet = {
       accessData.acctPathKey = encryptedAcctPathKey;
     }
 
+    // Derive shielded scan and spend keys if root key is available.
+    // Scan (account 1') and spend (account 2') use separate derivation paths from legacy (account 0')
+    // so the scan key only grants view access, not spending authority over legacy funds.
+    if (argXpriv.depth === 0) {
```
tuliomir, line 627: question: Another case to verify if it's really an issue, regarding asserting the `scan` and `spend` fields:

> ### `generateAccessDataFromXpriv` shielded branch: executed but never asserted, and the depth guard's two arms aren't pinned — Medium
> The new block at `src/utils/wallet.ts:578` derives shielded keys only when `argXpriv.depth === 0`. The existing `access data from xpriv` test calls it with a root xpriv (depth 0 → fields present) and with change/account xprivs (depth > 0 → fields absent), but every assertion uses `.toMatchObject`, which **ignores both the presence and the absence** of `scan*/spend*`. So:
> - the root-import shielded derivation is line-covered but unasserted, and
> - the "account-level import correctly has *no* shielded fields" invariant (documented in `types.ts`) is unverified.
> 
> There's also no cross-check that **a seed import and a root-xpriv import of the same seed yield identical shielded keys** — the kind of asymmetry (`deriveNonCompliantChild` vs. compliant derivation, captured in the code comments) that this stack is clearly sensitive to.

__tests__/utils/wallet.test.ts
```diff
@@ -687,3 +687,138 @@ test('getXprivFromData', () => {
...
+    expect(caught).toBeInstanceOf(Error);
+    expect((caught as Error).message).toMatch(/wallet password \(not PIN\)/);
+    expect((caught as Error & { cause?: unknown }).cause).toBeInstanceOf(InvalidPasswdError);
+    // Critical: no fields should have been assigned before the throw. If
+    // decryptData throws, we must not leave a half-migrated record.
+    expect(record.scanXpubkey).toBeUndefined();
+    expect(record.scanMainKey).toBeUndefined();
+    expect(record.spendXpubkey).toBeUndefined();
+    expect(record.spendMainKey).toBeUndefined();
+  });
```
tuliomir, line 896: question: Here Claude considers the possibility of an incomplete migration. Not sure this would be possible at all, but worth considering.

> ### `migrateShieldedAccessData` partial-record case is untested — Low
> The guard is `hasAll = scanXpubkey && scanMainKey && spendXpubkey && spendMainKey`. A record with **some but not all** four fields (e.g. an interrupted earlier migration, or a future schema that adds one field before the others) takes the re-derive path and overwrites everything. No test covers the partial state — M.1 tests all-present, M.2/M.3 test all-absent.

src/types.ts
pedroferreira1: Agreed on keeping the literal union. The deciding factor is exactly the one you spotted:
  IAddressInfo.addressType extends it with 'shielded-spend', and these values are persisted in storage
  records where plain strings serialize naturally. A union also gives the same exhaustiveness
  checking an enum would, without the import coupling. Keeping as-is.

src/utils/wallet.ts
pedroferreira1: Good catch — the lines were covered but nothing asserted the re-encryption, so an oldPin/newPin
  swap, a scan/spend copy-paste, or a dropped block all stayed green. Added assertions for each:
  checkPassword under the new pin (and explicitly NOT the old), decrypted xprivs unchanged
  (re-encryption, not re-derivation), and password change leaving the pin-encrypted keys untouched.

__tests__/utils/wallet.test.ts
pedroferreira1: Added M.7 (create with passphrase, migrate with the same one → shielded xpubs identical to
  fresh-create) and M.8 (different passphrase → silently different keys). M.8 deliberately pins that
  this is undetectable — BIP39 passphrases can't be validated — so the contract "callers must supply
  the wallet's original passphrase" is now both documented and tested.

src/utils/wallet.ts
pedroferreira1: About the second comment:

Took the hardening despite the low priority — migration now cross-checks the PIN
  against mainKey before writing anything, with a specific error and the original attached as cause
  (M.6 test). Note the passphrase cannot get the same treatment: BIP39 passphrases are unverifiable by
  design (any passphrase yields a valid seed), so that failure mode is documented instead — see M.8.

src/utils/wallet.ts
pedroferreira1: All three gaps closed: root-xpriv import now asserts the scan/spend xpubs at the exact
  documented paths (m/44'/280'/1'/0, /2'/0) and encrypted under the pin; change-level import asserts
  all four fields are absent (pinning the depth-guard's else arm); and a seed import vs root-xpriv
  import of the same wallet now must yield identical shielded keys — guarding exactly the
  compliant/non-compliant derivation asymmetry you mentioned.

__tests__/utils/wallet.test.ts
pedroferreira1: It's possible only via an interrupted write or manual tampering, but worth pinning: M.9 now
  covers it. The re-derive-and-overwrite behavior is actually the desired semantics — derivation is
  deterministic from the seed, so overwriting a partial record heals it to exactly the same keys
  (asserted via decrypt-equality with the original record).

src/new/sendTransaction.ts
pedroferreira1: Added — a SendTransaction.prepareTxData test with a real shielded address in a transparent
  output, asserting the rejection fires through the actual send pipeline (the utils-level
  getAddressType test already existed; this pins the integration point you highlighted).

tuliomir (dismissed):

src/models/address.ts
```diff
@@ -102,11 +128,28 @@ class Address {
...
+    // checks above are individually necessary but not jointly sufficient —
+    // without this, a crafted 25-byte address carrying the shielded version
+    // byte (or a 71-byte one carrying a legacy byte) validates, and the
+    // pubkey extractors would silently clamp `subarray` reads on the short
+    // buffer. Note this runs only when the network is known (not under
+    // `skipNetwork`), since the byte→length mapping is network-defined.
+    const expectedLength =
+      firstByte === this.network.versionBytes.shielded
+        ? SHIELDED_ADDRESS_SIZE_BYTES
+        : LEGACY_ADDRESS_SIZE_BYTES;
```
r4mmer: I think this should be a switch case or have an object mapping version_byte to the correct size

src/models/address.ts
```diff
@@ -138,11 +184,130 @@ class Address {
...
+   * invalid as a whole, so both are checked regardless of which part the
+   * caller wants.
+   *
+   * @throws {AddressError} If address is not shielded or a pubkey is not
+   *   a point on the secp256k1 curve
+   * @return {IShieldedAddressParts}
+   * @memberof Address
+   * @inner
+   */
+  parseShielded(): IShieldedAddressParts {
```
r4mmer, line 236: Should we cache the parsed value?

src/utils/address.ts
```diff
@@ -20,16 +20,37 @@ import { IMultisigData, IStorage, IAddressInfo } from '../types';
...
+ * Throws for shielded addresses — callers expecting an output script type
+ * should not receive shielded addresses directly.
  *
  * @param {string} address
  * @param {Network} network
  *
- * @returns {string} output type of the address (p2pkh or p2sh)
+ * @returns {'p2pkh' | 'p2sh'} output type of the address
  */
 export function getAddressType(address: string, network: Network): 'p2pkh' | 'p2sh' {
```
r4mmer, line 40: Why are we failing shielded addresses on this util?
If this is due to the send transaction facade I think we should have a `getLegacyAddressType` that covers only the old addresses, then keep this valid for all addresses.

src/utils/shieldedAddress.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+  }
+  const network = new Network(networkName);
+
+  const scanHdPub = new HDPublicKey(scanXpubkey);
+  const spendHdPub = new HDPublicKey(spendXpubkey);
+
+  // Standard BIP32 derivation for public keys.
+  // Note: private key derivation (in processing.ts) uses deriveNonCompliantChild
+  // due to a historical bitcore-lib bug. Do not align these — the asymmetry is intentional.
```
r4mmer, line 104: This is great! Makes a lot of sense to keep this bug in the legacy only paths

src/models/address.ts
pedroferreira1: Good call — took the object mapping. The ternary's implicit "anything not shielded is legacy" would have silently assigned 25 bytes to any future address family added to versionBytes; the map states each byte explicitly and throws on an unmapped one instead of defaulting.

src/models/address.ts
pedroferreira1: Considered it, but recommending not yet: there's no hot path (nothing calls it repeatedly on one instance — scanning works from storage records, not Address objects), the class recomputes decode() in every method today so caching just this one is inconsistent, and network is a mutable public field the parse validates against— a cached result could survive a network swap. If profiling ever shows this hot, caching decode() at the class level is the better fix (benefits all methods).

src/utils/address.ts
pedroferreira1: The intent of this util is narrower than its name suggests — it answers "which script does an output to this address use", and I checked: all 7 caller files feed the
  result into output building, so they all want the throw (a 71-byte address must never reach script builders). The general classifier already exists as Address.getType(),
  which returns all three families — a getLegacyAddressType twin would leave the general variant with zero callers. I've clarified the doc to state it returns the output
  script type and to point to Address.getType() for general classification. If you'd still prefer the explicit rename (getOutputScriptType), I'm not opposed — but it's
  exported API used on master (creation.ts), so I'd rather do that rename in its own PR.

src/models/address.ts
```diff
@@ -143,10 +143,22 @@ class Address {
...
+    // but a future address family added to the network without an entry
+    // here must fail loudly rather than silently default to a length.
+    const sizeByVersionByte: Record<number, number> = {
+      [this.network.versionBytes.p2pkh]: LEGACY_ADDRESS_SIZE_BYTES,
+      [this.network.versionBytes.p2sh]: LEGACY_ADDRESS_SIZE_BYTES,
+      [this.network.versionBytes.shielded]: SHIELDED_ADDRESS_SIZE_BYTES,
+    };
+    const expectedLength = sizeByVersionByte[firstByte];
+    if (expectedLength === undefined) {
+      throw new AddressError(
```
tuliomir, line 158: chore(non-blocking): The stricter `versionByte` validation is missing an error handling test

tuliomir (dismissed):

src/models/address.ts
pedroferreira1: Good catch — added a test for it. Worth noting that branch is defensive and unreachable through the public path today: isVersionByteValid only accepts the same three bytes the size map registers, so the lookup can't return undefined via a normal address. The test forces the divergence by stubbing isVersionByteValid to accept an unmapped byte, which both exercises the throw and documents the guard's purpose — protecting against a future version byte added to the network's validity check without a matching size entry.

r4mmer (approved):

tuliomir (approved):

PR #1088, opened 2026-05-14
feat(shielded): data model + storage layer for shielded addresses

pedroferreira1:
## Summary

This layer answers **"where shielded data lives"** — types, stores, indexes, key access. Purely additive: nothing consumes these capabilities yet (5b/5c do), so existing wallets behave exactly as before.

This is the first part of PR5 (5a).

```
master ─ PR1 ─ PR2 ─ PR3 ─ PR4 ─ PR5 ◀ PR6 ─ PR7
```

Base: `shielded/pr-4-address-derivation`.

This is the largest source-code PR — **2200 net additions across 17 files**. Suggested two reviewers (one for storage, one for the data-model ripple).

## Acceptance Criteria

- Shielded key material is reachable through storage: `getScanXPrivKey(pin)` / `getSpendXPrivKey(pin)` decrypt with the PIN; `getScanXPubKey()` / `getSpendXPubKey()` return the public chains; `shieldedCryptoProvider` has a field + setter.
- `MemoryStore.saveAddress` routes by `addressType`: `shielded` addresses index into the new `shieldedAddressIndexes` map; `shielded-spend` records are stored but not indexed (matched later by base58).
- Address-chain methods (`getAddressAtIndex`, `getCurrentAddress`, index getters/setters) accept `IAddressChainOptions` — `{ legacy: false }` selects the shielded chain; omitting it preserves legacy behavior.
- `IWalletData` tracks the shielded chain's loaded / used / current indices independently of the legacy chain.
- `IUtxo` can carry `shielded` / `blindingFactor` / `assetBlindingFactor`; `IUtxoFilterOptions.shielded` filters them.
- No behavior change for existing wallets — every new field and method param is optional.

## Release Notes

- **New Features**
  - Added chain-aware address iteration/indexing with legacy vs shielded filtering.
  - Added ordered transaction history iteration (ascending/descending).
  - Added direct UTXO lookup and improved UTXO selection filtering for transparent vs shielded.
  - Added shielded crypto provider support plus scan/spend key accessors.
- **Bug Fixes**
  - Improved wallet balance calculations by skipping incomplete shielded inputs.
  - Preserves user-provided shielded change addresses instead of converting them.
- **Tests**
  - Expanded coverage for legacy vs shielded behavior, ordering, and filtering.

### DIFF
--- a/__tests__/utils/address.test.ts
+++ b/__tests__/utils/address.test.ts
@@ -208,8 +208,6 @@ test('Get address from pubkey', async () => {
 });
 
 test('deriveShieldedAddressFromStorage returns null when shielded keys unavailable', async () => {
-  // HDPrivateKey comes from the top-level import (added by the pr-4 review
-  // tests); re-importing it here shadowed that declaration.
   const { deriveShieldedAddressFromStorage } = await import('../../src/utils/address');
   const { encryptData } = await import('../../src/utils/crypto');
   const store = new MemoryStore();
--- a/__tests__/utils/address.test.ts
+++ b/__tests__/utils/address.test.ts
@@ -257,20 +257,24 @@ test('deriveShieldedAddressFromStorage returns paired shielded + spend records',
   // Independent oracle: the same inputs through the pure derivation helper.
   const expected = deriveShieldedAddress(scanXpubkey, spendXpubkey, index, networkName);
 
-  // User-facing shielded receive record (71-byte encoded form).
+  // User-facing shielded receive record (71-byte encoded form). Its
+  // ctMappingAddress links to the paired on-chain spend P2PKH.
   expect(result!.shieldedAddress).toEqual({
     base58: expected.base58,
     bip32AddressIndex: index,
     publicKey: expected.scanPubkey,
     addressType: 'shielded',
+    ctMappingAddress: expected.spendAddress,
   });
 
-  // Internal on-chain spend-derived P2PKH record (HASH160(spend_pubkey)).
+  // Internal on-chain spend-derived P2PKH record (HASH160(spend_pubkey)). Its
+  // ctMappingAddress links back to the paired user-facing shielded address.
   expect(result!.spendAddress).toEqual({
     base58: expected.spendAddress,
     bip32AddressIndex: index,
     publicKey: expected.spendPubkey,
     addressType: 'shielded-spend',
+    ctMappingAddress: expected.base58,
   });
 
   // Both records share the BIP32 index; the spend address is the on-chain
--- a/__tests__/utils/address.test.ts
+++ b/__tests__/utils/address.test.ts
@@ -228,3 +228,54 @@ test('deriveShieldedAddressFromStorage returns null when shielded keys unavailab
   const result = await deriveShieldedAddressFromStorage(0, storage);
   expect(result).toBeNull();
 });
+
+test('deriveShieldedAddressFromStorage returns paired shielded + spend records', async () => {
+  const { deriveShieldedAddressFromStorage } = await import('../../src/utils/address');
+  const { deriveShieldedAddress } = await import('../../src/utils/shieldedAddress');
+  const { encryptData } = await import('../../src/utils/crypto');
+  const store = new MemoryStore();
+  const storage = new Storage(store);
+  storage.config.setNetwork('testnet');
+  const networkName = storage.config.getNetwork().name;
+
+  // Independent scan and spend public chains (the wallet derives both from
+  // hardened accounts; for this test any two xpubs exercise the same path).
+  const xpriv = new HDPrivateKey();
+  const scanXpubkey = new HDPrivateKey().xpubkey;
+  const spendXpubkey = new HDPrivateKey().xpubkey;
+  await store.saveAccessData({
+    xpubkey: xpriv.xpubkey,
+    mainKey: encryptData(xpriv.xprivkey, '123'),
+    walletType: 'p2pkh' as const,
+    walletFlags: 0,
+    scanXpubkey,
+    spendXpubkey,
+  });
+
+  const index = 5;
+  const result = await deriveShieldedAddressFromStorage(index, storage);
+  expect(result).not.toBeNull();
+
+  // Independent oracle: the same inputs through the pure derivation helper.
+  const expected = deriveShieldedAddress(scanXpubkey, spendXpubkey, index, networkName);
+
+  // User-facing shielded receive record (71-byte encoded form).
+  expect(result!.shieldedAddress).toEqual({
+    base58: expected.base58,
+    bip32AddressIndex: index,
+    publicKey: expected.scanPubkey,
+    addressType: 'shielded',
+  });
+
+  // Internal on-chain spend-derived P2PKH record (HASH160(spend_pubkey)).
+  expect(result!.spendAddress).toEqual({
+    base58: expected.spendAddress,
+    bip32AddressIndex: index,
+    publicKey: expected.spendPubkey,
+    addressType: 'shielded-spend',
+  });
+
+  // Both records share the BIP32 index; the spend address is the on-chain
+  // P2PKH while the shielded address is the user-facing encoded form.
+  expect(result!.shieldedAddress.base58).not.toEqual(result!.spendAddress.base58);
+});
--- a/__tests__/utils/address.test.ts
+++ b/__tests__/utils/address.test.ts
@@ -206,3 +206,25 @@ test('Get address from pubkey', async () => {
   expect(address.base58).toBe(base58);
   expect(address.validateAddress()).toBeTruthy();
 });
+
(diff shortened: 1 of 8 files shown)


pedroferreira1:
@coderabbitai do a full re-review

pedroferreira1:
@coderabbitai all comments were addressed. Can you re-review?

__tests__/utils/address.test.ts
```diff
@@ -206,3 +206,76 @@ test('Get address from pubkey', async () => {
   expect(address.base58).toBe(base58);
   expect(address.validateAddress()).toBeTruthy();
 });
+
+test('deriveShieldedAddressFromStorage returns null when shielded keys unavailable', async () => {
+  // HDPrivateKey comes from the top-level import (added by the pr-4 review
+  // tests); re-importing it here shadowed that declaration.
```
tuliomir: chore(non-blocking): Remove this comment? Seems obsolete.

src/storage/storage.ts
```diff
@@ -903,6 +946,50 @@ export class Storage implements IStorage {
...
+      throw new Error('Scan private key is not present on this wallet.');
+    }
+    return decryptData(accessData.scanMainKey, pinCode);
+  }
+
+  /**
+   * Get the spend chain xprivkey for shielded UTXO signing.
+   * Uses account 2' (m/44'/280'/2'/0).
+   */
+  async getSpendXPrivKey(pinCode: string): Promise<string> {
```
tuliomir, line 965: chore: Most methods here in the `storage` file are untested.

I know they will receive coverage when their respective callers are built, but it would be nice to have at least a smoke unit test for them here where they're being built. 

This is especially important for the error handling cases that will not be the focus of the caller methods - they will actually prefer to mock these methods entirely to test their own error handling branches.

src/utils/address.ts
```diff
@@ -173,3 +174,47 @@ export function getAddressFromPubkey(pubkey: string, network: Network): Address
...
+  // Stored separately so the wallet can match incoming transactions by decoded.address,
+  // since on-chain scripts reference this P2PKH, not the shielded address.
+  const spendAddress: IAddressInfo = {
+    base58: info.spendAddress,
+    bip32AddressIndex: index,
+    publicKey: info.spendPubkey,
+    addressType: 'shielded-spend',
+  };
+
+  return { shieldedAddress, spendAddress };
```
tuliomir, line 223: question(concept): Should we return both the scan and spend addresses in the same method call? Any use case needs them both at the same time?

This is just conceptual, I believe the derivation cost to be negligible. Just trying to avoid a sensible information returned at an unnecessary moment.

__tests__/utils/address.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1088/commits/be24e0cea6f574fa84202350efe8bcbc16b30698

src/storage/storage.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1088/commits/be24e0cea6f574fa84202350efe8bcbc16b30698

src/utils/address.ts
pedroferreira1: ▎ Good question. I checked all three callers — loadAddresses, HathorWallet.getAddressAtIndex, and
  ▎ the WS sync/stream queue — and each saves both the shielded receive address and its spend-derived
  ▎ P2PKH at the same index in the same step. They're a derivation pair (same index, same scan/spend
  ▎ xpubs) that's always persisted together, so there's no caller that wants only one. Returning both
  ▎ keeps it a single atomic derivation; splitting would mean two calls (or a duplicated xpub fetch +
  ▎ derive) for no gain. And there's no over-exposure concern — these are the wallet's own addresses
  ▎ returned within the storage layer, not surfaced externally.

tuliomir (dismissed):

src/types.ts
```diff
@@ -116,6 +117,14 @@ export interface IAddressInfo {
   addressType?: AddressType | 'shielded-spend';
 }
 
+/**
+ * Options for address methods that can operate on either the legacy or shielded address chain.
+ * Defaults to legacy (true) for backward compatibility.
+ */
+export interface IAddressChainOptions {
+  legacy?: boolean; // default: true
+}
```
r4mmer, line 135: Instead of being an argument to pass all storage calls, why not a flag on the storage itself?

r4mmer (approved):

src/storage/memory_store.ts
```diff
@@ -207,13 +220,42 @@ export class MemoryStore implements IStore {
...
+        // is the internal P2PKH the receive pipeline uses to match incoming
+        // shielded outputs — surfacing it under the legacy chain would leak
+        // an internal address).
+        if (addrInfo.addressType === 'shielded' || addrInfo.addressType === 'shielded-spend') {
+          continue;
+        }
+      } else if (addrInfo.addressType !== 'shielded') {
+        // Shielded chain: only the user-facing 'shielded' receive entry.
+        continue;
+      }
```
r4mmer: Instead of iterating over all addresses and ignoring the ones we do not want, why not use the `addressIndexes` or `shieldedAddressIndexes` to know which addresses we have and `return this.addresses.get(addr)`?

src/storage/memory_store.ts
```diff
@@ -456,21 +552,38 @@ export class MemoryStore implements IStore {
 
     this.history.set(tx.tx_id, tx);
 
-    let maxIndex = this.walletData.lastUsedAddressIndex;
+    let legacyMaxIndex = this.walletData.lastUsedAddressIndex;
+    let shieldedMaxIndex = this.walletData.shieldedLastUsedAddressIndex;
     for (const el of [...tx.inputs, ...tx.outputs]) {
```
r4mmer: Here we iterate over inputs and outputs, but is the `tx.shielded_outputs` confrmed to be added to the outputs array?

src/storage/memory_store.ts
```diff
@@ -656,6 +769,31 @@ export class MemoryStore implements IStore {
...
+      if (info?.addressType === 'shielded') {
+        for (const candidate of this.addresses.values()) {
+          if (
+            candidate.addressType === 'shielded-spend' &&
+            candidate.bip32AddressIndex === info.bip32AddressIndex
+          ) {
+            effectiveFilterAddress = candidate.base58;
+            break;
+          }
+        }
```
r4mmer: There is no `ct_address => spend_address` tracking? we need to iterate over all addresses?

src/storage/storage.ts
r4mmer, line 1033: This will not unsubscribe from all addresses, we need a second loop to unsubcribe to the shielded spend addresses

r4mmer (changes requested):
I know that the wallet-service is out of scope, but can you create an issue to track the interface drift since the `src/wallet/walletServiceStorageProxy.ts` does not implement the new storage interface?

Also there are some comments by coderabbit that need addressing, some of these comments seem strange but we should investigate, specially the ones about src/utils/transaction.ts

src/types.ts
pedroferreira1: As we discussed, we will have wallets loading ct and legacy addresses, so we will keep as a parameter for now

src/storage/memory_store.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1088/commits/bfc4f7b5f51d6baaa61782aca23f42d9f179a84d

src/storage/storage.ts
pedroferreira1: The address subscription after loadAddress is done in PR 5c but this is actually a bug there as well. I will fix there because in this PR we don't have any ct address subscribed

src/storage/memory_store.ts
pedroferreira1: Added a mapping as we discussed: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1088/changes/4703764b88ecb43ee660104334e9b6fab50430ee

src/storage/memory_store.ts
pedroferreira1: I did a huge refactor yesterday that separated the outputs from the shielded ones. I should've updated this PR as well, this is a bug. Fixed: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1088/commits/c8fb756e5eda5619b2f1c2af36f59837f0b2fa6e

src/storage/memory_store.ts
```diff
@@ -723,6 +862,10 @@ export class MemoryStore implements IStore {
     this.utxos.set(`${utxo.txId}:${utxo.index}`, utxo);
   }
 
+  async getUtxo(utxoId: IUtxoId): Promise<IUtxo | null> {
```
tuliomir, line 865: chore: Missing docstring

tuliomir (approved):

r4mmer (approved):

PR #1091, opened 2026-05-15
fix(nano): resolve current blueprint id for upgraded contracts

pedroferreira1:
## Summary

`getBlueprintId(ncId, wallet)` returned the *creation* blueprint id of a nano contract instead of its *current* one. For contracts that had been upgraded in-place via `syscall.change_blueprint`, every method call was validated against the original blueprint's signature, so any method whose argument types had changed across the upgrade would be rejected client-side before the request ever reached the network.

Switches the function to query the state endpoint first (which reports the current blueprint id) and falls back to `wallet.getFullTxById` only when the state endpoint cannot serve the contract — preserving support for unconfirmed/unindexed contracts, which by definition cannot have been upgraded yet.

## Acceptance Criteria

- Correctly get current blueprint id for any nano contract

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end integration tests for upgradable nano contracts and two new upgrade test blueprints (v1→v2); test setup now publishes those blueprints for use.
  * Added unit tests for blueprint-ID resolution behavior and updated an on-chain blueprint list assertion to be resilient to growth.
* **Bug Fixes**
  * Prefer retrieving a contract’s current blueprint ID from state first, falling back to transaction data with improved error reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-lib/pull/1091?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/__tests__/integration/nanocontracts/upgrade_contract.test.ts
+++ b/__tests__/integration/nanocontracts/upgrade_contract.test.ts
@@ -1,0 +1,121 @@
+import { isEmpty } from 'lodash';
+import { GenesisWalletHelper } from '../helpers/genesis-wallet.helper';
+import { generateWalletHelper, waitForTxReceived, waitTxConfirmed } from '../helpers/wallet.helper';
+import { NANO_CONTRACTS_INITIALIZE_METHOD } from '../../../src/constants';
+import ncApi from '../../../src/api/nano';
+import { bufferToHex } from '../../../src/utils/buffer';
+import helpersUtils from '../../../src/utils/helpers';
+import { getBlueprintId } from '../../../src/nano_contracts/utils';
+
+/**
+ * End-to-end test for the wallet-lib's blueprint-id resolution on upgraded
+ * nano contracts.
+ *
+ * Setup:
+ * - UpgradeTestV1 declares `set_value(addr: Address)` and `upgrade_to(BlueprintId)`.
+ * - UpgradeTestV2 declares `set_value(addr: CallerId)` (storage-compatible).
+ *
+ * Flow:
+ * 1. Create a contract from V1; call `set_value` with a base58 Address (matches V1).
+ * 2. Call `upgrade_to(V2)`; assert the state endpoint now reports V2 as the
+ *    blueprint id, while the creation tx still carries V1 in `nc_blueprint_id`.
+ * 3. Call `set_value` again with a 64-char hex (a ContractId-form CallerId).
+ *    This succeeds only when the wallet-lib resolves the *current* blueprint
+ *    via the state endpoint. Without that fix, it parses against V1's `Address`
+ *    schema and rejects the value with a Zod regex error.
+ */
+describe('Upgradable nano contract', () => {
+  /** @type HathorWallet */
+  let hWallet;
+  let address0;
+
+  beforeAll(async () => {
+    hWallet = await generateWalletHelper();
+    address0 = await hWallet.getAddressAtIndex(0);
+    await GenesisWalletHelper.injectFunds(hWallet, address0, 1000n);
+  });
+
+  afterAll(async () => {
+    await hWallet.stop();
+    await GenesisWalletHelper.clearListeners();
+  });
+
+  const checkTxValid = async (wallet, tx) => {
+    const txId = tx.hash;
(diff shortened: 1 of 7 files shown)


raul-oliveira (dismissed):

src/nano_contracts/utils.ts
```diff
@@ -359,6 +364,13 @@ export const getBlueprintId = async (ncId: string, wallet: HathorWallet): Promis
     throw new Error('Nano contract ID is not defined');
   }
 
+  const [stateError, stateResponse] = await getResultHelper(() =>
+    ncApi.getNanoContractState(ncId, [], [], [])
+  );
+  if (!stateError && stateResponse?.blueprint_id) {
```
tuliomir, line 374: suggestion(non-blocking): Add a small comment "header" to each of those return blocks. Nano contracts are complex, and it helps having small breadcrumbs throughout the code.


```suggestion
  // Early exit: presence of a blueprint_id means the contract has just been initialized
  if (!stateError && stateResponse?.blueprint_id) {
```

__tests__/integration/nanocontracts/upgrade_contract.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    address0 = await hWallet.getAddressAtIndex(0);
+    await GenesisWalletHelper.injectFunds(hWallet, address0, 1000n);
+  });
+
+  afterAll(async () => {
+    await hWallet.stop();
+    await GenesisWalletHelper.clearListeners();
+  });
+
+  const checkTxValid = async (wallet, tx) => {
```
tuliomir, line 43: suggestion(non-blocking): I think we have such method already on the test utils, and it could be reused here.

__tests__/integration/nanocontracts/upgrade_contract.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    // request only succeeds when the wallet-lib uses V2's signature for
+    // arg validation — a regression of the fix would throw a Zod regex
+    // error here.
+    const callerIdHex = '0000421da6413f181b86c3fcf731976ad6f7ec0983598000a6658f35a212e3e1';
+    const txSetV2 = await hWallet.createAndSendNanoContractTransaction('set_value', address0, {
+      ncId,
+      args: [callerIdHex],
+    });
+    await checkTxValid(hWallet, txSetV2);
+  });
```
tuliomir, line 117: question(non-blocking): Unrelated to this line, would it make sense to add edge-cases to this test suite, or they are already being handled enough by fullnode tests?

Examples: trying to upgrade an unconfirmed NC , trying to call state-checking methods while the upgrade isn't confirmed, trying to upgrade an invalid NC, validating value passes on from pristine or from dirty base-contracts ( set_value called at least once )...

tuliomir (dismissed):

__tests__/integration/nanocontracts/upgrade_contract.test.ts
pedroferreira1: These are mostly fullnode tests, not wallet tests in my opinion. I added the `getBlueprintId` fallback path (state endpoint can't serve the contract → fall back to `getFullTxById`). This one was a valid concern to be added.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1091/changes/65b638961c8d5e0e74588547f8c2019f42550a0a

__tests__/integration/nanocontracts/upgrade_contract.test.ts
pedroferreira1: The util test does not exist, this is following the same standard we have for the other blueprint integration tests files. Maybe this util was created in one of your branches that are still under review?

In that case, you could use it here after the rebase. What do you think?

src/nano_contracts/utils.ts
pedroferreira1: Much better, thanks: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1091/changes/0b878f9f376e8c2f6c6e6f076e1fdb4167ef30de

raul-oliveira (approved):

tuliomir (approved):

PR #1093, opened 2026-05-18
feat: web3 auth single key

raul-oliveira:
# Choose a pull request template

- [feature PR](?template=feature.md&quick_pull=1)
- [Release process](https://github.com/HathorNetwork/hathor-wallet-lib/compare/release...master?template=release.md&quick_pull=1)

## Summary by CodeRabbit

* **New Features**
  * Single-key wallet support (raw secp256k1) with single-address mode and optional external signing.

* **Bug Fixes / Behavior**
  * Stronger validation and guardrails for address derivation, indexing, sync-mode selection, and signing routing between single-key and HD wallets.

* **Documentation**
  * README updated with single-key usage, constraints, and external signing guidance.

* **Tests**
  * Comprehensive test coverage for single-key flows, guards, signing, pin changes, and regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-lib/pull/1093?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/new/types.ts
```diff
@@ -51,6 +51,36 @@ export interface HathorWalletConstructorParams {
   xpriv?: string;
   /** Extended public key (xpub) for read-only wallet */
   xpub?: string;
+  /**
+   * Raw secp256k1 private key (hex string) for single-key wallets — e.g.
+   * Web3Auth onboarding where no BIP32 chain code is available.
+   *
+   * When set:
+   * - `publicKey` and `preCalculatedAddresses[0]` are REQUIRED
```
r4mmer: both are required?

src/new/types.ts
```diff
@@ -51,6 +51,36 @@ export interface HathorWalletConstructorParams {
...
+   * - `seed` / `xpriv` / `xpub` MUST NOT be provided
+   * - the wallet has exactly one address (index 0)
+   * - transactions MUST be signed via `setExternalTxSigningMethod`
```
r4mmer: MUST be signed?
Why not use the privateKey when signing and for web3auth specifically the external signing method is a MUST?

src/new/wallet.ts
```diff
@@ -777,6 +801,12 @@ class HathorWallet extends EventEmitter {
    * @memberof HathorWallet
    */
   async hasTxOutsideFirstAddress(): Promise<boolean> {
+    // Wallets with SINGLE_ADDRESS policy have exactly one address (index 0), so
+    // by definition no transactions can exist outside of it.
+    if ((await this.storage?.getScanningPolicy?.()) === SCANNING_POLICY.SINGLE_ADDRESS) {
+      return false;
+    }
```
r4mmer: This method is exactly to protect starting a single wallet in a wallet with transactions outside the first address.

For instance, we use the seed in a gap-limit wallet then start in a single address wallet, this method is meant to protect against this move, so just returning false in single address wallets will simply remove this protection.

You need to create a separate guard for single wallets where we do not have the xpub/xpriv (to avoid calling `getAddressAtIndex` and trying to derive outside the first address)

src/new/wallet.ts
```diff
@@ -805,6 +835,11 @@ class HathorWallet extends EventEmitter {
    * @inner
    */
   async getAddressAtIndex(index: number): Promise<string> {
+    const accessData = await this.storage.getAccessData();
+    if (accessData?.singleKeyMode && index !== 0) {
+      throw new AddressError('Single-key wallets only support address index 0.');
+    }
```
r4mmer: You need to create a separate guard for single wallets where we do not have the xpub/xpriv.
A single address wallet with xpub still has the "capabilities" to derive other addresses (it is actually a requiremente for some protections)

src/storage/storage.ts
```diff
@@ -177,6 +177,15 @@ export class Storage implements IStorage {
...
       return this.txSignFunc(tx, this, pinCode);
     }
+
+    const accessData = await this.getAccessData();
+    if (accessData?.singleKeyMode) {
+      throw new Error(
+        'Single-key wallets cannot sign transactions without an external signer. ' +
+          'Register one via wallet.setExternalTxSigningMethod() before signing.'
+      );
+    }
```
r4mmer: Single key wallet MUST sign with external signers, just the web3auth that cannot do this.
But xpub enabled single address wallets MUST do this (and probably privkey enabled wallets should too)

src/storage/storage.ts
```diff
@@ -903,6 +918,57 @@ export class Storage implements IStorage {
...
+   * - the pin is incorrect (decryption fails)
+   *
+   * @param pinCode Pin used to decrypt the key at rest.
+   * @returns The raw 32-byte secp256k1 private key as a hex string.
+   */
+  async getSingleKeyPrivateKey(pinCode: string): Promise<string> {
+    const accessData = await this._getValidAccessData();
+    if (!accessData.singleKeyMode || !accessData.singleKeyPrivateKey) {
+      throw new Error('Single-key private key is not present on this wallet.');
+    }
```
r4mmer: What about single key wallets that are started with xpriv?

src/utils/storage.ts
```diff
@@ -62,17 +62,26 @@ export function getHistorySyncMethod(mode: HistorySyncMode): HistorySyncFunction
...
+    return [];
   }
-  if (walletType === WalletType.MULTISIG) {
-    return [HistorySyncMode.POLLING_HTTP_API];
+
+  // Single-key wallets cannot use xpub-based stream sync
+  const accessData = await storage.getAccessData();
+  if (accessData?.singleKeyMode) {
+    return modes.filter(m => m !== HistorySyncMode.XPUB_STREAM_WS);
   }
```
r4mmer: We can have a single address wallet with xpub

__tests__/new/singleKeyWallet.test.ts
```diff
@@ -0,0 +1,725 @@
...
+    // No signer registered — falls back to the local raw-key path. Per
+    // r4mmer's review: external signer is the right escape hatch for
+    // Web3Auth (key not exposed), not a hard requirement for every
+    // single-key wallet.
+    const tx = new Transaction([], []);
+    const result = await storage.getTxSignatures(tx, PIN);
+    expect(result.inputSignatures).toEqual([]);
+    expect(result.ncCallerSignature).toBeNull();
+  });
+});
```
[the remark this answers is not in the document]
raul-oliveira: Done. Renamed the section comment and the `describe` block to reflect the actual behaviour (`__tests__/new/singleKeyWallet.test.ts:601-606`):

```diff
-// 11. getTxSignatures rejects single-key wallets without external signer
+// 11. getTxSignatures local-signs single-key wallets without external signer

-describe('Storage.getTxSignatures — single-key wallets', () => {
+describe('Storage.getTxSignatures — single-key wallets (local-sign fallback)', () => {
```

src/utils/wallet.ts
```diff
@@ -702,6 +702,31 @@ const wallet = {
...
+    { pin }: { pin: string }
+  ): IWalletAccessData {
+    const encryptedPrivateKey = encryptData(privateKeyHex, pin);
+    return {
+      walletType: WalletType.P2PKH,
+      walletFlags: 0,
+      singleKeyMode: true,
+      singleKeyPublicKey: publicKeyHex,
+      singleKeyPrivateKey: encryptedPrivateKey,
+    };
```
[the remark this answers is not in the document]
raul-oliveira: Addressed. `generateAccessDataFromPrivateKey` now derives the public key from the private key via `new bitcore.PrivateKey(privateKeyHex).toPublicKey()` and throws when the supplied `publicKeyHex` does not match (`src/utils/wallet.ts:730-733`). We persist the derived value (`derivedPubKeyHex`) instead of the caller-supplied one so the stored access data is always internally consistent. Covered by `__tests__/new/singleKeyWallet.test.ts:758-773`.

src/new/types.ts
raul-oliveira: Good call — they didn't have to be required. Both are derivable from `privateKey` + the connection's network, and forcing the caller to pre-compute them was defensive without a justification.

Relaxed in two places:

- **Constructor** (`src/new/wallet.ts:343-356`): when `privateKey` is provided and `publicKey` is not, we derive it via `new bitcore.PrivateKey(privateKey).toPublicKey().toString('hex')` before storing it on the instance.
- **`start()`** (`src/new/wallet.ts:1712-1725`): when `preCalculatedAddresses` is absent for a single-key wallet, we derive the single address from `publicKey + network` via `getAddressFromPubkey` and persist it directly. We must populate storage upfront because, unlike HD wallets, there's no xpub to fall back on.

Validation is preserved when the caller passes both: `generateAccessDataFromPrivateKey` still enforces the pubkey↔privkey match (C1) — so a mismatched explicit `publicKey` still throws. When the caller passes a mismatched `preCalculatedAddresses[0]`, we honour it (caller responsibility), matching the existing HD-wallet contract.

Tests for the three new optionality combinations: `__tests__/new/singleKeyWallet.test.ts:134-164` (constructor) and `__tests__/new/singleKeyWallet.test.ts:259-316` (`start()` end-to-end).

src/new/types.ts
raul-oliveira: Agreed — external signing is the escape hatch for Web3Auth, not a hard requirement. Reworked to sign locally by default: `buildSigningKeyResolver` (`src/utils/transaction.ts:175-193`) handles single-key and HD in one place, and `Storage.getTxSignatures` (`src/storage/storage.ts:183-189`) is now just `txSignFunc ?? getSignatureForTx`. JSDoc on `privateKey` updated (`src/new/types.ts:54-63`): "signed locally by default; register an external signer to delegate (required when the key is held in a trusted environment outside wallet-lib)."

src/new/wallet.ts
raul-oliveira: Fixed — the guard is now capability-based, not policy-based. `hasTxOutsideFirstAddress` (`src/new/wallet.ts:804-840`) only short-circuits when `canDeriveAddresses()` is false. HD wallets in `SINGLE_ADDRESS` policy still hold an xpub, still derive 1..GAP_LIMIT, and the migration-history protection you described works exactly as before. Only raw single-key wallets (no chain code) short-circuit to false. Regression test: `__tests__/new/singleKeyWallet.test.ts:714-750`.

src/new/wallet.ts
raul-oliveira: Done — the guard is now in `src/new/wallet.ts:850-857` and rejects derivation only when the wallet has no derivation capability:

```ts
if (index !== 0 && !(await this.storage.canDeriveAddresses())) {
  throw new AddressError(
    'This wallet has no key material to derive addresses beyond index 0.'
  );
}
```

HD wallets with xpub pass even in SINGLE_ADDRESS policy — which they need for `hasTxOutsideFirstAddress` (C4) and the other safety checks you mentioned. Only raw single-key / Web3Auth wallets are blocked.

The `canDeriveAddresses` flag is set explicitly by every access-data generator: `true` in `generateAccessDataFromXpub` (`src/utils/wallet.ts:539`), `generateAccessDataFromXpriv` (`src/utils/wallet.ts:623`), and `generateAccessDataFromSeed` (`src/utils/wallet.ts:712`); `false` in `generateAccessDataFromPrivateKey` (`src/utils/wallet.ts:742`). Old access data without the flag falls back to `!!accessData.xpubkey` inside `Storage.canDeriveAddresses` (`src/storage/storage.ts:910-913`), so existing wallets keep their derivation capability without re-encryption.

src/storage/storage.ts
raul-oliveira: Agreed. Storage no longer branches on wallet type for signing: `getTxSignatures` (`src/storage/storage.ts:183-189`) is just `txSignFunc ?? getSignatureForTx`. HD/single-key routing lives in `buildSigningKeyResolver` (see C3). External signer remains available via `setExternalTxSigningMethod` and is the natural choice for Web3Auth (and any future case where the key lives outside wallet-lib). The "throw when single-key without signer" rule was removed; the corresponding test was inverted to assert local signing (`__tests__/new/singleKeyWallet.test.ts:601-617`).

src/storage/storage.ts
raul-oliveira: Good question — there's no such thing in the current design, and the contract is now explicit. A "single-key wallet" specifically means a wallet backed by a **raw secp256k1 key** (no BIP32 chain code). Wallets started from a seed or xpriv that happen to run under `SINGLE_ADDRESS` scanning policy are **still HD wallets**; they keep `mainKey`/`xpubkey`/`canDeriveAddresses: true` and continue to use `getMainXPrivKey` for signing.

`singleKeyMode` is only set by `generateAccessDataFromPrivateKey` (`src/utils/wallet.ts:725-744`); the seed/xpriv/xpub generators never set it.

`getSingleKeyPrivateKey` (`src/storage/storage.ts:949-974`) now has JSDoc making the boundary explicit and an error message pointing at the correct alternative:

```
'No raw single-key private key on this wallet. Use getMainXPrivKey for HD wallets.'
```

An HD caller that lands on this method by mistake gets a clear error and a redirection.

src/utils/storage.ts
raul-oliveira: Right — fixed in `src/utils/storage.ts:78-91`. The `XPUB_STREAM_WS` filter now checks `accessData.canDeriveAddresses ?? !!accessData.xpubkey`. HD wallets in `SINGLE_ADDRESS` policy keep `XPUB_STREAM_WS`; only wallets without derivation capability (raw single-key) are filtered. The `?? !!xpubkey` fallback covers pre-flag access data without re-encryption. I read `accessData` directly (not via the `Storage.canDeriveAddresses()` helper) because some tests mock `getWalletType` without persisting access data and this utility must not throw on partial state. Regression: `__tests__/new/singleKeyWallet.test.ts:734-738`.

src/types.ts
```diff
@@ -656,11 +670,25 @@ export interface IStorage {
   getMainXPrivKey(pinCode: string): Promise<string>;
   getAcctPathXPrivKey(pinCode: string): Promise<string>;
   getAuthPrivKey(pinCode: string): Promise<string>;
+  /**
+   * Resolve the private key for a given address index.
+   *
+   * Resolve the raw secp256k1 private key for a given address index.
```
r4mmer: duplicated explanation

src/new/wallet.ts
```diff
@@ -1657,11 +1709,29 @@ class HathorWallet extends EventEmitter {
...
+    } else if (this.preCalculatedAddresses) {
       for (const [index, addr] of this.preCalculatedAddresses.entries()) {
+        // Attach the pubkey for single-key wallets so consumers don't fall
+        // through to the xpub-based derivation path (which is absent here).
+        const pubkey = this.privateKey && index === 0 ? this.publicKey : undefined;
         await this.storage.saveAddress({
           base58: addr,
           bip32AddressIndex: index,
+          ...(pubkey ? { publicKey: pubkey } : {}),
         });
```
r4mmer, line 1722: Just an idea, if we are starting with the privateKey, maybe just save the first address and ignore the rest.

r4mmer (commented):
There are some issues flagged by coderabbit that are interesting, the checkPin/checkPassword/changePin methods will unconditionally fail on the new mode and they are required for the wallets to work.

src/types.ts
raul-oliveira: Fixed. Removed the duplicated sentence in `src/types.ts:673-682`; the JSDoc now leads with the more specific "raw secp256k1" wording and drops the redundant intro line.

src/new/wallet.ts
raul-oliveira: Decided to keep master's `preCalculatedAddresses` contract intact — it has always been treated as the caller's source of truth (every integration helper passes `walletData.addresses` straight through, master never filters). Silently dropping entries only for single-key wallets would split that contract.

The change in `src/new/wallet.ts:1712-1734` is purely additive: the iteration loop is unchanged from master (with the cached `publicKey` attached at index 0 so pubkey lookups don't fall through to the absent xpub); a new fallback derives the single address from `publicKey + network` when `preCalculatedAddresses` is omitted. Foreign-address risk at indexes > 0 stays the caller's responsibility, same as for HD wallets.

raul-oliveira:
> There are some issues flagged by coderabbit that are interesting, the checkPin/checkPassword/changePin methods will unconditionally fail on the new mode and they are required for the wallets to work.

@r4mmer — `checkPin` was the load-bearing bug. Fixed in `src/storage/storage.ts:1080-1090`: it now falls back to `singleKeyPrivateKey` when `mainKey` is absent. New test `Storage.checkPin — single-key wallets` in `__tests__/new/singleKeyWallet.test.ts` covers correct-pin and wrong-pin cases.

Verified the other two against current code:

- **`changePin`** already works — `walletUtils.changeEncryptionPin` (`src/utils/wallet.ts:755-789`) has a dedicated `singleKeyPrivateKey` branch, covered by the existing `changeEncryptionPin — single-key wallets` test.
- **`checkPassword`** validates `accessData.words`; single-key wallets have no seed, so throwing is semantically correct. `checkPin` is the right gate across both shapes.

Happy to add a clearer `checkPassword` error message if you think it's worth it.

src/utils/transaction.ts
```diff
@@ -163,18 +163,51 @@ const transaction = {
...
+  async buildSigningKeyResolver(
+    storage: IStorage,
+    pinCode: string
+  ): Promise<(addressIndex: number) => { privateKey: PrivateKey; pubkeyDER: Buffer }> {
+    const accessData = await storage.getAccessData();
+    if (accessData?.singleKeyMode) {
+      const rawPrivHex = await storage.getSingleKeyPrivateKey(pinCode);
+      const privateKey = new PrivateKey(rawPrivHex);
+      const pubkeyDER = privateKey.toPublicKey().toDER();
+      return () => ({ privateKey, pubkeyDER });
```
r4mmer, line 184: I'm thinking about this, should we have a guard for only index 0?
So we don't sign something wrongly

src/new/wallet.ts
```diff
@@ -1659,11 +1711,27 @@ class HathorWallet extends EventEmitter {
 
     if (this.preCalculatedAddresses) {
       for (const [index, addr] of this.preCalculatedAddresses.entries()) {
```
r4mmer, line 1713: When starting from a privateKey why not just get the first address and ignore the rest?

__tests__/integration/fullnode-specific/single-key-wallet.test.ts
```diff
@@ -0,0 +1,155 @@
...
+// Mirrors the derivation used by `generateAccessDataFromSeed`:
+// m/44'/280'/0'/0 (change path), then bitcore `.deriveChild(0)` for addr 0.
+const SEED =
+  'drink favorite jewel cage jar grief endorse judge kitten ten hurry where ' +
+  'project alley drill illness claim month era simple paddle crater menu problem';
+
+const network = new Network('testnet');
+const rootXpriv = new Mnemonic(SEED).toHDPrivateKey('', network.bitcoreNetwork);
+const changeXpriv = rootXpriv.deriveNonCompliantChild(P2PKH_ACCT_PATH).deriveNonCompliantChild(0);
+const addr0HDKey = changeXpriv.deriveChild(0);
```
r4mmer, line 49: Always use `deriveNonCompliantChild` this is a bug in bitcore-lib that the derivation algoritm had.
We are locked on the non-compliant derivation for now (for private key derivation).

src/new/wallet.ts
```diff
@@ -1747,19 +1825,28 @@ class HathorWallet extends EventEmitter {
...
+      if (addressIndex !== 0) {
+        throw new AddressError('Single-key wallets only support address index 0.');
+      }
+      const rawPrivHex = await this.storage.getSingleKeyPrivateKey(pinCode);
+      return new bitcore.PrivateKey(rawPrivHex);
+    }
 
-    return addressHDPrivKey;
+    const mainXPrivKey = await this.storage.getMainXPrivKey(pinCode);
+    return new bitcore.HDPrivateKey(mainXPrivKey).derive(addressIndex);
```
r4mmer, line 1849: Use `deriveNonCompliantChild`

src/new/wallet.ts
```diff
@@ -777,6 +802,24 @@ class HathorWallet extends EventEmitter {
    * @memberof HathorWallet
    */
   async hasTxOutsideFirstAddress(): Promise<boolean> {
+    // Wallets without derivation capability (raw single-key / Web3Auth) only
+    // have address index 0 by construction — they could never have received
+    // funds at any other index, so the check is trivially false. This guard
+    // also avoids attempting to derive addresses we cannot derive.
+    //
+    // HD wallets in SINGLE_ADDRESS policy still hold an xpub and CAN derive
+    // higher indexes; running this check is exactly the protection r4mmer
```
r4mmer, line 811: Oh hi!

src/storage/storage.ts
```diff
@@ -903,6 +945,65 @@ export class Storage implements IStorage {
...
+   */
+  async getAddressPrivKeyForIndex(
+    pinCode: string,
+    addressIndex: number
+  ): Promise<bitcore.PrivateKey> {
+    const accessData = await this._getValidAccessData();
+
+    if (accessData.singleKeyMode) {
+      if (addressIndex !== 0) {
+        throw new AddressError('Single-key wallets only support address index 0.');
```
r4mmer, line 997: Maybe add the `addressIndex` to the error to make it easier to debug

src/new/wallet.ts
```diff
@@ -805,6 +848,14 @@ class HathorWallet extends EventEmitter {
    * @inner
    */
   async getAddressAtIndex(index: number): Promise<string> {
+    // Only reject when the wallet has no material to derive other indexes.
+    // HD wallets running SINGLE_ADDRESS policy still hold an xpub and must
+    // be allowed to derive (it's a requirement of `hasTxOutsideFirstAddress`
+    // and other safety checks).
+    if (index !== 0 && !(await this.storage.canDeriveAddresses())) {
+      throw new AddressError('This wallet has no key material to derive addresses beyond index 0.');
```
r4mmer, line 856: Maybe add the `index` to the error to make it easier to debug

src/new/wallet.ts
```diff
@@ -1747,19 +1825,28 @@ class HathorWallet extends EventEmitter {
...
-    const mainXPrivKey = await this.storage.getMainXPrivKey(pinCode);
-    const addressHDPrivKey = new bitcore.HDPrivateKey(mainXPrivKey).derive(addressIndex);
+  async getAddressPrivKey(
+    pinCode: string,
+    addressIndex: number
+  ): Promise<bitcore.HDPrivateKey | bitcore.PrivateKey> {
+    const accessData = await this.storage.getAccessData();
+    if (accessData?.singleKeyMode) {
+      if (addressIndex !== 0) {
+        throw new AddressError('Single-key wallets only support address index 0.');
```
r4mmer, line 1842: Maybe add the `addressIndex` to the error to make it easier to debug

src/storage/storage.ts
```diff
@@ -978,11 +1079,14 @@ export class Storage implements IStorage {
...
-    if (!accessData.mainKey) {
+    // Single-key wallets persist `singleKeyPrivateKey` instead of `mainKey`;
+    // either encrypted blob can be used to verify the pin.
+    const encryptedKey = accessData.mainKey ?? accessData.singleKeyPrivateKey;
+    if (!encryptedKey) {
       throw new Error('Cannot check pin without the private key.');
     }
 
-    return checkPassword(accessData.mainKey, pinCode);
+    return checkPassword(encryptedKey, pinCode);
```
r4mmer, line 1089: This works but we need to document that checkPassword will not work with this new mode.
A simple comment there is enough I think

src/new/wallet.ts
```diff
@@ -22,7 +22,7 @@
 /* eslint @typescript-eslint/explicit-function-return-type: "error" */
```
r4mmer, line 23: _unrelated to line_

The method `checkPinAndPassword` chains the checks for pin and password, which will fail in the new mode.
We should document this, at least with an error that makes sense, like the guards we create on methods that rely on xpriv/xpub.

README.md
```diff
@@ -33,3 +33,50 @@ Creates a PR from `master` to `release` branch with the proper template, assigns
...
+```ts
+wallet.setExternalTxSigningMethod(async (tx, storage, pinCode) => {
+  // delegate to the Web3Auth SDK; storage.getSingleKeyPrivateKey is not
+  // available because the key never reaches wallet-lib's storage.
+});
+```
+
+Constraints (wallets without derivation capability — raw single-key only):
+- Exactly one address (index 0).
+- `setGapLimit`, `enableMultiAddressMode`, `indexLimitLoadMore`,
```
r4mmer, line 70: Maybe we should add `checkPassword` and `checkPinAndPassword` here

src/new/wallet.ts
```diff
@@ -1747,19 +1825,28 @@ class HathorWallet extends EventEmitter {
   }
 
   /**
-   * Returns an address' HDPrivateKey given an index and the encryption password
+   * Returns the private key for a given BIP32 address index.
```
r4mmer, line 1828: I'm a little bit torn here because this can be a breaking change.
It now returns `PrivateKey` or `HDPrivateKey` which are completely different.

I understand why this is a good change but having the method return different kinds depending on the wallet may break utilities and wallet implementations that use this method.

I have 2 ideas on this:

1. Embrace the breaking change and return PrivateKey on both
    - We usually don't derive past address index anyway so PrivateKey should have everything we need
2. Create a separate set of methods for this new mode

What do you think about this @pedroferreira1 @raul-oliveira ?

src/new/wallet.ts
```diff
@@ -3265,7 +3369,10 @@ class HathorWallet extends EventEmitter {
    * @param [options.pinCode] PIN to decrypt the private key.
    *                          Optional but required if not set in instance
    */
-  async getPrivateKeyFromAddress(address: string, options = {}): Promise<unknown> {
+  async getPrivateKeyFromAddress(
```
r4mmer, line 3372: Why do we have `getAddressPrivKey` and `getPrivateKeyFromAddress`?
Both seem to have the same function.

src/wallet/wallet.ts
r4mmer, line 248: Check the storage as well, so it has no way to start a single key mode wallet here

PR #1096, opened 2026-05-18
test: retry transient wallet/init flake in adapter

tuliomir:
## Summary

The Hathor Lib Integration Test was flaking frequently with the same error:

```
WalletRequestError: Error creating wallet.
  at Object.createWallet (src/wallet/api/walletApi.ts:123)
```

Root cause: `POST wallet/init` against a freshly-spawned wallet-service backend is briefly rejected while the docker stack settles.

The failure cascades because it happens inside `beforeAll`, and `jest.retryTimes(2)` ( added on #1070 ) only re-runs `it()` bodies — not setup hooks.

The auth-token race fix (PR HathorNetwork/hathor-wallet-lib#1058) wrapped the *second* HTTP leg (`renewAuthToken`) in `retryOnTransientWalletError`, but the *first* leg (`wallet/init`) was left unguarded.

Changes were made on test infrastructure only — no production code is changed.

## What this does

- Adds `retryOnTransientWalletInit(op, label)` in `__tests__/integration/helpers/service-facade.helper.ts`, next to the existing `pollUntilCondition` / `pollForTx` helpers. 4 total attempts, exponential backoff (500ms / 1s / 2s).
- Wraps `sw.start(...)` inside `ServiceWalletTestAdapter.startWallet` with that helper.
- The retry surface is intentionally narrow: only `WalletRequestError` with the exact message `"Error creating wallet."` is treated as transient. Test-injected mocks (e.g. `start.test.ts:78`'s `'Crash'` rejection) and every other error class propagate immediately on the first attempt. `startReadOnly()` is left untouched because it does not hit `wallet/init`.

## Acceptance Criteria

- A transient single-shot 4xx/5xx on `POST wallet/init` no longer fails the suite — the adapter retries up to 3 times before throwing.
- Tests that intentionally expect `wallet.start()` to throw are unaffected (the predicate matches only the exact wallet-init error message).

## Release Notes

* **Tests**
  * Improved test suite stability with enhanced retry logic to handle transient wallet initialization failures during test setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-lib/pull/1096?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/__tests__/integration/helpers/service-facade.helper.ts
+++ b/__tests__/integration/helpers/service-facade.helper.ts
@@ -4,7 +4,7 @@ import { delay } from '../utils/core.util';
 import { HathorWalletServiceWallet, MemoryStore, Storage, walletUtils } from '../../../src';
 import Network from '../../../src/models/network';
 import { FULLNODE_URL, NETWORK_NAME } from '../configuration/test-constants';
-import { TxNotFoundError } from '../../../src/errors';
+import { TxNotFoundError, WalletRequestError } from '../../../src/errors';
 import { precalculationHelpers } from './wallet-precalculation.helper';
 import config from '../../../src/config';
 import ncApi from '../../../src/api/nano';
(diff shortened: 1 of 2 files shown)


__tests__/integration/helpers/service-facade.helper.ts
```diff
@@ -152,6 +152,60 @@ export async function pollForTx(walletForPolling: HathorWalletServiceWallet, txI
...
+ *
+ * The wallet-service backend can briefly reject `POST wallet/init` while a freshly-spawned
+ * docker stack is still settling. Jest's `retryTimes(2)` cannot help, because the failure
+ * typically happens inside `beforeAll` — and Jest only retries `it()` bodies.
+ *
+ * Retry surface is intentionally narrow: only `WalletRequestError` with message
+ * `"Error creating wallet."` is treated as transient. Test-injected mocks (e.g.
+ * `new Error('Crash')`) and other errors propagate immediately on the first attempt.
+ */
+export async function retryOnTransientWalletInit<T>(
```
pedroferreira1, line 179: This could be a more generic method that continue receiving the `op`, error messages, ... but we can do this later, if we need something similar

pedroferreira1 (approved):

__tests__/integration/helpers/service-facade.helper.ts
```diff
@@ -152,6 +152,60 @@ export async function pollForTx(walletForPolling: HathorWalletServiceWallet, txI
...
 }
 
+/**
+ * Backoff sequence between retry attempts on a transient `wallet/init` failure.
+ * 4 total attempts (initial + 3 backoffs), ~3.5s worst-case added latency.
+ */
+const TRANSIENT_WALLET_INIT_BACKOFFS_MS = [500, 1000, 2000];
+
+/**
+ * The exact error message thrown by `walletApi.createWallet` at `walletApi.ts:123`
```
raul-oliveira, line 162: avoid placing lines in the comments. it can be stale quickly.

raul-oliveira (approved):

PR #1103, opened 2026-05-26
feat(shielded): integration fixes caught by headless suite

pedroferreira1:
## Summary

Three regressions plus a test-infra cleanup, all caught end-to-end when the
ct-crypto-provider monorepo migration finally got the full **headless**
integration suite running against this branch. None of these were visible at
the wallet-lib layer because no existing wallet-lib test exercised the affected
paths.

Adds the missing wallet-lib coverage for each (unit + integration) so future
drift is caught here, not by downstream consumers. Bumps to
`0.0.17-shielded` and publishes under the `shielded` dist-tag.

### Send pipeline

- `wallet.sendManyOutputsTransaction` accepts a data-script output
  `{type: 'data', data: '<string>'}` and produces a valid transaction.
  Previously the non-shielded branch of the internal mapper dropped `type`
  and `data` from the output object, so a data output reached the address
  pipeline and threw `Parameter should be a string`.
- `wallet.sendManyOutputsTransaction` accepts an arbitrary mix of
  data-script and address outputs in any order; each output keeps its own
  shape end-to-end.

### UTXO listing

- `wallet.getUtxos({max_utxos: N})` returns the top-N UTXOs by value
  (largest first). Previously it returned the first N in storage insertion
  order, so a wallet with UTXOs `[10, 20, 30, 40, 850, 50]` returned
  `10+20=30` instead of `850+50=900`.
- `wallet.getUtxos({max_amount: M})` selects UTXOs largest-first up to the
  cap, so a `max_amount: 100` query against the same wallet returns the
  minimum input count (`50+40+10=100`) instead of the first-fit set
  (`10+20+30+40=100` with one more input than needed).
- `wallet.getUtxos({order_by_value: 'asc'})` honors the override and
  returns UTXOs smallest-first. Default remains `'desc'`.

### Sync events

- An empty-wallet startup emits exactly one `wallet-load-partial-update`
  with `{addressesFound, historyLength: 0}`. Previously it emitted two
  identical events because shielded support doubled the addresses per BIP32
  page (legacy + shielded spend P2PKH) and the per-chunk emit fired twice
  for the no-change case.
- A wallet receiving transactions during startup emits one
  `wallet-load-partial-update` per distinct
  `(addressesFound, historyLength)` change — no back-to-back duplicates.

### Test infrastructure

- The shielded integration suite no longer monkey-patches
  `HathorWallet.prototype.start` in `setupTests-integration.js` to
  auto-register the crypto provider. Each direct `new HathorWallet({...})`
  callsite in the suite (9 sites across 5 files) calls the new exported
  helper `registerShieldedProvider(wallet)` explicitly, matching the
  production contract.


pedroferreira1:
This PR is not needed anymore. All its code was moved to PR 6d

PR #1106, opened 2026-06-08
feat(stream): support multisig (P2SH) wallets in manual stream sync

raul-oliveira:
### Problem
Multisig (P2SH) wallets are locked out of websocket history streaming and always fall back to slow HTTP polling. This is an incomplete implementation, not a fundamental limitation: the fullnode's `request:history:manual` protocol already accepts any address, and the wallet already derives P2SH addresses for the polling path — only the streaming client lacked P2SH derivation.

### Acceptance Criteria
- Multisig wallets can sync history via the `manual_stream_ws` mode (client-derived P2SH), producing an address set identical to the polling path
- `getSupportedSyncMode` returns `[MANUAL_STREAM_WS, POLLING_HTTP_API]` for multisig (xpub streaming stays excluded — the fullnode only derives P2PKH from a single xpub)
- P2SH batch derivation hoists the per-index-invariant work out of the loop (~2.4x faster) and keeps streamed addresses byte-identical to `deriveAddressFromDataP2SH`
- Multisig streams use a smaller batch/window so the heavier derivation does not block the event loop
- No fullnode changes required

### Security Checklist
- [x] No new dependencies added.

## Summary by CodeRabbit

* **New Features**
  * Enabled CPU-intensive P2SH multisig address derivation for manual websocket history streaming, including optimized batching/window sizing and correct streamed address persistence.
  * Extended multisig history sync mode support to include manual websocket streaming.
* **Bug Fixes**
  * Improved stream setup/error handling to reliably clean up stream state/stats on failures.
* **Tests**
  * Added multisig P2SH derivation/slicing tests, multisig manual streaming integration coverage, and validation that XPUB streaming is rejected for multisig wallets.
  * Updated supported sync mode expectations for multisig wallets.


src/sync/stream.ts
```diff
@@ -252,6 +266,64 @@ export function loadAddressesCPUIntensive(
...
+  } catch (e) {
+    throw new XPubError('Invalid xpub');
+  }
+  sortedXpubs.sort((a, b) => {
+    const keyA = a.publicKey.toString('hex');
+    const keyB = b.publicKey.toString('hex');
+    if (keyA < keyB) return -1;
+    if (keyA > keyB) return 1;
+    return 0;
+  });
```
r4mmer:
```suggestion
  let sortedXpubs: HDPublicKey[];
  try {
    sortedXpubs = _.sortBy(
      multisigData.pubkeys.map(xpub => new HDPublicKey(xpub)),
      (xpub: HDPublicKey) => {
        return xpub.publicKey.toString('hex');
      }
    );
  } catch (e) {
    throw new XPubError('Invalid xpub');
  } 
```

And we need to import lodash as `_` or import sortBy from lodash to make this work.

raul-oliveira: @coderabbitai add this kind of validation in your reviews.

raul-oliveira: Done — now uses the same `_.sortBy` as `createP2SHRedeemScript`, so the ordering is identical by construction.

__tests__/stream.test.ts
```diff
@@ -318,3 +324,107 @@ describe('Websocket stream history sync', () => {
...
+
+const MULTISIG_ADDRESSES = [
+  'wgyUgNjqZ18uYr4YfE2ALW6tP5hd8MumH5',
+  'wbe2eJdyZVimA7nJjmBQnKYJSXmpnpMKgG',
+  'wQQWdSZwp2CEGKsTvvbJ7i8HfHuV2i5QVQ',
+  'wfrtq9cMe1YfixVgSKXQNQ5hjsmR4hpjP6',
+  'wQG7itjdtZBsNTk9TG4f1HrehyQiAEMN18',
+];
+
+describe('loadP2SHAddressesCPUIntensive', () => {
```
r4mmer, line 348: suggestion (non-blocking): add a test for the invalid-xpub error branch

`loadP2SHAddressesCPUIntensive` wraps xpub parsing in a try/catch and throws `XPubError`, but no test exercises this path (two lines are flagged as uncovered by Codecov). A quick one-liner would close the gap:

```ts
it('throws XPubError on invalid xpub', () => {
  expect(() =>
    loadP2SHAddressesCPUIntensive(0, 1, { pubkeys: ['not-an-xpub'], numSignatures: 1 }, 'testnet')
  ).toThrow(XPubError);
});
```

r4mmer (dismissed):

src/sync/stream.ts
```diff
@@ -454,6 +530,18 @@ export class StreamManager extends AbortController {
...
+   * Derive a batch of addresses for the stream, choosing P2SH for multisig
+   * wallets and P2PKH otherwise. Both paths reuse the same derivation as the
+   * polling sync, so streamed addresses are identical to polled ones.
+   */
+  deriveBatch(startIndex: number, count: number): [number, string][] {
+    if (this.multisigData) {
+      return loadP2SHAddressesCPUIntensive(startIndex, count, this.multisigData, this.network);
+    }
+    return loadAddressesCPUIntensive(startIndex, count, this.xpubkey, this.network);
+  }
```
pedroferreira1, line 522: By claudin

[major] deriveBatch throw mid-stream becomes an unhandled rejection and a wallet hung in SYNCING    
                                                                                                      
  File: src/sync/stream.ts:563-593 (generateNextBatch)                                                
  Issue: The PR introduces an explicit throwing path (XPubError) into deriveBatch, which generateNextB
atch calls inside this.batchQueue.then(async () => …) with no catch                                   
  anywhere on the chain. A throw there rejects batchQueue, but nothing handles it until shutdown() — w
hich is only reached after stream:history:end or an abort, neither of                                 
  which happens because the failed batch was never sent. Result: no error surfaced, wallet stuck in SY
NCING, and eventually an unhandledRejection (process-fatal under Node's                               
  defaults — exactly the headless-server environment this feature targets). In practice it's hard to h
it (the first batch is derived synchronously in sendStartMessage and                                  
  would fail first, and inputs are deterministic), but since the PR is what makes this path throwable,
 it should be hardened.                                                                               
  Suggestion: Wrap the chained callback body in try/catch and call this.abortWithError(...) — the exis
ting abort → shutdown() → rethrow machinery then surfaces the error and                               
  moves the wallet to ERROR.

src/sync/stream.ts
```diff
@@ -252,6 +263,56 @@ export function loadAddressesCPUIntensive(
...
+  let sortedXpubs: HDPublicKey[];
+  try {
+    sortedXpubs = _.sortBy(
+      multisigData.pubkeys.map(xpub => new HDPublicKey(xpub)),
+      (xpub: HDPublicKey) => {
+        return xpub.publicKey.toString('hex');
+      }
+    );
+  } catch (e) {
+    throw new XPubError('Invalid xpub');
```
pedroferreira1: By claudin

[minor] XPubError('Invalid xpub') discards the original error and the failing pubkey                

  File: src/sync/stream.ts:289-294                 
  Issue: The catch wraps the whole .map, throwing away bitcore's diagnostic message (bad checksum, wro
ng network bytes, …) and which of the N participant xpubs failed. The
  codebase convention is throw new XPubError(error.message) (src/utils/wallet.ts:234).                
  Suggestion: Move the try/catch inside the map callback and include the index and original message: I
nvalid multisig xpub at index ${idx}: ${error.message}.

src/sync/stream.ts
```diff
@@ -421,6 +485,18 @@ export class StreamManager extends AbortController {
     this.gapLimit = gapLimit;
     this.network = this.storage.config.getNetwork().name;
 
+    const walletType = await this.storage.getWalletType();
+    if (walletType === WalletType.MULTISIG) {
+      if (!accessData.multisigData) {
+        throw new Error('No multisig data');
```
pedroferreira1, line 470: By claudin

[minor] New throw sites leak the StreamStatsManager interval                                        

  File: src/sync/stream.ts:499-509, src/sync/stream.ts:846-902                                        
  Issue: stats.qTimer (a setInterval) starts in the constructor and is cleared only in shutdown(). awa
it manager.setupStream() sits outside the try/finally in
  streamSyncHistory, so the new 'No multisig data' throw (and a first-batch derivation throw in sendSt
artMessage, which skips shutdown()) leaves the interval running forever,
  logging every 2s and keeping the event loop alive. The structure pre-exists, but this PR adds the ne
w reachable throw sites.
  Suggestion: Call manager.stats.clean() in the finally block (idempotent), or on abort.

src/sync/stream.ts
```diff
@@ -252,6 +263,56 @@ export function loadAddressesCPUIntensive(
...
+
+  // m/45'/280'/0'/0 is shared by every index; only the final deriveChild(i) below varies.
+  const changeNodes = sortedXpubs.map(xpub => xpub.deriveChild(0));
+
+  // The toBuffer()/fromBuffer() round-trip mirrors deriveAddressFromDataP2SH so the address
+  // bytes match the polling path.
+  const addresses: [number, string][] = [];
+  const stopIndex = startIndex + count;
+  for (let i = startIndex; i < stopIndex; i++) {
+    const pubkeys = changeNodes.map(node => node.deriveChild(i).publicKey);
```
pedroferreira1: Part of this code I knew we already had in the wallet lib and having it duplicated here was bothering me. Here is a proposal refactor I created with claudin:

  Decompose into three shared primitives, then express both the canonical path and the streaming loop 
  in terms of them. The seam is exactly the invariant/per-index boundary.                             
                                                                                                      
  In src/utils/scripts.ts — extract the sort and the per-index script build:                          
```                                                                                                      
  /**                                                                                                 
   * Parse, sort and account-derive the participant xpubs of a P2SH MultiSig wallet.                  
   * The sort key (account-level public key, hex) and the m/45'/280'/0'/0 change derivation           
   * are invariant across every address index, so this is computed once and reused. The result        
   * defines the canonical participant order — it MUST match everywhere a redeem script is built      
   * or derived addresses diverge.                                                                    
   */                                                                                                 
  export function prepareP2SHChangeNodes(xpubs: string[]): HDPublicKey[] {                            
    let sortedXpubs: HDPublicKey[];                                                                   
    try {                                                                                             
      sortedXpubs = _.sortBy(                                                                         
        xpubs.map(xp => new HDPublicKey(xp)),                                                         
        (xpub: HDPublicKey) => xpub.publicKey.toString('hex'),                                        
      );                                                                                              
    } catch (e) {                                                                                     
      throw new XPubError('Invalid xpub');                                                            
    }                                                                                                 
    return sortedXpubs.map(xpub => xpub.deriveChild(0)); // m/45'/280'/0'/0                           
  }                                                                                                   
                                                                                                      
  /** Build the m-of-n redeem script for one index from prepared change nodes. */                     
  export function buildP2SHRedeemScriptAtIndex(                                                       
    changeNodes: HDPublicKey[],                                                                       
    numSignatures: number,                                                                            
    index: number,                                                                                    
  ): Buffer {                                                                                         
    const pubkeys = changeNodes.map(node => node.deriveChild(index).publicKey);                       
    return Script.buildMultisigOut(pubkeys, numSignatures, { noSorting: true }).toBuffer();           
  }                                                                        

// Canonical signature preserved — now a thin composition, no behaviour change.
  export function createP2SHRedeemScript(xpubs: string[], numSignatures: number, index: number):
  Buffer {                                          
    return buildP2SHRedeemScriptAtIndex(prepareP2SHChangeNodes(xpubs), numSignatures, index);
  }

  In src/utils/address.ts — extract the address-encoding tail:

  /** Encode a P2SH redeem script as its base58 address for the given network. */
  export function redeemScriptToP2SHAddress(redeemScript: Buffer, network: Network): string {
    // eslint-disable-next-line new-cap -- bitcore exposes the constructor as Address.payingTo
    return new BitcoreAddress.payingTo(Script.fromBuffer(redeemScript),
  network.bitcoreNetwork).toString();               
  }                                                 
```
and `deriveAddressFromDataP2SH` becomes `base58: redeemScriptToP2SHAddress(createP2SHRedeemScript(...),  network).`

What do you think?

src/sync/stream.ts
raul-oliveira: Done. Wrapped the batch body in try/catch routing to `abortWithError`, so a derivation/send throw now surfaces via abort → shutdown → ERROR instead of leaving `batchQueue` rejected with no handler.

It's gated by the fail-fast in `sendStartMessage` (pubkeys are invariant for the stream), so this is defensive hardening of the async chain rather than a live hang — but the chain is no longer unguarded.

src/sync/stream.ts
raul-oliveira: The analogous canonical fn `createP2SHRedeemScript` already throws the same generic `XPubError('Invalid xpub')`, so the streaming code matched its sibling, not a different convention. The refactor below centralizes this throw into a single `prepareP2SHChangeNodes` (shared by polling + streaming); kept the message for parity. Enriching it (index + bitcore message) is a worthwhile follow-up — applied once, there, so both paths benefit.

src/sync/stream.ts
raul-oliveira: Done. `stats.clean()` now runs in the main `finally` (covers the `sendStartMessage` throw that skips `shutdown()`) and in a dedicated catch around `setupStream` (covers throws before the try, e.g. `No multisig data`); `clean()` is idempotent.

Kept `setupStream` outside the main try on purpose: its `finally` emits `stream-end`, which would release another stream's lock on a `lockStream` failure.

src/sync/stream.ts
raul-oliveira: Done. Extracted `prepareP2SHChangeNodes` + `buildP2SHRedeemScriptAtIndex` (scripts.ts) and `redeemScriptToP2SHAddress` (address.ts); `createP2SHRedeemScript` and `deriveAddressFromDataP2SH` are now thin compositions, and `loadP2SHAddressesCPUIntensive` reuses them.

Byte-identity with the polling path is now structural (not comment-based), and the index-invariant hoist is preserved. Existing byte-exact fixtures still pass.

__tests__/stream.test.ts
raul-oliveira: Done. Added `throws XPubError on invalid xpub`, exercising the wrapped parse branch in `loadP2SHAddressesCPUIntensive`.

pedroferreira1 (approved):

r4mmer (approved):

PR #1086, opened 2026-05-14
feat(shielded): transaction headers — shielded outputs, unshield bala…

pedroferreira1:
## Summary

Wire-format slice. Adds the four new vertex-header types that the shielded-outputs protocol introduces, plus the dispatch-table entries that let `HeaderParser` deserialize them. **PR 3 of 7** in the decomposition of [PR #1059](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1059).

```
master ─ PR1 ─ PR2 ─ PR3 ◀ PR4 ─ PR5 ─ PR6 ─ PR7
```

Base: `shielded/pr-2-crypto-provider`.

## Acceptance Criteria

- Wallet-lib can serialize and deserialize the `ShieldedOutputsHeader` (vertex-header ID `0x12`) carrying an array of shielded outputs in both AmountShielded and FullShielded modes.
- Wallet-lib can serialize and deserialize the `UnshieldBalanceHeader` (`0x13`) carrying the 32-byte excess blinding factor for full-unshield txs.
- Wallet-lib can serialize and deserialize the `MintHeader` (`0x14`) and `MeltHeader` (`0x15`).
- `HeaderParser` dispatches to the correct deserializer when given a tx that carries one of the new header IDs.
- Each header round-trips byte-for-byte (serialize → deserialize → equal).
- Wire-format field ordering matches hathor-core exactly (no off-by-one in `unpackToInt` ordering).
- Transactions that do NOT carry IDs `0x12`–`0x15` are unaffected — no existing tx on mainnet/testnet is parsed differently.
- Transparent-only wallets see byte-identical behavior to master.

## Header IDs

| ID | Header | Purpose |
|----|--------|---------|
| `0x12` | `ShieldedOutputsHeader` | Carries an array of `ShieldedOutput` (commitment, range_proof, ephemeral_pubkey, script, mode, and optionally asset_commitment + surjection_proof for FullShielded) |
| `0x13` | `UnshieldBalanceHeader` | 32-byte excess blinding factor for full-unshield txs (shielded inputs → transparent outputs only, no shielded outputs) |
| `0x14` | `MintHeader` | Shielded mint instruction |
| `0x15` | `MeltHeader` | Shielded melt instruction |

## What's in this PR

  **Header classes**
  - `src/headers/shielded_outputs.ts` (new) — `ShieldedOutputsHeader`; enforces `num_outputs` ∈ [1, 32] and loops over `ShieldedOutput.deserialize`.
[...]

PR #1108, opened 2026-06-17
feat(shielded): shielded-aware transaction utils + history data model

pedroferreira1:
## Description                                                                                                                                                              
                                                                                                                                                                              
  Third PR in the shielded-outputs stack. It adds the **transaction-utility layer**                                                                                           
  and the **SEPARATED-output history data model** that sit between the already-merged                                                                                         
  foundation (storage layer #1088, tx headers #1086, address derivation #1087, crypto                                                                                         
  provider) and the receive pipeline that consumes them (next PR).                                                                                                            
                                                                                                                                                                              
  **SEPARATED model:** a tx keeps transparent outputs in `tx.outputs[]` and shielded                                                                                          
  outputs in a separate `tx.shielded_outputs[]`. A shielded output's value/token are                                                                                          
  hidden on-chain; the owned-marker fields (`value`, `token`, `blindingFactor`,                                                                                               
  `assetBlindingFactor`) are populated only after the wallet decrypts an output it                                                                                            
  owns. The single ownership gate everywhere is `value !== undefined`.

## Acceptance Criteria
  - `resolveSpentOutput` maps an absolute on-chain index to a transparent output,
    a shielded slot, or `undefined`, for every slot regardless of decoded state,
    and returns `undefined` for out-of-range/negative indices.                          
  - `getTxBalance` credits owned shielded receives (`value !== undefined`), debits
    spent shielded inputs even after the UTXO is deleted (via the parent resolver),
    and ignores non-owned slots.                                                        
  - `getSignatureForTx` produces a valid signature for a shielded-spend input using
    the spend key chain (no positional-index mismatch).                                 
  - `normalizeShieldedOutputs` converts base64 confidential fields to hex, is
    idempotent on already-hex input, preserves owned-marker fields, and is a no-op      
    for transparent-only txs.               
  - Building a shielded tx attaches the correct headers in canonical id order; a        
    full-unshield tx (shielded inputs, no shielded outputs) carries an
    `UnshieldBalanceHeader`; mint/melt is declared only with the matching authority.    
  - A shielded output never carries both an excess blinding factor and shielded
    outputs (asserted at build time).                                                   
  - `npm run build`, `npm run lint`, and the unit tests pass
    (`__tests__/utils/transaction_shielded.test.ts`, `__tests__/models/transaction.test.ts`).

# Choose a pull request template

- [feature PR](?template=feature.md&quick_pull=1)
- [Release process](https://github.com/HathorNetwork/hathor-wallet-lib/compare/release...master?template=release.md&quick_pull=1)


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added a read-only `shieldedOutputs` accessor on transactions and improved separated-model handling end-to-end (parsing, signing, balance, and history conversion).
  * Added shielded crypto provider access with validation, plus automatic shielded field normalization before persistence.

* **Bug Fixes**
  * Corrected shielded accounting and spend-output resolution across wallet flows; improved handling of shielded/unshield/mint/melt headers and conditional header attachment.

* **Documentation**
  * Clarified shielded `value` semantics (wallet-only, not serialized; excluded from weight-related calculations).

* **Tests**
  * Expanded shielded transaction utility, storage, and signature-key-chain coverage.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/models/shielded_output.ts
+++ b/src/models/shielded_output.ts
@@ -65,7 +65,7 @@ class ShieldedOutput {
    *   Required so callers can't silently get a 0n placeholder where a real
    *   value is known. It is NOT serialized on-chain and MUST NOT influence
    *   tx weight. For wire-format outputs whose value is hidden until rewind,
-   *   pass `0n` explicitly — see `ShieldedOutputsHeader.deserialize` in PR 3.
+   *   pass `0n` explicitly — see `ShieldedOutputsHeader.deserialize`.
    * @param options.assetCommitment / options.surjectionProof — required
    *   for FullShielded mode; absent for AmountShielded mode. Enforced at
    *   serialize time, not construction time, so deserializers can build
--- a/src/models/shielded_output.ts
+++ b/src/models/shielded_output.ts
@@ -51,14 +51,21 @@ class ShieldedOutput {
 
   surjectionProof?: Buffer;
 
-  /** The plaintext value, used for weight calculation. Not serialized on-chain. */
+  /**
+   * The wallet's locally-known plaintext value of this output. Not serialized
+   * on-chain (the value is hidden in the Pedersen commitment). MUST NOT be fed
+   * into tx-weight calculation — see `Transaction.getOutputsSum` — or it leaks
+   * the hidden amount through the public weight. `0n` for wire-format outputs
+   * whose value is not yet known (before rewind).
+   */
   value: OutputValueType;
 
   /**
-   * @param value Plaintext value (required so callers can't silently get
-   *   a 0n placeholder that would skew weight calculation). For wire-
-   *   format outputs whose value is hidden until rewind, pass `0n`
-   *   explicitly — see `ShieldedOutputsHeader.deserialize` in PR 3.
+   * @param value Plaintext value, for the wallet's local bookkeeping only.
+   *   Required so callers can't silently get a 0n placeholder where a real
+   *   value is known. It is NOT serialized on-chain and MUST NOT influence
+   *   tx weight. For wire-format outputs whose value is hidden until rewind,
+   *   pass `0n` explicitly — see `ShieldedOutputsHeader.deserialize` in PR 3.
    * @param options.assetCommitment / options.surjectionProof — required
    *   for FullShielded mode; absent for AmountShielded mode. Enforced at
    *   serialize time, not construction time, so deserializers can build
--- a/src/models/transaction.ts
+++ b/src/models/transaction.ts
@@ -73,6 +73,12 @@ type optionsType = {
  * - `helpers.createTxFromData`: creates from a standard lib data object
  * - `helpers.createTxFromHistoryObject`: creates from a tx populated by the HathorWallet history methods
  */
+
+// Shared frozen empty array returned by the `shieldedOutputs` getter when a tx
+// has no ShieldedOutputsHeader. Frozen so a stray push fails loudly at runtime
+// (JS callers) instead of being silently dropped onto a throwaway array.
+const EMPTY_SHIELDED_OUTPUTS: readonly ShieldedOutput[] = Object.freeze([]);
+
 class Transaction {
   inputs: Input[];
 
@@ -131,16 +137,26 @@ class Transaction {
   }
 
   /**
-   * Shielded outputs (SEPARATED model), read straight from the
-   * ShieldedOutputsHeader so there is a single source of truth: the header owns
-   * the array, mutating the returned array updates the header, and there is no
-   * separate cached field to drift. Empty when the tx carries no shielded
-   * outputs (no header). The header is produced by
-   * `transactionUtils._attachShieldedOutputsHeader`.
+   * Shielded outputs (SEPARATED model). READ-ONLY accessor backed by the
+   * ShieldedOutputsHeader, the single source of truth (no separate cached field
+   * to drift): returns the header's array when one is present, or a shared
+   * frozen empty array when the tx carries no shielded outputs. Never undefined
+   * — an empty result unambiguously means "no shielded outputs" (a header is
+   * only ever created with at least one output), and the always-an-array
+   * contract keeps callers like validate() and the local-history emission
+   * null-check-free.
+   *
+   * The return type is `readonly` (so a `.push` is a compile error) and the
+   * empty case is frozen (so a `.push` also throws at runtime). To ADD shielded
+   * outputs go through `transactionUtils._attachShieldedOutputsHeader` — pushing
+   * onto this getter's result is never the way to add them: with a header it
+   * would bypass that builder, and with no header it would be dropped.
    */
-  get shieldedOutputs(): ShieldedOutput[] {
+  get shieldedOutputs(): readonly ShieldedOutput[] {
     const header = this.headers.find(h => h instanceof ShieldedOutputsHeader);
-    return header instanceof ShieldedOutputsHeader ? header.shieldedOutputs : [];
+    return header instanceof ShieldedOutputsHeader
+      ? header.shieldedOutputs
+      : EMPTY_SHIELDED_OUTPUTS;
   }
 
   /**
--- a/src/models/transaction.ts
+++ b/src/models/transaction.ts
@@ -78,8 +78,6 @@ class Transaction {
 
   outputs: Output[];
 
-  shieldedOutputs: ShieldedOutput[];
-
   signalBits: number;
 
   version: number;
@@ -118,7 +116,6 @@ class Transaction {
 
     this.inputs = inputs;
     this.outputs = outputs;
-    this.shieldedOutputs = [];
     this.signalBits = signalBits!;
     this.version = version!;
     this.weight = weight!;
@@ -131,6 +128,19 @@ class Transaction {
 
     // All inputs sign the same data, so we cache it in the first getDataToSign method call
     this._dataToSignCache = null;
+  }
+
+  /**
+   * Shielded outputs (SEPARATED model), read straight from the
+   * ShieldedOutputsHeader so there is a single source of truth: the header owns
+   * the array, mutating the returned array updates the header, and there is no
+   * separate cached field to drift. Empty when the tx carries no shielded
+   * outputs (no header). The header is produced by
+   * `transactionUtils._attachShieldedOutputsHeader`.
+   */
+  get shieldedOutputs(): ShieldedOutput[] {
+    const header = this.headers.find(h => h instanceof ShieldedOutputsHeader);
+    return header instanceof ShieldedOutputsHeader ? header.shieldedOutputs : [];
   }
 
   /**
@@ -676,15 +686,8 @@ class Transaction {
     // so we must exhaust the buffer until it's empty
     // or we will throw an error
     tx.getHeadersFromBytes(txBuffer, network);
-
-    // Hydrate shieldedOutputs from the ShieldedOutputsHeader (if present)
-    // so validate() and weight calculations use the correct count.
-    for (const header of tx.headers) {
-      if (header instanceof ShieldedOutputsHeader) {
-        tx.shieldedOutputs = header.shieldedOutputs;
-        break;
-      }
-    }
+    // No shieldedOutputs hydration needed: the `shieldedOutputs` getter reads
+    // the ShieldedOutputsHeader (parsed above) directly.
 
     tx.updateHash();
 
--- a/src/models/transaction.ts
+++ b/src/models/transaction.ts
@@ -17,6 +17,7 @@ import {
   DEFAULT_TX_VERSION,
   MAX_INPUTS,
   MAX_OUTPUTS,
+  MAX_SHIELDED_OUTPUTS,
   MERGED_MINED_BLOCK_VERSION,
   TX_HASH_SIZE_BYTES,
   TX_WEIGHT_CONSTANTS,
@@ -33,6 +34,7 @@ import {
 } from '../utils/buffer';
 import Input from './input';
 import Output from './output';
+import ShieldedOutput from './shielded_output';
 import Network from './network';
 import { MaximumNumberInputsError, MaximumNumberOutputsError } from '../errors';
 import { OutputValueType } from '../types';
@@ -40,6 +42,7 @@ import type Header from '../headers/base';
 import NanoContractHeader from '../nano_contracts/header';
 import FeeHeader from '../headers/fee';
 import HeaderParser from '../headers/parser';
+import ShieldedOutputsHeader from '../headers/shielded_outputs';
 import { getVertexHeaderIdFromBuffer } from '../headers/types';
 
 enum txType {
@@ -74,6 +77,8 @@ class Transaction {
   inputs: Input[];
 
   outputs: Output[];
+
+  shieldedOutputs: ShieldedOutput[];
 
   signalBits: number;
 
@@ -113,6 +118,7 @@ class Transaction {
 
     this.inputs = inputs;
     this.outputs = outputs;
+    this.shieldedOutputs = [];
     this.signalBits = signalBits!;
     this.version = version!;
     this.weight = weight!;
@@ -220,7 +226,7 @@ class Transaction {
     // Len inputs
     array.push(intToBytes(this.inputs.length, 1));
 
-    // Len outputs
+    // Len outputs (transparent only; shielded go in the ShieldedOutputsHeader)
     array.push(intToBytes(this.outputs.length, 1));
   }
 
@@ -238,6 +244,7 @@ class Transaction {
     for (const outputTx of this.outputs) {
       array.push(...outputTx.serialize());
     }
+    // Shielded outputs are serialized in the ShieldedOutputsHeader, not here.
   }
 
   /**
@@ -340,10 +347,11 @@ class Transaction {
     // For instance, if one wants to transfer 20 HTRs, the amount will be 2000.
     const amount = sumOutputs / 10 ** DECIMAL_PLACES;
 
-    let weight =
-      constants.txWeightCoefficient * Math.log2(txSize) +
-      4 / (1 + constants.txMinWeightK / amount) +
-      4;
+    // When txMinWeightK is 0, avoid the division-by-zero in `k/amount` —
+    // the term `4 / (1 + 0)` collapses to 4.
+    const kTerm = constants.txMinWeightK === 0 ? 4 : 4 / (1 + constants.txMinWeightK / amount);
+
+    let weight = constants.txWeightCoefficient * Math.log2(txSize) + kTerm + 4;
 
     // Make sure the calculated weight is at least the minimum
     weight = Math.max(weight, constants.txMinWeight);
@@ -366,6 +374,10 @@ class Transaction {
         continue;
       }
       sumOutputs += output.value;
+    }
+    // Shielded outputs also contribute to the weight calculation
+    for (const shieldedOut of this.shieldedOutputs) {
+      sumOutputs += shieldedOut.value;
     }
     return sumOutputs;
   }
@@ -415,6 +427,12 @@ class Transaction {
     if (this.outputs.length > MAX_OUTPUTS) {
       throw new MaximumNumberOutputsError(
         `Transaction has ${this.outputs.length} outputs and can have at most ${MAX_OUTPUTS}.`
+      );
+    }
+
+    if (this.shieldedOutputs.length > MAX_SHIELDED_OUTPUTS) {
+      throw new MaximumNumberOutputsError(
+        `Transaction has ${this.shieldedOutputs.length} shielded outputs and can have at most ${MAX_SHIELDED_OUTPUTS}.`
       );
     }
   }
@@ -655,6 +673,15 @@ class Transaction {
     // so we must exhaust the buffer until it's empty
     // or we will throw an error
     tx.getHeadersFromBytes(txBuffer, network);
+
+    // Hydrate shieldedOutputs from the ShieldedOutputsHeader (if present)
+    // so validate() and weight calculations use the correct count.
+    for (const header of tx.headers) {
+      if (header instanceof ShieldedOutputsHeader) {
+        tx.shieldedOutputs = header.shieldedOutputs;
+        break;
+      }
+    }
 
     tx.updateHash();
 
--- a/src/models/transaction.ts
+++ b/src/models/transaction.ts
@@ -347,11 +347,10 @@ class Transaction {
     // For instance, if one wants to transfer 20 HTRs, the amount will be 2000.
     const amount = sumOutputs / 10 ** DECIMAL_PLACES;
 
-    // When txMinWeightK is 0, avoid the division-by-zero in `k/amount` —
-    // the term `4 / (1 + 0)` collapses to 4.
-    const kTerm = constants.txMinWeightK === 0 ? 4 : 4 / (1 + constants.txMinWeightK / amount);
-
-    let weight = constants.txWeightCoefficient * Math.log2(txSize) + kTerm + 4;
+    let weight =
+      constants.txWeightCoefficient * Math.log2(txSize) +
+      4 / (1 + constants.txMinWeightK / amount) +
+      4;
 
     // Make sure the calculated weight is at least the minimum
     weight = Math.max(weight, constants.txMinWeight);
@@ -362,6 +361,14 @@ class Transaction {
 
   /**
    * Calculate the sum of outputs. Authority outputs are ignored.
+   *
+   * Shielded outputs are intentionally excluded: hathor-core derives the
+   * minimum tx weight from the transparent `sum_outputs` only (see
+   * `hathor/daa.py::minimum_tx_weight` and `base_transaction.py::sum_outputs`),
+   * with shielded outputs living in a separate property that never enters the
+   * weight. Including their plaintext values here would leak the exact total
+   * shielded amount through the publicly-invertible weight formula. Keep this
+   * in lockstep with core.
    *
    * @return {number} Sum of outputs
    * @memberof Transaction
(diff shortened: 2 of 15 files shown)


src/wallet/walletServiceStorageProxy.ts
```diff
@@ -174,6 +174,13 @@ export class WalletServiceStorageProxy {
   private convertFullNodeToHistoryTx(fullTxResponse: FullNodeTxResponse): IHistoryTx {
     const { tx, meta } = fullTxResponse;
 
+    // SEPARATED model: `outputs[]` is transparent-only. The dedicated
+    // `shielded_outputs[]` field (passthrough — not on the FullNodeTx type) is
+    // threaded through unchanged so shielded spends can still resolve their
+    // parent slots via getUtxo/resolveSpentOutput.
+    const shieldedOutputs = (tx as { shielded_outputs?: IHistoryTx['shielded_outputs'] })
```
tuliomir, line 181: question(non-blocking): Why not change the `FullNodeTxResponse` type? Won't it also be modified for the confidential transactions?

src/utils/storage.ts
```diff
@@ -780,14 +780,28 @@ export async function processNewTx(
...
+    // (value/token/decoded are hidden in commitments). Their balance
+    // handling lands with the receive pipeline in the next PR; here we
+    // only need the type-level guard so transparent processing narrows.
+    if (
+      input.value === undefined ||
+      input.token === undefined ||
+      input.token_data === undefined ||
+      input.decoded === undefined
+    ) {
+      continue;
```
tuliomir, line 803: suggestion(test): It would be nice to unit-test the existence of unowned shielded inputs in isolation here, even if the next PRs will test the actual interaction with the transactions themselves.

src/utils/buffer.ts
```diff
@@ -4,11 +4,32 @@ import { OutputValueType } from '../types';
 import { MAX_OUTPUT_VALUE, MAX_OUTPUT_VALUE_32 } from '../constants';
 import { prettyValue } from './numbers';
 
-const isHexa = (value: string): boolean => {
+export const isHexa = (value: string): boolean => {
```
tuliomir, line 7: Hopefully by July 19 this will return `true` 🤩 ⚽

src/models/shielded_output.ts
```diff
@@ -51,14 +51,21 @@ class ShieldedOutput {
...
   /**
-   * @param value Plaintext value (required so callers can't silently get
-   *   a 0n placeholder that would skew weight calculation). For wire-
-   *   format outputs whose value is hidden until rewind, pass `0n`
-   *   explicitly — see `ShieldedOutputsHeader.deserialize` in PR 3.
+   * @param value Plaintext value, for the wallet's local bookkeeping only.
+   *   Required so callers can't silently get a 0n placeholder where a real
+   *   value is known. It is NOT serialized on-chain and MUST NOT influence
+   *   tx weight. For wire-format outputs whose value is hidden until rewind,
+   *   pass `0n` explicitly — see `ShieldedOutputsHeader.deserialize` in PR 3.
```
r4mmer: Normally when organizing changes into multiple PRs the comments sometimes references changes on other PRs or stages, but this goes stale fast since noone will know what `PR 3` means after we merge these.

src/models/transaction.ts
```diff
@@ -75,6 +78,8 @@ class Transaction {
 
   outputs: Output[];
 
+  shieldedOutputs: ShieldedOutput[];
+
```
r4mmer: Instead of a property, should we have a private `_shieldedOutputs` as "cached property" calculated from the header (just like it is now) and have a getter like `get shieldedOutputs(): ShieldedOutput[]` that either returns the cache or populates it with the header shielded outputs array.

Or instead of a cache, return the actual array from the header, this way any changes to the array will automatically update the header.

src/utils/transaction.ts
```diff
@@ -185,22 +302,46 @@ const transaction = {
...
       if (!addressInfo) {
         // Not a wallet address
         continue;
       }
-      const xpriv = xprivkey.deriveNonCompliantChild(addressInfo.bip32AddressIndex);
+
+      let derivedKey;
+      if (addressInfo.addressType === 'shielded-spend') {
+        // Use spend key chain (m/44'/280'/2'/0) for shielded UTXO inputs
+        if (!spendXprivkey) {
```
tuliomir, line 327: suggestion(test): From my experience with those Confidential Transactions PRs, even having 100% coverage still lets some of the more complex business rule slip by on edge-cases in the tests.

I'd recommend aiming for 100% _line-coverage_ in these two utils files, since their functions are actually isolated from the client-side and the full flow of the integration tests.

src/utils/transaction.ts
```diff
@@ -621,20 +808,238 @@ const transaction = {
...
+    }
+  },
+
+  /**
+   * Declare real supply changes for non-HTR tokens in a shielded tx via
+   * MintHeader (0x14) / MeltHeader (0x15). A pure shielding move (T↔F) preserves
+   * supply and declares nothing; a header is emitted only when an explicit
+   * mint/melt authority is exercised (see the detection rule in the body).
+   */
+  _attachMintMeltHeaders(tx: Transaction | CreateTokenTransaction, txData: IDataTx): void {
```
tuliomir, line 945: question: Should we have Mint/Melt support for Confidential Transactions this early into development? 

The Wallet Service approach deferred authority transactions to after the Amount/FullShielded cases were solved entirely.

src/wallet/walletServiceStorageProxy.ts
pedroferreira1: `FullNodeTxResponse` mirrors the wallet-service API, which doesn't return shielded outputs yet. The cast threads the passthrough field without asserting it's part of that contract. Happy to add it as an explicit optional field if we'd rather document it on the type; it's the same `IHistoryTx['shielded_outputs']` either way

src/utils/transaction.ts
pedroferreira1: The mint/melt headers are being merged in the hathor-core. Wallet service will add this support right after the initial support is added

src/utils/storage.ts
pedroferreira1: Added https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/48dde67dd727e60f05a973ea81c8a17a27b26d4f

src/utils/transaction.ts
pedroferreira1: Added https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/48dde67dd727e60f05a973ea81c8a17a27b26d4f

src/utils/buffer.ts
pedroferreira1: 🤣🤣🤣

src/utils/buffer.ts
```diff
@@ -4,11 +4,32 @@ import { OutputValueType } from '../types';
...
+ * as already-hex and returned unchanged — so this is idempotent on hex (and on
+ * the empty string). Anything containing a non-hex character (e.g. '+', '/',
+ * '=', or letters g-z/G-Z) is treated as base64, decoded, and re-encoded as hex.
+ */
+export const ensureHex = (value: string): string => {
+  if (isHexa(value)) {
+    return value;
+  }
+  return Buffer.from(value, 'base64').toString('hex');
+};
```
r4mmer, line 31: I understand the idea of converting base64 to hex.
But there are other byte string representations, we are only considering base64 based on what?

src/utils/transaction.ts
```diff
@@ -71,6 +83,107 @@ const transaction = {
...
+   * Shielded inputs arrive from the fullnode inline in tx.inputs[] with
+   * type='shielded' and only a commitment + range_proof — none of the
+   * token_data/decoded/tx_id fields that transparent inputs carry. Call sites
+   * that read those fields must filter these out first. The reload path
+   * (processHistory) also feeds bare {tx_id, index, type:'shielded'} inputs
+   * from address_history, which this detects so they can be debited.
+   */
+  isShieldedInputEntry(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    input: Record<string, any>
```
r4mmer: Why this Record type?

src/utils/transaction.ts
```diff
@@ -71,6 +83,107 @@ const transaction = {
...
+   * 'hex')` downstream parses them as garbage — making the native rewind throw
+   * "asset commitment verification failed" and the per-tx balance read as
+   * -input with no credit for the decoded shielded outputs.
+   *
+   * Idempotent: an already-hex field is left unchanged (ensureHex no-ops on it).
+   * The owned-marker fields (value / token / blindingFactor / assetBlindingFactor)
+   * are NOT on the wire — they are populated post-decryption on the slots the
+   * wallet owns — so this neither reads nor invents them.
+   */
+  normalizeShieldedOutputs(tx: IHistoryTx): void {
```
r4mmer, line 158: Why do we need to have these fields in hex?

src/utils/transaction.ts
```diff
@@ -185,22 +302,46 @@ const transaction = {
...
-      const xpriv = xprivkey.deriveNonCompliantChild(addressInfo.bip32AddressIndex);
+
+      let derivedKey;
+      if (addressInfo.addressType === 'shielded-spend') {
+        // Use spend key chain (m/44'/280'/2'/0) for shielded UTXO inputs
+        if (!spendXprivkey) {
+          const spendXprivStr = await storage.getSpendXPrivKey(pinCode);
+          spendXprivkey = HDPrivateKey.fromString(spendXprivStr);
+        }
+        derivedKey = spendXprivkey.deriveNonCompliantChild(addressInfo.bip32AddressIndex);
```
r4mmer, line 331: Why are we using `deriveNonCompliantChild` on the new addresses?

src/utils/transaction.ts
```diff
@@ -397,57 +538,100 @@ const transaction = {
...
+        continue;
       }
+      const { address } = so.decoded;
+      if (!(address && (await storage.isAddressMine(address)))) {
+        continue;
+      }
+      const isLocked = this.isOutputLocked(so, { refTs: nowTs }) || isHeightLocked;
+      // token_data on a shielded output may be undefined (FullShielded hides
+      // the token); default to 0 so the authority check reads as funds.
+      accrue(so.token ?? NATIVE_TOKEN_UID, so.value, so.token_data ?? 0, isLocked, 1n);
```
r4mmer: Wait, if we default to 0 wont it wrongly count as HTR funds.

src/utils/transaction.ts
```diff
@@ -397,57 +538,100 @@ const transaction = {
...
+      let inputValue = input.value;
+      let inputTokenData = input.token_data;
+
+      if (!address || inputToken === undefined || inputValue === undefined) {
+        const spentTx = await storage.getTx(input.tx_id);
+        const resolved = spentTx ? this.resolveSpentOutput(spentTx, input.index) : undefined;
+        // Owned-marker gate: only a decrypted shielded slot (value defined) is
+        // a wallet spend we can debit; a non-owned/undecoded slot has value
+        // === undefined.
+        if (resolved?.kind !== 'shielded' || resolved.output.value === undefined) {
```
r4mmer, line 627: If the kind is not shielded we skip?
Also, maybe we should have a more clear comment on why we are skipping here, something like "We do not own this input, skipping"

src/utils/transaction.ts
```diff
@@ -397,57 +538,100 @@ const transaction = {
...
+          continue;
         }
-      } else {
-        balance[input.token].tokens.unlocked -= input.value;
+        const so = resolved.output;
+        address = so.decoded?.address;
+        inputToken = so.token ?? NATIVE_TOKEN_UID;
```
r4mmer: I do not like the default `NATIVE_TOKEN_UID`, the only place that sets the token should be where we resolve the shielded fields

src/utils/transaction.ts
```diff
@@ -719,6 +1173,31 @@ const transaction = {
...
+            const parsed = parseScript(so.script, storage.config.getNetwork());
+            return {
+              mode: so.mode,
+              commitment: so.commitment.toString('hex'),
+              range_proof: so.rangeProof.toString('hex'),
+              script: so.script.toString('hex'),
+              token_data: so.tokenData,
+              ephemeral_pubkey: so.ephemeralPubkey.toString('hex'),
+              asset_commitment: so.assetCommitment?.toString('hex'),
+              surjection_proof: so.surjectionProof?.toString('hex'),
```
r4mmer, line 1205: Why dont we use the ensureHex here?

src/utils/transaction.ts
```diff
@@ -780,6 +1260,111 @@ const transaction = {
...
+      // sum(r_in) - sum(r_out), independent of token (it lives on G).
+      // Include shielded inputs of every token so the G-term sum is
+      // correct; transparent inputs (valueBlindingFactor=0) contribute
+      // nothing to the sum but their value matters for the per-token
+      // balance the verifier checks separately.
+      for (const inp of txData.inputs) {
+        const utxo = await storage.getUtxo({ txId: inp.txId, index: inp.index });
+        if (!utxo) continue;
+        if (utxo.shielded) {
+          if (!utxo.blindingFactor) continue;
```
r4mmer, line 1310: Why this guard?

src/utils/transaction.ts
```diff
@@ -780,6 +1260,111 @@ const transaction = {
...
+      }
+
+      if (shieldedInputs.length > 0) {
+        const cryptoProvider = storage.shieldedCryptoProvider;
+        if (!cryptoProvider) {
+          throw new Error(
+            'Shielded crypto provider is not set. Cannot compute excess blinding ' +
+              'factor for a tx that spends shielded UTXOs without producing shielded outputs.'
+          );
+        }
```
r4mmer: Do you think we can put this on a helper method?
This way we have a "crypto provider getter" that throws when not available, instead of having to check everywhere.
Maybe even configure the fallback automatically if no provider has been set yet.

src/utils/storage.ts
r4mmer: Need to update this to the SEPARATED-model resolver.
We need to check `origTx.outputs.length + origTx.shieldedOutputs.length <= input.index` and we need to make sure the `origTx.shieldedOutputs` is calculated.

Maybe we need a sweep to see if there are any more of these checks we missed

r4mmer (changes requested):

src/models/shielded_output.ts
pedroferreira1: Agreed, dropped the "in PR 3" suffix; the `ShieldedOutputsHeader.deserialize` pointer stays, and the sibling reference a few lines down already cites the same symbol without a PR tag, so nothing's lost.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/6e357847c8f1381c3ab32114ca540f007704f498

src/models/transaction.ts
pedroferreira1: There's no actual drift today: shielded outputs serialize only through the `ShieldedOutputsHeader` (`serializeInputsOutputs` skips them), so `tx.shieldedOutputs` is a derived convenience field used only by `validate()`'s count check and the local-history emission — and both write sites already assign the header's exact array reference, so per-element mutations auto-sync. I like the header-backed `get shieldedOutputs()` getter (it deletes the `createFromBytes` hydration step and makes drift structurally impossible) and I'd skip the separate `_shieldedOutputs` cache (invalidation cost, no payoff).

src/utils/buffer.ts
pedroferreira1: the fullnode's `_shielded_output_to_json` uses a fixed per-field encoding — commitment / ephemeral_pubkey / asset_commitment as hex, range_proof / script / surjection_proof as base64 — the same on the HTTP and websocket paths, so base64 is the only non-hex form we can receive. Your comment also caught a doc bug: the docstring framed it as transport-based; corrected to per-field.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/b1ed98fd7b8469910805f28ce528d5ba4657b549

src/utils/transaction.ts
pedroferreira1: Agreed, the body only reads `.type` and returns a plain boolean, so `Record<string, any>` (+ the `no-explicit-any` disable) bought nothing. Tightened the param to `{ type?: string } | null | undefined` and dropped the cast and disable.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/c9bc8ddbd86f17842649d3aad85a11cc0f0e2a74

src/utils/transaction.ts
pedroferreira1: Hex is the canonical internal form for these fields: consumers read them as hex (`Buffer.from(so.script, 'hex')`), the history model stores them hex, and serialization emits hex, so normalizing the base64 wire payload to hex once keeps a single representation across the rewind / decryption / serialization paths. (Same per-field encoding detail as the `ensureHex` thread.)

src/utils/transaction.ts
pedroferreira1: For now we are using the same way of the legacy derivation. I have a note here to revisit all of them before the final merge. I agree with you, we can start using the compliant derivation for new keys. 

I will probably open a PR later with this change

src/utils/transaction.ts
pedroferreira1: The skip is correct: this block runs only for "bare" inputs (no inline value/token), which the fullnode emits only for shielded inputs, transparent inputs carry inline fields and take the earlier path. So we skip anything that isn't an owned, decrypted shielded slot. Reworded the comment to say exactly that.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/8d028b625e188490ae063929f3aa18c21da7d055

src/utils/transaction.ts
pedroferreira1: Agreed — removed the fallback. An owned (decrypted) shielded slot always carries its token (and the debit already asserts non-null), so the default was dead and would mislabel a custom token as HTR. `inputToken` now comes straight from the resolved slot.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/8d028b625e188490ae063929f3aa18c21da7d055

src/utils/transaction.ts
pedroferreira1: Those are Buffers (`ShieldedOutput` model fields), so `Buffer.toString('hex')` is the direct, correct encoding. `ensureHex` takes a string and auto-detects base64-vs-hex,  it's for normalizing wire payloads, not model Buffers, so it doesn't apply here.

src/utils/transaction.ts
pedroferreira1: This is the full-unshield `excess = Σr_in − Σr_out` loop; a shielded input contributes its value blinding factor `r_in`, which only exists if the wallet decrypted/owns the output. A shielded UTXO with no `blindingFactor` is one we can't supply `r_in` for, so we skip it. Happy to add a one-line comment there.

src/utils/storage.ts
pedroferreira1: Good catch — real bug. `input.index` is absolute in the SEPARATED model, so the positional `outputs[input.index]` check wrongly rejects a shielded input (`index >= outputs.length`). Fixed to use `resolveSpentOutput`, and swept storage.ts — the same positional read was in `processUtxoUnlock` (line 954), fixed there too. Shielded inputs/UTXOs are skipped in these transparent-spend paths (their lifecycle is the receive pipeline's).

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/c81b7693f3b4743a953cab049401cd994c6df8da

src/utils/transaction.ts
pedroferreira1: You're right — that was the same dead fallback as C9, just on the credit (output) side. An owned, decrypted shielded slot always carries its token: decryption recovers value and tokenUid together (IDecryptedShieldedOutput declares both required), and the `value !== undefined` ownership gate already ran above. So `so.token ?? NATIVE_TOKEN_UID` was dead and would mislabel a custom token as HTR. Removed it — reads `so.token` directly now, mirroring the input-debit path. Kept `token_data ?? 0`: token_data is legitimately optional even for owned slots (FullShielded hides it) and only feeds the authority check.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/efb2e7a05bfd6e3c99b8dbe3e13d53a140f0b565

src/utils/transaction.ts
pedroferreira1: Done — added `Storage.getShieldedCryptoProvider()` (returns the provider or throws) and used it at the call site. The check turned out to be a single site (transaction.ts:1319), not "everywhere", but the helper still reads better. No auto-fallback: a missing provider is a setup error and fails loudly.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/d43c25be5ff0d47dd359d135328c1160739bc08e

pedroferreira1:
@coderabbitai do a full re-review

src/models/transaction.ts
r4mmer: So, will we create the header-backed getter?

src/utils/transaction.ts
r4mmer: Should we create an issue to track this?

r4mmer (dismissed):

src/storage/storage.ts
```diff
@@ -180,6 +180,21 @@ export class Storage implements IStorage {
...
   }
 
+  /**
+   * Get the shielded crypto provider, or throw if it has not been configured.
+   * Confidential-transaction code paths require the provider; a missing one is a
+   * setup error and must fail loudly rather than silently degrade.
+   */
+  getShieldedCryptoProvider(): IShieldedCryptoProvider {
+    if (!this.shieldedCryptoProvider) {
+      throw new Error(
```
tuliomir, line 190: suggestion(test): It would be nice to have full unit test line-coverage on these isolated functions. They help reduce the amount of edge cases we have to think of on the integration tests on later PRs.

src/utils/storage.ts
```diff
@@ -434,11 +434,21 @@ export async function processSingleTx(
       continue;
     }
 
-    if (origTx.outputs.length <= input.index) {
+    // SEPARATED model: `input.index` is an absolute on-chain index spanning the
+    // transparent outputs then the shielded outputs, so resolve it via the
+    // arithmetic resolver rather than a positional `outputs[index]` read — which
+    // would wrongly reject (or misread) a shielded input whose index is
+    // >= outputs.length.
+    const resolved = transactionUtils.resolveSpentOutput(origTx, input.index);
```
tuliomir, line 442: suggestion(test): Same with those util functions. They can be unit tested for the edge cases.

If we decide to leave the validation entirely to the integration tests on later PRs it's fine, but I understand this higher unit coverage will help us later.

tuliomir (dismissed):

src/utils/transaction.ts
pedroferreira1: Better than that, I'm creating a PR with the changes, just going to review today. I will probably ask for review Monday

src/storage/storage.ts
pedroferreira1: Added unit tests for getShieldedCryptoProvider: returns the provider when one is set, and throws (never silently defaults) when it's unset or explicitly cleared.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/500fe7706817d57d99261aeb387fa5f2fddcb076

src/utils/storage.ts
pedroferreira1: Added unit tests for processSingleTx's SEPARATED-model resolution branches: it throws "Spending an unexistent output" on an out-of-range input index, and skips a shielded input (index resolving into the shielded range) without throwing. The resolveSpentOutput arithmetic itself is already covered in transaction_shielded.test.ts (transparent / idx===T / T+S-1 / T+S / negative / non-owned-middle / S=0), so this fills in the consumer side.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/500fe7706817d57d99261aeb387fa5f2fddcb076

src/models/transaction.ts
pedroferreira1: Done — shieldedOutputs is now a header-backed getter. The ShieldedOutputsHeader owns the array and the getter reads it directly, so there's a single source of truth: mutating the returned array updates the header, drift is structurally impossible, and the createFromBytes hydration step is gone. validate()'s count check and the local-history emission read through it unchanged.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1108/changes/6f368a194011a78a11e02fd7750022b7840fbf46

src/models/transaction.ts
```diff
@@ -133,6 +130,19 @@ class Transaction {
...
+   * ShieldedOutputsHeader so there is a single source of truth: the header owns
+   * the array, mutating the returned array updates the header, and there is no
+   * separate cached field to drift. Empty when the tx carries no shielded
+   * outputs (no header). The header is produced by
+   * `transactionUtils._attachShieldedOutputsHeader`.
+   */
+  get shieldedOutputs(): ShieldedOutput[] {
+    const header = this.headers.find(h => h instanceof ShieldedOutputsHeader);
+    return header instanceof ShieldedOutputsHeader ? header.shieldedOutputs : [];
+  }
```
r4mmer, line 160: > the header owns the array, mutating the returned array updates the header

Well, the only case where the header does not own the array is when there is no header, we return an empty array, owned by noone.

Should we return null/undefined if there is no header?

tuliomir (dismissed):

r4mmer (dismissed):

src/models/transaction.ts
pedroferreira1: The original shieldedOutputs property was always an array ([] default). Returning undefined changes that contract and forces null-checks at every read site (validate()'s .length, the local-history .map()) for no clear benefit.

r4mmer (approved):

tuliomir (approved):

PR #1110, opened 2026-06-18
refactor(tokens): integer math for token deposit/withdraw

raul-oliveira:
## Motivation

Mirrors hathor-core [PR #1711](https://github.com/HathorNetwork/hathor-core/pull/1711) on the client side.

HTR deposit/withdraw for deposit-based tokens applied the deposit percentage with float math, which loses precision above the JS safe-integer boundary (2^53) and produces wrong values for large amounts still within `MAX_OUTPUT_VALUE` (2^63). This PR uses integer math instead, with the percentage as a parts-per-billion fraction (`10^7 / 10^9 = 1%`) matching hathor-core. No behavior change for valid configurations.

### Acceptance Criteria

- Deposit/withdraw calculations are exact for amounts above 2^53 (within `MAX_OUTPUT_VALUE`).
- No behavior change for valid existing configurations.
- Backward compatible with fullnodes that don't expose the new `/version` fields.

## Follow-up (not in this PR)

End-to-end integration test against a fullnode image with #1711 is deferred to #1117: published images with #1711 are 85 commits ahead of the pinned baseline and need a privnet config migration first.

### Security Checklist

## Release Notes

* **Bug Fixes**
  * Corrected token deposit/withdraw fee calculations to use exact bigint fraction math, avoiding floating-point precision issues and ensuring correct deposit ceiling vs withdrawal flooring.
* **New Features**
  * Added integer-precision token deposit percentage support via numerator/denominator with legacy fallback.
  * Exposed wallet facade methods to compute deposit/withdraw amounts using the stored fraction.
  * Extended version data and API schemas to optionally provide the fraction fields (legacy field is deprecated).
* **Tests**
  * Added edge-case coverage for negative inputs and large bigint rounding behavior.


src/storage/storage.ts
```diff
@@ -205,6 +209,37 @@ export class Storage implements IStorage {
...
+   * that predate these fields), it falls back to deriving the numerator from the deprecated
+   * float percentage, keeping the standard 1e9 denominator, and ultimately to the defaults.
+   *
+   * @returns {{ numerator: bigint; denominator: bigint }}
+   */
+  getTokenDepositPercentageFraction(): { numerator: bigint; denominator: bigint } {
+    const numerator = this.version?.token_deposit_percentage_numerator;
+    const denominator = this.version?.token_deposit_percentage_denominator;
+    if (numerator != null && denominator != null) {
+      return { numerator: BigInt(numerator), denominator: BigInt(denominator) };
```
tuliomir, line 241: chore: we should have all flows this function tested in isolation by a unit test suite

src/utils/bigint.ts
```diff
@@ -66,6 +66,18 @@ export const JSONBigInt = {
...
 };
 
+/**
+ * Ceil division using integer math for non-negative operands, equivalent to `ceil(a / b)`.
+ *
+ * Avoids the precision loss of `Math.ceil(a / b)` with floats for large values.
+ */
+export function ceilDiv(a: bigint, b: bigint): bigint {
+  if (a < 0n || b <= 0n) {
+    throw new Error('ceilDiv expects a non-negative dividend and a positive divisor');
```
tuliomir, line 76: chore: we should unit test this error handler

raul-oliveira: Added a `ceilDiv` suite covering the guard (negative dividend, zero/negative divisor) plus zero-dividend, exact, and round-up cases.

src/storage/storage.ts
raul-oliveira: Added a `getTokenDepositPercentageFraction` suite covering every branch: fields preferred when both present, derivation from the float percentage (including when only one field is set), and the default fallback for unset/deposit-less versions.

src/template/transaction/executor.ts
```diff
@@ -626,8 +626,10 @@ export async function execCompleteTxInstruction(
     let deposit = 0n;
 
     if (ctx.tokenVersion === TokenVersion.DEPOSIT) {
-      const amount = ctx.balance.createdTokenBalance!.tokens;
-      deposit = interpreter.getHTRDeposit(amount);
+      // createdTokenBalance is negative, so we use its absolute value.
```
tuliomir, line 629: question(non-blocking): Should we really mutate the input to match our validation rules? Wouldn't it be better to throw instead?

tuliomir (dismissed):

src/wallet/api/schemas/walletApi.ts
```diff
@@ -237,6 +237,8 @@ export const fullNodeVersionDataSchema = z
     minTxWeightCoefficient: z.number(),
     minTxWeightK: z.number(),
     tokenDepositPercentage: z.number(),
```
pedroferreira1: This is deprecated and should be removed in the next node versions. We should prepare the wallet for it already, otherwise we would need a new update to stop requiring this field

src/wallet/types.ts
```diff
@@ -642,7 +642,12 @@ export interface FullNodeVersionData {
   minTxWeight: number;
   minTxWeightCoefficient: number;
   minTxWeightK: number;
+  /** @deprecated Use the numerator/denominator fraction below, which keeps integer precision. */
   tokenDepositPercentage: number;
```
pedroferreira1: Same comment as in the schema

src/types.ts
```diff
@@ -585,7 +585,12 @@ export interface ApiVersion {
   min_tx_weight: number;
   min_tx_weight_coefficient: number;
   min_tx_weight_k: number;
+  /** @deprecated Use the numerator/denominator fraction below, which keeps integer precision. */
   token_deposit_percentage: number;
```
pedroferreira1: Same comment as in the others

src/utils/tokens.ts
```diff
@@ -299,25 +301,28 @@ const tokens = {
...
+    depositDenominator: bigint = TOKEN_DEPOSIT_PERCENTAGE_DENOMINATOR
   ): OutputValueType {
-    // This conversion from mintAmount to Number may cause loss of precision for large amounts,
-    // but this is fully equivalent to the reference Python implementation, which does the same.
-    // It'll never be a problem for mainnet as no values can reach the precision boundary, but
-    // it may happen in custom networks.
-    return BigInt(Math.ceil(depositPercent * Number(mintAmount)));
+    // Coerce to bigint so callers passing a number keep working (the API previously used Number()).
+    const amount = BigInt(mintAmount);
+    if (mintAmount < 0) {
```
pedroferreira1, line 322:
```suggestion
    if (mintAmount <= 0) {
```

pedroferreira1 (commented):
This is a PR that adds breaking changes to the wallets.

The wallets desktop and mobile (maybe others as well) use the method `getDepositAmount` and `getWithdrawAmount` that changed their parameters.

If that's really the case, we need to merge this carefully and think about the clients updates.

Also, this reflects a bad design decision. I prefer having a method in the wallet facades `wallet.getDepositAmount` and `wallet.getWithdrawAmount` that will get the percentage parameters from `this` correctly and then call the utils method. The interface methods with the clients don't need parameter and will never change, so this breaking change wouldn't happen if it had been designed like this.

Maybe we should add these facade methods in this PR already to improve this. What do you think?

src/utils/tokens.ts
pedroferreira1: Or should it allow 0? If it should, we must just update the error message below

src/template/transaction/executor.ts
raul-oliveira: is this case, is needed for the calculations inside the builder. We use positives for inputs and negatives for outputs, so it's a rule that is already derived inside the executor.

src/wallet/api/schemas/walletApi.ts
raul-oliveira: Done — `tokenDepositPercentage` is now `z.number().optional()` + `@deprecated`, so validation survives fullnodes dropping it.

src/wallet/types.ts
raul-oliveira: Same fix on `FullNodeVersionData`: `tokenDepositPercentage?` optional + `@deprecated`.

src/types.ts
raul-oliveira: Same fix on `ApiVersion`: `token_deposit_percentage?` optional + `@deprecated`. Readers already fall back to the defaults, so nothing breaks.

src/utils/tokens.ts
raul-oliveira: Now allows 0 (`getDepositAmount(0n) === 0n`); only negatives throw, with the message updated to `mint amount should not be negative`.

raul-oliveira:
@pedroferreira1 agreed. Added `getDepositAmount(mintAmount)` / `getWithdrawAmount(meltAmount)` to both facades (`HathorWallet`, `HathorWalletServiceWallet`) and `IHathorWallet`. They pull the fraction from the wallet's `/version` storage and delegate to the utils, so the client signature takes only the amount and won't break on future percentage-format changes. Sync; they throw `WalletError('Wallet not ready')` instead of falling back to the 1% default. The internal deposit/withdraw sites now route through them.

Follow-up: migrate desktop/mobile off the direct `tokensUtils.getDepositAmount(amount, percent)` calls to the new facade methods.

pedroferreira1 (approved):

__tests__/storage/storage.test.ts
```diff
@@ -216,6 +218,65 @@ describe('config version', () => {
...
+        token_deposit_percentage_numerator: 3,
+        token_deposit_percentage_denominator: 100,
+      } as ApiVersion);
+      expect(storage.getTokenDepositPercentageFraction()).toEqual({
+        numerator: 3n,
+        denominator: 100n,
+      });
+    });
+
+    it('should derive the fraction from the float percentage when the fields are absent', async () => {
```
tuliomir, line 236: suggestion(non-blocking): Add a new test to prove that all works after the migration is completed and the `token_deposit_percentage` field is no longer offered by the backend.

tuliomir (approved):

PR #1111, opened 2026-06-19
feat(itest): provision integration wallets from the hathor-integration-test-helper

tuliomir:
Integrates the wallet-lib integration test suite with the dockerized **[hathor-integration-test-helper (ITH)](https://github.com/HathorNetwork/hathor-integration-test-helper)** service.

- **`feat(itest)`** — Replaces the static `precalculated-wallets.json` pool with on-demand `GET /simpleWallet` calls to the ITH `wallet-provider` service, so the pool can never run out. `WalletPrecalculationHelper.getPrecalculatedWallet` becomes async and the awaits propagate through `buildWalletInstance`/adapters to every call site (behavior-preserving).
- **`ci(itest)`** — Uses the published `hathornetwork/hathor-integration-test-helper` image in docker-compose instead of building from git — faster and more reproducible in CI.

### Acceptance Criteria
- Integration tests fetch precalculated wallets on demand from the dockerized ITH `wallet-provider` service instead of the static JSON pool.
- The wallet pool can no longer be exhausted across a test run.

### Security Checklist

## Summary by CodeRabbit

* **Tests**
  * Updated the integration test infrastructure to fetch precalculated wallets dynamically from a wallet-provider service (via HTTP) instead of relying on static local files.
  * Converted wallet construction and precalculation-related test flows to fully support async/await, ensuring proper initialization and data loading order across all integration suites.
* **Chores**
  * Adjusted integration Docker/test scripts to remove tmp precalculated-wallet file handling and align startup/teardown with the new provider-based approach.


__tests__/integration/configuration/docker-compose.yml
```diff
@@ -233,5 +233,19 @@ services:
...
+  # the image, matching the old static set.
+  # https://github.com/HathorNetwork/hathor-integration-test-helper
+  wallet-provider:
+    image:
+      ${HATHOR_LIB_INTEGRATION_TESTS_WALLET_PROVIDER_IMAGE:-hathornetwork/hathor-integration-test-helper:latest}
+    ports:
+      - "3020:3020"
+    networks:
+      - hathor-privnet
+    # TODO: add a healthcheck once the ITH exposes a readiness endpoint.
```
pedroferreira1, line 248: Do you have this on track? Should be easy to create a simple PR with it

__tests__/integration/fullnode-specific/start.test.ts
```diff
@@ -65,8 +65,8 @@ describe('[Fullnode-specific] start', () => {
     await adapter.stopAllWallets();
   });
 
-  it('should reject with invalid constructor parameters', () => {
-    const walletData = precalculationHelpers.test!.getPrecalculatedWallet();
+  it('should reject with invalid constructor parameters', async () => {
+    const walletData = await precalculationHelpers.test!.getPrecalculatedWallet();
```
pedroferreira1, line 69: Why don't you use `getPrecalculatedWallet` from the adapter? It has exactly this same code

The same for the other places in this file with this code

__tests__/integration/helpers/wallet-precalculation.helper.ts
```diff
@@ -246,31 +201,19 @@ export class WalletPrecalculationHelper {
...
+  async getPrecalculatedWallet(): Promise<PrecalculatedWalletData> {
+    const { data } = await axios.get(`${testConfig.walletProviderUrl}/simpleWallet`);
+    // Fail loudly with the raw payload if it ever returns an unexpected shape — otherwise a malformed
+    // response surfaces as a confusing error deep inside wallet construction.
+    if (!data?.words || !Array.isArray(data?.addresses)) {
+      throw new Error(`Wallet provider returned an unexpected response: ${JSON.stringify(data)}`);
     }
-    unusedWallet.isUsed = true; // We are using it right now. Marking it.
-    return unusedWallet;
+    return { isUsed: false, words: data.words, addresses: data.addresses };
```
pedroferreira1: Why did you change `isUsed` from `true` to `false`?

__tests__/integration/configuration/docker-compose.yml
tuliomir: Yes we do! The only reason it's not yet available is because I decided to ship the first version of the ITH early, with only a subset of the planned features ( healthcheck included ).

This is so that we can start the integration piece by piece and have a clear understanding of any unforeseen impacts it may have on the lib CI.

__tests__/integration/fullnode-specific/start.test.ts
tuliomir: The concept of the adapter is to be used only by the `shared` tests, as a kind of additional facade that allows small differences in the contracts of the `fullnode` and `wallet service` facades not to cause any issues in a shared test suite.

Once we're in the `*-specific` folders it's important to use the original facades directly and not use the adapter, so we can keep the adapter lean from methods not used in the shared tests and also the tests themselves are more "real", since they're not reaching any abstractions.

Having an adapter call here could confuse future developers/agents thinking they should test the adapter, so it's best not to have it even on the import list of this file.

__tests__/integration/helpers/wallet-precalculation.helper.ts
tuliomir: Good catch. Although this will have no impact anywhere in the code, it's conceptually wrong: this wallet is now being used. Reverted on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1111/commits/20ad1ce55c6e3fe2d2a3f4ea891988bbac17ffbc

This will soon be made irrelevant when we create a whole service file to interact with the ITH , but it's good to have the code consistent at all times.

__tests__/integration/helpers/wallet-precalculation.helper.ts
```diff
@@ -246,31 +201,19 @@ export class WalletPrecalculationHelper {
...
+   * about a pool running out.
+   * @returns {Promise<PrecalculatedWalletData>}
    */
-  getPrecalculatedWallet(): PrecalculatedWalletData {
-    const unusedWallet = this.walletsDb.find(w => !w.isUsed);
-    if (!unusedWallet) {
-      throw new Error('No unused wallet found');
+  // eslint-disable-next-line class-methods-use-this -- kept as an instance method to preserve the precalculationHelpers.test call sites
+  async getPrecalculatedWallet(): Promise<PrecalculatedWalletData> {
+    const { data } = await axios.get(`${testConfig.walletProviderUrl}/simpleWallet`);
```
r4mmer, line 211: suggestion(non-blocking): Add a timeout and retry strategy to this request.

Since this is a call dependent on another service maybe we should have some logs if any issues like a timeout on the call or continuous retries are slowing us down.

This may save some debug time later if we have any issues with the ITH service infra.

__tests__/integration/helpers/wallet-precalculation.helper.ts
```diff
@@ -246,31 +201,19 @@ export class WalletPrecalculationHelper {
...
+    // Fail loudly with the raw payload if it ever returns an unexpected shape — otherwise a malformed
+    // response surfaces as a confusing error deep inside wallet construction.
+    if (!data?.words || !Array.isArray(data?.addresses)) {
+      throw new Error(`Wallet provider returned an unexpected response: ${JSON.stringify(data)}`);
     }
-    unusedWallet.isUsed = true; // We are using it right now. Marking it.
-    return unusedWallet;
+    return { isUsed: true, words: data.words, addresses: data.addresses };
```
r4mmer, line 217: suggestion(non-blocking): Can we remove the `isUsed` property?

It would seem that `isUsed` is now a vestigial property with no real usage (and currently hardcoded), should we remove it?

r4mmer (approved):

__tests__/integration/helpers/wallet-precalculation.helper.ts
tuliomir: Yes we should!

My objective with this PR was to have as little a footprint as possible, but my next one will have a new `ithService` that should contain the updated/simplified logic and the removal of `wallet-precalculation` entirely.

__tests__/integration/helpers/wallet-precalculation.helper.ts
tuliomir: I agree. This service should be always available and extremely fast. Anything other than this represents that it is probably offline as a whole.

But as good practice we should have timeouts and retries. I'll add them when I build the `ithService` file, as a default to all requests made there.

pedroferreira1 (approved):

PR #1116, opened 2026-06-23
feat(shielded): receive pipeline — decrypt, persist & UTXO lifecycle

pedroferreira1:
## Description                                                                                                                                                              
                                                                                                                                                                              
   The receive side of the separated-output model. It detects, decrypts, and accounts for wallet-owned shielded outputs so they flow through the same balance / UTXO / history / address-gap bookkeeping as transparent outputs — no parallel shielded accounting path.

In the SEPARATED model the fullnode delivers transparent outputs in tx.outputs[] and shielded outputs in a parallel tx.shielded_outputs[]; the on-chain absolute index of shielded slot s is tx.outputs.length + s. A slot is wallet-owned only once ECDH decryption has written its decoded marker (value !== undefined) — that single gate drives every downstream decision. Nothing is appended into tx.outputs[] and no index is synthesized.

Targets the alpha-v4 node, which delivers shielded outputs separated on the wire with spent_by stamped identically to transparent outputs — so shielded spent_by is handled exactly like transparent throughout, with no v3-era special-casing.                                                                                            
                                                                                                                                                                                                                                                   
  ## Acceptance Criteria                                                                                                                                                      
  - Add processShieldedOutputs â a shielded receive pipeline that detects wallet-owned shielded outputs, ECDH-decrypts them, and writes the decoded fields back in place onto each shielded_outputs[] slot.
- Route owned shielded outputs through the same balance / UTXO / address-metadata / lock bookkeeping as transparent outputs (shared creditOutput), gated on the single value !== undefined ownership marker.
- Recover the FullShielded token UID and cross-check it against the on-chain asset commitment, rejecting a mismatch.
- Track the shielded address chain's gap limit independently from the legacy P2PKH chain (shieldedMaxIndexUsed), deriving the shielded + spend-P2PKH pair per index.
- Preserve shielded: true + blinding factors on confirmed shielded UTXOs across metadata events, via a save-shape shared with creditOutput (shieldedUtxoSaveFields).
- Handle shielded spent_by identically to transparent for the alpha-v4 wire (drop the v3 null-coercion and the processHistory deletion loop).
- Resolve shielded spent/locked outputs by absolute index T+s via resolveSpentOutput, not positional outputs[] reads.
- Decrypt the scan xpriv once per tx instead of once per owned output.
- Thread pinCode through processHistory/processNewTx/IStorage so history (re)processing can derive the scan key.
- Add unit coverage for the pipeline and storage paths (processing.test.ts, storage.test.ts, incl. dual-chain checkGapLimit).


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * History/transaction syncing now supports an optional PIN to decrypt protected wallet-owned shielded outputs.
  * Shielded output processing now decodes and reports decrypted fields directly during sync.
* **Bug Fixes**
  * Corrected shielded token UID resolution, improved skipping when required metadata is missing, and stricter validation for fully shielded outputs.
  * Fixed shielded/legacy index and UTXO spend handling, plus improved dual-chain gap-limit loading.
* **Tests**
  * Expanded Jest coverage for shielded decoding, storage processing, PIN-gated scenarios, and cross-check failures.
  * Improved WebSocket stream test stability.
* **Chores**
  * Increased integration test timeout to reduce flakiness.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


src/utils/transaction.ts
```diff
@@ -1554,7 +1554,6 @@ const transaction = {
       ephemeral_pubkey: ensureHex(so.ephemeral_pubkey),
       asset_commitment: so.asset_commitment ? ensureHex(so.asset_commitment) : undefined,
       surjection_proof: so.surjection_proof ? ensureHex(so.surjection_proof) : undefined,
-      spent_by: so.spent_by ?? null,
```
pedroferreira1, line 1557: The full node now sends `spent_by` in all API responses, so the line above `...so` already copies this value

pedroferreira1:
@coderabbitai I've just finished lot's of changes in this PR. Please do a full re-review

src/shielded/processing.ts
```diff
@@ -0,0 +1,282 @@
...
+} from './types';
+
+/**
+ * Resolve the token UID for a shielded output.
+ *
+ * Uses the same token_data convention as transparent outputs (via getTokenIndexFromData).
+ * Returns the 32-byte hex UID needed by the crypto layer (NATIVE_TOKEN_UID_HEX for HTR).
+ * For FullShielded outputs, the token is unknown until decrypted.
+ */
+export function resolveTokenUid(shieldedOutput: IShieldedOutput, tx: IHistoryTx): string {
```
r4mmer: Instead of having a `resolveTokenUid` only for shielded outputs, why not have one that works with all outputs?
Since we exclude fully shielded here it basically already should work with all outputs.

src/shielded/processing.ts
```diff
@@ -0,0 +1,282 @@
...
+function deriveScanChildPrivkey(
+  scanHdPrivKey: HDPrivateKey,
+  addressIndex: number,
+  logger: ILogger
+): Buffer | undefined {
+  try {
+    // deriveNonCompliantChild is required for private keys due to a historical
+    // bitcore-lib serialization bug. Public key derivation (in shieldedAddress.ts)
+    // uses standard deriveChild because it was always correct. Do not align these.
+    const childKey = scanHdPrivKey.deriveNonCompliantChild(addressIndex);
```
r4mmer, line 62: I still think we should not use `deriveNonCompliantChild` for the new addresses.

src/shielded/processing.ts
```diff
@@ -0,0 +1,282 @@
...
+  // first owned shielded output). The PBKDF2 unlock + HD construction are the
+  // expensive part; only the cheap per-address child derivation runs per output.
+  // Every wallet reloads its history from block 0, so this hot path walks every
+  // shielded tx on first sync.
+  let scanHdPrivKey: HDPrivateKey | undefined;
+
+  for (const [sIndex, shieldedOutput] of shieldedOutputs.entries()) {
+    const absoluteIndex = transparentCount + sIndex;
+    const address = shieldedOutput.decoded?.address;
+    if (!address) continue;
```
r4mmer, line 127: suggestion(nitpick): Add comment explaining what not having as address means here.

Usually not having an address means the output is not _to_ an address or does not have a canonical script.
The problem is that the fullnode has special validation for shielded addresses, I think that we _currently_ do not allow non-P2PKH scripts on a shielded output (_needs verification_)

src/shielded/processing.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    // Check if this address belongs to our wallet
+    const addressInfo = await storage.getAddressInfo(address);
+    if (!addressInfo) continue;
```
r4mmer: question(nitpick): Should we use `.isAddressMine()`?

src/shielded/processing.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    if (!scanHdPrivKey) {
+      try {
+        scanHdPrivKey = new HDPrivateKey(await storage.getScanXPrivKey(pinCode));
+      } catch (e) {
+        storage.logger.warn('Failed to unlock scan xpriv for shielded outputs', e);
+        continue;
```
r4mmer: question(minor): Should we continue?

Since we have a shielded output but we failed to fetch the scan private key it means the wallet balance and history will be wrong or incomplete, should we actually fail instead of trying to power through this error?

src/shielded/processing.ts
```diff
@@ -0,0 +1,282 @@
...
+          // on-chain assetCommitment hex so the failure is debuggable from
+          // logs alone.
+          storage.logger.error(
+            `FullShielded token UID cross-check failed for tx ${tx.tx_id} ` +
+              `output ${absoluteIndex} — asset commitment mismatch. ` +
+              `recovered tokenUid=${recoveredTokenUid}, ` +
+              `on-chain assetCommitment=${assetCommitment.toString('hex')}, ` +
+              `expected assetCommitment=${expectedAc.toString('hex')}`
+          );
+          continue;
```
r4mmer, line 204: Again, should we continue on errors?

src/shielded/processing.ts
```diff
@@ -0,0 +1,275 @@
...
+    // Derive this address's scan private key (ECDH) from the cached HD key.
+    // isAddressMine above guarantees the address is in storage, so getAddressInfo
+    // is non-null; we fetch it here only for the bip32 derivation index.
+    const addressInfo = await storage.getAddressInfo(address);
+    const privkey = deriveScanChildPrivkey(
+      scanHdPrivKey,
+      addressInfo!.bip32AddressIndex,
+      storage.logger
+    );
+    if (!privkey) continue;
```
tuliomir, line 151: question(non-blocking): Is this a defensive  clause too? If the address is ours, how could we not derive the scan child privkey? Maybe add a comment like you did on the `if (!address)` above.

src/utils/storage.ts
```diff
@@ -92,19 +98,39 @@ export async function loadAddresses(
...
-    if ((await storage.getWalletType()) === 'p2pkh') {
-      address = await deriveAddressP2PKH(i, storage);
     } else {
-      address = await deriveAddressP2SH(i, storage);
+      // derive legacy address at index i
+      let address: IAddressInfo;
+      if ((await storage.getWalletType()) === 'p2pkh') {
+        address = await deriveAddressP2PKH(i, storage);
+      } else {
+        address = await deriveAddressP2SH(i, storage);
```
tuliomir, line 109: question(non-blocking): Don't. we have any tests for P2SH legacy addresses? This is not related to this PR, but we surely should have an issue to this if this test is missing.

src/utils/storage.ts
```diff
@@ -158,7 +185,7 @@ export async function apiSyncHistory(
     itCount = loadMoreAddresses.count;
   }
   if (foundAnyTx && shouldProcessHistory) {
-    await storage.processHistory();
+    await storage.processHistory(pinCode);
```
tuliomir, line 188: chore: The lack of tests in such a central method indicate a possible bug on the test suites.

src/utils/storage.ts
```diff
@@ -364,14 +396,43 @@ export async function checkGapLimit(storage: IStorage): Promise<IScanPolicyLoadA
...
+  // Check both the legacy and shielded chains independently. loadAddresses
+  // derives a legacy + shielded pair at each BIP32 index, so we extend
+  // whichever chain is furthest behind its gap. Omitting the shielded chain
+  // here silently caps owned shielded-address discovery at the legacy chain's
+  // loaded range, so a wallet receiving on shielded indexes beyond the
+  // legacy-used range would never load (and thus never decrypt) them.
+  const legacyNeedMore = lastUsedAddressIndex + gapLimit > lastLoadedAddressIndex;
+  // Only check the shielded gap when shielded keys are available.
+  const hasShieldedKeys = !!(await storage.getAccessData())?.spendXpubkey;
+  const shieldedNeedMore =
```
tuliomir, line 409: thought: This effectively doubles the amount of listening channels we'll have with the fullnode, right?

Something to bear in mind, even though we're increasing the limit of subscriptions on the fullnode side.

src/utils/storage.ts
```diff
@@ -434,50 +524,56 @@ export async function processSingleTx(
...
+    // absolute on-chain index {input.tx_id, input.index}, the same key both are
+    // saved under. The store only holds UTXOs this wallet owns (every saveUtxo is
+    // gated on isAddressMine), so a non-null getUtxo IS the ownership check — and
+    // it deletes the real stored record rather than one reconstructed from the
+    // output. If the output isn't ours (or was already removed) getUtxo returns
+    // undefined and there is nothing to delete. Removing it stops the selector
+    // from re-offering a spent UTXO ("input already spent" on the next send).
+    const spentUtxo = await storage.getUtxo({ txId: input.tx_id, index: input.index });
+    if (spentUtxo) {
+      await store.deleteUtxo(spentUtxo);
```
tuliomir, line 548: chore: Again, the lack of tests in this important branch can conceal bugs.

src/shielded/processing.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/commits/11991484802cbbbea6c3f40606fa6ad33983db01

`resolveTokenUid` is now generic over any output and uses the same token_data → token-index convention as the transparent path (see hydrateIOWithToken in transaction.ts). It also fails loud on a missing token_data instead of silently resolving to the native slot, which would misattribute a custom token as HTR.

src/shielded/processing.ts
pedroferreira1: A shielded output with no decoded address is a data output or a non-canonical/non-P2PKH script, so it can't be an owned shielded output and we skip it. (Defensive: the fullnode is believed to reject non-P2PKH shielded scripts.)

Added comment https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/changes/11991484802cbbbea6c3f40606fa6ad33983db01

src/shielded/processing.ts
pedroferreira1: Agreed — switched the ownership check to storage.isAddressMine(address) in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/changes/11991484802cbbbea6c3f40606fa6ad33983db01; it reads clearer at the call site than testing getAddressInfo for null. getAddressInfo is now called only after that gate, purely to fetch the bip32 index for the child derivation.

src/shielded/processing.ts
pedroferreira1: Makes sense. The scan-xpriv unlock now propagates its error instead of being swallowed. A failure there (wrong PIN, or a missing/corrupted scan key) is systemic - it would fail for every owned shielded output â so failing loud is correct rather than silently under-counting the wallet's shielded balance.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/changes/11991484802cbbbea6c3f40606fa6ad33983db01 done

src/shielded/processing.ts
pedroferreira1: You're right, it's defensive — added a comment mirroring the if (!address) one above it https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/changes/96a49185c0e7f92e5c12a937c2b5310c60a07cd9. deriveScanChildPrivkey only returns undefined if bitcore's child derivation throws (it logs a warning internally), which shouldn't happen for a valid owned address at a valid index, so we skip rather than crash the sync.

src/utils/storage.ts
pedroferreira1: This is already covered — P2SH legacy address derivation is tested in __tests__/utils/address.test.ts ("Derive address p2sh" and "Derive p2sh address from data"), plus an end-to-end P2SH multisig fixture in stream.test.ts. Happy to add a specific case if there's one you had in mind that's missing.

src/utils/storage.ts
pedroferreira1: Added processHistory orchestration tests in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/changes/96a49185c0e7f92e5c12a937c2b5310c60a07cd9 — they assert it walks the history oldest-first (order: 'asc') and cleans stale metadata before reprocessing.

src/utils/storage.ts
pedroferreira1: Added tests for the spent-output delete path in processSingleTx https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/changes/96a49185c0e7f92e5c12a937c2b5310c60a07cd9, covering both a transparent input and a shielded input (resolved by absolute on-chain index) — each stores a UTXO, spends it, and asserts it's removed from the store.

src/utils/storage.ts
pedroferreira1: This is something that will be discussed with the business/core team for us to get to a good limit of addresses subscribed

src/shielded/processing.ts
pedroferreira1: In a future PR I added a refactor to this derivation of new keys. All of them will be normalized to use `deriveChild`. https://github.com/HathorNetwork/hathor-wallet-lib/commit/c53184abd609a3fc41cb8d58da2b9c90f93cdd79

tuliomir (dismissed):

__tests__/shielded/processing.test.ts
```diff
@@ -159,24 +160,25 @@ describe('processShieldedOutputs (SEPARATED model — write in place)', () => {
...
       getScanXPrivKey: jest.fn().mockRejectedValue(new Error('no key')),
       logger: { warn: jest.fn(), info: jest.fn(), debug: jest.fn() },
     } as any;
 
     const provider = makeMockProvider();
 
-    const result = await processShieldedOutputs(storage, tx, provider, 'pin');
-    expect(result).toEqual([]);
-    expect(storage.logger.warn).toHaveBeenCalled();
+    // C4: a scan-xpriv unlock failure is SYSTEMIC (wrong PIN / missing key), so
```
r4mmer: C4?? as in an explosive error?

src/shielded/processing.ts
pedroferreira1: Yes, here we intentionally skip-and-log rather than throw, and it's a deliberate split from the scan-key case. The rule across processShieldedOutputs is: systemic/wallet-level failures throw (e.g. the scan-xpriv unlock â a wrong PIN or missing key fails for *every* owned output, so failing loud is correct), while per-output failures on untrusted on-chain data skip that one output and log. The asset_commitment is network-supplied, so throwing would let a single malformed or forged output halt the entire wallet's history sync (which re-walks from block 0) â a per-output DoS. Skipping is safe for correctness because we never write the recovered fields, so `value` stays `undefined` and the output is excluded everywhere by the `value !== undefined` ownership gate (no bogus credit, no ghost UTXO â covered by the "reject (skip) â¦ cross-check fails" test). And it isn't silent: this branch logs at `error` with the recovered tokenUid + both commitments, since a mismatch means either forgery or a construction bug we'd want to see.

__tests__/stream.test.ts
```diff
@@ -194,12 +205,11 @@ describe('Websocket stream history sync', () => {
     } finally {
       // Stop wallet
       await wallet.stop({ cleanStorage: true, cleanAddresses: true });
-      mockServer.stop();
     }
-  }, 10000);
+  }, 30000);
```
r4mmer: suggestion(non-blocking): Move this to a constant.

Having 1 const that configures all tests would make it easier to configure or revert this later.
We could even make it a file-wide constant so only these tests you are creating/updating will use it.

src/shielded/processing.ts
```diff
@@ -131,26 +119,34 @@ export async function processShieldedOutputs(
   for (const [sIndex, shieldedOutput] of shieldedOutputs.entries()) {
     const absoluteIndex = transparentCount + sIndex;
     const address = shieldedOutput.decoded?.address;
+    // No decoded address means a data output or a non-canonical/non-P2PKH script,
+    // which can't be an owned shielded output — skip it. Shielded outputs are
+    // expected to carry a spend P2PKH address (the fullnode is believed to reject
+    // non-P2PKH shielded scripts — to confirm with core), so this is defensive.
```
r4mmer: "to confirm with core"?
Can we confirm?

r4mmer (dismissed):

__tests__/shielded/processing.test.ts
pedroferreira1: Opsss!! Comment updated: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/commits/9f6d08d5074c37244a6e341562bbf500b6769379

__tests__/stream.test.ts
pedroferreira1: Makes sense, added the constant: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/commits/9f6d08d5074c37244a6e341562bbf500b6769379

src/shielded/processing.ts
pedroferreira1: Confirmed and comment updated: https://github.com/HathorNetwork/hathor-wallet-lib/pull/1116/commits/9f6d08d5074c37244a6e341562bbf500b6769379

r4mmer (approved):

tuliomir (approved):

PR #1121, opened 2026-06-30
test(itest): migrate token-authority to shared suite

tuliomir:
## Motivation

Part of the adapter-driven shared-tests migration: token-authority coverage (`mintTokens`, `delegateAuthority`, `destroyAuthority`, and FEE-token mint) lived in the monolithic, fullnode-only `hathorwallet_facade.test.ts`. The goal is to run that same coverage against **both** the fullnode and wallet-service facades from a single shared suite.

A first migration pass was only green against the fullnode adapter — two shared tests failed on wallet-service — and it silently dropped several business-rule cases. This PR completes the migration so the full `token-authority` suite passes on both adapters (28/28) plus `mint-tokens`, with no coverage lost.

## Acceptance Criteria

- The `token-authority` and `mint-tokens` integration suites pass against **both** the fullnode and wallet-service adapters.
- No business-rule coverage is lost relative to the pre-migration `hathorwallet_facade.test.ts`.

## What changed

- **Migration**: token-authority blocks moved to `shared/token-authority.test.ts`, with the fullnode-only `data` / `unshiftData` mint cases in `fullnode-specific/mint-tokens.test.ts`. Adds `mintTokens` / `meltTokens` / `destroyAuthority` adapter methods (both facades).
- **Restored coverage** the first migration cut dropped:
  - delegating an authority the wallet does not hold is rejected;
  - melt-side delegate-without-keeping (source can no longer melt, recipient can);
  - FEE-token mint rejected when no HTR covers the flat fee;
  - FEE-token mint tx shape + flat fee header (fullnode-specific);
  - external-routed mint authority actually lands at the target address.
- **Fixes surfaced by running the suite on wallet-service**:
  - the service `delegateAuthority` adapter hung on `createAnother: false` — it waited for a source UTXO carrying the delegation txId, which only exists when an authority is kept; it now settles on the available-count drop;
  - the HTR-shortage rejection regex did not cover the wallet-service `selectUtxos` message.

## Breaking changes

None — every change is under `__tests__/`; no production code is touched.

## Summary by CodeRabbit

* **New Features**
  * Added fullnode-specific `mintTokens` integration coverage for data-script output behavior and strict flat **1 HTR** fee-header validation for fee-token mints.
* **Bug Fixes**
  * Improved integration adapter synchronization so minting, delegating, melting, and destroying authority now waits for wallet-side propagation; supports cross-wallet settlement when a recipient wallet is specified.
* **Tests**
  * Expanded shared authority lifecycle tests across mint/delegate/destroy for both adapter facades, including consistent “insufficient HTR” error matching and cross-wallet assertions.
  * Reorganized integration test placement between shared and fullnode-specific suites.


__tests__/integration/adapters/service.adapter.ts
```diff
@@ -393,27 +402,166 @@ export class ServiceWalletTestAdapter implements IWalletTestAdapter {
...
+    return { hash: result.hash };
+  }
+
+  async mintTokens(
+    wallet: FuzzyWalletType,
+    tokenUid: string,
+    amount: bigint,
+    options?: MintTokensAdapterOptions
+  ): Promise<MintTokensResult> {
+    const sw = this.concrete(wallet);
```
raul-oliveira: question: meaning of sw?

__tests__/integration/fullnode-specific/mint-tokens.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import { TOKEN_MINT_MASK, TOKEN_AUTHORITY_MASK } from '../../../src/constants';
+import { TokenVersion } from '../../../src/types';
+import FeeHeader from '../../../src/headers/fee';
+import Header from '../../../src/headers/base';
+
+// `script` of a 'foobar' data output: [len(6), f, o, o, b, a, r, OP_CHECKSIG(172)]
+const FOOBAR_DATA_SCRIPT = Buffer.from([6, 102, 111, 111, 98, 97, 114, 172]);
+
+/** Asserts the tx carries exactly one fee header with a single entry of `expectedFee`. */
+function validateFeeAmount(headers: Header[], expectedFee: bigint) {
```
raul-oliveira, line 41: If I'm not wrong we already have a method exactly like this, does make sense to have a utils for the tests?

__tests__/integration/adapters/service.adapter.ts
tuliomir: Thanks for the flag. Fixed on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1121/commits/a8057233facb92b7cbfad1fa0f7208e0b1350fda

__tests__/integration/adapters/service.adapter.ts
```diff
@@ -456,27 +425,180 @@ export class ServiceWalletTestAdapter implements IWalletTestAdapter {
...
+
+  async destroyAuthority(
+    wallet: FuzzyWalletType,
+    tokenUid: string,
+    type: AuthorityType,
+    count: number
+  ): Promise<DestroyAuthorityResult> {
+    const serviceWallet = this.concrete(wallet);
+
+    // Capture the current authority count so we can wait for the UTXO index to
```
raul-oliveira, line 588: suggestion(non-blocker):
What about checking the balance here too?
I mean, verifying that the balance remains the same even after destroying.

__tests__/integration/shared/token-authority.test.ts
```diff
@@ -0,0 +1,478 @@
...
+import { FullnodeWalletTestAdapter } from '../adapters/fullnode.adapter';
+import { ServiceWalletTestAdapter } from '../adapters/service.adapter';
+
+const adapters: IWalletTestAdapter[] = [
+  new FullnodeWalletTestAdapter(),
+  new ServiceWalletTestAdapter(),
+];
+
+const htrBalance = (wallet: FuzzyWalletType): Promise<bigint> =>
+  wallet.getBalance(NATIVE_TOKEN_UID).then(b => b[0].balance.unlocked);
```
raul-oliveira, line 44: suggestion: reuse the tokenBalance method below here too.

```js
const htrBalance = (wallet: FuzzyWalletType): Promise<bigint> => tokenBalance(wallet, NATIVE_TOKEN_UID);
```

__tests__/integration/shared/token-authority.test.ts
```diff
@@ -0,0 +1,478 @@
...
+        // Minting with the facade default (createAnotherMint) leaves a mint authority.
+        const mintAuthorities = await adapter.getAuthorityUtxos(
+          wallet,
+          token.hash,
+          AuthorityType.MINT
+        );
+        expect(mintAuthorities.length).toBeGreaterThanOrEqual(1);
+        mintAuthorities.forEach(u => expect(u.authorities).toBe(TOKEN_MINT_MASK));
+      } finally {
+        await adapter.stopWallet(wallet);
```
raul-oliveira, line 82: I feel this stopWallet could be in the afterEach hook. What do you think about it?

__tests__/integration/shared/token-authority.test.ts
```diff
@@ -0,0 +1,478 @@
...
+      }
+    });
+
+    it('should mint tokens to a chosen mint authority address', async () => {
+      const { wallet } = await adapter.createWallet();
+      try {
+        const addr0 = (await wallet.getAddressAtIndex(0))!;
+        const authAddr = (await wallet.getAddressAtIndex(5))!;
```
raul-oliveira, line 90: rename to `authorityAddr` or `authorityAddress`

__tests__/integration/shared/token-authority.test.ts
```diff
@@ -0,0 +1,478 @@
...
+        await adapter.injectFunds(wallet, addr0, 10n);
+        const token = await adapter.createToken(wallet, 'MintAddrToken', 'MAT', 100n);
+
+        await adapter.mintTokens(wallet, token.hash, 100n, { mintAuthorityAddress: authAddr });
+
+        const authAtAddr = await adapter.getAuthorityUtxos(wallet, token.hash, AuthorityType.MINT, {
```
raul-oliveira, line 96: same here `authorityAtAddr` or `authorityAtAddress`

raul-oliveira (approved):
Looks good. The only thing that I really want you to address is the authority variables names. It led me to `auth` (from authentication/authorization) automatically

__tests__/integration/shared/token-authority.test.ts
tuliomir: @coderabbitai please add this as a rule for your reviews: in this repository we have "authority" as a common word and variable name. We should name it accordingly to avoid ambiguity with any eventual "authentication" or "authorization".

r4mmer (approved):

PR #1122, opened 2026-07-02
tests: precalculated shielded addresses for integration tests

pedroferreira1:
### Motivation

The shielded receive pipeline (#pr-5c) derives a shielded address pair at every
BIP32 index during address loading. Under jest's vm sandbox, bitcore's pure-JS
EC math runs ~40–58× slower than production (~360ms vs ~6ms per index —
profiled), so with ~120 wallet starts × 20 indexes plus re-walks from index 0
on every sync/reload, the integration suite regressed from <70 min to over 90
min (timing out). This is a **jest-environment cost, not a production
regression** — 20 derivations take ~124ms in plain node.

The fix follows the structure the integration tests always used for the same
problem on the legacy chain: **precalculate the addresses and inject them, so
nothing is re-derived at test time**. Derivation is NOT disabled — injected
indexes are byte-identical to live-derived ones, indexes past the injected
window still derive live, and a drift test re-derives the fixtures from the
seeds on every run.

### Acceptance Criteria
- Derive each shielded address index once per wallet: address loading skips
  indexes already present in storage (mirroring the legacy skip), re-pushing
  the paired spend address so subscriptions are unchanged.
- Parse the scan/spend parent xpubs once per `loadAddresses` call instead of
  twice per index, and skip the redundant on-curve re-validation when encoding
  pubkeys derived internally.
- Add a `preCalculatedShieldedAddresses` wallet parameter (test tooling,
  mirrors `preCalculatedAddresses`) that injects precalculated shielded pairs
  at wallet start.
- Commit shielded address fixtures for the fixed integration seeds (genesis,
  miner, ocb and the five multisig wallets — 22 indexes each) and wire them
  through the helpers, adapter and direct wallet-construction sites.
- Add a drift test that re-derives the fixtures live from the same seeds
  (full window for genesis, index 0 for the others), so fixtures can never
  silently rot and the live-derivation path stays exercised.
- Forward the optional `shieldedAddresses` field from the wallet-provider
  payload when present (older providers omit it and wallets derive live —
  fully backward compatible; provider-side support lands separately in
  hathor-integration-test-helper).

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for preloaded **shielded** address data during wallet startup via `preCalculatedShieldedAddresses`.
  * Shielded address loading can now persist and reuse precomputed shielded/spend address pairs.
* **Bug Fixes**
  * Improved shielded address derivation and storage behavior to avoid redundant work and prevent duplicate saves.
  * Ensures shielded address derivation consistently includes the paired spend address and ct-mapping.
* **Tests**
  * Added integration and lifecycle tests covering precalculated shielded fixtures and derivation caching.
  * Increased default integration test timeout to reduce CI false failures.

### DIFF
--- a/__tests__/integration/configuration/precalculated-shielded-addresses.test.ts
+++ b/__tests__/integration/configuration/precalculated-shielded-addresses.test.ts
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Drift check for the committed shielded address fixtures.
+ *
+ * The fixtures spare every fixed-seed wallet (genesis, multisig, ocb, miner)
+ * the per-index EC derivation at start, which jest's vm sandbox slows down
+ * ~40-58x. This suite re-derives them live from the same seeds and asserts
+ * equality, so the fixtures can never silently rot: if the derivation logic
+ * or a seed in test-constants changes, this fails and the fixtures must be
+ * regenerated (see the fixture file header for instructions).
+ *
+ * No network access is needed — this is pure derivation.
+ */
+
+import walletUtils from '../../../src/utils/wallet';
+import { deriveShieldedAddress } from '../../../src/utils/shieldedAddress';
+import { PRECALCULATED_SHIELDED_ADDRESSES } from './precalculated-shielded-addresses';
+import { NETWORK_NAME, WALLET_CONSTANTS } from './test-constants';
+import { multisigWalletsData } from '../helpers/wallet-precalculation.helper';
+
+const EXPECTED_WINDOW = 22; // mirrors the legacy precalculated address window
+
+// Access-data generation is expensive under jest (hardened derivations +
+// PBKDF2) — do it once per seed, not once per compared index.
+const accessDataCache = new Map<string, { scanXpubkey: string; spendXpubkey: string }>();
+function getShieldedXpubs(words: string) {
+  let cached = accessDataCache.get(words);
+  if (!cached) {
+    const accessData = walletUtils.generateAccessDataFromSeed(words, {
+      pin: '123',
+      password: '123',
+      networkName: NETWORK_NAME,
+    });
+    cached = { scanXpubkey: accessData.scanXpubkey!, spendXpubkey: accessData.spendXpubkey! };
+    accessDataCache.set(words, cached);
+  }
+  return cached;
+}
+
+/** Derive the fixture entry shape live, straight from the wallet's own code paths. */
+function deriveLiveEntry(words: string, index: number) {
+  const { scanXpubkey, spendXpubkey } = getShieldedXpubs(words);
+  const info = deriveShieldedAddress(scanXpubkey, spendXpubkey, index, NETWORK_NAME);
+  return {
+    bip32AddressIndex: index,
+    shieldedBase58: info.base58,
+    spendBase58: info.spendAddress,
+    scanPubkey: info.scanPubkey,
+    spendPubkey: info.spendPubkey,
+  };
+}
+
+describe('pre-calculated shielded address fixtures', () => {
+  it('covers every fixed seed with a full window of well-formed entries', () => {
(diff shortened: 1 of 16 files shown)


src/types.ts
```diff
@@ -141,6 +141,26 @@ export interface IAddressChainOptions {
...
+ *
+ * Mirrors the pre-calculated legacy addresses passed at wallet construction:
+ * carries exactly the fields needed to reconstruct the two storage records
+ * that live derivation (deriveShieldedAddressPair) would produce, so injected
+ * indexes skip the expensive EC derivation in loadAddresses.
+ */
+export interface IPrecalculatedShieldedAddress {
+  bip32AddressIndex: number;
+  /** The user-facing 71-byte shielded address (scan + spend pubkeys) */
+  shieldedBase58: string;
```
r4mmer, line 155: question(nitpick): Should we enforce a naming system for the new addresses?

I was thinking about this, a "shielded address" can be a little bit ambiguous since we have the 71-byte address that is required to create shielded outputs, but we also have the "spend address" which is the address publicly available on the shielded output.

We have the names _scan_ and _spend_ that determine the usage of the keys and addresses, which we also tie with the bip32 derivation account.
Maybe we should name the full address "confidential transaction address" or "CT address" to avoid the  ambiguous "shielded address".

My idea is to avoid the `shielded_address` name altogether so we have 0 doubt of what the address is.

In this type we could have `ctAddressB58` and `spendAddressB58`

r4mmer (approved):

__tests__/integration/helpers/wallet.helper.ts
```diff
@@ -315,8 +326,14 @@ export async function waitForTxReceived(
   }
 
   if (timeoutReached) {
-    // Throw error in case of timeout
-    throw new Error(`Timeout of ${timeoutPeriod}ms without receiving the tx with id ${txId}`);
+    // The timeout can fire while the loop is inside its 1s delay tick, right as
+    // the tx lands and finishes processing — re-check before declaring a
+    // timeout, so a received tx is never misreported as missing.
+    storageTx = (await hWallet.getTx(txId)) as IHistoryTx;
```
tuliomir, line 332: Such a generous gatekeeper, giving out a last-second chance for the tx 😁

__tests__/integration/configuration/precalculated-shielded-addresses.test.ts
```diff
@@ -0,0 +1,104 @@
...
+ * No network access is needed — this is pure derivation.
+ */
+
+import walletUtils from '../../../src/utils/wallet';
+import { deriveShieldedAddress } from '../../../src/utils/shieldedAddress';
+import { PRECALCULATED_SHIELDED_ADDRESSES } from './precalculated-shielded-addresses';
+import { NETWORK_NAME, WALLET_CONSTANTS } from './test-constants';
+import { multisigWalletsData } from '../helpers/wallet-precalculation.helper';
+
+const EXPECTED_WINDOW = 22; // mirrors the legacy precalculated address window
```
tuliomir, line 27: suggestion: Use the actual value from the fixture

```ts
const EXPECTED_WINDOW = WALLET_CONSTANTS.genesis.addresses.length;
```

__tests__/utils/loadAddressesShielded.test.ts
```diff
@@ -0,0 +1,151 @@
...
+      const liveShielded = await liveStorage.getAddressAtIndex(i, { legacy: false });
+      const injectedShielded = await injectedStorage.getAddressAtIndex(i, { legacy: false });
+      expect(injectedShielded).toEqual(liveShielded);
+      const liveSpend = await liveStorage.store.getAddress(liveShielded!.ctMappingAddress!);
+      const injectedSpend = await injectedStorage.store.getAddress(
+        injectedShielded!.ctMappingAddress!
+      );
+      expect(injectedSpend).toEqual(liveSpend);
+    }
+  }, 30000);
```
tuliomir, line 112: chore: Replace these magic number `30000` with a constant.
```suggestion
  }, DEFAULT_DERIVATION_TEST_TIMEOUT);
```

tuliomir (approved):

PR #1089, opened 2026-05-14
feat(shielded): wallet lifecycle, send pipeline, schema relaxation

pedroferreira1:
## Summary

Wires PRs 1–5 into the wallet's transaction-building and lifecycle paths. The integration slice: **receive** (`onNewTx` → `normalizeShieldedOutputs` → `processShieldedOutputs` → `saveAddress`) and **send** (`sendManyOutputs` → `buildShieldedOutputs` → `balanceBlindingFactors` → `attachHeader`) become end-to-end usable. **PR 6 of 7** in the decomposition of [PR #1059](https://github.com/HathorNetwork/hathor-wallet-lib/pull/1059).

```
master ─ PR1 ─ PR2 ─ PR3 ─ PR4 ─ PR5 ─ PR6 ◀ PR7
```

Base: `shielded/pr-5-storage-layer`.

Suggested **two reviewers**: one for wallet lifecycle + WS event ordering, one for send-side blinding factor arithmetic.

## Acceptance Criteria

- `wallet.start()` auto-loads the shielded crypto provider when `@hathor/ct-crypto-node` is installed; absence leaves the slot undefined and every downstream branch short-circuits.
- `wallet.start()` migrates pre-shielded access-data records to shielded when given both pinCode and password; xpub-only / xpriv-imported wallets bypass the migration.
- WS-driven receive (`onNewTx`) normalizes both shielded wire-shape variants (a) inline-in-outputs and (b) separated, then decrypts wallet-owned shielded outputs.
- A rejected `newTxPromise` no longer poisons the WS event chain — subsequent events continue to be processed (chain-resilience fix).
- `getShieldedUnblindingForTx` is available so audit/explorer tools can verify a shielded tx with browser-side primitives.
- `setPinCode` is available for runtime pin configuration.
- `sendManyOutputs` accepts shielded outputs (`ProposedOutput.shielded`) and a `changeShieldedMode` option so HTR fee-change can be emitted shielded, matching the recipient's privacy mode.
- The send pipeline builds shielded outputs with correct blinding-factor balancing (uses PR 2's `createShieldedOutputs`).
- The send pipeline attaches `ShieldedOutputsHeader` or `UnshieldBalanceHeader` correctly — the two are mutually exclusive.
- The send pipeline applies the correct per-output shielded fee (`FEE_PER_AMOUNT_SHIELDED_OUTPUT` or `FEE_PER_FULL_SHIELDED_OUTPUT`).
[...]

PR #1125, opened 2026-07-09
feat(shielded): wire format, schemas & parsing for separated shielded outputs

pedroferreira1:
### Description

First of three PRs that split the former "wallet lifecycle" PR (#1089) into reviewable, stacked layers. This one adds the data-model foundation for confidential (shielded) outputs in the separated-output model: the wire vocabulary, its validation schemas, and the normalization/guard utilities. No wallet-facing behaviour change beyond parsing — the lifecycle/read integration and send pipeline land in the two PRs stacked on top.

Stacking: base master → 6a (this PR) → 6b → 6c

### Acceptance Criteria
- Add the shielded-output wire type with a required mode discriminator (1 = AmountShielded, 2 = FullShielded), classified directly from the first wire byte.
- Add shielded-output validation schemas to the IHistoryTx schema set.
- Add addressType to input signatures and 'shielded' | 'transparent' to the history-input type so downstream layers can distinguish shielded spends.
- Add normalizeShieldedOutputs (base64→hex), isShieldedOutputIndex, shielded-output relocation, and sender-local-insert enrichment in the transaction utils.
- Reject shielded output indices in the transparent-only partial-transaction flow with a clear error.
- Reject shielded output indices in the transaction-template flow (context + executor) with a clear error.
- Export the shielded module namespace from the library entrypoint.
- Harden the JSONBigInt reviver against a missing parse context.
- Add unit coverage for the schemas, normalization, and the partial-tx / template guards.

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Enhanced history parsing/validation with shielded output modeling (including mode-aware wire shape) and support for `shielded_outputs`.
  * Exposed the `shielded` module as a top-level named export.

* **Bug Fixes**
  * Rejected shielded inputs earlier with clear, consistent error messages across partial transactions and transaction templates.
  * Improved shielded output normalization/signing metadata, and safely skips shielded processing when `ephemeral_pubkey` is missing.

* **Tests**
  * Added/expanded test coverage for shielded input rejection, schema discriminator behavior, normalization idempotency, and updated signature expectations.


src/models/partial_tx.ts
```diff
@@ -480,6 +480,12 @@ export class PartialTx {
               return resolve(false);
             }
 
+            // Partial transactions are a transparent-only flow; reject an input
+            // whose index lands in the shielded range with a clear error rather
+            // than silently resolving false (which would look like a mismatch).
+            if (transactionUtils.isShieldedOutputIndex(data.tx, input.index)) {
+              return reject(new Error('Shielded inputs are not supported in partial transactions'));
+            }
```
pedroferreira1, line 488: We will add support for partial tx, tx template, and multisig wallets later with shielded outputs

pedroferreira1:
@coderabbitai all your comments were addressed. Please resolve them and run a re-review

src/schemas.ts
```diff
@@ -81,9 +88,49 @@ export const IHistoryInputSchema: ZodSchema<IHistoryInput> = z
...
+    commitment: z.string().optional(),
+    script: z.string().optional(),
+    decoded: IHistoryOutputDecodedSchema.optional(),
+    value: bigIntCoercibleSchema.optional(),
+    token_data: z.number().optional(),
+    token: z.string().optional(),
+  })
+  .passthrough();
+
+// A plain union, NOT z.discriminatedUnion: older nodes omit `type` on
```
r4mmer: We could add the type field when it is missing, this way older inputs are always transparent AND we get a discriminated union which has a lot of typing benefits.

src/shielded/types.ts
```diff
@@ -43,19 +43,20 @@ export type {
...
+  // hex, 33 bytes. Protocol-optional: on-chain the field is all-zeros when
+  // absent and the fullnode omits the JSON key. Without it the output can
+  // never be rewound via ECDH (see shielded/processing.ts).
+  ephemeral_pubkey?: string;
   // FullShielded only:
   asset_commitment?: string; // hex, 33 bytes
   surjection_proof?: string; // hex, variable
 }
 
 export interface IShieldedOutput extends IShieldedOutputProofs {
```
r4mmer, line 55: Why not use the types from the zodiac schemas?
It i basically the same

src/shielded/processing.ts
```diff
@@ -118,6 +118,13 @@ export async function processShieldedOutputs(
 
   for (const [sIndex, shieldedOutput] of shieldedOutputs.entries()) {
     const absoluteIndex = transparentCount + sIndex;
+    // No ECDH hint (the on-chain field was all-zeros, so the fullnode omitted
+    // it): the output can never be rewound by this wallet — treat as non-owned.
+    // Checked first: it needs no storage access or key material, and skipping
+    // here means no scan child key is ever derived for the slot (the derived
+    // key is only zeroed by the rewind try/finally below).
+    if (!shieldedOutput.ephemeral_pubkey) continue;
```
r4mmer, line 126: Can we add a test with this case?

r4mmer (commented):
The bigint revival hardening from the acceptance criteria is not on the PR, was it deferred?

Also, should we have a test that does not accept invalid output modes?

src/schemas.ts
pedroferreira1: Done. Added a `z.preprocess` that normalizes the discriminant — any non-'shielded' value (including a missing type from older nodes, or an unexpected string) becomes 'transparent' — then switched to `z.discriminatedUnion`. Keeps the "never fail-closed on an unknown discriminator" behavior and gains the narrowing/perf/error-message benefits. Added a test for the unknown-type→transparent case too.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1125/commits/ccbf68ff3a39a2932ab4e751d52e464ead4af21b

src/shielded/types.ts
pedroferreira1: They do look almost identical, but two things block reusing the schema types here: `schemas.ts` already does `import { ShieldedOutputMode } from './shielded/types'`, so inferring `IShieldedOutput` from the schema module would create a circular import; and `IShieldedOutput` is the domain type at the `@hathor/ct-crypto-provider` boundary (it extends `IShieldedOutputProofs` and types mode as the native enum), intentionally independent of the wire-validation layer. I'd keep them separate for those reasons. What do you think?

src/shielded/processing.ts
pedroferreira1: Added https://github.com/HathorNetwork/hathor-wallet-lib/pull/1125/commits/ccbf68ff3a39a2932ab4e751d52e464ead4af21b

pedroferreira1:
> The bigint revival hardening from the acceptance criteria is not on the PR, was it deferred?

Actually my bad. I removed the file change because it was not necessary but forgot to update the acceptance criteria.

> Also, should we have a test that does not accept invalid output modes?

Added!

src/shielded/types.ts
r4mmer: Great!

r4mmer (dismissed):

r4mmer (approved):

tuliomir (approved):

PR #1127, opened 2026-07-09
feat(shielded): wallet lifecycle, sync & read/accounting

pedroferreira1:
### Description

Second of three PRs splitting the former "wallet lifecycle" PR (#1089). This one integrates shielded outputs into the HathorWallet read path: address derivation & storage, sync, receive handling, and all balance/UTXO/history reporting. After this PR the wallet can hold, derive, sync, and report shielded outputs it owns — it cannot yet send (that is 6c).

Stacking: master → 6a (wire foundation) → 6b (this PR) → 6c (send, formerly #1089).

### Acceptance Criteria
- Derive and store the shielded scan/spend address pair per BIP32 index (storage loading + memory-store shielded-index map + address utils).
- Sync received shielded outputs via their on-chain spend-derived P2PKH addresses (stream sync).
- Parse the fullnode /transaction shielded-output schema and hydrate getTxById correctly:  prefer the locally-decrypted transaction for balance so owned shielded receives aren't credited as 0.
- Normalize and process shielded outputs on new-transaction delivery in the SEPARATED model (onNewTx), including re-delivery/void handling.
- Account for owned shielded outputs in balance, transaction history, per-address info, and UTXO listing — transparent-only by default so consolidation never spends a shielded UTXO.
- Add getShieldedUnblindingForTx, exposing the wallet's unblinding data (cleartext value/token + value/asset blinding factors) only for outputs/inputs the wallet owns.
- Add setShieldedCryptoProvider and thread the pin through syncHistory / reloadStorage.
- Widen the address-getter interfaces (IAddressChainOptions) and expose the read-side types (UtxoOptions.shielded, ShieldedOpening / ShieldedOpeningEntry, history headers / shielded_outputs).
- Add unit coverage for unblinding, onNewTx shielded handling, and shielded address accounting.

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for shielded transaction inputs/outputs across balances, history, address discovery, and unblinding.
  * Added PIN-aware history sync/processing and optional PIN per websocket update.
  * Added chain/path options for address derivation and a configurable shielded crypto provider hook.
  * Added a helper to derive output-script addresses from shielded addresses.
* **Bug Fixes**
  * Improved partially decoded shielded handling and idempotent shielded output processing.
  * Prevented untrusted shielded decode-only fields from impacting accounting; preserved decoded shielded data across re-deliveries/reloads.
  * Hardened against forged shielded input confidential fields.
* **Tests**
  * Expanded coverage for shielded accounting and websocket/on-new-transaction behavior.

### DIFF
--- a/jest-integration.config.js
+++ b/jest-integration.config.js
@@ -25,10 +25,10 @@ module.exports = {
     },
     // We need a high coverage for the HathorWallet class
     './src/new/wallet.ts': {
-      statements: 90,
-      branches: 83,
-      functions: 90,
-      lines: 90,
+      statements: 75,
+      branches: 65,
+      functions: 75,
+      lines: 75,
     },
   },
 };
--- a/src/new/sendTransaction.ts
+++ b/src/new/sendTransaction.ts
@@ -545,7 +545,11 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
                 // This just returns if the transaction is not a CREATE_TOKEN_TX
                 await addCreatedTokenFromTx(transaction as CreateTokenTransaction, storage);
                 // Add new transaction to the wallet's storage.
-                wallet.enqueueOnNewTx({ type: '', history: historyTx }); // FIXME: Add a type here
+                // Pass the pin used for this send so the wallet can decrypt and
+                // credit its own shielded change even when it holds no stored
+                // pinCode (per-call-pin flow).
+                // FIXME: Add a type for the message
+                wallet.enqueueOnNewTx({ type: '', history: historyTx }, this.pin ?? undefined);
               })(this.wallet, this.storage, this.transaction);
             }
             this.emit('send-tx-success', this.transaction);
--- a/src/new/sendTransaction.ts
+++ b/src/new/sendTransaction.ts
@@ -548,8 +548,14 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
                 // Pass the pin used for this send so the wallet can decrypt and
                 // credit its own shielded change even when it holds no stored
                 // pinCode (per-call-pin flow).
-                // FIXME: Add a type for the message
-                wallet.enqueueOnNewTx({ type: '', history: historyTx }, this.pin ?? undefined);
+                // The type labels the WS event this sender-local insert emulates
+                // (the fullnode will deliver the real one for this tx shortly);
+                // only the handleWebsocketMsg router inspects it, and this path
+                // bypasses the router.
+                wallet.enqueueOnNewTx(
+                  { type: 'wallet:address_history', history: historyTx },
+                  this.pin ?? undefined
+                );
               })(this.wallet, this.storage, this.transaction);
             }
             this.emit('send-tx-success', this.transaction);
--- a/src/new/types.ts
+++ b/src/new/types.ts
@@ -104,6 +104,13 @@ export interface UtxoOptions {
   amount_bigger_than?: bigint;
   max_amount?: bigint;
   only_available_utxos?: boolean;
+  /**
+   * Whether to include shielded UTXOs. Defaults to `false` (transparent-only):
+   * `getUtxos` feeds consolidation, which spends its results as TRANSPARENT
+   * inputs, so a shielded UTXO must never leak into that listing. Pass `true`
+   * to explicitly include shielded UTXOs.
+   */
+  shielded?: boolean;
 }
 
 /**
@@ -348,6 +355,23 @@ export interface ProposedOutput {
 }
 
 /**
+ * The unblinding data the wallet holds for a single shielded output/input it
+ * owns: the cleartext `value`/`token` plus the value blinding factor (`vbf`)
+ * and, for FullShielded slots, the asset blinding factor (`abf`).
+ */
+export interface ShieldedOpening {
+  value: bigint;
+  token: string;
+  vbf: string;
+  abf?: string;
+}
+
+/** A {@link ShieldedOpening} tagged with the on-chain absolute slot index. */
+export interface ShieldedOpeningEntry extends ShieldedOpening {
+  index: number;
+}
+
+/**
  * Proposed input for a transaction
  * @property txId Transaction ID of the input
  * @property index Index of the output being spent
@@ -442,6 +466,19 @@ export interface GetTxHistoryFullnodeFacadeReturnType {
   ncMethod?: string;
   ncCaller?: Address; // Address type
   firstBlock?: string;
+  /**
+   * Tx-level headers (FeeHeader, etc.) carried through unchanged from
+   * the on-chain payload so callers can compute network fees without
+   * a second `getTx` round-trip.
+   */
+  headers?: unknown[];
+  /**
+   * Shielded output entries from `tx.shielded_outputs[]`. Each entry's
+   * `mode` (1 = AmountShielded, 2 = FullShielded) is the only field
+   * the privacy-fee accumulator needs; others are passed through for
+   * future use.
+   */
+  shielded_outputs?: { mode?: number }[];
 }
 
 /**
--- a/src/new/types.ts
+++ b/src/new/types.ts
@@ -117,10 +117,14 @@ export interface UtxoOptions {
  * Options for filtering available UTXOs
  * @property token Search for UTXOs of this token UID
  * @property filter_address Address to filter the utxos
+ * @property shielded Include shielded UTXOs. Defaults to `false`
+ *   (transparent-only) because the results feed transparent-only input
+ *   selection; pass `true` to explicitly include shielded UTXOs.
  */
 export interface GetAvailableUtxosOptions {
   token?: string;
   filter_address?: string;
+  shielded?: boolean;
 }
 
 /**
--- a/src/new/wallet.ts
+++ b/src/new/wallet.ts
@@ -110,6 +110,7 @@ import { IHistoryTxSchema } from '../schemas';
 import GLL from '../sync/gll';
 import { TransactionTemplate, WalletTxTemplateInterpreter } from '../template/transaction';
 import Address from '../models/address';
+import type { HistoryTransactionOutput } from '../models/types';
 import type { IShieldedCryptoProvider } from '../shielded/types';
 import { ConnectionState, FullNodeVersionData, IHathorWallet, Utxo } from '../wallet/types';
 import Transaction from '../models/transaction';
@@ -836,14 +837,17 @@ class HathorWallet extends EventEmitter {
     let address = await this.storage.getAddressAtIndex(index, opts);
 
     if (address === null) {
+      // Derivation on a storage miss is a pure PEEK: the address is returned
+      // without being persisted. Persisting belongs to loadAddresses /
+      // scanAddressesToLoad, which manage the loaded window, its tracking
+      // cursors and the WS subscriptions coherently — saving here would leave
+      // holes in the window and claim ownership of an unwatched address.
       if (opts?.legacy === false) {
-        // Shielded address not yet derived at this index — derive and save
+        // Shielded address not yet derived at this index
         const result = await deriveShieldedAddressFromStorage(index, this.storage);
         if (!result) {
           throw new Error('Shielded keys not available');
         }
-        await this.storage.saveAddress(result.shieldedAddress);
-        await this.storage.saveAddress(result.spendAddress);
         return result.shieldedAddress.base58;
       }
       // Legacy address derivation
@@ -852,7 +856,6 @@ class HathorWallet extends EventEmitter {
       } else {
         address = await deriveAddressP2SH(index, this.storage);
       }
-      await this.storage.saveAddress(address);
     }
     return address.base58;
   }
@@ -867,20 +870,19 @@ class HathorWallet extends EventEmitter {
    * @inner
    */
   async getAddressPathForIndex(index: number, opts?: IAddressChainOptions): Promise<string> {
+    let acctPath: string;
     if (opts?.legacy === false) {
       // Shielded addresses are formed by scanPubkey (account 1') + spendPubkey (account 2').
       // We return the spend path since it's used for on-chain scripts and signing.
-      return `${SHIELDED_SPEND_ACCT_PATH}/0/${index}`;
-    }
-
-    const walletType = await this.storage.getWalletType();
-    if (walletType === WalletType.MULTISIG) {
+      acctPath = SHIELDED_SPEND_ACCT_PATH;
+    } else if ((await this.storage.getWalletType()) === WalletType.MULTISIG) {
       // P2SH
-      return `${P2SH_ACCT_PATH}/0/${index}`;
+      acctPath = P2SH_ACCT_PATH;
+    } else {
+      // P2PKH
+      acctPath = P2PKH_ACCT_PATH;
     }
-
-    // P2PKH
-    return `${P2PKH_ACCT_PATH}/0/${index}`;
+    return `${acctPath}/0/${index}`;
   }
 
   /**
@@ -1261,11 +1263,15 @@ class HathorWallet extends EventEmitter {
         value: bigint,
         decodedAddress: string | undefined,
         outputToken: string,
+        // Absent on entries the wallet inserted locally (spent_by cannot be
+        // reconstructed there); the wire always stamps it (hex or null).
         spentBy: string | null | undefined,
-        lockable: Parameters<typeof transactionUtils.isOutputLocked>[0]
+        // Only `decoded` is consumed (timelock check) — passing the full
+        // output object at the call sites is fine.
+        lockable: Pick<HistoryTransactionOutput, 'decoded'>
       ): void => {
         if (decodedAddress !== address || token !== outputToken) return;
-        const is_spent = (spentBy ?? null) !== null;
+        const is_spent = spentBy != null;
         const is_locked =
           transactionUtils.isOutputLocked(lockable) ||
           transactionUtils.isHeightLocked(tx.height, nowHeight, rewardLock);
@@ -1294,13 +1300,9 @@ class HathorWallet extends EventEmitter {
       // carry an opening; the rest are skipped.
       for (const so of tx.shielded_outputs ?? []) {
         if (so.value === undefined) continue; // not owned / not yet decrypted
-        accrue(
-          so.value,
-          so.decoded?.address,
-          so.token ?? NATIVE_TOKEN_UID,
-          so.spent_by ?? null,
-          so
-        );
+        // The `value` gate above proves the slot was decrypted, and decode
+        // writes value+token together — so `token` is always present here.
+        accrue(so.value, so.decoded?.address, so.token!, so.spent_by, so);
       }
     }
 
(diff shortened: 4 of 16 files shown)


src/new/wallet.ts
```diff
@@ -1038,6 +1091,99 @@ class HathorWallet extends EventEmitter {
...
+   * commitment at that index.
+   *
+   * SEPARATED model: shielded outputs live in `tx.shielded_outputs[]` (the
+   * full on-chain-ordered list); the wallet writes the owned-marker fields
+   * (`value`/`token`/`blindingFactor`/`assetBlindingFactor`) IN PLACE on the
+   * slots it decrypts. The single ownership gate is `value !== undefined`;
+   * the on-chain absolute index of `shielded_outputs[s]` is
+   * `tx.outputs.length + s`.
+   */
+  async getShieldedUnblindingForTx(
```
pedroferreira1, line 1121: This method is used by the wallet mobile. The explorer has a feature that we can send blinding factor for ONLY a specific tx. This method create these blinding factors, so you can click on the mobile wallet and open the explorer unblinded

src/new/sendTransaction.ts
```diff
@@ -545,7 +545,17 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+                // credit its own shielded change even when it holds no stored
+                // pinCode (per-call-pin flow).
+                // The type labels the WS event this sender-local insert emulates
+                // (the fullnode will deliver the real one for this tx shortly);
+                // only the handleWebsocketMsg router inspects it, and this path
+                // bypasses the router.
+                wallet.enqueueOnNewTx(
+                  { type: 'wallet:address_history', history: historyTx },
+                  this.pin ?? undefined
+                );
```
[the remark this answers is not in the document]
pedroferreira1: the PIN is threaded via the constructor, not run, so this.pin is the effective per-call PIN on the send path — shielded change is credited. (And optionally: confirm the token paths can't produce shielded change, else thread the pin there too.)

pedroferreira1:
@coderabbitai all your comments were addressed already. Can you close all of them and re-run a review?

src/wallet/types.ts
```diff
@@ -359,12 +360,15 @@ export interface IHathorWallet {
...
-    markAsUsed: boolean;
-  }): AddressInfoObject | Promise<AddressInfoObject>; // FIXME: Should have a single return type
-  getNextAddress(): AddressInfoObject | Promise<AddressInfoObject>; // FIXME: Should have a single return type;
+  getCurrentAddress(
+    options?: {
+      markAsUsed: boolean;
+    },
+    opts?: IAddressChainOptions
+  ): AddressInfoObject | Promise<AddressInfoObject>; // FIXME: Should have a single return type
+  getNextAddress(opts?: IAddressChainOptions): AddressInfoObject | Promise<AddressInfoObject>; // FIXME: Should have a single return type;
```
r4mmer, line 373: **issue (non-blocking):** IHathorWallet only partially adopted the new shielded read API — three additions are missing.

The interface updated `getAddressAtIndex`/`getCurrentAddress`/`getNextAddress` to take `IAddressChainOptions`, but:

1. `getAllAddresses` still lacks the opts arg (line 345)
2. `getAddressPathForIndex` still lacks the opts arg (line 438)
3. `setShieldedCryptoProvider` is entirely absent, even though the PR description says clients MUST call it to enable shielded support

Consumers typed against `IHathorWallet` (test doubles, DI containers, wallet-service adapters) can't reach the shielded API without downcasting to the concrete class. The fix is three signature additions in this file.

src/new/types.ts
```diff
@@ -104,16 +104,27 @@ export interface UtxoOptions {
   amount_bigger_than?: bigint;
   max_amount?: bigint;
   only_available_utxos?: boolean;
+  /**
+   * Whether to include shielded UTXOs. Defaults to `false` (transparent-only):
+   * `getUtxos` feeds consolidation, which spends its results as TRANSPARENT
+   * inputs, so a shielded UTXO must never leak into that listing. Pass `true`
+   * to explicitly include shielded UTXOs.
+   */
+  shielded?: boolean;
```
r4mmer, line 113: **suggestion (non-blocking):** Override `shielded: false` in `prepareConsolidateUtxosData` so consolidation can never select shielded UTXOs.

`UtxoOptions` newly exposes `shielded?: boolean` (documented as "a shielded UTXO must never leak into that listing"), but the chain `consolidateUtxos` → `consolidateUtxosSendTransaction` → `prepareConsolidateUtxosData` → `getUtxos` forwards options unchanged, so a caller passing `{ shielded: true }` bypasses the documented invariant.

The runtime effect is a hard build-time error (SEPARATED-model input index falls past `spentTx.outputs.length`), not silent data loss, so this is defense-in-depth. A one-line override (`shielded: false`) inside `prepareConsolidateUtxosData` covers it.

Note: CodeRabbit flagged this same file/line earlier and it appears to have been bulk-resolved without a code change.

src/new/wallet.ts
```diff
@@ -2754,9 +3078,16 @@ class HathorWallet extends EventEmitter {
...
+      // spend-derived decoded.address is on the wire for every output.
+      const resolved = transactionUtils.resolveSpentOutput(spentTx, input.index);
+      const spentAddress = resolved?.output?.decoded?.address;
+      if (!spentAddress) {
+        continue;
+      }
+      const addressInfo = await this.storage.getAddressInfo(spentAddress);
       if (addressInfo === null) {
         continue;
       }
```
r4mmer, line 3101: **issue (non-blocking):** Three call sites invoke `getAddressPathForIndex` without opts, so shielded-spend addresses report the legacy P2PKH path.

`getAddressPathForIndex` selects the shielded-spend BIP32 path (`m/44'/280'/2'`) only when `opts.legacy === false`, but these call sites omit opts even when the owned address is a shielded-spend record, so `addressPath` reports `m/44'/280'/0'` instead:

- `getSignatures` (src/new/wallet.ts:3132) via the address-info fetch
- `getWalletInputInfo` (src/new/wallet.ts:3094) via the address-info fetch
- `getAvailableUtxos` (src/new/wallet.ts:1408) when a caller passes `{ shielded: true }`

External-signing / hardware-wallet consumers that derive a pubkey from the reported path will get the legacy pubkey instead of the spend-chain pubkey that actually locks/signs the output — signature verification against the reported path fails.

Fix per site: fetch `storage.getAddressInfo(address)` and pass `{ legacy: addressInfo?.addressType !== 'shielded-spend' ? undefined : false }`. `getSignatures` can also read `sigData.addressType` directly, which `getSignatureForTx` already propagates for exactly this purpose.

src/sync/stream.ts
```diff
@@ -629,7 +632,32 @@ export class StreamManager extends AbortController {
...
+              await this.storage.saveAddress(shieldedResult.spendAddress);
+            }
+          }
+        } catch (e) {
+          this.logger.error(
+            'Failed to derive shielded address at index',
+            addr.bip32AddressIndex,
+            e
+          );
+        }
```
r4mmer, line 656: **issue (blocking):** Streaming sync modes never scan or subscribe the shielded-spend chain, so shielded-only txs are silently missed.

The `MANUAL_STREAM_WS` / `XPUB_STREAM_WS` paths never tell the fullnode to scan the wallet's shielded-spend P2PKH chain and never subscribe those addresses for WS notifications. A wallet receiving a shielded-only tx while using either streaming mode misses it entirely — no historical scan, no realtime delivery — diverging from `POLLING_HTTP_API`, which handles both.

To fix: include the spend-derived P2PKH in `StreamManager.deriveBatch`'s outbound batches and call `connection.subscribeAddress` on each newly derived spend address. For xpub-stream mode, either extend the protocol to accept the shielded-spend xpub or fall back to manual for shielded-capable wallets.

src/new/sendTransaction.ts
```diff
@@ -545,7 +545,17 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
               })(this.wallet, this.storage, this.transaction);
```
r4mmer, line 559: **issue (blocking):** The sender-local insert loses the race to WS delivery, so the wallet's own shielded change is never decrypted on per-call-pin wallets.

The insert at line 555 is documented as a defense to decrypt the wallet's own shielded change on wallets constructed with `pinCode=null`, but WS's `handleWebsocketMsg` (wallet.ts:936) enqueues the same tx first with no pin. `processNewTx`'s `pinCode !== undefined` gate (storage.ts:1107) then skips `processShieldedOutputs`, and by the time this IIFE's per-call pin arrives, `isNewTx=false` routes through `processMetadataChanged`, which never decrypts.

The self-send-change symptom is dormant in 6b (no shielded-output creation until 6c), but the plumbing is being merged broken. Fix options:

a) detect `isNewTx=false && hasUndecodedSlot && txPin !== undefined` and route through a decrypt-only helper, or
b) have `handleWebsocketMsg` defer WS events for tx_ids with a pending sender-local enqueue.

r4mmer (commented):
Ultrareview findings, formatted as [conventional comments](https://conventionalcomments.org/): 2 blocking issues, 2 non-blocking issues, 1 suggestion. Details inline.

jest-integration.config.js
```diff
@@ -25,10 +25,10 @@ module.exports = {
...
     // We need a high coverage for the HathorWallet class
     './src/new/wallet.ts': {
-      statements: 90,
-      branches: 83,
-      functions: 90,
-      lines: 90,
+      statements: 75,
+      branches: 65,
+      functions: 75,
+      lines: 75,
```
pedroferreira1, line 31: They are restored in PR 7 https://github.com/HathorNetwork/hathor-wallet-lib/pull/1090/changes/6f353bfb8bafcfd2f476ad238adfc2eaab4524c2

Where we add the integration tests

src/wallet/types.ts
tuliomir: Thanks for bringing this up, r4mmer. This is certainly happening because the Lib's test files are not under Typescript coverage, so they never trigger those changes.

It's a project that should after the Shared Integration Tests finish.

src/utils/transaction.ts
```diff
@@ -205,6 +205,89 @@ const transaction = {
...
+   * can be legitimately known by the fullnode (they are hidden in the
+   * commitment), so a hostile source could pre-fill them to forge a credit or a
+   * debit that skips the ECDH-rewind verification. Ownership is re-established
+   * ONLY from the wallet's own rewind or its own storage. Call this at every
+   * wire-ingress point before persisting. No-op for honest deliveries (which
+   * carry these bare) and for transparent txs.
+   */
+  clearUntrustedShieldedData(tx: IHistoryTx): void {
+    for (const so of tx.shielded_outputs ?? []) {
+      // eslint-disable-next-line no-param-reassign
```
tuliomir: suggestion: Simplify the `eslint` command
```suggestion
/* eslint-disable no-param-reassign */
for (const so of tx.shielded_outputs ?? []) {
  ...
  all changes
  ...
}
/* eslint-enable no-param-reassign */  
```

src/utils/storage.ts
```diff
@@ -335,6 +335,10 @@ export async function* loadAddressHistory(
         for (const tx of result.history) {
           foundAnyTx = true;
           if (saveTxs) {
+            // Wire ingress: never trust decode-only shielded fields off the
+            // address-history response; addTx restores them from our own
+            // storage and processHistory re-derives the rest via ECDH rewind.
+            transactionUtils.clearUntrustedShieldedData(tx);
```
tuliomir, line 341: ⭐ That's a great defensive move!

src/utils/address.ts
```diff
@@ -51,6 +51,28 @@ export function getAddressType(address: string, network: Network): 'p2pkh' | 'p2
   return addrType;
 }
 
+/**
+ * Resolve an address to the form usable as a transparent output script.
+ *
+ * Shielded addresses have no direct output script (on-chain they use the
+ * spend-derived P2PKH), so this returns the spend address base58 for a shielded
+ * input and the address unchanged otherwise. Use it before `getAddressType` /
+ * output building wherever a caller-supplied address may be shielded — e.g. a
```
tuliomir, line 60: question: Should we consider adding a call to `resolveOutputScriptAddress` inside `getAddressType`? That way callers would not need to worry about calling both methods.

tuliomir (dismissed):

src/wallet/types.ts
pedroferreira1: Fixed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1127/changes/e753707ef0d1ef7363ff266a9a3a3400a0cf26ec: added `opts?: IAddressChainOptions` to `getAllAddresses` and
`getAddressPathForIndex`, and declared `setShieldedCryptoProvider(provider?): void`
on `IHathorWallet`. Since `HathorWalletServiceWallet` also implements the interface
and the wallet-service backend has no shielded support, it gets a `Not implemented.`
stub consistent with its other fullnode-only methods.

src/new/types.ts
pedroferreira1: Good defense-in-depth catch. Confirmed the flag threads through
`consolidateUtxos → consolidateUtxosSendTransaction → prepareConsolidateUtxosData →
getUtxos`, so a caller passing `{ shielded: true }` would reach selection. As you
note it's a loud build-time failure (a SEPARATED-model shielded input index past
`outputs.length`), not silent loss — but consolidation spends its inputs as
transparent, so there's never a reason for it to select shielded UTXOs.

Fixed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1127/changes/e753707ef0d1ef7363ff266a9a3a3400a0cf26ec: `prepareConsolidateUtxosData` now forces `shielded: false` on
the `getUtxos` call (literal after the spread, so it wins over any caller value).

src/sync/stream.ts
pedroferreira1: Confirmed the gap — streaming's `processQueue` derives the shielded+spend pair but
only `saveAddress`s it locally; it never `subscribeAddresses` or sends it to the
node for historical scan, unlike the polling path. Thanks for flagging.

On the fix, though: the contained "fall back to polling for shielded-capable
wallets" turns out to disable streaming for essentially **every seed wallet** —
`scanXpubkey`/`spendXpubkey` are populated for any root-key wallet
(`utils/wallet.ts:627 if (argXpriv.depth === 0)`), so `getSpendXPubKey()` is defined
for legacy and multisig wallets too, not just "shielded" ones. That guard silently
drops all seed wallets to polling and breaks 6 existing `stream.test.ts` cases. The
real fix is scoped work: for `MANUAL_STREAM_WS`, add the spend-derived P2PKH to
`deriveBatch`'s outbound scan window and `subscribeAddress` it; for `XPUB_STREAM_WS`
it's a fullnode-protocol limitation (the node can't reach the hardened
`m/44'/280'/2'` account from the single main xpub), so that mode needs a protocol
extension to accept the shielded-spend xpub or a documented fallback.

Given that, I'd rather handle this as a focused follow-up PR than bolt
a regressing guard into 6b. Mitigation meanwhile: the default sync mode is
`POLLING_HTTP_API` (which handles both), so only wallets that explicitly opt into a
streaming mode are exposed. What do you think?

src/utils/transaction.ts
pedroferreira1: Good call — this method has 7 `// eslint-disable-next-line no-param-reassign`
across its two loops (blanking wire-supplied `value`/`token`/blinding/`decoded`
fields in place is the method's whole job). Switching to the block form —
`/* eslint-disable no-param-reassign */` … `/* eslint-enable no-param-reassign */`
around the reassignments.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1127/changes/bfa88bfd7a964183b1ae062b058fa6445104e7b8

src/utils/address.ts
pedroferreira1: Good question. I'd lean toward keeping them separate:

- `getAddressType` is deliberately a strict *output-script* classifier — it throws
  for shielded so a 71-byte address can't silently slip into a transparent
  script-building path that doesn't support it. Callers rely on that guard: e.g.
  nano-contract action/authority outputs (`nano_contracts/builder.ts`) aren't
  shielded-capable, so a shielded address there should be rejected loudly, not
  quietly converted.
- It returns only the *type*, not the resolved address — a caller building the
  output still needs the resolved P2PKH string, so it would call
  `resolveOutputScriptAddress` anyway. Folding it in wouldn't remove the second
  call, it would just weaken the guard.
- Resolution is context-dependent — correct only where shielded output-building is
  actually supported. Today that's one site: the shielded change in token creation
  (`tokens.ts:508`), which resolves once into `cAddress` and uses that single value
  for both the type and the output, so there's no type/address mismatch.

For general 'shielded'-aware classification the doc already points to
`new Address(addr, { network }).getType()`. If pairing them ever gets error-prone at
a new shielded-output site, I'd add a small local `resolve → { type, address }`
helper there rather than change the shared classifier's contract. Any specific call
site you're worried about?

src/new/wallet.ts
pedroferreira1: Confirmed the mechanism, and agree it matters for external/hardware signers that
re-derive the pubkey from the reported path. (Internal signing is unaffected —
`getSignatureForTx` already signs shielded-spend inputs on the spend chain, so no
signature is actually produced wrong; only the reported `addressPath` string was.)

Scope: these three sites are unchanged context on this PR — the fix lands in the send
layer (the next PR), where `getWalletInputInfo` and `getSignatures` are introduced with
the SEPARATED-model spend-key resolution. On the next PR those two already pass
`{ legacy: addressType !== 'shielded-spend' }` (`getSignatures` reads
`sigData.addressType` directly, as you suggested). The only remaining site was
`getAvailableUtxos`, now fixed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/ff27558ebb72ffd0f934843096ba47d559551fac by looking up
`getAddressInfo(utxo.address)` and passing the same `legacy` flag.

src/new/sendTransaction.ts
pedroferreira1: Great diagnosis — confirmed end to end: a WS `new-tx` with no pin persists the tx
first, the `pinCode !== undefined` gate skips decryption, and the later
per-call-pin insert then routes through `processMetadataChanged` (`isNewTx=false`),
which never decrypted. And you're right it's dormant in this PR (no shielded outputs are
created until the next one).

Fixed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/ff27558ebb72ffd0f934843096ba47d559551fac with your option (a), framed as the ordering-independent
invariant you suggested: `processMetadataChanged` now takes the pin and runs
`processShieldedOutputs` on any still-undecoded owned slots (same per-slot gate as
`processNewTx`), so the wallet's own shielded change is decrypted whichever event
wins the race. Avoided option (b), since deferring/suppressing the WS event would
drop its `first_block`/height metadata. The end-to-end regression test (self-shield
send on a `pinCode=null` wallet with the WS event injected before the sender-local
insert) will go in the integration suite in PR 7.

src/sync/stream.ts
r4mmer: Accepting the deferral — you're right about the guard: verified that `spendXpubkey` is derived for every root-key wallet (`utils/wallet.ts` populates it whenever `argXpriv.depth === 0`), so gating on `getSpendXPubKey()` would silently drop all seed wallets to polling. A focused follow-up PR (spend-derived P2PKH in `deriveBatch` + `subscribeAddress` for manual mode; protocol extension for xpub mode) is the right scope, and the `POLLING_HTTP_API` default keeps the exposure opt-in meanwhile.

Resolving this thread; withdrawing the blocking status in favor of the follow-up.

r4mmer (approved):

tuliomir (approved):

PR #1128, opened 2026-07-09
test: migrate tx-voiding itests to fullnode-specific

tuliomir:
Relocates the `processing transaction metadata changes` describe block (transaction-voiding history/balance recomputation) out of the monolithic `hathorwallet_others.test.ts` into a dedicated `fullnode-specific/metadata-voiding.test.ts`, as part of the shared-tests migration.

It lands in `fullnode-specific/` (not `shared/`) because voiding is exercised via fullnode-only internals with no wallet-service equivalent: the tests spy-capture the raw `onNewTx` websocket payload, flip `is_voided`, replay it, and assert on `MemoryStore` internals — none of which the wallet-service facade exposes.

Pure relocation — no assertions or coverage changed (confirmed by an adversarial diff review).

**Acceptance Criteria**

- [ ] The voiding block runs under `fullnode-specific/` with identical assertions.
- [ ] No behavioral/coverage change vs. the previous monolith location.

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for wallet behavior when transactions are marked as voided.
  * Verified balance, token, transaction, and UTXO state restoration across incoming, self-sent, externally sent, and token-creation transactions.
  * Removed unused test imports.


__tests__/integration/fullnode-specific/metadata-voiding.test.ts
```diff
@@ -0,0 +1,483 @@
...
+import { IHistoryTx } from '../../../src/types';
+
+/**
+ * Returns the most recent `onNewTx` spy call that carried the given txId, or
+ * `undefined` when the spy never saw it. Used to recover the websocket payload
+ * for a specific transaction so the test can replay it with `is_voided` set.
+ */
+function findLastCallFor(
+  txId: string,
+  wsSpy: jest.SpiedFunction<typeof HathorWallet.prototype.onNewTx>
```
raul-oliveira, line 36: since it's fullnode specify, why the name `wsSpy`?

raul-oliveira: is it for websocket?

raul-oliveira (approved):

__tests__/integration/fullnode-specific/metadata-voiding.test.ts
tuliomir: That's right: a it's a spy for the `onNewTx`, that is called via websocket.

r4mmer (approved):

PR #1131, opened 2026-07-15
feat(shielded): send pipeline — build, sign & send shielded outputs

pedroferreira1:
### Summary

Third and final split of the former "wallet lifecycle" PR. Builds on the wire-format layer (6a, merged) and the lifecycle/read layer (6b) to complete the shielded-outputs wallet integration: the wallet can now build, sign and send transactions that create shielded outputs.

Changes:
- Send pipeline (src/new/sendTransaction.ts): the shielded path through SendTransaction — assembling, signing and pushing txs that emit shielded outputs.
- Wallet facade (src/new/wallet.ts): sendManyOutputsSendTransaction maps ProposedOutput → ISendOutput with changeShieldedMode; assemblePartialTransaction / getWalletInputInfo / getSignatures resolve spend keys under the SEPARATED output model.
- Token creation (src/utils/tokens.ts): shielded change-address resolution when minting/creating tokens.
- Types (src/new/types.ts): send-path type surface for the above.
- Tests (__tests__/new/sendTransaction.test.ts): unit coverage for the shielded send pipeline.

### Acceptance Criteria

- Add a shielded send pipeline in SendTransaction that builds, signs and pushes transactions creating shielded outputs.
- Add wallet-facade send support mapping ProposedOutput → ISendOutput with changeShieldedMode.
- Add SEPARATED-model spend-key resolution in assemblePartialTransaction, getWalletInputInfo and getSignatures.
- Add shielded change-address resolution to token creation.

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for shielded recipient outputs in transactions and send-many flows.
  * Added `changeShieldedMode` to convert transparent HTR change into shielded HTR change with correct fee balancing and output ordering.
* **Bug Fixes**
  * Improved spent-output resolution and shielded spend validation for more reliable address/token handling.
  * Hardened wallet history rebuild to skip failing shielded decodes without aborting.
  * Improved handling of universal (71-byte) change addresses and aligned derivation for shielded-spend signing.
  * Prevented incompatible shielded UTXOs from being used in transparent partial-tx proposals.
* **Documentation**
  * Refreshed `resolveOutputScriptAddress` wording for universal formats.
* **Tests**
  * Expanded coverage for shielded routing and HTR change conversion edge cases (including failure and fee-short scenarios).


src/new/sendTransaction.ts
```diff
@@ -745,6 +1138,81 @@ export async function prepareSendTokensData(
...
+export async function convertHtrChangeIfRequested(
+  partialHtrTxData: Pick<IDataTx, 'inputs' | 'outputs'>,
+  shieldedOutputDefs: IResolvedShieldedOutputDef[],
+  mode: ShieldedOutputMode | null,
+  wallet: HathorWallet | null,
+  network: ReturnType<IStorage['config']['getNetwork']>
+): Promise<{ addedFee: bigint }> {
+  if (!mode) return { addedFee: 0n };
+  if (!wallet) return { addedFee: 0n };
+  if (shieldedOutputDefs.length === 0) return { addedFee: 0n };
```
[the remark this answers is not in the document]
pedroferreira1: Thanks — but keeping the change transparent at the cap would contradict a deliberate design decision for this parameter: changeShieldedMode must never silently downgrade to a transparent change. If the requested privacy can't be honored, the send fails loudly rather than quietly leaking the change alongside an otherwise-private send.

src/new/sendTransaction.ts
```diff
@@ -207,15 +322,55 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+          // Non-owned / undecoded shielded slot with no stored UTXO — the
+          // wallet cannot spend it (no value/blinding to sign with).
+          const err = new SendTxError(ErrorMessages.INVALID_INPUT);
+          err.errorData = { txId: input.txId, index: input.index };
+          throw err;
+        }
+        spentValue = so.value ?? utxo!.value;
+        spentToken = so.token ?? utxo!.token;
+        spentAddress = so.decoded?.address ?? utxo!.address;
+        spentAuthorities = utxo ? utxo.authorities ?? 0n : 0n;
```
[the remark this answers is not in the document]
pedroferreira1: After the earlier change to this block (slot-only resolution, dropping the stored-UTXO fallback), the three paths are consistent for every reachable send, so there's no invalid-proof scenario left to fix:

pedroferreira1:
@coderabbitai all your comments were addressed. Close all of them and run a re-review

src/utils/storage.ts
```diff
@@ -1119,15 +1119,18 @@ export async function processNewTx(
...
+      // Rethrow so onNewTx does not flip the tx to FINISHED with owned shielded
+      // outputs left uncredited: leaving it PROCESSING keeps it retryable on the
+      // next reload/sync, and the WS queue survives via enqueueOnNewTx's .catch.
       storage.logger.error(
         'Unexpected error processing shielded outputs for tx',
         tx.tx_id,
         '- wallet may be missing shielded funds.',
         e
       );
+      throw e;
```
tuliomir: issue: This seems like an important edge case to unit test, and not delegate to next PRs, as it has the potential to stop the wallet from serving the client entirely.

Also, the rethrow's blast radius reaches processHistory, not just onNewTx

The comment above reasons only about the `onNewTx` path — "the WS queue survives via enqueueOnNewTx's .catch". That holds. But `processNewTx` is also called from `processHistory` (src/utils/storage.ts:561), inside an unguarded `for await` over the entire history.

The systemic causes named here — wrong PIN, missing/corrupt scan key — are exactly the ones that would hit *every* owned shielded tx during a full history load. The first one now aborts the whole walk: `updateWalletMetadataFromProcessedTxData` never runs, and the wallet is left with partially-processed state.

Consider catching at the `processHistory` call site (log + continue, so one bad tx doesn't strand the load) and letting the rethrow serve only the `onNewTx` retry path it was designed for. Either way the comment should state which callers it has reasoned about.

src/new/sendTransaction.ts
```diff
@@ -306,13 +567,214 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+      valueBlindingFactor: Buffer;
+      generatorBlindingFactor: Buffer;
+    }> = [];
+    // Transparent non-authority inputs collected separately. Needed only for the
+    // excess-blinding-factor calc on full-unshield txs where the wallet holds a
+    // mix of transparent + shielded UTXOs of the same token: the fullnode's
+    // balance verifier sums ALL inputs (transparent + shielded) against all
+    // outputs, so computeBalancingBlindingFactor must see the transparent
+    // inputs too or it returns a bf that doesn't satisfy the equation (the
+    // fullnode then panics when it tries to build the excess commitment).
```
tuliomir: suggestion: The word `panic` here is not very descriptive of what actually happens in the fullnode. Does it spend excessive time computing and does not find a valid solution? Or it just rejects it?

I'd suggest rewording this part for more clarity.

src/new/sendTransaction.ts
```diff
@@ -245,13 +408,68 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+            {},
+            { legacy: false }
+          );
+          const addressObj = new Address(shieldedAddress, { network });
+          if (!addressObj.isShielded()) {
+            throw new SendTxError(
+              'Wallet did not return a shielded address for custom-token change conversion.'
+            );
+          }
+          shieldedOutputDefs.push({
```
tuliomir, line 470: issue: converted fee-token change escapes `FEE_PER_OUTPUT`

Explicit shielded outputs push a phantom into `txData.outputs` (line 305), so `Fee.calculate` counts them *and* they get the shielded fee. This converted change is removed before `Fee.calculate`, so it's charged only the shielded fee.

Same kind of output, two fee treatments. If the fullnode charges `FEE_PER_OUTPUT` on shielded fee-token outputs, this underpays by exactly that and the tx is rejected.

**suggestion:** confirm the fullnode rule, then align both paths. Either way, add a test with a `TokenVersion.FEE` custom token plus `changeShieldedMode`.

src/new/sendTransaction.ts
```diff
@@ -745,6 +1207,145 @@ export async function prepareSendTokensData(
...
+    // (changeValue + pulled) > additionalFee, leaving a positive shielded
+    // value after subtracting the fee.
+    const deficit = additionalFee - changeValue;
+    const pulledInputs: IDataInput[] = [];
+    let pulledSum = 0n;
+    const selectOptions: IUtxoFilterOptions = {
+      token: HTR_UID,
+      authorities: 0n,
+      only_available_utxos: true,
+      order_by_value: 'desc',
```
tuliomir: issue: `desc` ordering sweeps the largest UTXO into shielded change

The deficit here is at most `FEE_PER_FULL_SHIELDED_OUTPUT`, but `order_by_value: 'desc'` grabs the biggest HTR UTXO available. All of it flows into `changeValue` and lands in the shielded output at line 1339.

A user shielding a trivial change can have most of their HTR balance silently shielded.

**suggestion:** use `'asc'` so the smallest sufficient UTXO covers the deficit.

src/new/sendTransaction.ts
```diff
@@ -745,6 +1207,145 @@ export async function prepareSendTokensData(
...
+        'HTR change is too small to fund its shielded-output fee and no additional ' +
+          'HTR is available to cover the difference.'
+      );
+    }
+
+    partialHtrTxData.inputs.push(...pulledInputs);
+    changeValue += pulledSum;
+  }
+
+  const { address: shieldedAddress } = await wallet.getCurrentAddress({}, { legacy: false });
```
tuliomir, line 1382: issue: every change output shares one shielded address

`getCurrentAddress` is called here and again inside the custom-token loop at line 448, with nothing advancing the index between calls. N change outputs all resolve to the same shielded address.

For a privacy feature this is a meaningful leak — it links the change outputs to each other.

**suggestion:** advance the address after each conversion, or derive a fresh one per change output.

src/new/sendTransaction.ts
```diff
@@ -745,6 +1207,145 @@ export async function prepareSendTokensData(
...
+    };
+    for await (const utxo of storage.selectUtxos(selectOptions)) {
+      pulledInputs.push(helpers.getDataInputFromUtxo(utxo));
+      pulledSum += utxo.value;
+      if (pulledSum > deficit) {
+        break;
+      }
+    }
+
+    if (pulledSum <= deficit) {
```
tuliomir, line 1365: suggestion (non-blocking): consider a fallback instead of aborting

When no spare HTR exists, the whole send fails. An alternative is to skip shielding this change and leave it transparent — the tx still balances, the user just loses privacy on the change.

Worth deciding explicitly: silent privacy downgrade vs. hard failure. If the hard failure is intentional, a short comment stating why would help.

__tests__/new/sendTransaction.test.ts
```diff
@@ -543,3 +643,638 @@ describe('releaseUtxos', () => {
...
+    // Transparent change removed.
+    expect(partialHtrTxData.outputs).toHaveLength(0);
+    // Shielded HTR change appended.
+    expect(defs).toHaveLength(3);
+    const htrChange = defs[2];
+    expect(htrChange.token).toBe(NATIVE_TOKEN_UID);
+    expect(htrChange.shieldedMode).toBe(ShieldedOutputMode.FULLY_SHIELDED);
+    expect(htrChange.value).toBe(100n - FEE_PER_FULL_SHIELDED_OUTPUT);
+    // scanPubkey is hex-encoded 33 bytes (66 hex chars).
+    expect(htrChange.scanPubkey).toMatch(/^[0-9a-f]{66}$/);
```
tuliomir: issue: scanPubkey is asserted by shape only — a scan/spend swap survives the suite

This regex is the only `scanPubkey` assertion in the file, and no test asserts the converted def's `address` or `shieldedAddress` at all. `buildShieldedAddress()` (line 658-662) generates both keys and discards them, so nothing binds the emitted pubkey to the address the wallet returned.

Swap `getScanPubkey()` and the spend pubkey at src/new/sendTransaction.ts:1338/1341 (or 459/462). Both are 33-byte compressed pubkeys, so both match `/^[0-9a-f]{66}$/`, values and def counts still pass, and the suite stays green — while every shielded change output in production becomes undetectable by the receiving wallet. The change would be permanently unrecoverable.

Build the address from known keys and assert `def.scanPubkey === knownScanPubkey.toString('hex')` and `def.address === Address(knownShieldedAddr).getSpendAddress().base58`. Cheapest fix here, highest consequence if wrong.

src/new/types.ts
```diff
@@ -399,12 +402,23 @@ export interface SendTransactionFullnodeOptions {
  * @property changeAddress Address for change output
  * @property startMiningTx Boolean to trigger start mining (default true)
  * @property pinCode Pin to decrypt xpriv information
+ * @property changeShieldedMode If set, the transparent HTR change output
+ *   (the surplus over any HTR being sent plus ALL fees — fees are always
+ *   charged in HTR, including per-shielded-output fees) is rewritten as a
+ *   shielded HTR output in the given mode. Mirrors the user-selected
+ *   privacy mode of the transaction so HTR change doesn't leak the
+ *   sender alongside an otherwise-private send. Only the HTR change is
```
tuliomir: issue: the changeShieldedMode JSDoc contradicts the implementation on two counts

"Only the HTR change is converted — custom-token change always stays transparent" is false: custom-token change is converted at src/new/sendTransaction.ts:437-471.

"the conversion is best-effort: change too small to fund its own shielded-output fee is kept transparent" is also false, and this is the problem one — src/new/sendTransaction.ts:1314-1319 *throws* a `SendTxError`. A caller who read this doc will not expect `sendManyOutputs` to reject outright on a small-change send with no spare HTR.

This is the public API doc for the option. Either fix the doc to match the throwing behavior, or make the code match the doc.

src/utils/tokens.ts
```diff
@@ -501,7 +501,14 @@ const tokens = {
 
       // get output change
       if (foundAmount > requiredAmount) {
-        const cAddress = await storage.getChangeAddress({ changeAddress });
+        // A shielded change address has no direct output script; resolve it to
+        // its spend-derived P2PKH (same as explicit shielded outputs in the send
+        // pipeline). Without this, getAddressType throws for a shielded
+        // changeAddress and token creation with a shielded change fails.
+        const cAddress = resolveOutputScriptAddress(
```
tuliomir, line 511: issue: a shielded change address is flattened to a transparent spend-chain P2PKH

This resolves the crash, but the resulting HTR deposit-change output is a plain transparent P2PKH at the spend-derived address — not a shielded output. A caller passing a shielded change address to `prepareMintTxData` reasonably expects shielded change and gets the opposite.

Worse than a normal change address: it publishes a spend-chain address in the clear on-chain, which is linkable to that wallet's shielded spends. The comment's analogy to "explicit shielded outputs in the send pipeline" doesn't hold — there the P2PKH is a *phantom* replaced by a real shielded output at line 573; here it is the final output.

Either reject a shielded `changeAddress` here with a clear error, or route it through the same conversion the send pipeline uses.

src/new/sendTransaction.ts
```diff
@@ -306,13 +567,214 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+        );
+      }
+      if (shieldedOutputDefs.length > MAX_SHIELDED_OUTPUTS) {
+        throw new SendTxError(
+          `Cannot create more than ${MAX_SHIELDED_OUTPUTS} shielded outputs per transaction ` +
+            `(requested ${shieldedOutputDefs.length}).`
+        );
+      }
+
+      shieldedOutputs = await createShieldedOutputs(
```
r4mmer, line 704: question: Should we shuffle the shielded outputs?

The idea of shuffling is to make it harder to identify which address belongs to the wallet of the sender.
Since shielded outputs make it impossible to know the amount (or tokens) even if we know the address I think it is safe to not shuffle but if the possibility of knowing which addresses belong to a wallet is unwanted maybe we should.

src/new/sendTransaction.ts
```diff
@@ -306,13 +567,214 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+    // inputs too or it returns a bf that doesn't satisfy the equation (the
+    // fullnode then panics when it tries to build the excess commitment).
+    const transparentInputEntries: Array<{
+      value: bigint;
+      valueBlindingFactor: Buffer;
+      generatorBlindingFactor: Buffer;
+    }> = [];
+
+    for (const inp of allInputs) {
+      const utxo = await this.storage.getUtxo({
```
r4mmer, line 626: question: This means all inputs MUST be from our wallet, correct?

If an input is from another's wallet this will return null or undefined I believe.

src/new/sendTransaction.ts
```diff
@@ -745,6 +1207,145 @@ export async function prepareSendTokensData(
...
+    const pulledInputs: IDataInput[] = [];
+    let pulledSum = 0n;
+    const selectOptions: IUtxoFilterOptions = {
+      token: HTR_UID,
+      authorities: 0n,
+      only_available_utxos: true,
+      order_by_value: 'desc',
+      filter_method: (utxo: IUtxo) => !usedUtxos.has(`${utxo.txId}:${utxo.index}`),
+    };
+    for await (const utxo of storage.selectUtxos(selectOptions)) {
```
r4mmer, line 1357: When the user supplies the inputs we do not choose more utxos, even if the transaction should fail.
Here this contract is broken and we may end up choosing more utxos.

Should we enforce this "implicit" rule?

src/new/sendTransaction.ts
```diff
@@ -187,11 +268,52 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+            `Shielded output requires a shielded address, got '${output.address}'.`
+          );
+        }
+        const spendBase58 = shieldedAddressObj.getSpendAddress().base58;
+        shieldedOutputDefs.push({
+          address: spendBase58,
+          value: output.value,
+          token: output.token,
+          scanPubkey: shieldedAddressObj.getScanPubkey().toString('hex'),
+          shieldedMode: output.shieldedMode,
```
r4mmer, line 301: This value is never validated from the user input, only the "existence" of it through `isShieldedOutput`.
Maybe we should throw if a `shieldedMode` other than the accepted are given.

r4mmer (changes requested):

src/utils/storage.ts
pedroferreira1: Good catch — you're right the rethrow reaches processHistory, not just onNewTx. processHistory now wraps each processNewTx in a try/catch that logs and continues, so a systemic decode failure (wrong PIN / corrupt scan key) can't abort the whole reload after cleanMetadata() has wiped the metadata. The rethrow stays, but its comment now scopes it to the realtime caller it was designed for (processSingleTx ← onNewTx's isNewTx path); the reload callers (processHistory: startup / scanAddressesToLoad / voided-flip reprocess) catch per-tx and continue. Added a regression test that walks a two-tx history whose shielded decode fails systemically and asserts the walk resolves and skips both txs instead of aborting.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/commits/4d58b8c9911fc7f8fc50595af99327e5b79bb85d

src/new/sendTransaction.ts
pedroferreira1: Reworded to "the fullnode rejects the tx: its balance verification fails when it builds the excess commitment."

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/commits/4d58b8c9911fc7f8fc50595af99327e5b79bb85d

src/new/sendTransaction.ts
pedroferreira1: Fixed — `order_by_value` is now 'asc'. The loop pulls smallest-first and stops as soon as the pulled sum exceeds the deficit (at most FEE_PER_FULL_SHIELDED_OUTPUT), so it shields the least extra HTR rather than sweeping the largest UTXO into shielded change. Kept the accumulate-until-sufficient loop instead of a min-value filter, which would spuriously fail when no single UTXO covers the deficit but several small ones would.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/commits/4d58b8c9911fc7f8fc50595af99327e5b79bb85d

src/new/sendTransaction.ts
pedroferreira1: This is the wallet's existing change-address behavior, not something the shielded path introduces. getChangeAddress resolves to getCurrentAddress() (non-advancing), and normal transparent change uses the exact same changeAddress || getCurrentAddress() — so multiple change outputs already share the current address regardless of shielding. Making shielded change advance the index while transparent change doesn't would be inconsistent across the wallet.

src/new/sendTransaction.ts
pedroferreira1: This hard failure is deliberate: the caller opted into shielding the change, so silently leaving it transparent to rescue the send would publish the change (and link the sender) against their stated privacy intent. We fail loudly so the caller can decide (send less, consolidate HTR, or drop changeShieldedMode). Added a comment at the throw site stating this, and updated the changeShieldedMode JSDoc to document the throwing behavior.

__tests__/new/sendTransaction.test.ts
pedroferreira1: Agreed — cheapest fix, highest consequence. HTR change and custom-token change now build the address from known scan/spend keys and assert def.scanPubkey === knownScanPubkey.toString('hex') and def.address === Address(knownShieldedAddr).getSpendAddress().base58, binding both conversion sites (1338/1341 and 459/462). Verified the binding catches the exact swap you described: swapping getScanPubkey() → the spend pubkey at both sites turns both tests red; reverting turns them green.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/commits/4d58b8c9911fc7f8fc50595af99327e5b79bb85d

src/new/types.ts
pedroferreira1: Fixed — the JSDoc now states that (1) custom-token change is converted when the tx already carries explicit shielded outputs, (2) on a pure-transparent send it's a no-op, and (3) the send throws SendTxError when the HTR change can't fund its shielded-output fee with no spare HTR (no silent transparent downgrade).

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/commits/4d58b8c9911fc7f8fc50595af99327e5b79bb85d

src/utils/tokens.ts
pedroferreira1: The premise doesn't quite hold given the address model: the 71-byte address is the wallet's universal format used for both transparent and shielded outputs, and the transaction layer resolves any output address to its script uniformly at build time (see getChangeAddress, storage.ts). Token-creation deposit change is transparent by design — shielded token-creation change is explicitly out of scope for this PR — so resolving the 71-byte change address to its spend-derived P2PKH for a transparent output is the intended uniform behavior, not a shielded→transparent downgrade; there's no shielded output being flattened. You're right the comment was misleading, though — it framed this as mirroring the send-pipeline's shielded-output conversion (where the P2PKH is a phantom replaced by a real shielded output). I reworded both resolveOutputScriptAddress's docstring and the token-creation comment to describe it as the uniform address→script resolution producing a transparent output.

src/new/sendTransaction.ts
pedroferreira1: Good point — `isShieldedOutput` only checks the field exists. The value is now validated against the accepted ShieldedOutputMode values with a clear SendTxError at prepare time, instead of failing opaquely in the crypto layer.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/f81e53289622af5c8237160ce1b0c2b720afe5b9

src/new/sendTransaction.ts
pedroferreira1: Agreed. `convertHtrChangeIfRequested` now takes whether HTR auto-selection is enabled; when the HTR inputs were user-supplied it throws instead of pulling extra UTXOs, honoring the "user supplies inputs → wallet chooses nothing more" contract. Auto-select sends are unchanged. Test asserts that even with a spare UTXO available, a user-supplied send throws rather than pulling it.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/f81e53289622af5c8237160ce1b0c2b720afe5b9

src/new/sendTransaction.ts
pedroferreira1: Correct, and they already are — enforced upstream of this loop. A user-supplied transparent input whose output isn't isAddressMine is rejected by checkUnspentInput ("Output [N] of transaction [T] is not from the wallet"), and a non-owned/undecoded shielded input is rejected at the resolve step. So by the time this loop calls getUtxo, every input is wallet-owned and getUtxo is always found — the assumption holds, and it fails loudly (not a silent skip) if violated. Added a test covering the shielded path.

src/new/sendTransaction.ts
pedroferreira1: Done — shielded outputs are now shuffled before creation, matching the transparent-output shuffle; a test asserts the shuffle result is actually used.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/f81e53289622af5c8237160ce1b0c2b720afe5b9

src/new/sendTransaction.ts
tuliomir: Agreed that we should keep the code consistent. Opened #1136 to discuss changing the behavior of the entire wallet on this regard.

src/utils/storage.ts
```diff
@@ -558,16 +558,30 @@ export async function processHistory(
...
+        nowTs,
+        currentHeight,
+        pinCode,
+      });
+      legacyMaxIndexUsed = Math.max(legacyMaxIndexUsed, processedData.legacyMaxAddressIndex);
+      shieldedMaxIndexUsed = Math.max(shieldedMaxIndexUsed, processedData.shieldedMaxAddressIndex);
+      for (const token of processedData.tokens) {
+        tokens.add(token);
+      }
+    } catch (e) {
```
tuliomir, line 575: issue: this catch is unscoped — it swallows every processNewTx failure, not just the shielded-decode rethrow it was added for.

`processNewTx` also throws for store write failures inside creditOutput (saveUtxo, editTokenMeta/editAddressMeta) and for corrupt nano/OCB history entries. All of those now become silent per-tx skips.

Failure scenario: an IndexedDB quota or lock error during reload makes every saveUtxo throw — every tx is "skipped", processHistory resolves cleanly, and the wallet goes READY with empty balances and only log lines. That is the exact stranded-empty-wallet outcome this commit set out to prevent, except now it is silent instead of loud.

Suggestion: throw a typed error (e.g. ShieldedDecodeSystemicError) at the systemic-rethrow site and skip here only on instanceof; rethrow everything else.

src/utils/storage.ts
```diff
@@ -558,16 +558,30 @@ export async function processHistory(
...
+      // rethrows a SYSTEMIC shielded-decode failure (wrong PIN / missing scan
+      // key) for the realtime onNewTx path so it stays retryable; but here we
+      // are re-walking the whole history after cleanMetadata() already wiped the
+      // metadata, so aborting would leave the wallet with empty balances. Log
+      // and skip this tx — the rest of the history still rebuilds its metadata,
+      // and the skipped tx is re-processed on the next reload (once the
+      // PIN/scan key is available). Skipping here loses that tx's transparent
+      // credit too, but only until the next reload, whereas aborting loses
+      // everything now.
+      storage.logger.error('Error processing tx during history reload, skipping', tx.tx_id, e);
```
tuliomir: issue: a skip leaves no surfaced signal — the wallet stays wrong indefinitely while reporting READY.

The comment above says the skipped tx "is re-processed on the next reload", but nothing guarantees one: there is no per-tx PROCESSING marker on this path (that retry mechanism exists only on the realtime path), no emitted event, and no PIN-unlock trigger that re-runs processHistory. A reload with a wrong PIN skips every shielded tx and the wallet reports READY with understated balances until the user happens to restart.

Suggestion: count skips and, when greater than zero, surface it — emit a wallet event or set a "partial history" flag — instead of relying on logger.error alone.

src/utils/storage.ts
```diff
@@ -1119,15 +1133,23 @@ export async function processNewTx(
         await store.saveTx(tx);
       }
     } catch (e) {
-      // processShieldedOutputs handles per-output rewind failures internally.
-      // Reaching here means something unexpected went wrong at the
-      // infrastructure level.
+      // processShieldedOutputs handles per-output rewind failures internally, so
+      // a throw here is a SYSTEMIC failure (wrong PIN, missing/corrupt scan key).
+      // Rethrow for the REALTIME single-tx caller (processSingleTx <- onNewTx's
```
tuliomir: issue: with the reload catch unscoped, the skip is not atomic for non-decode errors.

The shielded-decode throw fires before any crediting, so the intended skip is clean. But creditOutput persists per output, and the nano throws fire after balance edits but before the numTransactions bumps — so a mid-loop saveUtxo failure leaves outputs 0..k credited, the tx "skipped", and metadata internally inconsistent (partial credit, missing counters), which is worse than merely missing a tx.

This falls out of the catch-scoping issue: narrowing the reload catch to the pre-credit decode failure eliminates it.

src/utils/storage.ts
```diff
@@ -558,16 +558,30 @@ export async function processHistory(
...
-    for (const token of processedData.tokens) {
-      tokens.add(token);
+    try {
+      const processedData = await processNewTx(storage, tx, {
+        rewardLock,
+        nowTs,
+        currentHeight,
+        pinCode,
+      });
+      legacyMaxIndexUsed = Math.max(legacyMaxIndexUsed, processedData.legacyMaxAddressIndex);
```
tuliomir, line 570: suggestion: a skipped tx silently drops its address-index and token contributions, shrinking discovery.

A skipped tx's legacyMaxAddressIndex / shieldedMaxAddressIndex / tokens never reach updateWalletMetadataFromProcessedTxData. If the skipped tx is the only one touching the highest used address index, gap-limit scanning caps short and funds on later indexes are never loaded — and a mixed tx failing only on the shielded decode also loses its transparent addresses' index advance.

Worth folding into whatever surfacing mechanism the skip gets: a skip that also shrank the scan range is strictly more urgent to retry.

__tests__/utils/storage.test.ts
```diff
@@ -216,6 +216,73 @@ describe('processHistory — orchestration', () => {
     await processHistory(storage);
     expect(cleanSpy).toHaveBeenCalled();
   });
+
+  it('skips (does not abort on) a tx whose shielded decode fails systemically, and keeps walking', async () => {
```
tuliomir: suggestion: this test pins the mechanism (skip + continue + log) but not the rationale the code comment promises.

Both fixture txs have empty outputs, so the "loses that tx's transparent credit too" trade-off is unexercised; and both txs fail, so "the rest of the history still rebuilds its metadata" is never asserted — no healthy tx follows a failed one with an assertion that its balance and UTXOs survived.

Also, no test specifies what happens to a NON-shielded error during reload, leaving the catch's breadth entirely unpinned. Adding a healthy-tx-after-failed-tx case and a non-decode-error case would close both gaps.

src/new/sendTransaction.ts
```diff
@@ -747,6 +1242,167 @@ export async function prepareSendTokensData(
...
+    // Ascending value: the whole pulled sum flows into the shielded change
+    // (changeValue += pulledSum), so pulling smallest-first — and stopping as
+    // soon as pulledSum > deficit (a tiny value, <= FEE_PER_FULL_SHIELDED_OUTPUT)
+    // — shields the least extra HTR. 'desc' would sweep the largest UTXO into
+    // the shielded change, silently shielding most of the wallet's HTR balance.
+    const selectOptions: IUtxoFilterOptions = {
+      token: HTR_UID,
+      authorities: 0n,
+      only_available_utxos: true,
+      order_by_value: 'asc',
```
tuliomir, line 1354: suggestion: the desc-to-asc fix is itself unpinned — a silent revert stays green.

order_by_value appears nowhere in the test suite, and the pull-loop fixtures offer a single pullable UTXO, so neither the ordering nor the break is observable. A regression back to 'desc' — the behavior that swept the wallet's largest HTR UTXO into the shielded change — would not fail any test.

Suggestion: one fixture with several pullable UTXOs of different values, asserting the smallest sufficient set was chosen.

src/utils/storage.ts
pedroferreira1: Valid and the subtlest one.

- **Signal (done):** the `shieldedDecodeSkippedTxIds` flag + summary log now flag that discovery may be short — this is the "fold into the signal" fallback you suggested, so a skip that shrank the scan range is at least visible.

After we log the error, the wallet can retry it.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/2d52cc1b39f3a7c76adfafead95bb3145b64f81d

src/new/sendTransaction.ts
pedroferreira1: Agreed. Added `H.4e — pulls the SMALLEST sufficient HTR UTXO (order_by_value asc), not the largest`: the mock storage now honors `order_by_value`, the fixture offers several pullable UTXOs of different values, and the test asserts the smallest-sufficient one is pulled — it fails on a silent revert to `'desc'`.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/2d52cc1b39f3a7c76adfafead95bb3145b64f81d

__tests__/utils/storage.test.ts
pedroferreira1: Agreed. Strengthened the block:

- `processes a healthy tx that follows a skipped one — the rest of the history still rebuilds` — a healthy transparent tx after a skipped shielded one, asserting its UTXO survives (the "rest rebuilds" claim).
- `rethrows (aborts) a NON-decode error during reload — store failures are not swallowed` — a generic store failure aborts instead of being skipped, directly validating ①'s scoping.
- the systemic-skip test now also asserts `shieldedDecodeSkippedTxIds` (②).

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/2d52cc1b39f3a7c76adfafead95bb3145b64f81d

src/utils/storage.ts
pedroferreira1: Agreed, and the same commit resolves it. The shielded-decode throw fires **before** any `creditOutput`, so the scoped skip is atomic — nothing is half-credited. The mid-loop store/nano failures that could leave outputs `0..k` credited are no longer in the skip path at all: they're not a `ShieldedDecodeSystemicError`, so they rethrow and abort the walk instead of being swallowed. Noted this in the catch comment.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/2d52cc1b39f3a7c76adfafead95bb3145b64f81d

src/utils/storage.ts
pedroferreira1: Agreed — a silent skip with no signal is the wrong-PIN-reload footgun from the known deferred real-time-decode limitation. Added `storage.shieldedDecodeSkippedTxIds` (on `IStorage`): after the walk it holds the list of skipped tx ids, or `null` when there were none, plus a summary error log. That gives the wallet a "partial history" flag to surface, so an understated-balance reload is visible instead of silent. (Emitting a wallet event off this flag can ride on the same follow-up as the real-time-decode fix.)

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/2d52cc1b39f3a7c76adfafead95bb3145b64f81d

src/utils/storage.ts
pedroferreira1: Agreed — the catch-all was the wrong call, and your store-write scenario is exactly the failure mode: every `saveUtxo` throw would be swallowed as a per-tx "skip", the walk would still resolve, and the wallet would go READY with empty balances and only log lines — the precise stranded-wallet outcome this skip was meant to prevent.

Fixed: the shielded-decode rethrow site now throws a typed `ShieldedDecodeSystemicError` (`src/errors.ts`), and `processHistory` catches **only** `instanceof ShieldedDecodeSystemicError` — every other error (store write, corrupt nano/OCB entry) is rethrown, restoring loud-fail. New test `rethrows (aborts) a NON-decode error during reload — store failures are not swallowed` pins it.

https://github.com/HathorNetwork/hathor-wallet-lib/pull/1131/changes/2d52cc1b39f3a7c76adfafead95bb3145b64f81d

tuliomir (approved):

r4mmer (approved):

PR #1132, opened 2026-07-15
fix(shielded): compliant deriveChild for shielded keys + per-address balance Map

pedroferreira1:
### Summary

Two focused fixes to the shielded wallet foundation. This PR is part of the shielded-outputs stack and is based on PR 6c (shielded/pr-6-wallet-lifecycle).

1. Compliant deriveChild for shielded scan/spend keys.
The shielded scan (account 1') and spend (account 2') private keys were derived with bitcore's deriveNonCompliantChild, while their public counterparts already used the compliant deriveChild. That asymmetry is silently wrong for the ~1/256 of indices affected by the historical bitcore serialization bug: at those indices the private key's pubkey does not match the address, breaking the ECDH scan keypair and the spend signing key (an unspendable / undetectable shielded output). Shielded keys are new — there is no legacy on-chain material to stay bug-compatible with — so both the public and private sides now use deriveChild. Legacy P2PKH/P2SH chains are untouched and keep deriveNonCompliantChild for backward compatibility with existing funds.

2. Per-address balance Map (getDefaultAddressMeta).
export const DEFAULT_ADDRESS_META created a single balance Map at module load, and every { ...DEFAULT_ADDRESS_META } shallow-spread copied that same Map reference into each address's metadata. A balance.set(...) for one address therefore mutated a Map shared by every address — and every wallet instance in the process — leaking balances across addresses and surviving wallet reloads. Replaced with a getDefaultAddressMeta() factory that returns a fresh Map per call.

### Acceptance Criteria

- Shielded scan and spend keys derive with compliant BIP32 (deriveChild) on both the public and private sides, so the derived keypair always matches the shielded address (no index-dependent asymmetry).
- Legacy P2PKH/P2SH and wallet-service auth derivations are unchanged (deriveNonCompliantChild), preserving existing-fund compatibility.
- Each address's metadata owns a private balance Map, so per-address balances no longer leak across addresses or across wallet instances.

### Security Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shielded wallet key derivation to use the standard BIP32 method, ensuring consistent scan and spend keys.
  * Fixed address metadata initialization so balances are kept separate across addresses and wallet instances.
  * Updated shielded access-data migration and transaction signing to use the corrected key derivation behavior.


src/shielded/processing.ts
```diff
@@ -56,10 +56,11 @@ function deriveScanChildPrivkey(
...
   try {
-    // deriveNonCompliantChild is required for private keys due to a historical
-    // bitcore-lib serialization bug. Public key derivation (in shieldedAddress.ts)
-    // uses standard deriveChild because it was always correct. Do not align these.
-    const childKey = scanHdPrivKey.deriveNonCompliantChild(addressIndex);
+    // Compliant BIP32 derivation, matching the scan PUBLIC key put into the
+    // shielded address (shieldedAddress.ts). Shielded keys are new — there is no
+    // legacy key material to stay bug-compatible with — so unlike the legacy
+    // P2PKH chain (still on deriveNonCompliantChild) they use the correct method.
+    const childKey = scanHdPrivKey.deriveChild(addressIndex);
```
tuliomir, line 63: suggestion(non-blocking): add a regression test pinning the priv/pub derivation agreement this fix restores.

The bug being fixed — scan private key derived non-compliantly while the address's scan pubkey derived compliantly, diverging at ~1/256 indices — has no test asserting the invariant, so reintroducing the asymmetry on the private side would keep the suite green while making outputs undetectable/unspendable at the divergent indices.

A unit test asserting, across a window of indices (or at a seed/index engineered to hit the divergence), that this child privkey's pubkey equals `deriveShieldedAddress(...).scanPubkey` — and likewise for the spend key in `src/utils/transaction.ts` — would pin the private side too.

src/storage/storage.ts
```diff
@@ -67,10 +67,22 @@ import {
...
+ * Build a fresh address-metadata object.
+ *
+ * The `balance` Map MUST be a new instance per call. The previous
+ * `export const DEFAULT_ADDRESS_META` created ONE Map at module load, and every
+ * `{ ...DEFAULT_ADDRESS_META }` shallow-spread copied that same Map REFERENCE
+ * into each address's metadata — so a `balance.set(...)` for one address mutated
+ * a Map shared by every address AND every wallet instance in the process,
+ * leaking balances across addresses and surviving wallet reloads.
+ */
+export function getDefaultAddressMeta(): IAddressMetadata {
```
tuliomir, line 80: suggestion(non-blocking): add a regression test for the balance-Map aliasing this factory fixes.

No test references `getDefaultAddressMeta`, so a future revert to a shared module-level Map (or a `{ ...someDefault }` spread of one) would land green while re-leaking balances across addresses and wallet instances.

A cheap pin: obtain metadata for two addresses, `balance.set(...)` on one, assert the other's Map is unaffected — plus assert two `getDefaultAddressMeta()` calls return distinct `Map` instances.

src/utils/wallet.ts
```diff
@@ -709,10 +709,10 @@ const wallet = {
 
     // Derive shielded scan (account 1') and spend (account 2') keys.
     // Separate from legacy (account 0') so scan key only grants view access.
-    const scanAcctXpriv = rootXpriv.deriveNonCompliantChild(SHIELDED_SCAN_ACCT_PATH);
-    const scanXpriv = scanAcctXpriv.deriveNonCompliantChild(0);
-    const spendAcctXpriv = rootXpriv.deriveNonCompliantChild(SHIELDED_SPEND_ACCT_PATH);
-    const spendXpriv = spendAcctXpriv.deriveNonCompliantChild(0);
+    const scanAcctXpriv = rootXpriv.deriveChild(SHIELDED_SCAN_ACCT_PATH);
```
tuliomir, line 712: issue: the committed shielded address fixtures were not regenerated — the drift-check suite fails on this branch.

Switching this derivation to `deriveChild` changes the scan/spend xpubs for the fixed test seeds, but `__tests__/integration/configuration/precalculated-shielded-addresses.ts` still holds the old-derivation values. Running the drift-check locally at head 2315427c fails 2 of 3 tests — e.g. genesis index 0 expects `K3LV1px1…` (fixture) but derives `K3MdUaNx…` live.

This lands red silently: the stacked PR's CI ran only CodeRabbit, so no test workflow executed. Please regenerate the fixtures per the instructions in the fixture file's header and include them in this PR.

src/utils/transaction.ts
```diff
@@ -430,7 +430,7 @@ const transaction = {
           const spendXprivStr = await storage.getSpendXPrivKey(pinCode);
           spendXprivkey = HDPrivateKey.fromString(spendXprivStr);
         }
-        derivedKey = spendXprivkey.deriveNonCompliantChild(addressInfo.bip32AddressIndex);
+        derivedKey = spendXprivkey.deriveChild(addressInfo.bip32AddressIndex);
```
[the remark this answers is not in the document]
pedroferreira1: Fixed. Switched both `spendXpriv.deriveNonCompliantChild(idx)` assertions (the pubkey check and the signature-verify check) to `deriveChild(idx)`, matching the shielded-spend signing path. Left the `mainXpriv.deriveNonCompliantChild(idx)` comparison untouched — that's the legacy/main chain, which stays non-compliant, and that assertion is exactly "the shielded pubkey is NOT the legacy one."

src/shielded/processing.ts
pedroferreira1: Added — a new describe block in `__tests__/utils/shieldedAddress.test.ts` asserting the private scan/spend child pubkey (as processing.ts / transaction.ts derive it) equals `deriveShieldedAddress(...).scanPubkey`/`.spendPubkey` across a window of indices.

One clarification from tracing the mechanism: `deriveChild` and `deriveNonCompliantChild` are identical for **non-hardened** children — they only diverge at **hardened** steps, so the divergence is per-**seed** (at the hardened account path `m/44'/280'/1'`), not "per index". To give the test real teeth I picked a seed (fill byte `0xA4`) whose hardened scan/spend account paths DO diverge between the two methods, plus an assertion that the non-compliant account derivation yields a different xpub — so a revert can't sneak through green. Mutation-tested: reverting the production shielded derivation to `deriveNonCompliantChild` turns both `wallet.test.ts` and the drift-check red (the drift-check fails exactly 2/3, matching your original report).

Framing note: this path is new/unreleased, so this isn't a production bug fix or a compat concern — `deriveChild` is simply bitcore's correct method, used from the start (the legacy chains keep `deriveNonCompliantChild` because they have deployed key material). The regression test just guards a future edit from re-introducing the asymmetry.

src/storage/storage.ts
pedroferreira1: Added — a new test in `__tests__/storage/storage.test.ts`: two `getDefaultAddressMeta()` calls return distinct `balance` Map instances, and a `balance.set(...)` on one is not visible on the other. A revert to a shared module-level Map (or a `{ ...someDefault }` spread of one) fails it instead of silently re-leaking balances across addresses/wallets.

src/utils/wallet.ts
pedroferreira1: Confirmed and fixed. Verified on the branch: live derivation gives `K3LV1px1…` for genesis[0] while the committed fixture was `K3MdUaNx…`, so the drift-check failed 2/3 (the structural test passes; the two live-comparison tests fail). Regenerated all 8 seeds × 22 indices from live derivation — the drift-check now passes 3/3.

On "lands red silently": the drift-check is genuinely network-free (I ran it under a bare jest config with no integration setup and it passes), but it lives in `__tests__/integration/`, so the unit config ignores it and only the docker-gated integration workflow runs it — which didn't run (CI = CodeRabbit only). Regenerating fixes this instance; a solid follow-up is to **move the drift-check into the unit suite** so any future derivation change is caught on every PR.

tuliomir (dismissed):

r4mmer (dismissed):

r4mmer (approved):

tuliomir (approved):

PR #1135, opened 2026-07-22
refactor: extract reusable signTxInputs from getSignatureForTx

pedroferreira1:
### Description:

External signers that plug in via the `txSignFunc` hook (`setExternalTxSigningMethod`) — e.g.
a passkey signer that derives keys from a biometric ceremony instead of decrypting a stored
key — currently have to reimplement `transactionUtils.getSignatureForTx`'s per-input loop,
because `getSignatureForTx` sources its signing key from `storage.getMainXPrivKey(pin)`, which a
keyless wallet cannot provide. Such copies drift: the Hathor mobile passkey signer's copy has
already fallen behind (no shielded-spend key chain; the positional spent-output lookup this code
explicitly warns is wrong for shielded inputs).

### Acceptance Criteria
- `transactionUtils.signTxInputs(tx, storage, getXpriv)` is exported and produces
      `ITxSignatureData` (input signatures + nano/OCB caller signature).
- `getSignatureForTx` delegates to `signTxInputs` and its behavior is unchanged — all existing
      signing tests pass, including the shielded-spend and non-caller nano tests.

### Security Checklist

## Summary by CodeRabbit

- **New Features**
  - Token creation can proceed without a PIN when an external transaction-signing method is configured, including nano contract token creation.
  - External signing now supports token creation for xpub-only wallets.

- **Bug Fixes**
  - Token creation no longer fails early due to a missing PIN when external signing is available (PIN still required otherwise).
  - Transaction signing now correctly derives/signs inputs by type (e.g., `shielded-spend`) using resolver-provided keys.

- **Tests**
  - Added integration coverage for fullnode token creation with no PIN under external signing.
  - Added unit tests for PIN-handling and correct key selection/signature metadata.


src/new/wallet.ts
```diff
@@ -2168,7 +2168,10 @@ class HathorWallet extends EventEmitter {
     };
```
pedroferreira1, line 2207: This file changes was because I (claudin) forgot to do these changes in the previous PR (https://github.com/HathorNetwork/hathor-wallet-lib/pull/1133). So I added here because they are just small changes and I wanted to speed up the review/merge process without waiting for another CI to run

src/new/wallet.ts
```diff
@@ -2207,7 +2210,7 @@ class HathorWallet extends EventEmitter {
         tokenVersion: newOptions.tokenVersion,
       }
     );
-    return transactionUtils.prepareTransaction(txData, pin, this.storage, {
+    return transactionUtils.prepareTransaction(txData, pin ?? '', this.storage, {
```
r4mmer, line 2252: Can we add a test for signTx:true + external signer + no PIN?

r4mmer (dismissed):

src/new/wallet.ts
```diff
@@ -2168,7 +2168,10 @@ class HathorWallet extends EventEmitter {
     };
 
     const pin = newOptions.pinCode || this.pinCode;
-    if (!pin) {
+    // The pin is only used to decrypt the local key for signing. When an external tx-signing
+    // method is registered (e.g. a passkey signer), signing does not use it, so it is optional
+    // — mirrors signTx.
+    if (!pin && !this.storage.hasTxSignatureMethod()) {
```
tuliomir, line 2213: suggestion(non-blocking): pin-optional relaxation covers only 2 of the ~7 signing entry points

This PR relaxes the guard here and in `createNanoContractCreateTokenTransaction`, but the sibling entry points still hard-require a pin with no `hasTxSignatureMethod()` escape: `createNanoContractTransaction`, `createOnChainBlueprintTransaction`, and the mint/melt/delegate/destroy prepare methods.

An external-signer wallet therefore gets an inconsistent surface — it can create a token but not mint it, and call the nano create-token method but not a plain nano method — surfacing downstream as an inexplicable `Pin is required.`

If the narrow scope is deliberate (only what the passkey wallet needs today), stating that in the PR description would help; otherwise the remaining `ERROR_MESSAGE_PIN_REQUIRED` / `PinRequiredError` sites could get the same `!pin && !this.storage.hasTxSignatureMethod()` pattern.

src/new/wallet.ts
```diff
@@ -3586,7 +3589,9 @@ class HathorWallet extends EventEmitter {
     }
     const newOptions = { pinCode: null, signTx: true, ...options };
     const pin = newOptions.pinCode || this.pinCode;
-    if (!pin) {
+    // Optional when an external tx-signing method is registered (e.g. a passkey signer) —
+    // mirrors signTx.
+    if (!pin && !this.storage.hasTxSignatureMethod()) {
```
tuliomir, line 3645: suggestion(non-blocking): no test covers the pin-optional path of this method

Both new wallet-level tests exercise `prepareCreateNewToken`; this guard change ships untested — reverting the condition to `if (!pin)` keeps the suite green.

A mirror of the new external-signer test targeting `createNanoContractCreateTokenTransaction` (with `signTx: false`) would pin it.

src/utils/transaction.ts
```diff
@@ -375,18 +375,26 @@ const transaction = {
...
-  async getSignatureForTx(
+  async signTxInputs(
     tx: Transaction,
     storage: IStorage,
-    pinCode: string
+    getXpriv: (chain: 'legacy' | 'spend') => Promise<HDPrivateKey>
   ): Promise<ITxSignatureData> {
-    const xprivstr = await storage.getMainXPrivKey(pinCode);
-    const xprivkey = HDPrivateKey.fromString(xprivstr);
+    const xprivkey = await getXpriv('legacy');
```
tuliomir: thought(non-blocking): `getXpriv('legacy')` fires even when no input needs the legacy chain

For a tx whose wallet-owned inputs are all shielded-spend and which has no nano/OCB header, the legacy xpriv is fetched but never used. That's free for the stored-key wrapper, but for the motivating consumer — a passkey signer where each resolver call may be a biometric ceremony — an unnecessary `'legacy'` call may become a real UX cost.

Deferring the fetch to first use (the same lazy pattern already used for `'spend'`) would avoid it; the docstring's "called at most once per chain ('legacy' always)" would need updating either way.

tuliomir (dismissed):

src/new/wallet.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1135/commits/a40f44452e087edafa3e84449bef5677ab631607

src/new/wallet.ts
pedroferreira1: Good catch — the narrow scope is deliberate. The passkey wallet only needs send (relaxed in #1133) and create-token (this PR) today, so I relaxed exactly those. The !pin && !this.storage.hasTxSignatureMethod() pattern is mechanical, but I'd rather extend it to the remaining sign-requiring entry points (prepareMint/Melt/Delegate/DestroyTokensData, createNanoContractTransaction, createOnChainBlueprintTransaction) as those flows actually come online for the passkey wallet — each paired with a test like the one just added — than relax them un-exercised here. I'll note the scope in the PR description and track the rest as a follow-up.

src/new/wallet.ts
pedroferreira1: Good catch — added in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1135/changes/4535ad22091086811ed6702e9816b2ff014c59a5. Two unit tests in hathorwallet.test.ts mirroring the prepareCreateNewToken ones: with an external tx-signing method + no pin the guard doesn't fire (build fails later on the fake nano contract, but not with Pin is required.); without one it still rejects Pin is required.. Reverting the condition to if (!pin) now fails the suite.

src/utils/transaction.ts
pedroferreira1: Great point — fixed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/1135/changes/6de82394a016278f699cc7910b007ddfce09d221. signTxInputs now resolves each chain lazily and at most once: 'legacy' only when a legacy input, the nano caller, or an OCB pubkey needs it (same laziness 'spend' already had). An all-shielded tx with no nano/OCB header no longer triggers a 'legacy' call, so a passkey signer isn't asked for a biometric prompt it doesn't need. Docstring updated, and the "sources keys from the resolver" test now asserts 'legacy' isn't requested for the all-shielded case.

tuliomir (dismissed):

r4mmer (dismissed):

tuliomir (approved):

PR #1142, opened 2026-07-24
refactor(wallet): unify pre-calculated legacy + shielded addresses into one param

pedroferreira1:
Unify the two pre-calculated-address injection parameters into one.

Today a wallet is seeded with two independent, position-coupled arrays:
`preCalculatedAddresses` (legacy `string[]`) and `preCalculatedShieldedAddresses`
(`IPrecalculatedShieldedAddress[]`). Because they can disagree, `start()` needs a
coverage check + derive-fill for the "legacy present, shielded missing" case.

This folds both chains for one index into a single entry:

```ts
interface IPrecalculatedAddress {
  bip32AddressIndex: number;
  base58: string;                 // legacy P2PKH
  shielded?: { shieldedBase58; spendBase58; scanPubkey; spendPubkey };
}
preCalculatedAddresses?: string[] | IPrecalculatedAddress[] | null;
```

- **Back-compat:** a plain `string[]` is still accepted (legacy-only; shielded
  pairs are derived at start, as before).
- **Removed** `preCalculatedShieldedAddresses`.

### Acceptance Criteria

- Accept a unified `IPrecalculatedAddress[]` for `preCalculatedAddresses`, each
  entry carrying an index's legacy address and its optional shielded pair.
- Keep accepting a legacy `string[]` (back-compat, legacy-only).
- Remove the separate `preCalculatedShieldedAddresses` construction option.

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added unified `preCalculatedAddresses` support that accepts per-index entries combining legacy base58 data with optional shielded address pairs.
  * Enhanced fullnode test adapter with token/NFT helpers and an address ownership check.
* **Bug Fixes**
  * Improved wallet startup persistence for injected pre-calculated addresses, including stricter shielded/legacy consistency validation.
  * Rejected invalid or non-string base58 values when saving to in-memory storage.
* **Tests**
  * Expanded normalization, merge/save utilities, and end-to-end start/persistence coverage with additional edge-case assertions.


src/new/wallet.ts
```diff
@@ -2002,19 +2018,22 @@ class HathorWallet extends EventEmitter {
...
+    // legacy-only; the unified IPrecalculatedAddress[] carries each index's
+    // legacy address plus an optional shielded pair. Any index whose pair is
+    // omitted is derived below, once the scan/spend xpubs are in access data.
+    const injectedAddresses = normalizePreCalculatedAddresses(this.preCalculatedAddresses);
+    for (const entry of injectedAddresses) {
+      await this.storage.saveAddress({
+        base58: entry.base58,
+        bip32AddressIndex: entry.bip32AddressIndex,
+      });
     }
```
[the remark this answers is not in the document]
pedroferreira1: Good catch, and confirmed: MemoryStore.saveAddress does throw 'Already have this address' on any duplicate base58, and MemoryStore is the only IStore implementation left, so there's no store that tolerates it. stop() also defaults to cleanAddresses: false, so a stop() → start() on the same instance would hit it.
▎
▎ Two corrections to the framing, then I'll fix it: this isn't a regression from this PR — master's loop is byte-for-byte the same unguarded saveAddress (only the iteration source changed), and master already had the guarded shielded helper, so the asymmetry predates the refactor. And the blast radius is narrow: preCalculatedAddresses is test tooling, and no in-tree caller does a same-instance restart.
▎
▎ I'm fixing it anyway since this PR is rewriting those exact lines and leaving a correctly-guarded sibling four lines below makes the asymmetry look accidental. But I'll use an index-keyed guard rather than isAddressMine: isAddressMine resolves to a base58-only addressExists, so it would silently swallow the case where the same base58 is stored under a different bip32AddressIndex, turning a loud "your precalculated list disagrees with storage" into a silent index mismatch. getAddressAtIndex(entry.bip32AddressIndex) also matches what loadAddresses already does for the legacy chain (getAddressAtIndex(i) !== null), so it's the closer sibling to the code that actually runs during derivation.

__tests__/integration/helpers/wallet-precalculation.helper.ts
```diff
@@ -12,7 +12,38 @@ import { NETWORK_NAME } from '../configuration/test-constants';
...
+ * legacy-only and have their pair derived at wallet start.
+ */
+export function mergePrecalculatedAddresses(
+  legacyAddresses: string[] | undefined | null,
+  shieldedAddresses?: IPrecalculatedShieldedAddress[] | null
+): IPrecalculatedAddress[] {
+  const shieldedByIndex = new Map(
+    (shieldedAddresses ?? []).map(entry => [entry.bip32AddressIndex, entry])
+  );
+  return (legacyAddresses ?? []).map((base58, bip32AddressIndex) => {
```
tuliomir: issue(blocking): every shielded pair the legacy list doesn't cover is silently dropped — already live in the ocb path

The merge iterates the *legacy* array and looks shielded pairs up from that walk. There is no reverse pass over `shieldedByIndex` and no signal on leftovers, so `merge(undefined, [22 fixtures])` returns `[]` — all 22 discarded.

This is not hypothetical. `__tests__/integration/nanocontracts/bet.test.ts:824` calls `generateWalletHelper({ seed })` with `WALLET_CONSTANTS.ocb.seed`. That seed has no legacy list in `WALLET_CONSTANTS.ocb`, but it does have 22 committed shielded fixtures. Before this PR both params travelled independently and all 22 were injected; now the merge returns `[]` and `loadAddresses` re-derives every pair live.

That is exactly the 40–58x jest-sandbox EC cost this fixture file's own header says it exists to eliminate, reintroduced silently. The suite stays green — it just gets slower, so nothing surfaces it. Same shape is latent at `__tests__/integration/adapters/fullnode.adapter.ts:543-547`.

Key the merge on the union of both index sets rather than on legacy array position, and throw (or at least warn) on a shielded entry with no legacy counterpart. This helper is now the single funnel every integration wallet start routes through and it has zero tests — worth cases for `merge(undefined, nonEmpty)`, `merge([], nonEmpty)`, and shielded index ≥ legacy length.

src/new/wallet.ts
```diff
@@ -2002,19 +2018,18 @@ class HathorWallet extends EventEmitter {
...
+    // Persist the injected pre-calculated addresses exactly as provided — the
+    // caller's list is authoritative. Back-compat: a string[] is legacy-only;
+    // the unified IPrecalculatedAddress[] carries each index's legacy address
+    // plus an optional shielded pair. An index with no shielded pair persists
+    // legacy-only; nothing is derived here.
+    const injectedAddresses = normalizePreCalculatedAddresses(this.preCalculatedAddresses);
+    await savePrecalculatedLegacyAddresses(this.storage, injectedAddresses);
+    const injectedShieldedPairs = injectedAddresses
+      .filter(entry => entry.shielded)
+      .map(entry => ({ bip32AddressIndex: entry.bip32AddressIndex, ...entry.shielded! }));
```
tuliomir: issue(non-blocking): spread order lets a nested `bip32AddressIndex` hijack the entry's real index

The spread comes last, so any `bip32AddressIndex` structurally present on `shielded` wins. `Omit<IPrecalculatedShieldedAddress, 'bip32AddressIndex'>` removes it from the type but not from the value — excess-property checks fire only on fresh literals, so assigning a whole `IPrecalculatedShieldedAddress` compiles clean.

That is the natural caller move, since the repo's own fixtures are `Record<string, IPrecalculatedShieldedAddress[]>`. Verified at runtime: `{bip32AddressIndex: 7, shielded: {bip32AddressIndex: 0, …}}` emits index `0`.

The consequence is worse than a mislabel. Index 0's slot in `shieldedAddressIndexes` is then held by index 7's pair, so `loadAddresses` takes the "already injected" branch at i=0, never derives index 0's real pair, and subscribes index 7's spend address in its place — funds sent to index 0's shielded address become invisible to the wallet.

```suggestion
      .map(entry => ({ ...entry.shielded!, bip32AddressIndex: entry.bip32AddressIndex }));
```

Worth noting `mergePrecalculatedAddresses` hand-copies the four fields, which suggests the hazard was already felt — the defence belongs here rather than only in test code.

src/types.ts
```diff
@@ -165,6 +165,26 @@ export interface IPrecalculatedShieldedAddress {
   spendPubkey: string;
 }
 
+/**
+ * A pre-calculated address for a single BIP32 index, injected at wallet start
+ * to skip live EC derivation. Carries the legacy P2PKH address and, for
+ * shielded wallets, its shielded pair. An index whose `shielded` block is
+ * omitted persists its legacy address only — the injected list is persisted
```
tuliomir: issue(non-blocking): "nothing extra is derived at start" is false for shielded wallets

`loadAddresses` unconditionally enters `if (hasShieldedKeys)` at `src/utils/storage.ts:136`. An injected legacy-only entry never populates `shieldedAddressIndexes` (`memory_store.ts:344-349` routes only `addressType: 'shielded'` there), so `getAddressAtIndex(i, { legacy: false })` returns `null` and `deriveShieldedAddressPair` runs for that index in the same `start()`.

So for a shielded wallet, injecting a `string[]` — the documented back-compat shape — derives the entire shielded chain live, which is precisely what this sentence tells the reader it avoids. Indexes past the injected window also derive live, per the scan policy.

The docstring this replaced was accurate on that second point ("indexes past the injected window still derive live"); the rewrite traded a true caveat for a false absolute. Suggested: *"Injected indexes skip live EC derivation for the chains actually supplied. An entry with no `shielded` block persists legacy-only; a shielded wallet still derives that index's pair during address loading, as do indexes past the injected window."*

src/new/types.ts
```diff
@@ -71,14 +71,17 @@ export interface HathorWalletConstructorParams {
...
-   * Pre-calculated shielded address pairs to load into storage (test tooling,
-   * mirrors preCalculatedAddresses). Injected indexes skip live EC derivation
-   * in address loading; indexes past the injected window still derive live.
+   * Pre-calculated addresses to load into storage, skipping live EC derivation
+   * for the injected indexes. Two accepted shapes:
+   *   - `string[]` — legacy-only, index = array position (back-compat).
+   *   - `IPrecalculatedAddress[]` — unified: each index carries its legacy
+   *     address and, for shielded wallets, its shielded pair.
+   *
+   * The persisted set is exactly what is passed — the caller's list is
```
tuliomir: issue(non-blocking): same false "nothing extra is derived" claim, on the option a consumer reads first

Same defect as the one on `src/types.ts:172` — `loadAddresses` derives the full shielded chain for any legacy-only entry during the same `start()`. This copy matters more, because `HathorWalletConstructorParams` is what lands in the published `.d.ts`. Worth fixing both together.

Secondary: "the caller's list is authoritative" is now inverted by this PR's own idempotency guard. On a conflict `savePrecalculatedLegacyAddresses` throws rather than overwriting, so storage wins, not the caller.

__tests__/integration/helpers/wallet.helper.ts
```diff
@@ -105,12 +109,19 @@ export async function generateWalletHelper(param) {
...
+    // separate shielded param is gone.
+    preCalculatedAddresses: mergePrecalculatedAddresses(
+      walletData.addresses,
+      walletData.shieldedAddresses
+    ),
     scanPolicy: getGapLimitConfig(),
   };
   if (param) {
-    Object.assign(walletConfig, param);
+    // param's precalc (already folded into the merged array above) is stripped
```
tuliomir: suggestion: "already folded into the merged array" is only true when the caller supplied a legacy list

`param.preCalculatedAddresses` really is folded in at `:99`, so stripping `_pca` is correct. `param.preCalculatedShieldedAddresses` is not: it reaches `walletData.shieldedAddresses` at `:102-103` and is then dropped by the merge whenever the legacy list is absent or shorter.

Stripping it on the claim that it was "already folded" is what makes the ocb regression invisible — see the blocking comment on `mergePrecalculatedAddresses`.

Also, the stated reason only applies to `_pca`. `preCalculatedShieldedAddresses` is no longer a constructor param at all, so leaving it in `rest` could not clobber anything — it would simply be ignored.

__tests__/integration/fullnode-specific/start.test.ts
```diff
@@ -213,7 +216,10 @@ describe('[Fullnode-specific] start', () => {
       connection: generateConnection(),
       password: DEFAULT_PASSWORD,
       pinCode: DEFAULT_PIN_CODE,
-      preCalculatedShieldedAddresses: getPrecalculatedShieldedForSeed(multisigWalletsData.words[0]),
+      preCalculatedAddresses: mergePrecalculatedAddresses(
```
tuliomir, line 224: suggestion: this test became tautological — it now asserts a constant equals itself

At the merge base this config passed **only** `preCalculatedShieldedAddresses`, so the legacy P2SH addresses were derived live from `pubkeys`/`numSignatures`, and the loop at `:236-240` genuinely checked `deriveAddressP2SH` output against `WALLET_CONSTANTS.multisig.addresses`.

Now those same constants are injected, and `getAddressAtIndex` reads storage first (`src/new/wallet.ts:866`) while `loadAddresses` skips any index already present. The assertion reads back what the test wrote.

The regression that survives: break multisig redeem-script or P2SH address derivation and this test stays green. `__tests__/utils/address.test.ts:190` still pins `deriveAddressP2SH` at unit level against a different fixture, so this is assertion loss rather than blindness — but the end-to-end multisig path (config → accessData → `loadAddresses` → storage) is now unverified anywhere.

Either drop the legacy injection here, or keep it and add a second wallet started without it. One single test should always execute the slow path to ensure the correctness of the entire suite.

src/new/wallet.ts
```diff
@@ -2002,19 +2018,18 @@ class HathorWallet extends EventEmitter {
...
-    if (this.preCalculatedShieldedAddresses) {
-      // Mirrors the legacy injection above: pre-populates the shielded chain so
-      // loadAddresses skips the per-index EC derivation for these indexes.
-      await savePrecalculatedShieldedAddresses(this.storage, this.preCalculatedShieldedAddresses);
+    // Persist the injected pre-calculated addresses exactly as provided — the
+    // caller's list is authoritative. Back-compat: a string[] is legacy-only;
+    // the unified IPrecalculatedAddress[] carries each index's legacy address
+    // plus an optional shielded pair. An index with no shielded pair persists
+    // legacy-only; nothing is derived here.
+    const injectedAddresses = normalizePreCalculatedAddresses(this.preCalculatedAddresses);
```
tuliomir, line 2028: suggestion: nothing starts a wallet with a unified array, so this split wiring is unpinned

`normalizePreCalculatedAddresses` and `savePrecalculatedLegacyAddresses` are each tested in isolation; the reshaping *between* them is not. No unit test anywhere passes `preCalculatedAddresses` to a real `start()` — the only `start()` in the unit suite is `__tests__/stream.test.ts:171`, which passes none.

Regressions that survive a green suite: invert the `.filter`, drop `bip32AddressIndex` from the mapped object, or revert to a plain `saveAddress` loop — which resurrects the very `'Already have this address'` bug the last commit was written to fix.

The shielded half is worse, since injection is a pure optimisation: even the integration suite would stay green, only slower. The `mergePrecalculatedAddresses` finding is that failure mode already realised.

A `MemoryStore` plus stubbed-connection test with a two-entry unified array (one with `shielded`, one without), asserting `getAddressAtIndex(0)`, `getAddressAtIndex(0, { legacy: false })` and `isAddressMine(spendBase58)`, closes all of it.

src/new/wallet.ts
```diff
@@ -162,6 +163,25 @@ const ERROR_MESSAGE_PIN_REQUIRED = 'Pin is required.';
...
+ * `IPrecalculatedAddress[]` shape (each index's legacy address plus an optional
+ * shielded pair) passes through unchanged.
+ */
+export function normalizePreCalculatedAddresses(
+  input: string[] | IPrecalculatedAddress[] | null
+): IPrecalculatedAddress[] {
+  if (!input || input.length === 0) {
+    return [];
+  }
+  if (typeof input[0] === 'string') {
```
tuliomir: suggestion: the union isn't expressible for dynamically-built lists, which pushes callers into the cast that corrupts storage

`typeof input[0] === 'string'` classifies the whole array from element 0. A mixed array — which `string[] | IPrecalculatedAddress[]` blocks for literals but not for `as` casts or JS consumers — silently corrupts: `['addrA', {bip32AddressIndex: 1, base58: 'addrB'}]` yields `base58` as an *object*, and `MemoryStore.saveAddress` only rejects falsy `base58`, so it is keyed into the address map and every later `isAddressMine` returns garbage.

The push toward that cast is structural: `(string | IPrecalculatedAddress)[]`, a `.map()` returning either shape, and `[...legacy, {…}]` all fail to compile against the current union. So "unified where I have a fixture, plain string otherwise" — the obvious use case — has no way to be typed.

Widening to `(string | IPrecalculatedAddress)[] | null` and normalising per element is strictly more permissive, fully back-compat, and deletes both casts:

```ts
input.map((e, i) => (typeof e === 'string' ? { bip32AddressIndex: i, base58: e } : e))
```

src/types.ts
```diff
@@ -165,6 +165,26 @@ export interface IPrecalculatedShieldedAddress {
...
+ *
+ * This is the unified successor to passing a `string[]` of legacy addresses
+ * alongside a separate `IPrecalculatedShieldedAddress[]`: both chains for one
+ * index now travel together. A plain `string[]` is still accepted (legacy-only,
+ * back-compat).
+ */
+export interface IPrecalculatedAddress {
+  bip32AddressIndex: number;
+  /** The legacy P2PKH address (base58). */
+  base58: string;
```
tuliomir: suggestion: shielded-only injection is no longer expressible, though that was the removed option's whole purpose

`base58` is required, and `injectedShieldedPairs` is derived by filtering `injectedAddresses`. A caller holding shielded fixtures but no legacy list therefore has no valid option: omit the entry and the pair is dropped; supply a placeholder `base58` and it either persists a bogus address or trips the new mismatch throw at `src/utils/storage.ts:206`.

`__tests__/integration/fullnode-specific/start.test.ts:213` was exactly this caller before the PR — which is why it had to gain a legacy list it doesn't otherwise need.

Either make `base58` optional and skip the legacy save when absent, or keep an explicit shielded-only path.

src/utils/storage.ts
```diff
@@ -176,6 +177,46 @@ export async function loadAddresses(
...
+ * default) and started again. The skip is keyed on the BIP32 index rather than
+ * on the address, matching how loadAddresses guards the legacy chain
+ * (`getAddressAtIndex(i) !== null`); an index already holding a DIFFERENT
+ * address means the injected list disagrees with storage, which stays a loud
+ * error instead of being silently skipped.
+ *
+ * @param storage The wallet storage instance
+ * @param entries Pre-calculated addresses, one per BIP32 index
+ */
+export async function savePrecalculatedLegacyAddresses(
```
tuliomir, line 198: suggestion: the shielded sibling still skips a mismatched pair silently

This attaches to the new function, but the concern is `savePrecalculatedShieldedAddresses` at `src/utils/storage.ts:250-257`, outside the diff.

The docblock calls these two siblings, but they disagree on the case that matters. This one throws when an index already holds a *different* address, and its test documents why — "precisely so a genuine disagreement stays loud". The shielded one guards on `isAddressMine(base58)` instead, so an injected pair that disagrees with what storage holds at that index isn't detected at all: the address is new, so it saves, and `shieldedAddressIndexes.set(i, …)` overwrites the existing mapping while the stale record stays in `this.addresses`.

That is the same silent corruption the legacy guard was just hardened against, and it is on the chain where a wrong `ctMappingAddress` costs address-history visibility. Applying the index-keyed guard there too would make the "sibling" claim true.

src/utils/storage.ts
```diff
@@ -176,6 +177,46 @@ export async function loadAddresses(
...
+  storage: IStorage,
+  entries: IPrecalculatedAddress[]
+): Promise<void> {
+  for (const entry of entries) {
+    const existing = await storage.getAddressAtIndex(entry.bip32AddressIndex);
+    if (existing !== null) {
+      if (existing.base58 !== entry.base58) {
+        throw new Error(
```
tuliomir: nitpick: bare `Error` rather than one of the repo's typed errors

`src/errors.ts` already exports `AddressError` and ~25 siblings. As written, a downstream consumer can only match on the message string to tell "your pre-calculated list disagrees with storage" apart from any other startup failure.

src/types.ts
```diff
@@ -165,6 +165,26 @@ export interface IPrecalculatedShieldedAddress {
...
+ * exactly as given, nothing extra is derived at start.
+ *
+ * This is the unified successor to passing a `string[]` of legacy addresses
+ * alongside a separate `IPrecalculatedShieldedAddress[]`: both chains for one
+ * index now travel together. A plain `string[]` is still accepted (legacy-only,
+ * back-compat).
+ */
+export interface IPrecalculatedAddress {
+  bip32AddressIndex: number;
+  /** The legacy P2PKH address (base58). */
```
tuliomir: nitpick: "legacy P2PKH address" is wrong for multisig wallets

`loadAddresses` derives P2SH when `walletType === 'p2sh'` (`src/utils/storage.ts:122-125`), and this PR's own multisig call sites feed P2SH addresses through this exact field (`wallet.helper.ts:218`, `start.test.ts:220`). `:170` repeats the claim.

Suggest "the legacy chain address (P2PKH, or P2SH for multisig wallets)". Contrast `IPrecalculatedShieldedAddress.spendBase58` at `:160`, where "P2PKH" is genuinely correct.

__tests__/integration/helpers/wallet.helper.ts
```diff
@@ -105,12 +109,19 @@ export async function generateWalletHelper(param) {
     connection: generateConnection(),
     password: DEFAULT_PASSWORD,
     pinCode: DEFAULT_PIN_CODE,
-    preCalculatedAddresses: walletData.addresses,
-    preCalculatedShieldedAddresses: walletData.shieldedAddresses,
+    // Both chains for each index travel together in the unified array; the
+    // separate shielded param is gone.
```
tuliomir: nitpick: "the separate shielded param is gone" is a diff annotation, not a description of the code

It describes a parameter that no longer exists anywhere in the repo, so it is unresolvable for a reader without memory of this PR. The first sentence stands alone.

Same class, lower stakes elsewhere: `src/types.ts:175` "the unified successor to…" (`IPrecalculatedShieldedAddress` is still very much alive — `savePrecalculatedShieldedAddresses` takes it — so "successor" also misleads about the sibling's status); `savePrecalculatedLegacyAddresses.test.ts:47` "used to throw … before startup could complete" (unanchored — pre-PR master, or an earlier commit here?); and `precalculated-shielded-addresses.ts:43` "did" → "does".

__tests__/utils/savePrecalculatedLegacyAddresses.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    networkName: 'testnet',
+  });
+
+  async function makeStorage(): Promise<Storage> {
+    const storage = new Storage(new MemoryStore());
+    await storage.saveAccessData(accessData);
+    return storage;
+  }
+
+  const entries: IPrecalculatedAddress[] = [
```
tuliomir, line 29: nitpick: the fixture always pre-seeds accessData, inverting `start()`'s real precondition

In `start()` the injection at `:2027` runs *before* `getAccessData()` at `:2035` and before accessData is generated at `:2044` — on a fresh wallet there is none.

It works because `Storage.getAddressAtIndex` (`src/storage/storage.ts:323-328`) delegates straight to the store with no validation, unlike its immediate sibling `getAddressPubkey` at `:336-347`, which does call `_getValidAccessData()`. One case on a bare `new Storage(new MemoryStore())` would keep that asymmetry pinned.

src/new/wallet.ts
```diff
@@ -162,6 +163,25 @@ const ERROR_MESSAGE_PIN_REQUIRED = 'Pin is required.';
...
 
+/**
+ * Normalize the `preCalculatedAddresses` option into per-index entries.
+ *
+ * Back-compat: a `string[]` is legacy-only, index = array position. The unified
+ * `IPrecalculatedAddress[]` shape (each index's legacy address plus an optional
+ * shielded pair) passes through unchanged.
+ */
+export function normalizePreCalculatedAddresses(
+  input: string[] | IPrecalculatedAddress[] | null
```
tuliomir: nitpick: `undefined` is neither in the signature nor in the tests, though the option is optional

This declares `… | null` while `HathorWalletConstructorParams.preCalculatedAddresses` is optional. `undefined` works today via `!input`, but a refactor to `input === null` would break the optional path with nothing failing — `normalizePreCalculatedAddresses.test.ts:6-7` pins only `null` and `[]`.

Widening the signature plus a one-line assertion closes it.

__tests__/integration/helpers/wallet-precalculation.helper.ts
pedroferreira1: Confirmed and fixed. I ran it before touching anything: getPrecalculatedShieldedForSeed(WALLET_CONSTANTS.ocb.seed) returns 22 pairs, and mergePrecalculatedAddresses(WALLET_CONSTANTS.ocb.addresses, …) returned length 0 — ocb has only a seed, and bet.test.ts:824 calls generateWalletHelper({ seed }) exactly as you said. A shorter legacy list truncates the same way (merge(2 legacy, 22 pairs) → 2).
▎
▎ The merge is now keyed on the union of both index sets and returns entries in ascending index order. The helper had zero tests; it has seven, covering merge(undefined, nonEmpty), merge([], nonEmpty), a shorter legacy list, and a shielded index past its end.

src/new/wallet.ts
pedroferreira1: Confirmed, and worse than a mislabel exactly as you described. Reproduced through a real start(): with {bip32AddressIndex: 0, shielded: <a whole IPrecalculatedShieldedAddress declaring 7>}, getAddressAtIndex(7, {legacy: false}) came back holding index 0's pair. Applied your suggestion, and the commit carries the test — which also covers your point in the other comment
▎
▎ Your closing observation was the giveaway: mergePrecalculatedAddresses hand-copies the four fields, so the hazard had been felt once already. The merge now says so in a comment rather than leaving it as folklore.

src/types.ts
pedroferreira1: Correct, and measured: a shielded wallet with 4 legacy-only entries injected calls deriveShieldedAddressPair 4 times during the same start() — the whole injected window. deriveAddressP2PKH is called 0 times, so the injection buys exactly the legacy half and the sentence claimed both.
▎
▎ Rewrote it to your framing: injection skips live EC derivation for the chains actually supplied; whatever an entry omits is derived during address loading, as are indexes past the injected window.

src/new/types.ts
pedroferreira1: Fixed together with the previous comment, and you're right that this copy matters more — it lands in the published .d.ts.
▎
▎ On the secondary point: agreed the wording was inverted, with one refinement. On a conflict nobody "wins" — savePrecalculatedLegacyAddresses throws, so start() aborts loudly rather than storage silently taking precedence. The genuinely wrong phrase was "the persisted set is exactly what is passed": it's the union of what storage held and what was injected, injection over a populated index is a no-op, and a disagreement is a hard error. That's what it says now.

__tests__/integration/helpers/wallet.helper.ts
pedroferreira1: Right on both halves - preCalculatedShieldedAddresses reaches walletData.shieldedAddresses and was then dropped by the merge whenever the legacy list was absent or shorter, so the claim was doing real damage by making the first comment look accounted for. And the stated reason only ever applied to _pca; preCalculatedShieldedAddresses isn't a constructor param at all, so leaving it in rest would have been ignored. Comment now says exactly that.

__tests__/integration/fullnode-specific/start.test.ts
pedroferreira1: Confirmed by execution: I injected 5 deliberately wrong base58 addresses and ran loadAddresses(0,5) — deriveAddressP2SH was called 0 times, no error, and the wrong addresses read straight back. Assertion loss, precisely as you framed it.
▎
▎ Took the "drop the legacy injection" option, which the comment after fix makes available again: the config injects the shielded pairs alone, so the legacy P2SH chain derives live from pubkeys/numSignatures and the loop is a real check once more — while the expensive half stays pre-calculated.
▎
▎ Added a unit-level lock too, since this path deserves coverage that doesn't need a fullnode: rebuild WALLET_CONSTANTS.multisig.addresses from the seed, pubkeys and numSignatures and compare.

src/new/wallet.ts
pedroferreira1: Agreed, and it's the reason #2 shipped undetected. Added the MemoryStore + stubbed-connection test you sketched, driving a real start() with a two-entry unified array (one with shielded, one without) and asserting getAddressAtIndex(0), getAddressAtIndex(0, { legacy: false }) and isAddressMine(spendBase58) — plus the index-hijack case, which fails on the old mapping. It's in the same commit as the second comment fix, verified failing first.

src/new/wallet.ts
pedroferreira1: Confirmed, including the corruption: the mixed list normalised to base58: {…}, savePrecalculatedLegacyAddresses saved it without complaint, and isAddressMine('addrB') then returned false — the index is poisoned. (One small refinement: it's a silent ownership miss rather than garbage from isAddressMine, and only string-first corrupts silently; object-first fails loud in saveAddress.)
▎
▎ Adopted your widening and per-element normalisation — both as casts are gone. Also tightened MemoryStore.saveAddress to reject a non-string base58, so this can't be reached from an untyped caller that never goes through normalisation.
▎
▎ One deliberate consequence: a unified array is no longer returned by reference, so the identity assertion became toEqual plus per-entry toBe — entries are still passed through untouched.

src/types.ts
pedroferreira1: Correct, and this was the root cause behind the first comment. Both escape hatches fail as you said: omitting drops the pair, and a placeholder base58 either becomes that index's legacy address permanently (loadAddresses returns a stored address without re-deriving) or trips the mismatch throw on a restart.

 Took the type route: IPrecalculatedAddress is now a union, so an index may carry the legacy address, the shielded pair, or both, and "neither" stays a compile error. The legacy save skips an entry with no base58, leaving that index to live derivation. start.test.ts:213 is a shielded-only caller again as a result.

src/utils/storage.ts
pedroferreira1: Confirmed — and it's my own guard that created the asymmetry, so thanks for holding it to the same standard. Reproduced: injecting a different pair at an occupied index resolved without error, flipped shieldedAddressIndexes, and left the previous record owned but unreachable from the chain.
▎
▎ Applied the index-keyed guard, and also compare ctMappingAddress, since shielded-spend records aren't index-addressable and a re-pointed spend address is otherwise invisible. The isAddressMine saves stay as the fall-through so a partially stored pair still heals — there's a test for that.

src/utils/storage.ts
pedroferreira1: Fixed. Both new shielded mismatches and the legacy one are AddressError now, and the test asserts the type rather than the message so it stays that way.

src/types.ts
pedroferreira1: Correct — loadAddresses derives P2SH for walletType === 'p2sh', and this PR's own multisig call sites feed P2SH through that exact field. It reads "the legacy chain address (P2PKH, or P2SH for multisig wallets)" now. It landed in the union-type commit rather than the docs one, since I was rewriting that line anyway to make base58 optional.

__tests__/integration/helpers/wallet.helper.ts
pedroferreira1: Agreed on all of them. Dropped "the separate shielded param is gone"; dropped "the unified successor to…" (and you're right that it misdescribed the sibling's status — IPrecalculatedShieldedAddress is very much alive); re-anchored the test docstring so it states the current behaviour instead of "used to"; and did → does.

__tests__/utils/savePrecalculatedLegacyAddresses.test.ts
pedroferreira1: Confirmed, including the mechanism: savePrecalculatedLegacyAddresses works fine on a bare new Storage(new MemoryStore()) while getAddressPubkey rejects on that same storage. Added the case, asserting both sides — so if getAddressAtIndex ever gains _getValidAccessData() for symmetry, this fails instead of fresh-wallet startup breaking silently. (Worth noting Storage.saveAddress is equally unvalidated, so the helper leans on two methods staying that way, not one.)

src/new/wallet.ts
pedroferreira1: Fixed - the signature accepts it and there's an assertion for it, so a refactor to input === null can't slip through.

tuliomir (dismissed):

__tests__/integration/helpers/wallet-precalculation.helper.ts
```diff
@@ -28,8 +28,8 @@ import { IPrecalculatedAddress, IPrecalculatedShieldedAddress } from '../../../s
  * put every one of its pairs back on the slow derivation path.
  */
 export function mergePrecalculatedAddresses(
-  legacyAddresses?: string[] | null,
-  shieldedAddresses?: IPrecalculatedShieldedAddress[] | null
+  legacyAddresses: string[] | undefined,
```
tuliomir: suggestion(non-blocking): simplify
```ts
legacyAddresses?: string[]
```

tuliomir (dismissed):

__tests__/integration/helpers/wallet-precalculation.helper.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1142/commits/932c80daad3030dd8f79c85b505be360eb06240d

tuliomir (approved):

raul-oliveira (approved):

PR #1143, opened 2026-07-24
feat: PIN-optional signing for all external-signer entry points

pedroferreira1:
### Summary

Follow-up to #1135, which made signing PIN-optional for an external tx-signing method (e.g. a passkey signer) — but only for `prepareCreateNewToken` and `createNanoContractCreateTokenTransaction`. Every **other** sign-requiring entry point still hard-required a PIN, with no escape when an external signer is registered. That left an external-signer (xpub-only) wallet with an inconsistent surface — it could create a token but not mint it, call the nano create-token method but not a plain nano method — surfacing downstream as an inexplicable `Pin is required.`

This applies the same `!pin && !this.storage.hasTxSignatureMethod()` escape to **all** of them. Each already signs through `storage.getTxSignatures`, so once the guard lets it through, the external signer produces the signatures:

- `sendManyOutputsSendTransaction`
- `prepareMintTokensData` / `prepareMeltTokensData`
- `prepareDelegateAuthorityData` / `prepareDestroyAuthorityData`
- `getSignatures`
- `createNanoContractTransaction`
- `buildTxTemplate`
- `createOnChainBlueprintTransaction`

### Readonly fix (bundled)

`createNanoContractTransaction` and `createOnChainBlueprintTransaction` guarded on `this.storage.isReadonly()` — which is `true` for any xpub-only wallet regardless of external signing — instead of the wallet-level `this.isReadonly()` (which returns `false` when an external signer is registered). So they'd throw `WalletFromXPubGuard` **before** ever reaching the PIN relaxation, making it unreachable for a passkey wallet. Switched both to `this.isReadonly()`, matching every other write method (and the same fix already applied to `createNanoContractCreateTokenTransaction`).

### Acceptance Criteria
- With an external tx-signing method registered and **no** PIN, none of the relaxed entry points throws `Pin is required.` — they proceed to sign via the external method (send, mint, melt, delegate, destroy, getSignatures, plain nano, on-chain blueprint, buildTxTemplate).
- Without an external signer and no PIN, each of those still rejects with `Pin is required.` (behavior unchanged for normal wallets).
- An xpub-only wallet with an external signer clears `WalletFromXPubGuard` on `createNanoContractTransaction` and `createOnChainBlueprintTransaction` (previously blocked by `storage.isReadonly()`); a truly readonly wallet (no external signer) still rejects with `WalletFromXPubGuard`.
- The relaxation is a no-op for the `signTx: false` (build-only) paths and only affects the PIN requirement — no change to input selection, signing, or encoding.
- Regression: normal seed+PIN wallets are unaffected (all methods still require a PIN when no external signer is set).
- New unit tests cover the with-signer and without-signer paths for mint, plain nano, and on-chain blueprint; full suite is green; `tsc` and ESLint pass.

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added PIN-free transaction preparation and signing when an external signing method is configured.
  * Enabled PIN-free token, authority, nano-contract, and on-chain blueprint workflows for compatible read-only wallets.

* **Bug Fixes**
  * Preserved PIN enforcement when no external signer is configured.
  * Improved externally signed flows to avoid unnecessary PIN and wallet restrictions.

* **Tests**
  * Added coverage for PIN-free operations with and without external signing, including fullnode integration scenarios.


__tests__/new/hathorwallet.test.ts
```diff
@@ -1530,6 +1534,147 @@ describe('prepare transactions without signature', () => {
...
+    // any unrelated breakage. End-to-end signing for the relaxed entry points is covered by the
+    // external-signer integration tests.
+    const err = await hWallet
+      .createNanoContractTransaction('noop', fakeAddress.base58, {
+        ncId: 'a'.repeat(64),
+        args: [],
+        actions: [],
+      })
+      .catch(e => e);
+    expect(err).toBeInstanceOf(NanoContractTransactionError);
```
tuliomir, line 1634: suggestion(non-blocking): nano and blueprint signing with an external signer is never executed

Both new nano/blueprint tests stop at caller-address validation, so `prepareNanoSendTransaction(nc, pin ?? '', …)` at `src/new/wallet.ts:3584` and `:3965` never runs. The integration test covers send/mint/melt/delegate/destroy but not nano or on-chain blueprint. If a pin requirement were reintroduced anywhere below these guards, every test here would still pass while passkey nano calls broke at runtime. Consider extending the integration test with one nano or blueprint call under the external signer.

src/new/wallet.ts
```diff
@@ -3174,10 +3174,10 @@ class HathorWallet extends EventEmitter {
       throw new WalletFromXPubGuard('getSignatures');
     }
     const pin = pinCode || this.pinCode;
-    if (!pin) {
+    if (!pin && !this.storage.hasTxSignatureMethod()) {
```
tuliomir: suggestion(non-blocking): two relaxed entry points got no test at all

`getSignatures` (:3177) and `buildTxTemplate` (:3886) both had their pin guard relaxed, but neither appears in the new unit or integration tests. `getSignatures` is a public API returning signature material, so its external-signer path is worth pinning. A one-line unit test per site — external signer registered, no pin, assert it does not throw `'Pin is required.'` — would close the gap.

src/new/wallet.ts
```diff
@@ -1929,7 +1929,7 @@ class HathorWallet extends EventEmitter {
     };
 
     const pin = newOptions.pinCode || this.pinCode;
-    if (!pin) {
+    if (!pin && !this.storage.hasTxSignatureMethod()) {
```
tuliomir: nitpick: the relaxed guards drop the explanatory comment the two precedents carry

The two prior relaxations both explain themselves inline — `signTx` at :3211-3213 ("The pin is only used to decrypt the local private key…") and `createNanoContractCreateTokenTransaction` at :3643-3644. The 11 sites added here carry no comment. For a security-relevant guard, the next reader benefits from the same one-liner; a shared constant or a single comment on the first occurrence would do. Same applies to the `storage.isReadonly()` → `this.isReadonly()` switch at :3542 and :3940, which reads as a no-op refactor without it.

__tests__/new/hathorwallet.test.ts
```diff
@@ -1530,6 +1534,147 @@ describe('prepare transactions without signature', () => {
...
+
+    // Must clear the xpub guard (was storage.isReadonly() → now this.isReadonly()) and the pin
+    // guard. As with the nano test above, the build then reaches caller-address validation, which
+    // this minimal mock storage fails; pin that exact failure rather than accepting any error, so
+    // an unrelated breakage cannot masquerade as the guards being cleared.
+    const err = await hWallet
+      .createOnChainBlueprintTransaction('0123abcd', fakeAddress.base58)
+      .catch(e => e);
+    expect(err).toBeInstanceOf(NanoContractTransactionError);
+    expect(err.message).toContain('does not belong to the wallet');
```
tuliomir, line 1676: nitpick: `createOnChainBlueprintTransaction` has no negative counterpart test

Every other relaxed entry point in this file got a positive/negative pair. `createOnChainBlueprintTransaction` got only the positive, so nothing pins that it *still* rejects a pinless call with no external signer.

tuliomir (dismissed):

__tests__/new/hathorwallet.test.ts
pedroferreira1: Confirmed and fixed. Both unit tests do stop at caller-address validation, so prepareNanoSendTransaction is unreached — and it has no unit test anywhere, so a pin requirement reintroduced inside it would have left the whole suite green.

One narrowing worth recording: the shared dispatch chain below those guards (getTxSignatures → signTxInputs, including the caller-signature branch) is exercised end to end by the transparent flows, so a regression there would be caught. What was genuinely uncovered is the nano-specific slice. 

Added a nano call under the external signer to the integration test. Cheaper than expected — global.FEE_BLUEPRINT_ID is already exported to every integration file by the shared setup, so no blueprint deployment or new bootstrap was needed; the existing xpub-only wallet with signTxInputs registered signs it unchanged. It asserts the nano header carries a caller signature.

src/new/wallet.ts
pedroferreira1: Correct, and both now have a positive and a negative case.

The positive ones assert more than "doesn't throw": they check the external signer is actually invoked, and invoked with ''. That's the part that was really unpinned — the pin ?? '' changes are most of this PR's diff and nothing exercised them, so reverting one to a bare pin was invisible.

src/new/wallet.ts
pedroferreira1: Agreed, and I went with your "shared" option rather than pasting the one-liner nine times.

All the pin guards now route through a single pinIsRequired(pin) whose docstring carries the rationale. Two small corrections to the numbers: there are three precedents, not two — prepareCreateNewToken also carries one — and this PR adds nine new sites, not eleven. The extra precedent strengthens the point, so I folded all three into the helper and removed their now-duplicated inline comments; twelve copies of the same sentence was the thing worth avoiding.

__tests__/new/hathorwallet.test.ts
pedroferreira1: Right, added. It sets up non-readonly storage so the pin guard is what fires rather than the xpub guard — otherwise the test would pass for the wrong reason.

src/new/wallet.ts
```diff
@@ -1929,7 +1940,7 @@ class HathorWallet extends EventEmitter {
     };
 
     const pin = newOptions.pinCode || this.pinCode;
-    if (!pin) {
+    if (this.pinIsRequired(pin)) {
```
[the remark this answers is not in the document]
pedroferreira1: Checked this one by running it rather than reading it: with pin: null and an external signer registered, the signer receives "", not null. SendTransaction.signTx normalizes with pin ?? this.pin ?? '' before signTransaction -> getTxSignatures -> the signer, so the empty-string contract is already honoured -> just one layer down.

That's why this path is written differently from the other seven: they hand a string straight to prepareTransaction, which does no normalization, so they need the ?? ''. This one delegates to SendTransaction, which owns it.

Worth flagging that applying the change would be actively harmful rather than neutral. this.pin has a second, non-signing consumer -> the post-send shielded-change insert passes this.pin ?? undefined so a wallet with no stored pin reports "none". Because ?? only catches null/undefined, an empty string passes through: enqueueOnNewTx would get "" instead of undefined, and onNewTx's txPin ?? this.pinCode ?? undefined would resolve to "", skipping the stored-pin fallback. So signing wouldn't change and shielded decryption would get a bogus empty pin.

tuliomir (approved):

raul-oliveira (approved):

PR #1145, opened 2026-07-24
fix(shielded): exclude shielded phantom outputs from the transparent fee calc

pedroferreira1:
### Description

Fixes tuliomir's review comment on #1131: a shielded fee-token output was
over-charged `FEE_PER_OUTPUT`.

An explicit shielded output pushes a transparent **phantom** output into
`txData.outputs` so UTXO selection accounts for the shielded value (the phantom
is stripped from the final tx). But that phantom was still present in the
outputs passed to `Fee.calculate`, so a `TokenVersion.FEE` shielded output was
charged `FEE_PER_OUTPUT` (via the phantom) **on top of** its shielded fee.

Confirmed against the fullnode (`experimental/shielded-outputs-alpha-v4`):

- A shielded output is charged **only** the shielded fee
  (`FEE_PER_AMOUNT_SHIELDED_OUTPUT` / `FEE_PER_FULL_SHIELDED_OUTPUT`), never
  `FEE_PER_OUTPUT` — shielded outputs live in a separate `tx.shielded_outputs`
  list and are never in `chargeable_outputs`
  (`transaction_verifier.calculate_shielded_fee`, `token_info.calculate_fee`).
- The declared fee is validated for an **exact match**
  (`verify_token_rules`: `expected_fee != fees_from_fee_header → raise`), so an
  over-declared fee is **rejected outright**, not merely over-paid.
- A FEE token whose value goes entirely into shielded outputs (transparent
  inputs, no transparent output) is still charged the flat melt fee — the melt
  rule is unchanged — which the wallet already mirrors once the phantom is gone.

**Fix:** exclude phantoms from the outputs handed to `Fee.calculate`. After
this, the wallet's transparent fee matches the node in every case: transparent
outputs → `FEE_PER_OUTPUT`; a FEE token with inputs but no transparent output →
flat melt fee; shielded outputs → shielded fee only. (The `changeShieldedMode`
custom-token change path was already correct — it removes the output before
`Fee.calculate` — this brings the explicit-shielded path in line with it.)

Adds a regression test (`sendTransaction.test.ts`) that fails without the fix
(2 chargeable FEE-token outputs) and passes with it (1). Validated locally:
`npm run build` 0, `npm run lint` 0, `sendTransaction.test.ts` 31/31.

### Acceptance Criteria

- A shielded FEE-token output is charged only the shielded fee, not
  `FEE_PER_OUTPUT`.
- Phantom outputs are excluded from the transparent fee calculation while still
  used for UTXO selection.
- Regression coverage for a `TokenVersion.FEE` token with a shielded output.

### Security Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected transaction fee calculations for shielded outputs.
  * Prevented internal placeholder outputs from being incorrectly counted toward fees.
  * Fixed exact-match fee rejection in transactions involving the fee-version token.


src/new/sendTransaction.ts
```diff
@@ -485,11 +485,26 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+    // FEE_PER_OUTPUT, and counts just the real transparent outputs. Leaving a
+    // phantom in would over-declare FEE_PER_OUTPUT for a FEE-token shielded
+    // output; since the node validates the declared fee for an exact match, the
+    // tx would be rejected. Excluding them keeps the wallet in lockstep with the
+    // node: transparent outputs → FEE_PER_OUTPUT (or the flat melt fee when a
+    // FEE token has inputs but no transparent output), shielded → shielded fee.
+    const feeOutputs =
+      phantomOutputs.size > 0
+        ? partialOutputs.filter(out => !phantomOutputs.has(out))
+        : partialOutputs;
```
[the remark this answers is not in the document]
pedroferreira1: Thanks — but I believe the premise is inverted here, so I'm going to leave the filter as-is.
▎
▎ The flat melt-branch FEE_PER_OUTPUT isn't a wallet-side overcharge — the fullnode charges it too. In TokenInfoDict.calculate_fee the node computes fee = shielded_fee, then per token: chargeable_outputs * FEE_PER_OUTPUT if there are chargeable outputs, else a flat FEE_PER_OUTPUT when chargeable_inputs > 0. And chargeable_outputs is only ever incremented while iterating the transparent self.outputs list — shielded outputs live in a separate header and never enter that loop.
▎
▎ So for exactly the scenario you describe (FEE-token transparent inputs, only shielded outputs), the node's expected fee is shielded_fee + FEE_PER_OUTPUT — which is what the wallet now declares. Since verify_token_rules compares for an exact match (expected_fee != fees_from_fee_header → InputOutputMismatch), suppressing the melt charge as suggested would make the wallet under-declare and every such tx would be rejected. The over-declaration this PR fixes was the mirror-image failure (the phantom added a second FEE_PER_OUTPUT), which is why the fix is to exclude phantoms rather than to special-case the melt branch.
▎
▎ That said, your instinct is pointing at something real, just one input over: when the FEE-token inputs are themselves shielded UTXOs, the node skips them for token accounting (if resolved.mode() != OutputMode.TRANSPARENT: continue, before chargeable_inputs += 1), so it expects shielded_fee alone — while the wallet's bestUtxoSelection deliberately selects shielded UTXOs and they reach Fee.calculate as ordinary IDataInputs, so hasInputs is true and we'd over-declare. That's a genuine bug, but it's pre-existing and orthogonal to this PR (present before and after), and the correct discriminator is input transparency, not per-token shielded-output presence — keying on the latter would break the legitimate case above.

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +186,89 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
...
+  // (no provider) — which is enough to inspect what Fee.calculate received.
+  await expect(sendTransaction.prepareTxData()).rejects.toThrow(
+    /Shielded crypto provider is not set/
+  );
+
+  expect(calcSpy).toHaveBeenCalledTimes(1);
+  const feeOutputs = calcSpy.mock.calls[0][1] as { token: string }[];
+  // Only the real transparent '02' output is fee-charged; the shielded output's
+  // phantom is filtered out. Without the fix this would be 2 (phantom included).
+  expect(feeOutputs.filter(out => out.token === '02')).toHaveLength(1);
```
tuliomir: suggestion(non-blocking): the assertion can't tell the phantom from the real output

This counts `'02'` outputs but never checks *which* one survived. Both `'02'` outputs in this tx — the real transparent one and the phantom — carry `token: '02'`, so the count is 1 either way.

Concretely: invert the new filter in `sendTransaction.ts` to `partialOutputs.filter(out => phantomOutputs.has(out))` and `feeOutputs` becomes `[phantom]` — still length 1, suite still green. The mutation this test exists to catch is the one it can't kill.

Assert identity rather than cardinality. The phantom sits at the spend-derived P2PKH while the real output is at the explicit address:

```suggestion
  const fee02 = feeOutputs.filter(out => out.token === '02');
  expect(fee02).toHaveLength(1);
  expect(fee02[0].address).toBe('WgKrTAfyjtNK5aQzx9YeQda686y7nm3DLi');
```

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +186,89 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
   );
 });
 
+test('prepareTxData does not charge FEE_PER_OUTPUT for a shielded FEE-token output', async () => {
```
tuliomir, line 190: suggestion(non-blocking): the melt-branch half of this fix is untested, and self-masking at one shielded output

This test keeps a real transparent `'02'` output, so dropping the phantom moves `outputCount` from 2 to 1 and stays inside the multiply branch. The shape the fix actually *changes* is the other one: a FEE token whose value goes entirely to shielded outputs.

There, with the phantom gone `outputCount` hits 0 and `Fee.calculate` falls through to the melt branch instead (`src/utils/fee.ts:85-87`). With **two** shielded `'02'` outputs and no transparent one the transparent fee moves `2n → 1n`; with **one** it stays `1n`, because the flat melt fee happens to equal the single `FEE_PER_OUTPUT` this PR removes.

That coincidence is the same hazard as `FEE_PER_OUTPUT` and `FEE_PER_AMOUNT_SHIELDED_OUTPUT` both being `1n` — the path is invisible in the smallest fixture and only diverges at two or more. Worth a second case: two shielded FEE-token outputs, no transparent output, asserting `Fee.calculate` receives zero `'02'` outputs and that the melt fee is charged exactly once.

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +186,89 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
...
+      {
+        address: shieldedAddress,
+        value: 10n,
+        token: '02',
+        shieldedMode: ShieldedOutputMode.FULLY_SHIELDED,
+      },
+    ],
+  });
+
+  const calcSpy = jest.spyOn(Fee, 'calculate');
```
tuliomir, line 252: nitpick: this spy is never restored

`jest.config.js` sets no `restoreMocks`/`resetMocks` and this file has no `afterEach`, so the spy stays installed on the module-level `Fee` class for every test that runs after this one.

Harmless today since it calls through, but `mock.calls` keeps accumulating — the next test that spies the same method inherits this call, and its own `toHaveBeenCalledTimes(1)` would then mean something else.

Adding `calcSpy.mockRestore()` at the end of the test (or an `afterEach(() => jest.restoreAllMocks())` for the file) closes it.

tuliomir (dismissed):

__tests__/new/sendTransaction.test.ts
pedroferreira1: You're right, and I reproduced it before fixing: I inverted the filter in sendTransaction.ts to partialOutputs.filter(out => phantomOutputs.has(out)) exactly as you described, and the test still passed. The one mutation it exists to catch was the one it couldn't kill — a count over token === '02' is satisfied by either output.
▎
▎ Applied your suggestion. With the identity assertion in place, that same inverted filter now fails with:
▎
▎ Expected: "WgKrTAfyjtNK5aQzx9YeQda686y7nm3DLi"
▎ Received: "WRict31TWrLrTLYKmnDkBwnt9poUWPKSfb"
▎
▎ — the phantom's spend-derived P2PKH, precisely where you said it would sit.

__tests__/new/sendTransaction.test.ts
pedroferreira1: Confirmed, and the coincidence is exact. I measured Fee.calculate directly with a FEE token that has inputs and N transparent outputs:
▎
▎ | transparent '02' outputs | fee                    |
▎ |--------------------------|------------------------|
▎ | 0                        | 1n ← melt branch       |
▎ | 1                        | 1n ← per-output branch |
▎ | 2                        | 2n                     |
▎ | 3                        | 3n                     |
▎
▎ So with a single shielded output the transparent fee is 1n before and after the fix — the fixture is completely blind to the branch, exactly as you said. Two outputs are the first shape where it diverges (2n → 1n).
▎
▎ Added the case: two fully-shielded FEE-token outputs, no transparent output, asserting that Fee.calculate receives zero '02' outputs and that its result is exactly FEE_PER_OUTPUT — i.e. the melt fee charged once, not twice. I put the "why two, not one" reasoning in a comment on the test so nobody later 'simplifies' it back to a single output and silently loses the coverage.
▎
▎ Both tests now fail without the fix (Expected length: 1 / Received: 2 and Expected length: 0 / Received: 2).

__tests__/new/sendTransaction.test.ts
pedroferreira1: Confirmed on all counts — no restoreMocks/resetMocks in jest.config.js, nothing in setupTests.js either, and no afterEach in this file. Worth adding that the file's own convention already goes the other way: the existing tests call mockRestore() explicitly (lines 431-432, 690, 1476), so the new test was the odd one out.
▎
▎ Took your first option — calcSpy.mockRestore() at the end of the test — since it matches what the rest of the file does, and applied it to the new melt-branch test as well.

tuliomir (dismissed):

src/new/sendTransaction.ts
```diff
@@ -485,11 +485,26 @@ export default class SendTransaction extends EventEmitter implements ISendTransa
...
+    // Phantom outputs are transparent proxies for shielded outputs — they exist
+    // ONLY so UTXO selection accounts for the shielded value, and are stripped
+    // from the final tx (below). They must NOT reach Fee.calculate: the fullnode
+    // charges a shielded output only the shielded fee (added below), never
+    // FEE_PER_OUTPUT, and counts just the real transparent outputs. Leaving a
+    // phantom in would over-declare FEE_PER_OUTPUT for a FEE-token shielded
+    // output; since the node validates the declared fee for an exact match, the
+    // tx would be rejected. Excluding them keeps the wallet in lockstep with the
+    // node: transparent outputs → FEE_PER_OUTPUT (or the flat melt fee when a
+    // FEE token has inputs but no transparent output), shielded → shielded fee.
```
raul-oliveira: Three statements of the same rationale, and two clauses duplicate comments that already exist in this file: "they exist ONLY so UTXO selection accounts for the shielded value" is at `:255` and `:307`, and "are stripped from the final tx (below)" is at `:586`. The closing sentence restates the two before it.

```suggestion
    // Phantoms exist only so UTXO selection accounts for the shielded value.
    // The fullnode counts just the real transparent outputs for FEE_PER_OUTPUT
    // and rejects any fee that isn't an exact match, so a phantom here would
    // over-declare. Shielded outputs are charged separately, in shieldedFee.
```

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +187,181 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
   );
 });
 
+test('prepareTxData does not charge FEE_PER_OUTPUT for a shielded FEE-token output', async () => {
+  // Regression: a shielded output pushes a transparent "phantom" into
+  // txData.outputs so UTXO selection accounts for its value. That phantom must
+  // NOT reach Fee.calculate — the fullnode charges a shielded output only the
+  // shielded fee, never FEE_PER_OUTPUT, and counts just the real transparent
+  // outputs. Leaving the phantom in over-declares FEE_PER_OUTPUT and the node's
+  // exact-match fee check rejects the tx.
```
raul-oliveira: This repeats the source comment almost verbatim, and the test name already carries the *what*.

```suggestion
  // Regression: the phantom pushed for UTXO selection must not reach
  // Fee.calculate — the node charges a shielded output only the shielded fee.
```

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +187,181 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
...
+    outputs: [
+      // A real TRANSPARENT FEE-token output → exactly one chargeable output.
+      {
+        type: OutputType.P2PKH,
+        address: 'WgKrTAfyjtNK5aQzx9YeQda686y7nm3DLi',
+        value: 10n,
+        token: '02',
+      },
+      // A SHIELDED FEE-token output → charged the shielded fee, never
+      // FEE_PER_OUTPUT. Its phantom must be excluded from Fee.calculate.
```
raul-oliveira: Third restatement of the same point, now inline on the fixture. The sibling comment on the transparent output earns its place by naming the expected count; this one doesn't add anything the header hasn't said.

```suggestion
```

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +187,181 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
...
+  await expect(sendTransaction.prepareTxData()).rejects.toThrow(
+    /Shielded crypto provider is not set/
+  );
+
+  expect(calcSpy).toHaveBeenCalledTimes(1);
+  const feeOutputs = calcSpy.mock.calls[0][1] as { token: string; address: string }[];
+  // Only the real transparent '02' output is fee-charged; the shielded output's
+  // phantom is filtered out. Assert WHICH output survived, not just how many:
+  // the phantom carries token '02' too, so a count alone passes even if the
+  // filter is inverted and the phantom is the one that reaches Fee.calculate.
```
raul-oliveira: Keep this one — that a count assertion alone would survive an inverted filter is genuinely non-obvious. Just shorter:

```suggestion
  // Assert WHICH output survived, not just the count: the phantom also carries
  // token '02', so a count alone passes even with the filter inverted.
```

__tests__/new/sendTransaction.test.ts
```diff
@@ -185,6 +187,181 @@ test('prepareTxData resolves 71-byte shielded addresses internally', async () =>
...
+  // The companion test above keeps a real transparent '02' output, so removing
+  // the phantom only moves outputCount 2 -> 1 and stays inside Fee.calculate's
+  // per-output branch. This case exercises the branch the fix actually changes:
+  // with no transparent '02' output, dropping the phantoms takes outputCount to
+  // 0, so Fee.calculate falls through to the flat melt fee instead.
+  //
+  // TWO shielded outputs, deliberately: with one, the transparent fee is 1n
+  // before AND after the fix (the flat melt fee equals the single
+  // FEE_PER_OUTPUT being removed), so a single-output fixture cannot see this
+  // branch at all. With two it moves 2n -> 1n.
```
raul-oliveira: The two-shielded-outputs choice is the part worth documenting; the paragraph comparing against the companion test can shrink.

```suggestion
  // Exercises the branch the companion test can't reach: with no transparent
  // '02' output, outputCount hits 0 and Fee.calculate falls through to the flat
  // melt fee. TWO shielded outputs deliberately — with one, the melt fee equals
  // the single FEE_PER_OUTPUT being removed, so the fee is 1n before and after.
```

raul-oliveira (changes requested):
Verified the fix against the fullnode source on `experimental/shielded-outputs-alpha-v1` and it's correct: `_update_token_info_from_outputs` only ever increments `chargeable_outputs` over `self.outputs` (transparent), shielded outputs are priced separately by `calculate_shielded_fee`, and `verify_sum` / `verify_token_rules` compare `expected_fee != fees_from_fee_header` for an exact match — so an over-declared `FEE_PER_OUTPUT` really does make the tx unspendable. Both new tests are genuinely red without the `feeOutputs` change, and lint/format are clean.

Two things I'd like to see before merging: a few comment trims (left inline as suggestions) and one extra test case.

## The same bug still fires when the input is shielded

The fix filters phantom **outputs**, but inputs that spend shielded UTXOs still count. The node skips those inputs entirely — `transaction.py::_get_token_info_from_inputs`:

```python
if resolved.mode() != OutputMode.TRANSPARENT:
    # Shielded input: skip for token info (amount is hidden)
    continue
...
if token_version == TokenVersion.FEE:
    token_info.chargeable_inputs += 1
```

The `continue` lands before both `chargeable_inputs += 1` and `token_dict[token_uid] = token_info`, so a FEE token whose inputs are all shielded never enters `token_dict` at all.

The wallet doesn't make that distinction. `sendTransaction.ts` resolves a shielded spend from the owned marker and pushes it into `txData.inputs` with `token` and `authorities: 0n` set, indistinguishable from a transparent input. I ran a probe on this branch — FEE token, one shielded input of 20n, two shielded outputs of 10n:

```
inputs seen by Fee.calculate:  [{"token":"ababab...","authorities":"0"}]
outputs seen by Fee.calculate: []
transparent fee = 1n
```

Zero outputs plus a counted input trips `fee.ts`'s "melt without outputs" branch. Node expects `0` there, wallet declares `1n` → `InputOutputMismatch`.

Where it matters:

| FEE-token inputs | Transparent outputs | Wallet (this branch) | Node | Match |
|---|---|---|---|---|
| transparent | ≥1 | N × FEE_PER_OUTPUT | same | ✅ |
| transparent | 0 (all shielded) | flat melt fee | flat melt fee | ✅ |
| shielded | ≥1 (partial unshield) | N × FEE_PER_OUTPUT | same | ✅ |
| **shielded** | **0 (shielded → shielded)** | **flat melt fee** | **0** | ❌ |

The broken row is the shielded→shielded transfer. It's also reachable via `changeShieldedMode`: converting the custom-token change to a shielded output (`sendTransaction.ts` ~449-483) zeroes out the token's transparent outputs.

Worth noting this PR strictly improves the case — it was `2n` before (both phantoms counted) and is `1n` now — it just doesn't reach the `0n` the node wants.

## Suggested third test case

Same shape as the two you added, so it slots in right after `prepareTxData charges the flat melt fee once…`:

| Test | FEE-token input | Transparent outputs | Covers |
|---|---|---|---|
| #1 (yours) | transparent | 1 | phantom isn't charged |
| #2 (yours) | transparent | 0 | flat melt-fee branch |
| **#3 (new)** | **shielded** | **0** | **shielded input shouldn't be charged either** |

Fixture: a parent tx with zero transparent outputs and one *owned* shielded slot of the FEE token (`value` / `token` / `decoded.address` populated, which is what makes it spendable), spent by absolute index 0. Then:

```ts
expect(feeInputs.filter(i => i.token === FEE_TOKEN)).toHaveLength(1);
expect(feeOutputs.filter(o => o.token === FEE_TOKEN)).toHaveLength(0);
await expect(calcSpy.mock.results[0].value).resolves.toBe(0n);
```

Pairing it with a control that swaps the shielded input for a transparent one is what pins the defect to the input's shielded-ness rather than to the zero-output branch: on this branch the control passes at `FEE_PER_OUTPUT` and the shielded case fails at `expected 0n, received 1n`. Happy to push the full test if useful.

## Fix that makes it pass

Mirrors the node's own `continue`, three lines:

```diff
 # src/types.ts — IDataInput
+  isShielded?: boolean;

 # src/new/sendTransaction.ts — where the input is built
+        isShielded: resolved.kind === 'shielded',

 # src/utils/fee.ts
-    const nonAuthorityInputs = Fee.groupTokenElementsByTokenUid(inputs);
+    const nonAuthorityInputs = Fee.groupTokenElementsByTokenUid(
+      inputs.filter(input => !('isShielded' in input && input.isShielded))
+    );
```

Full unit suite stays green with it (1232/1232), including your two tests and `__tests__/utils/fee.test.ts`. One caveat: filtering inside `groupTokenElementsByTokenUid` drops the token from `tokensSet` when *all* its inputs are shielded, which is exactly what the node does (the token never enters `token_dict`) — but it means the two sides have to move together if the node ever starts counting shielded inputs.

Fine by me if you'd rather land this PR as-is and track the input side separately — the two are independent.

## Smaller notes

- `calcSpy.mockRestore()` sits at the end of the test body, so it's skipped when an `expect` throws and the next test's `jest.spyOn` reuses the same mock. I hit this validating the PR: with the fix reverted, test 2 failed with `toHaveBeenCalledTimes(1) → received 2` instead of its own fee assertion. A `try/finally` (like `shuffleSpy` at ~1569) keeps the failure honest.
- The two tests share ~30 identical lines (same `selectUtxoMock`, same four `jest.spyOn` calls) — a small local helper would cover both.
- `as { token: string; address: string }[]` could be `IDataOutputWithToken[]`, the actual type, instead of redeclaring the shape.

src/new/sendTransaction.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1145/changes/b1a8a672b0d113e275bb22010850aea2084b04f0

__tests__/new/sendTransaction.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1145/changes/b1a8a672b0d113e275bb22010850aea2084b04f0

__tests__/new/sendTransaction.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1145/changes/b1a8a672b0d113e275bb22010850aea2084b04f0

__tests__/new/sendTransaction.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1145/changes/b1a8a672b0d113e275bb22010850aea2084b04f0

__tests__/new/sendTransaction.test.ts
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-lib/pull/1145/changes/b1a8a672b0d113e275bb22010850aea2084b04f0

raul-oliveira (approved):
LGTM

tuliomir (approved):

PR #1146, opened 2026-07-25
test: migrate address-info and checkAddressesMine tests

tuliomir:
## Summary
Migrates the remaining address-info coverage out of the monolithic `hathorwallet_others.test.ts` into the shared-tests structure, and promotes `checkAddressesMine` to the shared suite.

- New shared case: `__tests__/integration/shared/addresses.test.ts` — `checkAddressesMine` maps wallet addresses to `true`, a foreign address to `false`, and an empty query to `{}`. The two facades reach that empty result differently (the wallet-service short-circuits before its HTTP call, the fullnode falls out of its per-address loop), so the contract is pinned in the shared suite rather than in either facade-specific one.
- New fullnode-specific file: `__tests__/integration/fullnode-specific/address-info.test.ts` — `getAddressInfo` (4 cases), `getTxAddresses`, and the malformed-address case of `checkAddressesMine`.
- Adapter addition: `checkAddressesMine` on `IWalletTestAdapter` and both implementations. The result is typed as `CheckAddressesMineResult`, an alias of the library's own `WalletAddressMap` — which `IHathorWallet.checkAddressesMine` already declares for both facades.
- Removed from `hathorwallet_others.test.ts`: the `getAddressInfo`, `getTxAddresses` and `checkAddressesMine` describe blocks (259 lines). The file is now down to the index-limit scanning-policy block.
- Removed from `service-specific/addresses.test.ts`: the `checkAddressesMine` case, subsumed by the new shared test plus the existing `isAddressMine` gap-window loop — `HathorWalletServiceWallet.isAddressMine` delegates straight to `checkAddressesMine([address])`, so that loop already drives the same code path across the whole window.

Why address-info stays fullnode-specific: `getAddressInfo()` and `getTx()` throw `Not implemented` on `HathorWalletServiceWallet`, `getTxAddresses()` does not exist there at all, and the malformed-address case cannot round-trip the wallet-service response schema — `z.record(AddressSchema, z.boolean())` admits base58 keys only, and `parseSchema` throws rather than stripping.

The migrated `getAddressInfo` tests previously shared one `beforeAll` wallet and silently depended on balances left behind by earlier tests in the block, with no teardown. Each now builds its own wallet and its own funds, with `afterEach(stopAllWallets)`, and the assertion values were re-derived for the new setup.

## Acceptance Criteria
- `checkAddressesMine` is exercised against both facades from the shared suite.
- The fullnode-specific address-info tests cover only behavior the wallet-service facade cannot express.

## Summary by CodeRabbit

- **Tests**
  - Expanded integration coverage for address ownership checks, including empty-list handling and malformed-address behavior, across wallet facades.
  - Added fullnode facade integration tests for `getAddressInfo`, `getTxAddresses`, and `checkAddressesMine`, with assertions for expected accounting fields.
  - Simplified and reorganized existing address-related test suites, removing redundant coverage and updating shared test documentation/context.


__tests__/integration/fullnode-specific/address-info.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      total_amount_sent: 0n,
+      total_amount_available: 4n,
+    });
+  });
+
+  it('should return correct values for locked utxos', async () => {
+    const hWallet = await generateWalletHelper();
+    const addr0 = await hWallet.getAddressAtIndex(0);
+    const timelock1 = Date.now().valueOf() + 5000; // 5 seconds of locked resources
+    const timelockTimestamp = dateFormatter.dateToTimestamp(new Date(timelock1));
```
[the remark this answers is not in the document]
tuliomir: Not taking the 60s bump — the tight margin is intentional. These two steps really should complete in well under five seconds, and if they ever stop doing so we'd rather have this fail loudly than have the slowdown hidden behind a padded timelock.

That said, the comment did catch a real regression, just not the one described. The `5000` is verbatim from the original test in `hathorwallet_others.test.ts`, but its context changed in this migration: the original ran against a wallet already funded by a shared `beforeAll`, so `now + 5000` was followed immediately by the locking send. Moving the case to a self-funding wallet put a full `injectFunds` genesis round-trip *inside* that five-second budget.

Fixed by computing the timelock after funding, so the window covers only the locking tx and the assertion — same 5s, measured from the right moment: https://github.com/HathorNetwork/hathor-wallet-lib/commit/e12a46076774564611a7f9641d154b6e6be0cf25

Re-ran the suite after the change: 7/7 pass, with the locked-utxos case at 12.9s total (most of which is now outside the window).

__tests__/integration/hathorwallet_others.test.ts
```diff
@@ -1,263 +1,6 @@
...
-    ).toStrictEqual({
-      [address1]: true,
-      [address2]: true,
-      [address3]: true,
-      'invalid-address': false,
-    });
-  });
-});
-
 // getAuthorityUtxos tests moved to shared/authority-utxos.test.ts and fullnode-specific/authority-utxos.test.ts
```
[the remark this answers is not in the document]
tuliomir: Good catch on the file, though the remedy is inverted — the `getAuthorityUtxos` note is accurate and should stay. It was added by the authority-UTXO PR (`f3e15467`), and both files it points to exist; it documents a different migration, not this one.

The actual defect is a *missing* note. Leaving a breadcrumb where a migrated block used to live is a convention across this series — `hathorwallet_facade.test.ts` carries nine of them, `walletservice_facade.test.ts` three — and this PR removed three describe blocks from here without adding any. (It was inconsistent even within the PR: `service-specific/addresses.test.ts` did get its pointer to `shared/addresses.test.ts`.)

Added the two missing notes, keeping the authority one: https://github.com/HathorNetwork/hathor-wallet-lib/commit/54b083fb880cb0c211e357ba229614f19dc4141f

pedroferreira1 (approved):

__tests__/integration/adapters/types.ts
```diff
@@ -368,6 +372,20 @@ export interface IWalletTestAdapter {
...
 
+  /**
+   * Maps each queried address to whether it belongs to the wallet.
+   * Both facades expose `checkAddressesMine()` with the same result shape —
+   * the fullnode resolves it locally from storage, the wallet-service via its
+   * REST API. Callers must pass well-formed base58 addresses: the
+   * wallet-service response schema only admits base58 keys, so malformed
+   * address strings are a fullnode-only scenario
+   * (see `fullnode-specific/address-info.test.ts`).
+   */
```
raul-oliveira, line 383: How each facade resolves the call is exactly what the adapter abstracts away, and it pins this file to `src/` internals. Neighbouring docs here run 1–3 lines.

```suggestion
  /**
   * Maps each queried address to whether it belongs to the wallet.
   * Callers must pass well-formed base58 addresses — malformed strings are a
   * fullnode-only scenario (see `fullnode-specific/address-info.test.ts`).
   */
```

__tests__/integration/adapters/types.ts
```diff
@@ -622,3 +640,14 @@ export interface AdapterAddress {
...
+
+/**
+ * Result of an adapter `checkAddressesMine` query: each queried address mapped
+ * to whether it belongs to the wallet.
+ *
+ * Aliases the library's own {@link WalletAddressMap} rather than restating its
+ * shape: `IHathorWallet.checkAddressesMine` already declares that type as the
+ * return value for both facades, and the fullnode's `Record<string, boolean>`
+ * is structurally identical to it.
+ */
```
raul-oliveira, line 652: Nine lines of docblock on a one-line alias. The second paragraph argues for a decision `= WalletAddressMap` already states.

```suggestion
/** Each queried address mapped to whether it belongs to the wallet. */
```

__tests__/integration/fullnode-specific/address-info.test.ts
```diff
@@ -0,0 +1,287 @@
...
+    const addr2 = await hWallet.getAddressAtIndex(2);
+    const addr3 = await hWallet.getAddressAtIndex(3);
+
+    // Ensure both start as empty addresses
+    expect((await hWallet.getAddressInfo(addr2)).total_amount_received).toStrictEqual(0n);
+    expect((await hWallet.getAddressInfo(addr3)).total_amount_received).toStrictEqual(0n);
+
+    // Fund addr2 with all the amounts the assertions below track. injectFunds
+    // already waits past the funding tx's timestamp before returning, so the
+    // spend below needs no extra wait.
```
raul-oliveira, line 116: First sentence restates the call below. The wait behaviour is the real point — it's what justifies dropping the old `delay(500)`.

```suggestion
    // injectFunds waits past the funding tx's timestamp, so the spend needs no extra wait
```

__tests__/integration/fullnode-specific/address-info.test.ts
```diff
@@ -0,0 +1,287 @@
...
+      total_amount_available: 4n,
+    });
+  });
+
+  it('should return correct values for locked utxos', async () => {
+    const hWallet = await generateWalletHelper();
+    const addr0 = await hWallet.getAddressAtIndex(0);
+
+    // injectFunds already waits past the funding tx's timestamp before
+    // returning, so the spend below needs no extra wait.
```
raul-oliveira, line 144: Same note as the test above — stating it once is enough.

```suggestion
```

__tests__/integration/fullnode-specific/address-info.test.ts
```diff
@@ -0,0 +1,287 @@
...
+
+describe('[Fullnode] checkAddressesMine', () => {
+  afterEach(async () => {
+    await stopAllWallets();
+  });
+
+  // The general (well-formed addresses) behavior is covered for both facades in
+  // shared/addresses.test.ts. Only the fullnode facade accepts an arbitrary
+  // string here: it resolves ownership from local storage, while the
+  // wallet-service response schema only admits base58 keys.
```
raul-oliveira, line 276: The base58 reasoning is already spelled out in this file's header, ~150 lines up.

```suggestion
  // Well-formed addresses are covered for both facades in shared/addresses.test.ts.
```

raul-oliveira (commented):
Comment-level review only — the restructuring itself looks good.

One theme: the "wallet-service schema only admits base58, so malformed addresses are fullnode-only" rationale is written out in **five** places (`adapters/types.ts`, both file headers, and two inline blocks). The suggestions below keep it in full in the `fullnode-specific/address-info.test.ts` header and reduce the rest to pointers.

__tests__/integration/shared/addresses.test.ts
```diff
@@ -180,4 +183,39 @@ describe.each(adapters)('[Shared] addresses — $name', adapter => {
...
+      // Indices 1..3 are comfortably inside every facade's address window
+      // (see SHARED_ADDRESS_SAMPLE).
+      const address1 = await adapter.getAddressAtIndex(wallet, 1);
+      const address2 = await adapter.getAddressAtIndex(wallet, 2);
+      const address3 = await adapter.getAddressAtIndex(wallet, 3);
+
+      // A well-formed base58 address that belongs to another wallet. Malformed
+      // address strings are deliberately NOT exercised here: the wallet-service
+      // response schema only admits base58 keys, so that case is fullnode-only
+      // (see fullnode-specific/address-info.test.ts).
```
raul-oliveira, line 200: `foreignAddress` + `WALLET_CONSTANTS.genesis` already say "a valid address from another wallet".

```suggestion
      // Malformed strings are fullnode-only — see fullnode-specific/address-info.test.ts
```

__tests__/integration/shared/addresses.test.ts
```diff
@@ -180,4 +183,39 @@ describe.each(adapters)('[Shared] addresses — $name', adapter => {
...
+        [address1]: true,
+        [address2]: true,
+        [address3]: true,
+        [foreignAddress]: false,
+      });
+
+      // Empty in, empty out. Both facades agree on the result but reach it
+      // differently — the wallet-service short-circuits before its HTTP call,
+      // the fullnode falls out of its per-address loop — so the contract is
+      // worth pinning here rather than in either facade-specific suite.
```
raul-oliveira, line 215: Why this lives in the shared suite is the file header's job, and the short-circuit/loop detail ties the test to `src/` internals.

```suggestion
      // Empty in, empty out — same contract, different code path per facade
```

__tests__/integration/hathorwallet_others.test.ts
```diff
@@ -1,264 +1,9 @@
...
+// getAddressInfo and getTxAddresses tests moved to fullnode-specific/address-info.test.ts
+// checkAddressesMine tests moved to shared/addresses.test.ts and fullnode-specific/address-info.test.ts
```
raul-oliveira, line 6: Optional: git already records the moves, and this list grows a line per migration PR.

```suggestion
// Tests previously in this file now live in the shared/ and *-specific/ suites.
```

raul-oliveira (approved):
Approving — the comment-trimming suggestions above are all non-blocking.

__tests__/integration/hathorwallet_others.test.ts
tuliomir: The end goal of this project is to remove this file entirely. So having it reduced to a list of comments it a good indicator of this progress.

We'll keep this structure for now.

PR #1152, opened 2026-07-28
fix(nano): honor the caller's change address for the fee change output

pedroferreira1:
### Description

The HTR change output created when selecting utxos to pay the fee was always sent to the wallet's current address, ignoring every caller-supplied change address, while the deposit change output in the same transaction correctly honored its action-level one.

Adds a transaction-level change address to the nano contract builder and to CreateNanoTxOptions, wired from both wallet facades:

- a deposit's change output uses the action changeAddress, then the transaction changeAddress, then a wallet address
- the fee change output uses the transaction changeAddress, then a wallet address. It deliberately does not fall back to a deposit action's address: the fee belongs to the transaction and there may be any number of deposit actions, so no single one owns it.

The transaction-level address is validated once in verify(), so a bad address fails before any utxo is selected rather than from inside fee selection.

The regression test uses a deposit whose change is smaller than the fee, so the deposit change cannot absorb the whole fee. Otherwise the fee change output is never created and the defect stays hidden.

### Acceptance Criteria
- add the transaction-level changeAddress option (both facades)
- add `setChangeAddress()` to the builder
- fix the fee change output address
- define the deposit-change resolution order
- define the transaction-change resolution order, with the rationale for not falling back to a deposit action's address
- add the integration regression test

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for setting a transaction-level change address when creating nano contract transactions.
  * Deposit change uses the deposit-specific address when provided, with wallet address fallback.
  * Fee change is sent to the configured transaction-level address, or the wallet’s current address by default.
  * Invalid or wallet-unowned change addresses are rejected during transaction validation.

* **Tests**
  * Added coverage for fee and deposit change-address selection, address precedence, and validation errors.


src/new/wallet.ts
```diff
@@ -3579,6 +3579,11 @@ class HathorWallet extends EventEmitter {
       builder.setContractPaysFees(newOptions.contractPaysFees);
     }
 
+    // Set the transaction change address if declared
+    if (newOptions.changeAddress) {
+      builder.setChangeAddress(newOptions.changeAddress);
+    }
```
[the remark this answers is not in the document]
pedroferreira1: Any createNanoCreateTx method should not be considered in this PR

src/nano_contracts/types.ts
```diff
@@ -233,6 +233,12 @@ export type CreateNanoTxOptions = {
   maxFee?: OutputValueType;
   /** If the contract will pay the transaction fees (for FEE tokens) */
   contractPaysFees?: boolean;
+  /**
+   * Address for the change outputs created for the transaction itself, notably
+   * the change of the HTR selected to pay the fee. A deposit action's own
+   * `changeAddress` takes precedence for that action's change output.
+   */
+  changeAddress?: string | null;
```
tuliomir, line 246: issue(non-blocking): `changeAddress` is accepted but silently ignored on nano create-token txs

Both `createNanoContractCreateTokenTransaction` methods take `CreateNanoTxOptions`, so `changeAddress` is now type-accepted there and dropped — `selectFeeInputs` is shared, so their fee change still goes to the current address. Wiring is out of scope per your reply, but the type should not advertise an option that does nothing: either `Omit<CreateNanoTxOptions, 'changeAddress'>` on those two signatures, or a doc line here saying it applies only to `createNanoContractTransaction`. Note `CreateNanoTokenTxOptions` already has its own `changeAddress`, so a follow-up must define precedence between the two.

src/nano_contracts/builder.ts
```diff
@@ -399,7 +422,7 @@ class NanoContractTransactionBuilder {
     // If there's a change amount left in the utxos, create the change output
     if (utxosData.changeAmount) {
       const changeAddressStr =
-        changeAddressParam || (await this.wallet.getCurrentAddress()).address;
+        changeAddressParam || this.changeAddress || (await this.wallet.getCurrentAddress()).address;
```
tuliomir, line 425: suggestion: the new deposit-change fallback changes existing behavior and is untested

The added middle rung changes callers who pass a transaction `changeAddress` but no action-level one: their deposit change moves from the wallet's current address to the transaction address. That's the documented intent, but the integration test always supplies `changeAddress` on both deposit actions, so only the first and third rungs are exercised. A case with a deposit action that omits `changeAddress` would pin the rung and prove the precedence order, which is the part of this PR most likely to be quietly reverted later.

src/nano_contracts/builder.ts
```diff
@@ -640,6 +663,12 @@ class NanoContractTransactionBuilder {
       throw new NanoContractTransactionError(message);
     }
 
+    // Validated once here rather than at each use site, so a bad address fails
+    // before any utxo is selected instead of from inside fee selection.
+    if (this.changeAddress && !(await this.wallet.isAddressMine(this.changeAddress))) {
+      throw new NanoContractTransactionError('Change address must belong to the same wallet.');
```
tuliomir: nitpick: the two change-address validations throw an identical message

`executeDeposit` throws the same `'Change address must belong to the same wallet.'` at line 363. With two independent change addresses now in play, a caller who gets this error cannot tell which one was rejected. Worth distinguishing — e.g. `'Transaction change address must belong to the same wallet.'`.

src/wallet/wallet.ts
```diff
@@ -3048,6 +3048,11 @@ class HathorWalletServiceWallet extends EventEmitter implements IHathorWallet {
       builder.setContractPaysFees(newOptions.contractPaysFees);
     }
 
+    // Set the transaction change address if declared
+    if (newOptions.changeAddress) {
+      builder.setChangeAddress(newOptions.changeAddress);
```
tuliomir, line 3053: suggestion: two added lines are without coverage — and one is the whole wallet-service path

Some lines are without coverage:
- `src/nano_contracts/builder.ts` line 669;
- `src/wallet/wallet.ts` line 3053.

Only two, but note what the second one is — this wiring is the wallet-service half of the fix, and nothing executes it.

The new integration test proves the fullnode facade only, so the two facades are not equally evidenced despite carrying identical code. Line 669 is the new `verify()` rejection, reachable from a unit test with a mock storage and no network. Both are cheap; the wallet-service one is the one that matters.

src/nano_contracts/types.ts
pedroferreira1: Confirmed: only createNanoContractTransaction wires it into the builder; the create-token methods drop it. Wiring the create-token path stays out of scope for this PR, so I took the Omit<> route — the type should not advertise an option that does nothing.

The surface was wider than the comment suggests: it's six sites, not two — the createAndSend wrappers on both facades as well, plus the two IHathorWallet declarations. All six narrowed.

Also documented on the option itself that the create-token methods deliberately don't accept it, and that a follow-up wiring them must settle precedence against the change address those methods already take in their create-token options — the point you raised at the end.

src/nano_contracts/builder.ts
pedroferreira1: The coverage half is exactly right, and I'll push back on the other half.

Untested: confirmed empirically. Exactly one test anywhere passes a transaction-level changeAddress, and both of its deposit actions carry their own, so the first rung short-circuits. Deleting this.changeAddress || leaves the entire suite green.

 "Changes existing behavior": not so. changeAddress on CreateNanoTxOptions is introduced by this PR — master's type has only pinCode/maxFee/contractPaysFees/signTx. There are no existing callers passing it, so no one's deposit change moves. It's a new precedence rule that wasn't pinned, rather than a silent change to shipped behaviour.

Pinned it with a three-row precedence table (action → transaction → current address) in a new builder unit test file — the module had no unit tests at all. Unit, not integration, as you suggested: the deposit path only touches five wallet members, so no network is needed.

src/nano_contracts/builder.ts
pedroferreira1: Agreed, applied your wording for the new one. Checked first that nothing is coupled to the string — the message appears twice in the builder and zero times under tests, so no test churn.

Left the deposit-action message as-is: it's shipped, and once the transaction-level one is qualified, an unqualified "Change address…" unambiguously means the action-level one. Happy to qualify both if you'd rather.

src/wallet/wallet.ts
pedroferreira1: Both confirmed, and you're right that the wallet-service one is what matters. setChangeAddress has three references repo-wide and none in tests, and even the wallet-service nano integration suite only ever passes action-level or create-token change addresses — never the new option. Two facades carrying identical wiring with only one of them evidenced is exactly the shape that drifts silently.

Both covered by unit tests, no network. The wallet-service one spies build() and captures what actually landed on the builder rather than asserting the setter fired, with a negative twin for the omitted case. Mutation-verified: deleting the wiring now fails with Expected "WSFK…" / Received: null, where previously nothing caught it.

The builder rejection is covered too — using initialize with a preset blueprint id to keep verify() off the network, and empty actions so it can't pass for the wrong reason via the action-level check.

pedroferreira1:
@coderabbitai do your review

tuliomir (approved):

raul-oliveira (approved):

